PR #103, opened 2019-11-29
adding recipe module to map regionprops bbox to sizes

barentine:


barentine:
Any thoughts @David-Baddeley ?

David-Baddeley:
In general these feel as though they are a) very special case and b) quite granular (this applies to both the older overlap rois and the new bounding box stuff). My gut feeling is that they might be too granular and/or special case to be recipe modules on their own and that we should do some pretty extensive refactoring. I'm almost certain that IdentifyOverlappingROIs doesn't belong in measure. 

I think I've been a bit too lax about allowing additions to the core recipe modules as we are effectively defining a language which makes things really hard to alter in the future whilst maintaining backwards compatibility.

A couple of questions:
- What does MapBoundingBoxSize do, and what is the use case? 
- Where do the ROIs come from? (if it's Measure2D then arguably there is a bug in that Measure2D is not providing an output in sensible units).

I don't feel that MapBoundingBoxSize is something which is likely to be re-used much.

David-Baddeley:
More generally, we shouldn't be using the pyramid metadata within recipe modules (with the exception of the one special case one which remaps coordinates from the super-tiles back to the pyramid) - recipe modules shouldn't care if they are dealing with a super-tile or not (we might need to look at what is reported as voxel sizes for the super-tiles).

barentine:
I’ll add a bounding box special-case to fix Measure2D and change the pyramid to update voxelsize depending on layer selection instead of pyramid pixel size (will log the base tile pixel size elsewhere). To round things out we can drop the x_um completely and put it all in nm, can we not?

David-Baddeley:
Would be good to think about the use cases before rushing in with changes - if we fix measure2d I kind of want to do it properly.

barentine:
This is still broken, and I've resorted to using a plugin instead in the meantime. I think we should give up on fixing measure2d 'right' and just fix it.

barentine:
closing as region_props has changed since writing

PR #104, opened 2019-12-04
add .gitattributes file and normalize line-endings

barentine:
Occasionally the PRs I have been preparing will have a commit where the eol characters change for the entire file (as I'm developing in linux, and e.g. microscope.py currently has CRLF). Submitting this PR with a gitattributes file and renormalizing the eol endings so that we hopefully don't have to deal with this in the future.

See [here](https://help.github.com/en/github/using-git/configuring-git-to-handle-line-endings) for details


David-Baddeley:
If your commit is changing line endings for the whole file, your editor is
configured incorrectly (in general an editor should not change the line
endings in a file - most modern editors detect what is currently used and
run with that).

Is there any real need to have uniform line endings across files? Although
"nice", python doesn't care and I don't see any reason we should change
just for the sake of it.

In general I'm going to be pretty resistant to any changes like this which
make purely cosmetic and unnecessary changes to large fractions of the code
base as they obscure the authorship of files and make it hard to give
people the credit they deserve for their work.  We should discuss this
further, but if I do accept such changes I'd want them to be attributed to
a "reformatting bot" (or similar) rather than a specific user.

barentine:
It is a git setting, not Pycharm 2019. If I instead set `git config --global core.autocrlf false` I actually get a bit shouted at
image
but if I "fix" things, then I also "fix" every CRLF file I checkout/change/commit

I figured doing this in a single commit is probably the least offensive way of doing it if we're going to do it, and with the gitattributes file it shouldn't be a problem again afterwords.

Reformatting bot is fine by me, and I agree it would also be worth having a follow-up conversation about author crediting while you are in town.

David-Baddeley:
Hmm - that looks like a pycharm error, rather than a git error. Git should
be perfectly capable of checking out as is and committing as is without
messing with things.

barentine:
Yes, a pycharm warning. If you take a look at the git docs they suggest `input` rather than `false` where `input` "tell[s] Git to convert CRLF to LF on commit but not the other way around". Why they suggest this rather than just telling git to ignore it, I'm not completely sure - it could be a legacy editor issue as you alluded to earlier. That global setting option is "old", and the recommendations are to us a gitattribute file (which has the advantage of getting everyone on the same page regardless of their global git settings). [for more reading](https://adaptivepatchwork.com/2012/03/01/mind-the-end-of-your-line/)

barentine:
We've moved past this, and will probably never normalize line endings, and that's OK

PR #105, opened 2019-12-05
Cameramap selection to PYMEAcquire and startacq metadata

barentine:
- Allows camera calibration maps to be set / saved into metadata following the same database model as for the pixel size
- removes explicitly printed u's and displays pixel size out to 1 nm precision

### DIFF
--- a/PYME/Acquire/acquiremainframe.py
+++ b/PYME/Acquire/acquiremainframe.py
@@ -79,7 +79,8 @@ class PYMEMainFrame(AUIFrame):
         self.AddMenuItem('Camera', 'Toggle ROI\tF8', self.OnMCamRoi)
         self.miCamBin = self.AddMenuItem('Camera', 'Turn binning on', self.OnMCamBin, itemType='check')
         self.AddMenuItem('Camera', 'Set pixel size', self.OnMCamSetPixelSize)
-        
+        self.AddMenuItem('Camera', 'Set camera maps', self.OnMCamSetCameraMaps)
+
         #Acquire menu
         self.AddMenuItem('Acquire', 'Snapshot\tF5', self.OnMAquireOnePic)
               
@@ -530,6 +531,11 @@ class PYMEMainFrame(AUIFrame):
         dlg = voxelSizeDialog.VoxelSizeDialog(self, self.scope)
         dlg.ShowModal()
 
+    def OnMCamSetCameraMaps(self, event):
+        from PYME.Acquire.ui.camera_map_dialog import CameraMapDialog
+
+        dlg = CameraMapDialog(self, self.scope)
+        dlg.ShowModal()
 
     def OnMCamRoi(self, event):
         logging.debug('Stopping frame wrangler to set ROI')
--- a/PYME/Acquire/microscope.py
+++ b/PYME/Acquire/microscope.py
@@ -415,7 +415,11 @@ class microscope(object):
         if not 'StartupTimes' in tableNames:
             self.settingsDB.execute("CREATE TABLE StartupTimes (component TEXT, time REAL)")
             self.settingsDB.execute("INSERT INTO StartupTimes VALUES ('total', 5)")
-            
+        if not 'CameraMaps' in tableNames:
+            self.settingsDB.execute("CREATE TABLE CameraMaps (ID INTEGER PRIMARY KEY, name TEXT, dark_path TEXT, flat_path TEXT, var_path TEXT)")
+        if not 'CameraMapHistory' in tableNames:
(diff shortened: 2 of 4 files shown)


barentine:
Note that the camera map dialog layout may leave a bit to be desired, but it works for now

David-Baddeley:
A couple of quick comments - this breaks backwards compatibility as the unicode string prefixes are required on python2.7. 

It also partly duplicates and may not play nicely with existing camera map metadata propagation (currently the camera class is responsible for writing camera map metadata - see AndorZyla.py, and seeing as the maps are an intrinsic property of the camera this is probably a more natural place to put ). It's hard to know which of the two will win out  in practice.

barentine:
I prioritized py3 cosmetics over py2, being not terribly worried about the display of mu on py2. I won't put up a fight about changing it back, but I don't think we need to be very concerned about it.

I generate camera maps at the frame rate (and roi size) I plan on imaging at. This doesn't change often, but neither does my pixel size, so I thought it was a reasonable way to implement things. Note that this PR also answers the todo item in AndorZyla to make the cameramap metadata setting compatible with the cluster, which is obviously a priority of mine. I could imagine something complicated to automatically switch maps when intended, similar to how one might automate switching of pixel sizes if they have a motorized turret, etc., but I think this is a bit messy and not really worth the time.

The microscope object itself appends its GenStartMetadata method to the list during its initialization, i.e. first. The andor genstartmetadata will be called after, meaning that if we left both in place the camera class method should override the setting. This is probably unexpected for a user, and would eventually be nice to deprecate if we choose the database approach.

David-Baddeley:
You keep forgetting that the high-throughput system is a bit of a special
case and python3 currently represents a small minority of installations -
breaking the python2 version for the sake of it is going to be unacceptable
for the foreseeable future. Likewise, the majority of sCMOS users change
integration times and ROIs quite frequently, meaning that accomodating this
is important. The existing code does this in a relatively simple way (whole
chip maps at a range of integration times) and breaking this for existing
users is similarly not acceptable.

The new code does not solve the FIXME in the AndorZyla.py - all that would
be required here is looking in PYME-CLUSTER:\\CALIBRATION as well as
<PYMEDATADIR>\CALIBRATION. I still maintain that the correct way of doing
things is to have the camera handle the camera map metadata, as it handles
the rest of the camera metadata (the other potential alternative would be
the LMAnalysis module). I definitely don't think that it belongs in the
microscope object (pixel sizes are handled here because they derive from a
combination of hardware properties and are thus a property of the
microscope as a whole).

There are a few potential issues with the current approach and small ROIs /
fast integration times - for your case of fixed ROIs the workaround of
taking the map on the smaller ROI and inserting it into a blank full field
image should work, although this does not solve the general case. It would
also make sense to factor some of the code out of the Andor code and into a
more general sCMOS camera mixin.

I'm also not super-keen on adding more stuff to the settings database. In
many ways, it is a bit of a headache (specifically in multi-user situations
where getting the access permissions right under windows is a pain) and
needs a bit of a re-think. Ultimately I'd like to move much of this to the
PYME.config mechanism (although this would require a way to write to
PYME.config).

barentine:
Discussing a mu display may not the best use of our time. You should have write-access to this PR - please do as you like.

My camera maps are padded out to the full size - I can analyze whole-chip images with my 800 Hz maps, I'm sorry I didn't make that clear. This PR doesn't break anything related to camera map size, just allows easily switching between preferred camera maps if you do change roi / integration times.

Did I not fix the fixme? As you said, simply toss pyme-cluster://<rel path> in the settings dialog.

Unless you're doing away with the settings database completely, this doesn't add anything to the current permissions situation. We run our microscopes under a single user, which I thought was fairly standard.

David-Baddeley:
The existing code automatically selects the appropriate map for a given
integration time - your PR breaks this (and hence does not fix the fixme).

The long term goal would be to do away with the database completely, and
PYME is designed to run multi-user (this makes file management much easier
when users come to save data - it's also useful when you let people use it
unsupervised so they don't do silly things like installing itunes - before
you say this is unlikely, I've had it happen). Generally running single
user is a cop out  - and the fact that it is so common in commercial
microscope software should be an embarrassment. While adding more to the
setting database does not really make the situation worse, it does add to
the things which depend on the status quo and therefore makes it harder to
fix in the future.

I've fixed the FIXME so that it should now work for the cluster, and also
done a bit of refactoring to make it more easily pluggable into other
cameras.

barentine:
Thanks David. Don't love fixed filenames, but they are reasonable.

Closing - will submit separate pr using the mixin on the orca.

David-Baddeley:
The fixed filenames are a bit of a pain - but arguably less error prone than having to have the user remember to change things when they change the integration time (and/or enter arbitrary text strings in a GUI).

PR #107, opened 2019-12-05
htsms acquisition changes

barentine:
- add protocol for taking calibration stacks (no focus lock enabling or auto analysis. shorter steps. camera restart on each step)
- change spoolcontroller not to open every series if it is not auto localized.
- add a button to open series without localizing

### DIFF
--- a/PYME/Acquire/SpoolController.py
+++ b/PYME/Acquire/SpoolController.py
@@ -268,17 +268,6 @@ class SpoolController(object):
     def StopSpooling(self):
         """GUI callback to stop spooling."""
         self.spooler.StopSpool()
-
-    def OpenSeries(self):
-        """
-        GUI callback to open current series in dh5view without starting analysis
-        """
-        from PYME.Acquire import QueueSpooler, HTTPSpooler
-        if isinstance(self.spooler, QueueSpooler.Spooler):
-            subprocess.Popen('%s -q %s QUEUE://%s' % (self._dh5view_cmd.strip(' -g'),
-                                                      self.spooler.tq.URI, self.queueName), shell=True)
-        elif isinstance(self.spooler, HTTPSpooler.Spooler):
-            subprocess.Popen('%s %s' % (self._dh5view_cmd.strip(' -g'), self.spooler.getURL()), shell=True)
         
     def SpoolStopped(self, **kwargs):
         self.seriesCounter +=1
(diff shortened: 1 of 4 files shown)


David-Baddeley:
This is unfortunately going to be another problematic one, particularly as I try and shift people over from the old-style analysis to using the cluster code locally. In doing this I want as little to change in the UI as possible.

What use case are you trying to accomodate here that is not already met? The new "Open" button largely duplicates the current function of the "Analyse" button. We can debate what that button should be called, but  I'm not sure adding another button that in most cases does the same thing helps. 

The only unsupported use case I can think of is opening a series which you are automatically streaming and analysing on the cluster. How often do you need to do this?

Your proposed changes also break the current workflow for non-high throughput imaging using the cluster (or a local cluster). Will comment on this where it occurs in code.

PYME/Acquire/SpoolController.py
```diff
@@ -306,8 +317,6 @@ def LaunchAnalysis(self):
                 if self._analysis_launchers.full():
                     self._analysis_launchers.get().join()
                 self._analysis_launchers.put(t)
-            else:
```
David-Baddeley, line 309: This has to remain - otherwise the workflow for non-high throughput imaging with the cluster is broken. Current workflow is as follows:

click "Series", protocol launches dh5view when enough frames are in to start analysis, check parameters in dh5view, click "go". 

If your issue is that you are getting loads of dh5view windows popping up when you spool to the cluster without autoanalysis, then change the protocol so that it doesn't launch analysis.

barentine:
Thanks David, does the latest change seem workable? I backed out any hdfspoolframechanges and simply added an optional parameter to LaunchAnalysis.

barentine:
I should clarify that indeed the only use case I was trying to accommodate was imaging many series with a protocol that would normally launch analysis, without launching analysis (or opening 10,000 dh5view windows). I don't care about the "open" button, I was just trying to maintain what I perceived as the use case for opening dh5view when analysis isn't actually being launched.

David-Baddeley:
How important is that you use the same protocol when automatically analysing as not? Running `LaunchAnalysis` from a protocol when you don't actually want to launch analysis seems a bit counterintuitive. 

The `allow_open` option also feels a bit weird

barentine:
Sorry to be brief - most of the time I want auto analysis, but since I’d realized detection wasn’t working right I wanted to just acquire the data while I sorted it out and localize after.

I guess there are a couple things that we might want to look at refactoring, otherwise I could easily have 4 or 5 protocols. Number of frames per z stack, for instance, might be nice to have in the z stack settings GUI panel

For now I’ll make another protocol, but only if I need to image again before fixing detection

barentine:
declining PR, will for now use additional protocols for imaging without auto analysis

PR #108, opened 2019-12-06
Adding / correcting camera intensity-related unit docs

barentine:
also includes trivial fix to a `plt.cm` missing the `plt`, and updating the noise properties of the htsms camera.

### DIFF
--- a/PYME/localization/remFitBuf.py
+++ b/PYME/localization/remFitBuf.py
@@ -228,6 +228,19 @@ class CameraInfoManager(object):
             return mp
 
     def correctImage(self, md, img):
+        """
+        Parameters
+        ----------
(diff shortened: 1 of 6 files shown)


PYME/localization/remFitBuf.py
```diff
@@ -545,6 +558,29 @@ def __call__(self, gui=False, taskQueue=None):
 
     @classmethod
     def calcSigma(cls, md, data):
+        """
+        Estimate a per-pixel noise standard deviation (overestimate).
```
David-Baddeley: What do you mean by overestimate?

barentine: Sorry about that - for some reason I'd thought you told me that once, but I was just looking at the paper (yours and the IEEE) and you're right that it's not justified. Thanks for catching it.

PR #109, opened 2019-12-07
spoolframe quantization settings fixes

barentine:
Typo in electronsPerCount is self explanatory. Other fix is someone one can encounter with a noise-prop-less camera, e.g. fakecam in the simulator, see below. Raising an error gives the user a chance to manually set the offset.

```ERROR:root:Exception whilst putting files
  File "/home/smeagol/code/python-microscopy/PYME/Acquire/HTTPSpooler.py", line 176, in _queuePoll
    pzf = PZFFormat.dumps(frame, sequenceID=self.sequenceID, frameNum = imNum, **self.compSettings)
  File "/home/smeagol/code/python-microscopy/PYME/IO/PZFFormat.py", line 255, in dumps
    header['QuantOffset'] = quantizationOffset
ValueError: could not convert string to float: 'auto'
Exception in thread Thread-33:
  File "/home/smeagol/anaconda2/envs/py36/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/smeagol/anaconda2/envs/py36/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/smeagol/code/python-microscopy/PYME/Acquire/HTTPSpooler.py", line 176, in _queuePoll
    pzf = PZFFormat.dumps(frame, sequenceID=self.sequenceID, frameNum = imNum, **self.compSettings)
  File "/home/smeagol/code/python-microscopy/PYME/IO/PZFFormat.py", line 255, in dumps
    header['QuantOffset'] = quantizationOffset
ValueError: could not convert string to float: 'auto'```

### DIFF
--- a/PYME/Acquire/ui/HDFSpoolFrame.py
+++ b/PYME/Acquire/ui/HDFSpoolFrame.py
@@ -393,9 +393,7 @@ class PanSpool(afp.foldingPane):
             try:
                 q_offset = self.scope.cam.noiseProps['ADOffset']
             except AttributeError:
-                msg = "Camera doesn't define noise properties, manually set the desired quantization offset"
-                ans = wx.MessageBox(msg, 'Error', wx.OK)
-                raise AttributeError(msg)
+                ans = wx.MessageBox("Camera doesn't define noise properties, manually set the desired quantization offset" , 'Error', wx.OK)
         else:
             q_offset = float(q_offset)
 
--- a/PYME/Acquire/ui/HDFSpoolFrame.py
+++ b/PYME/Acquire/ui/HDFSpoolFrame.py
@@ -400,7 +400,7 @@ class PanSpool(afp.foldingPane):
 
         #quantization scale in GUI is in units of sigma, convert to ADU
         try:
-            q_scale = float(self.tQuantizeScale.GetValue())/self.scope.cam.GetElectrPerCount()
+            q_scale = float(self.tQuantizeScale.GetValue())/self.scope.cam.GetElectronsPerCount()
         except (AttributeError, NotImplementedError):
             print("WARNING: Camera doesn't provide electrons per count, using qscale in units of ADUs instead")
             q_scale = float(self.tQuantizeScale.GetValue())
--- a/PYME/Acquire/ui/HDFSpoolFrame.py
+++ b/PYME/Acquire/ui/HDFSpoolFrame.py
@@ -393,7 +393,9 @@ class PanSpool(afp.foldingPane):
             try:
                 q_offset = self.scope.cam.noiseProps['ADOffset']
             except AttributeError:
-                ans = wx.MessageBox("Camera doesn't define noise properties, manually set the desired quantization offset" , 'Error', wx.OK)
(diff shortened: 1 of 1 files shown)


David-Baddeley:
This currently fails softly at this point so as not to unnecessarily impede old school spooling (specifically for EMCCD cameras for which the AD offset varies with EM gain and temperature and is not consistent enough to put in the camera noise properties).  As it stands, spooling to file and queue currently work without quantization (and hence continue fine if you don't change the field from auto). Status quo is far from ideal, and was a temporary compromise. I currently prefer the status quo (marginally) as I don't really want users to have to think about quantization unless they need to, and it currently works in an "if I ignore it it goes away" sense, rather than having the user have to know that they need to enter a meaningless number in the offset field to carry on. That said, we should fix it properly for all use cases.

Moving forwards, there are two possible approaches (not necessarily mutually exclusive) 
1) move the hard failure into `HTTPSpooler` so that it fails slightly later on `StartSpooling()` (this should still avoid failing on every frame).
2) test which spooler we are using and only populate the compression settings if needed.

barentine:
I have 24 hr of worthless data after two late nights due to settings of 100 ADU offset and 0.5 scale somehow reverting to the pzf internal defaults of 0 ADU and 1.0. I'll continue looking into it, but for me the experience makes me want to avoid "ignore it it goes away" meets quantization, even if there's a case where its harmless. 

I like option 2. I assume that is in addition to this PR (i.e. adding a hard-fail if 'auto' doesn't work and you're spooling with compression)

David-Baddeley:
Option 1 is the safest option when it comes to preventing mishaps. My preference would thus be to revert the first bit of this PR and implement 1, in the short term, but also doing 2 later.

To be a bit more specific about what I propose with 1, it would be adding sanity checks in `HTTPSpooler` (somewhere around lines 153-158 where we set the spooler compression settings), to check if the compression values are sane and raising an exception if not. We might also want to change the default offset to something that is definitely going to fail the saneness test (e.g. -ve something large) to ensure that data is not going to be accidentally streamed with the wrong settings.

David-Baddeley:
Following up on my last comment, your PR would likely not have prevented the bad data as I suspect it was spooled using the action manager which doesn't touch the UI code. The error needs to be raised in `SpoolController.StartSpooling()` (or in something called from here) for it to protect you when doing automated acquisition.

David-Baddeley:
Just pushed a fix for 1) to `HTTPSpooler`. We might still have issues with the compression settings when spooling from the action manager - I believe this will use the default compression settings. These are now safe (quantization disabled by default for now) but not necessarily what you would want.

David-Baddeley:
To properly fix your issue, we need to add compression settings to the args on line 210 in `ActionUI`, ideally reading them from the spooling GUI. A clean solution here is not super simple. 

Best option here would probably be to move the compression "state" to SpoolController and update it every time you change one of the compression UI fields, unfortunately this requires a bunch of wiring and GUI code (or the use of traits). 

A temporary work-around to get things running quickly on the HT system would be to hard-code compression settings into ActionUI and come back to this (ideally soon) for a proper solution.

PR #115, opened 2019-12-12
Define scope "camera position"

barentine:
Set center of (0, 0) pixel as (0, 0) xy stage position, and call this (0, 0) "camera position". 

"Camera position" is more omniscient than stage position, as it tracks the stage position as well as camera ROI changes.


barentine:
reference frame is already the camera chip origin, and there are now helper functions to make this clear. In the future what might be useful is a position GUI slider/go-to in these units so people can easily get to, e.g. tileviewer positions.

PR #120, opened 2019-12-15
allow all remfitbuf prefit steps to be done on the GPU

barentine:
Corresponds to the varunits branch of the pyme-warp-drive repository. 

Note that depending on what we decide to do with remFitBuf, we will need to fix/remove functionality for AstigGaussGPUFitFR accepting a background parameter as an ndarray.


David-Baddeley:
Reinstated a comment and reverted `FitFactory.__init__(...)` signature so that fit factories still use a common signature.

PR #124, opened 2019-12-19
small fixes

barentine:
1. type error in tileviewer toggle_roi
2. use next function instead of method so things run py2 and 3 instead of just 2. Change to multiview is tested, as is the non-iterator usage in PYME/DSView/modules/blobMeasure.getNewID. PYME/contrib/lzw.py change is to an iterator so this should be fine but I did not explicitly test.
3. recent PYME.IO.tabular changes have removed the h5f attribute of HDFSource's that opened a file. Not sure if there is a cleaner way to open an HDFSource and MetaDataHandler, but this is at least functional.

### DIFF
--- a/PYME/recipes/multiview.py
+++ b/PYME/recipes/multiview.py
@@ -91,8 +91,9 @@ class ShiftCorrect(ModuleBase):
 
         try:  # try loading shift map as hdf file
             with unifiedIO.local_or_temp_filename(loc) as f:
-                shift_map_source = tabular.HDFSource(f, 'shift_map')
-                shift_map_source.mdh = HDFMDHandler(shift_map_source.h5f)
+                t = tables.open_file(f)
+                shift_map_source = tabular.HDFSource(t, 'shift_map')  # todo - is there a cleaner way to do this?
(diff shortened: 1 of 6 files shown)


PYME/recipes/multiview.py
```diff
@@ -91,15 +91,18 @@ def execute(self, namespace):
 
         try:  # try loading shift map as hdf file
             with unifiedIO.local_or_temp_filename(loc) as f:
-                shift_map_source = tabular.HDFSource(f, 'shift_map')
-                shift_map_source.mdh = HDFMDHandler(shift_map_source.h5f)
+                t = tables.open_file(f)
+                shift_map_source = tabular.HDFSource(t, 'shift_map')  # todo - is there a cleaner way to do this?
```
David-Baddeley, line 95: This is probably fine - although ensuring that the file gets closed (does `tables.open_file()` return a context manager that you can use with `with`?) on an error would be slightly cleaner. The reference in to the h5 file in HDFSource was removed to avoid dangling references keeping files open (I think this was causing issues with the high throughput analysis).

PR #125, opened 2019-12-20
Modifications to triangle_mesh tests, deletions where needed

zacsimile:


David-Baddeley:
Currently on hold until I can get the rest of the tests passing again and so can verify that this has really fixed things.

zacsimile:
I would argue we're safe to merge this now. "pytest python-microscopy/tests/PYME/experimental/test_triangle_mesh.py" now passes all tests, where as it did not before. There are no dependencies on other parts of PYME, other than _octree, which passes all of its tests.

David-Baddeley:
What's holding me up is that the CI checks are failing (this is likely due to other factors), but as skipping these tests was needed to get the CI up in the first place, I'd like to get back there before enabling them again. We could merge with the `pytest.mark.skip(...)`  bit uncommented and then look at this again once everything else is working.

zacsimile:
OK, I put pytest.mark.skip back in.

PR #127, opened 2020-01-10
Hardware/microscope_adapter: add adapter for Python microscope laser devices

carandraug:
This provides a Python microscope laser adapter for PyME. It should enable PyME to use all the lasers that Python microscope currently supports. I have tested with a Toptica iBeam and fake laser.

Configuration wise, something like this can be used (for the Toptica case):

```
@init_hardware('Lasers')
def lasers(scope):
    from microscope.lasers.toptica import TopticaiBeam
    from PYME.Acquire.Hardware.microscope_adapter import MicroscopeLaser
    l = TopticaiBeam('/dev/ttyUSB0')
    l.initialize()
    scope.l488 = MicroscopeLaser(l, name='mslaser')
    scope.l488.register(scope)
```

or like this for the test lasers on microscope side:

```
@init_hardware('Lasers')
def lasers(scope):
    from microscope.testsuite.devices import TestLaser
    from PYME.Acquire.Hardware.microscope_adapter import MicroscopeLaser
    l = TestLaser()
    l.initialize()
    scope.l488 = MicroscopeLaser(l, name='mslaser')
    scope.l488.register(scope)
```

### DIFF
--- a/PYME/Acquire/Hardware/microscope_adapter.py
+++ b/PYME/Acquire/Hardware/microscope_adapter.py
@@ -1,0 +1,62 @@
+#!/usr/bin/python
+
+##################
+# Copyright David Miguel Susano Pinto, 2020
+# david.pinto@bioch.ox.ac.uk
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+##################
+
+"""Adapters to the Python microscope package devices.
+
+This module provides adapter classes to devices controlled with the
+Python microscope package <https://pypi.org/project/microscope/>.
+"""
+
+import PYME.Acquire.Hardware.lasers
+
+
+class MicroscopeLaser(PYME.Acquire.Hardware.lasers.Laser):
+    """
+    Adapter for Python microscope laser classes.
+
+    Parameters
+    ----------
+        laser: an object, already initialized, implementing the
+            `microscope.devices.LaserDevice` interface.
+    """
+    def __init__(self, laser, *args, **kwargs):
+        # We can't call super() at the start because the parent class
+        # may try to call self.TurnOn().
+        self._laser = laser
+        self.powerControlable = True
+        self.MIN_POWER = self._laser.get_min_power_mw()
+        self.MAX_POWER = self._laser.get_max_power_mw()
+        super().__init__(*args, **kwargs)
+
+    def IsOn(self):
+        return self._laser.get_is_on()
+
+    def TurnOn(self):
+        self._laser.enable()
+
+    def TurnOff(self):
+        self._laser.disable()
+
+    def SetPower(self, power):
+        self._laser.set_power_mw(power)
+
+    def GetPower(self):
+        return self._laser.get_power_mw()


David-Baddeley:
Can probably merge, but a couple of quick thoughts:

1) verbosity: There is a fair bit of boilerplate in the examples. There might be a case for having the adapter at least call `initialize()` on the laser, possibly even have it instantiate the laser. If we expect this to be widely used, it might be reasonable to have proxies in the adapter class for each of the lasers from the `microscope` package, so initialisation would look like:

```python
@init_hardware('Lasers')
def lasers(scope):
    from PYME.Acquire.Hardware.oxford_micron import TopticaiBeam
    scope.l488 = TopticaiBeam('/dev/ttyUSB0', name='mslaser')
    scope.l488.register(scope)
```

This would be much more consistent for a user accessing different lasers

2) naming: I think `microscope_adapter` is potentially confusing to someone looking at the code. Something that better establishes the provenance - e.g. `oxford_micron` is probably better.

3) long term it might be better to either re-implement the code in `microscope` or to use the postulated common interface that is being talked about in https://github.com/python-data-acquisition (this is still early days and a common interface is some ways off). The `microscope` package introduces a dependency we don't want (pyro) and has some, in my opinion, problematic architectural choices (mainly related to baking inter-process communication in at such a low level - as well as attracting undesirable dependencies this will throttle performance - particularly when talking to cameras etc... pyro also brings name-resolution headaches).

David-Baddeley:
following up on the proxy stuff - there is probably a way of doing this which doesn't result in too much boilerplate within the adapter code

carandraug:
> Can probably merge, but a couple of quick thoughts:
>
>     1. verbosity: There is a fair bit of boilerplate in the
>     examples. There might be a case for having the adapter at least
>     call `initialize()` on the laser, possibly even have it
>     instantiate the laser.

I have thought the same about initialize.  The reason to not do it is
that it is possible to get an already constructed and initialized
device.  Actually, the recommendation on Python microscope is to use a
device server in which case the device will indeed already have been
initialized.  I kept it out of the example in this PR to keep things
as self contained and easy to test as possible.

>     If we expect this to be widely used, it
>     might be reasonable to have proxies in the adapter class for
>     each of the lasers from the `microscope` package, [...]

Not sure how well this would scale as we add support for more devices
in microscope, but if you think it makes things easier, sure.

>     2. naming: I think `microscope_adapter` is potentially confusing
>     to someone looking at the code. Something that better
>     establishes the provenance - e.g. `oxford_micron` is probably
>     better.

Python Microscope is also being worked on by people in Bielefeld,
Osaka, Montpellier, and Porto, so it shouldn't be renamed to
`oxford_micron`.

The name `microscope_adapter` marks the module as being an adapter of
Microscope device classes to PYME device classes.  I understand that
the name might cause complications if you already have a Microscope
class on PYME though, but I hope that the module subpackage might give
a good hint on what it does.  But feel free to change its name to
whatever you think is best.

>     2. long term it might be better to either re-implement the code
>     in `microscope` or to use the postulated common interface that
>     is being talked about in
>     https://github.com/python-data-acquisition (this is still early
>     days and a common interface is some ways off). The `microscope`
>     package introduces a dependency we don't want (pyro) and has
>     some, in my opinion, problematic architectural choices (mainly
>     related to baking inter-process communication in at such a low
>     level - as well as attracting undesirable dependencies this will
>     throttle performance - particularly when talking to cameras
>     etc... pyro also brings name-resolution headaches).

This PR only provides an adapter for Microscope classes, there's no
need to make it an hard dependency.  Also, the inter-process stuff in
Microscope is separated from the actual device interface and not
baked-in at all.

David-Baddeley:
The `devices.py` package which defines the device base classes  imports `Pyro4` and there's pyro stuff all through the `DataDevice` base class - that would seem pretty "baked in" to me :).

Just to be clear, the musings in comment 3 are mostly just brain farts, thoughts for the future, and a little bit of a "here be dragons" warning - your adapter is very, very welcome, it just depends on the goals. If your goal is to support the devices in the `microscope` package, then it's great. If it's to add support for, e.g., a Toptica laser, or a different camera it might be fine but there might be better ways and should warrant some thought.

After using Pyro extensively I've got cheesed off enough with some of the limitations that removing / minimising any Pyro dependency is a fairly high priority goal for PYME.

David-Baddeley:
As to naming, we have a `microscope.py` as a core part of PYMEAcquire, which is part of the naming confusion. The "microscope" package is currently hosted on the micron github, hence my micron_oxford suggestion. Would it's github address (micron_microscope) be a reasonable compromise?

The initialisation stuff was also mostly just thoughts - things potentially being initialised is a good argument for keeping things as is. The `from microscope import ... ` might still be problematic, however due to naming conflicts (I suspect that this might be OK on py3 because you'd need to do a `from .microscope` to get the local file, but not on py2).

carandraug:
> The devices.py package which defines the device base classes imports Pyro4 and there's pyro stuff all through the DataDevice base class - that would seem pretty "baked in" to me :)

I see your point. But there's only two places, inside DataDevice, and they're conditional. You can pass a normal `Queue` as a client and it will work in the same manner. We are in the process of even removing that (see issue MicronOxford/microscope#49 ).

> [...] If your goal is to support the devices in the microscope package, then it's great.  [...]

This is the long term goal. I started with lasers because they're relatively simple and it already has a defined interface on PyME.  I have since spoken with @barentine which showed me some more. I plan on extending to other device types if you find this welcome. Extending to other devices types would be made easier by having PyME defined interfaces for other device types (also already spoke with @barentine  about this and will open an issue to discuss that later).

> The "microscope" package is currently hosted on the micron github, hence my micron_oxford suggestion. Would it's github address (micron_microscope) be a reasonable compromise?

I would really prefer to not have Micron or Oxford on the name. Those things change, and indeed we have even created a [python-microscope](https://github.com/python-microscope) github organization, we just never got around to fix some of our infrastructure to do the repo transfer. Would `python_microscope_adapter` do it?

David-Baddeley:
Re defined interfaces.

PYME also has a defined interfaces for cameras (explicit through the `PYME.Acquire.Hardware.Camera.Camera` base class), and positioning devices (piezos, stages, etc ..., through `PYME.Acquire.Hardware.Piezos.piezo_base.BasePiezo` class). The latter is probably poorly named as it's the base class / interface for translation stages as well, and could use a lot of work on it's docstrings. 

Note that all light sources (e.g. arc lamps) and shutters currently use the laser interface. 

Of the other devices we currently support (e.g. filter wheels, DMDs, power meters) we only have one example of each, making the current implementation the default reference.

There are some wrinkles in the current interface definitions in that PYME will sometimes check for a function or attribute which is not present in the base class and use it if present (these features of the interface are hence defined implicitly). None of this is core functionality, but doing some refactoring to clarify this is also on the TODO list.

carandraug:
That's neat that there's already an interest on having a defined interface on the PyME side.  That will make future work on other device types adapter easier.

But regarding this PR, other than the module name change (and to what then?), is there any change needed?

David-Baddeley:
basically it's just the name stuff that is holding it up at this point (and it's still unclear to me what this should be). I can probably merge for now and then rename later. A shim / translation layer for the micromanager c-api is also on my todo list, and there might be a case for moving all shims to a common folder / subpackage under Hardware, which might in turn resolve the naming issues.

Ultimately it would also be nice if the configuration examples in the PR could make their way into the module docstring.

PR #128, opened 2020-01-15
py3 byte-str fix in clusterUI recipe views

barentine:
bytes.strip(str) fails and there's no reason to encode in the first place since query-dicts should have str keys

### DIFF
--- a/PYME/cluster/clusterUI/recipes/views.py
+++ b/PYME/cluster/clusterUI/recipes/views.py
@@ -22,7 +22,7 @@ def recipe_template(request):
 def get_input_glob(request):
     from PYME.IO import clusterIO
     
-    filepaths = clusterIO.cglob(request.GET.get('glob').encode().lstrip('/'))
+    filepaths = clusterIO.cglob(request.GET.get('glob').lstrip('/'))
     
     return render(request, 'recipes/input_list.html', {'filepaths' : filepaths,'serverfilter' : server_filter})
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I'm trying to recall why this was `encode()`ed in the first place. Looking back it would seem that  the `encode` is there from the first commit (so pre-dates any py3 work). Only thing I can think of is if django was giving us unicode rather than string.

David-Baddeley:
This is one of the huge annoyances with the 2->3 transition, on py2, encode returns string. 

It's not the dictionary keys we are encoding, it's the values stored at those keys (which may have been unicode).

David-Baddeley:
I suspect that the original `.encode()` was there as a "force this to be a string, whatever it is" to make sure that it had a suitable encoding for use in a URI. URIs are by definition UTF8 encoded c-strings (i.e. 1 byte per character) not unicode (which potentially has a variable number of bytes per character). The more I have to deal with it, the more annoying python3 strings are (in python2, there was a 1:1 mapping between python strings, c strings, and whatever you might write to a file of internet socket, this is unfortunately not the case in py3).

I suspect we could probably merge with no adverse effects, but it makes the code less *safe*.

barentine:
I agree the str/bytes fixes can be annoying, but I think we're mostly done. Re:safe, I suspect very few people using clusterUI recipe views will be running py2. 

Are we all set to merge?

David-Baddeley:
Currently it's around 50:50 for py2/py3 on the recipe views. Will merge with the proviso that we might need to revert if this proves a problem.

PR #131, opened 2020-01-23
store datasource uri in cluster localization results metadata

barentine:
store datasource name in localization results metadata for convenience in linking. Particularly helpful for recipes which require both localizations and the datasource they came from.

### DIFF
--- a/PYME/cluster/HTTPTaskPusher.py
+++ b/PYME/cluster/HTTPTaskPusher.py
@@ -159,6 +159,7 @@ class HTTPTaskPusher(object):
         self.taskQueueURI = _getTaskQueueURI()
 
         self.mdh = metadata
+        self.mdh['DataSourceID'] = self.dataSourceID  # for convenient linking after results files get downloaded/moved
 
         #load data source
         if dataSourceModule is None:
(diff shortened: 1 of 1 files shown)


David-Baddeley:
This would seem to partially duplicate `Analysis.DataFileID` (ID/hash of the data file) and `imageID` (UUID/hash identifying a particular raw series and all associated analyses.

Your use case might be slightly different (it appears you want the filename, whereas you'd need to do a lookup through a database which you don't have set up do do the reverse search based on IDs). 

I can see an argument for also storing the filename / URI, but we should try and keep things consistent naming wise. Maybe `Analysis.DataFileURI` ??

David-Baddeley:
PS - rationale for suggesting a different metadata key is the ID's are always hashes / UUIDs / database keys (i.e. not URIs). We should probably change the signature in `HTTPTaskPusher` as well, but this is lower priority. I'm a bit more anal about trying to make sure new metadata keys are sensible because they become a defacto standard which is really hard to change later.

barentine:
I can get behind 'Analysis.DataFileURI', 'Analysis.ImageSeriesURI', or 'Analysis.ImageStackURI'. I might like 'Localization.ImageStackURI' best - potentially most specific so it isn't misused/clobbered downstream.

Do you have a preference?

David-Baddeley:
'Analysis.DataFileURI' is probably most consistent with 'Analysis.DataFileID', so I'd probably argue for that. I'm pretty sure that recipes tend to nest metadata of the original file, which should hopefully avoid clobbering.

barentine:
Cool, all set @David-Baddeley

PR #133, opened 2020-01-24
keep track of which points are coalesced in multiview merge function

barentine:
This PR aims to enable going back to raw blinking frames after looking at merged localizations. Currently this is a bit muddled by taking the average of multiviewChannel keys during merge, and not recording which points are merged. Both of these issues are addressed in this PR.

### DIFF
--- a/PYME/Analysis/points/multiview.py
+++ b/PYME/Analysis/points/multiview.py
@@ -2,29 +2,42 @@
 import numpy as np
 from PYME.Analysis.points.DeClump import pyDeClump
 
-def coalesce_dict_sorted(inD, assigned, keys, weights_by_key):  # , notKosher=None):
+def coalesce_dict_sorted(inD, assigned, keys, weights_by_key, n_coalesced_key='n_coalesced'):
     """
-    Agregates clumps to a single event
-    Note that this will evaluate the lazy pipeline events and add them into the dict as an array, not a code
-    object.
-    Also note that copying a large dictionary can be rather slow, and a structured ndarray approach may be preferable.
-    DB - we should never have a 'large' dictionary (ie there will only ever be a handful of keys)
+    Aggregates clumps to a single event. Input dictionary should have every key sorted by the 'assigned' input. Output
+    dictionary will additionally have a key:value pair of the number of points used to produce each output point.
 
-    Args:
-        inD: input dictionary containing fit results
-        assigned: clump assignments to be coalesced
-        keys: list whose elements are strings corresponding to keys to be copied from the input to output dictionaries
-        weights_by_key: dictionary of weights.
+    Parameters
+    ----------
+        inD: dict
+            input dictionary containing fit results
+        assigned: 1darray, list
+            clump assignments to be coalesced
+        keys: list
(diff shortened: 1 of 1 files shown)


PYME/Analysis/points/multiview.py
```diff
@@ -2,30 +2,43 @@
...
     """
     from PYME.Analysis.points.DeClump import deClump
 
-    NClumps = int(np.max(assigned) + 1)  # len(np.unique(assigned))  #
-
     clumped = {}
 
+    NClumps = int(np.max(assigned) + 1)
+    # record number of points being coalesced for each clump
+    uni, counts = np.unique(assigned.astype(int), return_counts=True)
```
David-Baddeley, line 38: Do we need to worry about performance issues here?

PYME/Analysis/points/multiview.py
```diff
@@ -2,30 +2,43 @@
...
+    clumped[n_coalesced_key] = np.repeat(uni, counts)
```
David-Baddeley: Doesn't this generate an array of the original length, not the clumped length?

PYME/Analysis/points/multiview.py
barentine: '''
timeit.timeit('np.unique(test, return_counts=True)', setup='import numpy as np; test = np.random.randint(0, 100, int(1e6))', number=100) / 100
0.03026972654974088
'''
The 'right' thing to do would be piggy backing on the loop being run in deClump.findClumpsN, and have it return a counts-per group. As to where this fits in our/my priority queue, maybe we leave it for now and come back if it bothers us/others downstream?

PYME/Analysis/points/multiview.py
barentine: Yes, which is bad! Sorry about that.

barentine: Now fixed.

PYME/Analysis/points/multiview.py
David-Baddeley: The query was because I thought it was already piggy-backed for the standard (not multi-view) declumping. It was actually more the np.repeat that I have performance worries about.

Agree it's not high priority.

PYME/Analysis/points/multiview.py
```diff
@@ -2,30 +2,43 @@
...
     from PYME.Analysis.points.DeClump import deClump
 
-    NClumps = int(np.max(assigned) + 1)  # len(np.unique(assigned))  #
-
     clumped = {}
 
+    # record number of points being coalesced for each clump
+    uni, counts = np.unique(assigned.astype(int), return_counts=True)
+    NClumps = int(uni[-1] + 1)  # +1 for the zero clump
+    clumped[n_coalesced_key] = np.concatenate([[0], counts]) # the zero clump, unclumped, is not included in assigned
```
David-Baddeley, line 40: Are you sure that the zero clump is not included in `assigned`? I was pretty sure that `assigned` has the same length as the input data (which by inference means that it must include the zero clump).

barentine:
closing as this PR is stale - can readdress later if this becomes important

PR #135, opened 2020-01-28
removing problematic en-/de-codings in recipe outputs

barentine:
jinja2.Environment.render returns a (unicode) str, and unless we change open(filename) to open(filename, 'r+b'), which we currently don't, in unifiedIO.read, we get a str out:

```
  File "/home/smeagol/code/python-microscopy/PYME/recipes/output.py", line 368, in get_source
    source = unifiedIO.read(os.path.join(os.path.dirname(__file__), template)).decode('utf-8')
AttributeError: 'str' object has no attribute 'decode'
```

Even on py2
```
Python 2.7.16 |Anaconda custom (64-bit)| (default, Mar 14 2019, 21:00:58) 
[GCC 7.3.0] on linux2
>>> with open('/home/smeagol/code/localization-survey/localization_survey/reports/localization_survey.html') as f:
...     s = f.read()
... 
>>> type(s)
<type 'str'>
```

### DIFF
--- a/PYME/recipes/output.py
+++ b/PYME/recipes/output.py
@@ -365,9 +365,9 @@ class UnifiedLoader(jinja2.BaseLoader):
         from PYME.IO import unifiedIO
         try:
             if os.path.exists(os.path.join(os.path.dirname(__file__), template)):
-                source = unifiedIO.read(os.path.join(os.path.dirname(__file__), template)).decode('utf-8')
+                source = unifiedIO.read(os.path.join(os.path.dirname(__file__), template))
             else:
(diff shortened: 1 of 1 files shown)


David-Baddeley:
remember that py2 string == py3 bytes - the `.decode()` is absolutely required on py2. Solution would be to change `unifiedIO.read` to  use `open(filename, 'rb')`, as the alternative HTTP GET option for read should return bytes (everything is bytes in transit).

**NOTE:** `'rb'` **NOT** `'r+b'` as the latter opens in append mode, not read mode.

David-Baddeley:
Fixed unified IO

Issue #134, opened 2020-01-28
How should one add jinja filters to recipes.output.ReportOutput environment?


barentine:
I've written a plugin including a report template. I currently have my filter defined/added to the environment in `myplugin/reports/__init__.py`, which works fine if I import and `reports.env.get_template()`. My filter is very plugin-specific and does not belong (as I assume most will not) in PYME.recipes.output - how should I add it to the standard environment?

zacsimile:
Seeing as this is merged, can we close this issue?

PR #137, opened 2020-01-30
add config tree support for reports (jinja2 templates/filters)

barentine:
As raised issue #134, this PR:

- adds reports folder to pyme config tree, with subfolders for templates and filters
- refactors jinja2 environment/loader/template/etc out of `PYME.recipes.outputs.py` and into `PYME.reports` by way of the __init__ which pulls from `PYME.reports.templates`, `PYME.reports.filters`, which themselves pull from the config tree to populate the jinja2 environment
- opens report output file in `wb` mode for py3 compatibility


PYME/config.py
```diff
@@ -31,6 +31,11 @@
       |     |
       |     |- recipes
       |           |- anothermodule.txt
+      |     |- reports
+      |     |     |- templates
+      |     |     |     |-somemodule.txt
```
David-Baddeley: Need examples / documentation of what the templates and filters config files will look like

PYME/recipes/output.py
```diff
@@ -357,47 +357,6 @@ def default_view(self):
                     [Item(tn) for tn in params] + [Item('_'), ] +
                     [Item(tn) for tn in outputs], buttons=['OK', 'Cancel'])
 
-#from jinja2 import Environment, PackageLoader
-import jinja2
-import os
-class UnifiedLoader(jinja2.BaseLoader):
```
David-Baddeley, line 363: Moving this into a new module means that the existing templates under `recipes/templates` won't be found.

PYME/config.py
```diff
@@ -279,6 +284,45 @@ def get_plugins(application):
...
+        keys are the filename + extension, e.g. template.html, and values are the full path
+    """
+    templates = {}
+    for config_dir in config_dirs:
+        template_glob = glob.glob(os.path.join(config_dir, 'plugins', 'reports', 'templates/*.txt'))
+
+        for fn in template_glob:
+            with open(fn, 'r') as f:
+                for directory_listed in f:
+                    found = glob.glob(os.path.join(directory_listed, '*.html'))
```
David-Baddeley: I'd prefer not to force templates to be html. There's nothing about jinja which forces things to be html, and adding, e.g. a PDF output using templated latex and latex2pdf is on the wishlist (reasonably far down, but I've done similar things in the past and don't want our design to preclude this). Even if we stick to HTML, there are at least 3 valid file extensions (.htm, .html, .xhtml).

David-Baddeley: Probably the best way of solving this is to just have `UnifiedLoader` try all the directories we have listed as options. 

The other option is to stick with the status quo and require an explicit path to templates not stored in `recipes/templates`.

David-Baddeley: In either case, looking up from a global dictionary which is keyed by template name (with no path) is likely to problematic, as the likelihood of name collisions will be high. The chance that two plugins will elect to both call their template "report.html" is much greater than 0.

I'd suggest that plugin templates should need to be accessed through the module namespace, e.g. `nep_fitting/report.html`.

PYME/config.py
barentine: Do we want that here, or is it OK to leave it just in the template plugin repository?

PYME/recipes/output.py
barentine: Good catch

barentine: I moved the omnipresent report templates from recipes/templates to reports/templates, which makes more sense and fixes this issue

PYME/config.py
barentine: Now added

PYME/config.py
barentine: Things are now setup (and documented) to access templates through the module namespace with no limitations on extensions.

PR #138, opened 2020-01-30
fix blobFinding issues

csoeller:
When trying to use blob finding the other day I came across two buglets that kept me from using the module successfully. One is in the main blobFinding module and seems to be from using a now discontinued wx flag, the other was an invalid print statement from within ofind3d. I removed the invalid flag; I deleted the print statement since I felt this was a debugging hack that is most likely not needed any more.


David-Baddeley:
Thanks!

PS - you might want to add the email address used for earlier commits to your github account so that you get recognized (I coalesced them all to one email address when doing the git conversion, but I can't remember which one.

csoeller:
Thanks, I managed to add the exeter email address so that contribs for past commits are now appearing.

PR #139, opened 2020-01-30
Adding circular tile scanning

barentine:
- adds `pointScanner.CircleScanner` which inherits from `PointScanner` what it can. `CircleScanner` scans starting from a center position radially outwards in what should be the optimal path (rois are assumed to be square and spaced accordingly) 
image In retrospect building a zig-zagged circle approximation would probably allow for less overlap between tiles.

- adds `tiler.CircularTiler` which inherits from `tiler.Tiler` and `CircleScanner`
- adds `CircularTilePanel` which inherits from `tile_panel.TilePanel`, and allows scan radius to be define in microns
- allows `tile_pyramid.ImagePyramid` generation by `Tiler` and `CirclularTiler` to set origin regardless of start position(/shape)


David-Baddeley:
I think I've got it to a point where I'm happy to merge

Issue #143, opened 2020-02-06
Cameras: GetElectronsPerCount, GetNoiseProperties, or both?


barentine:
`HDFSpoolFrame` looks calls `scope.cam.GetElectronsPerCount` when setting quantization scale. This is only implemented on the ucam480 stuff, while `GetNoiseProperties` is a method of the base Camera class. 

I'll submit a PR to instead call getnoiseproperties in hdfspoolframe. Should we also change the calls in the ucam code?

David-Baddeley:
as discussed in #144 things are now unified around the `noise_properties` attribute (which should have been in the base class from the start). Can probably close this now.

PR #144, opened 2020-02-06
Use Camera base class method to get e- per ADU in spoolframe. camelCase kill in orca code

barentine:
See issue #143

### DIFF
--- a/PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
+++ b/PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
@@ -77,7 +77,7 @@ DCAMPROP_TRIGGERSOURCE_SOFTWARE = 3
 
 DCAMCAP_START_SEQUENCE = ctypes.c_int32(int("-1",0))
 
-noiseProperties = {
+noise_properties = {
 '100233' : {
         'ReadNoise': 3.51,
         'ElectronsPerCount': 0.47,
@@ -107,7 +107,7 @@ class HamamatsuORCA(HamamatsuDCAM, CameraMapMixin):
     def __init__(self, camNum):
         HamamatsuDCAM.__init__(self, camNum)
 
-        self.noiseProps = {}
+        self.noise_properties = {}
         self.waitopen = DCAMWAIT_OPEN()
         self.waitstart = DCAMWAIT_START()
         self.initialized = False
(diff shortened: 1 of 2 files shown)


David-Baddeley:
This is actually surprisingly hard due to the possibility of a camera having multiple noise characteristics depending on e.g. gain settings or readout modes. I'm not sure there is a consistent way of handling this between cameras.

Putting the noise properties in the init script is attractive from this perspective (different cameras can still handle things differently), but has the potential to break down badly when you e.g. swap cameras between microscopes, and adds a lot of verbosity / bloat if you have to pass noise properties for multiple different readout modes. You'd also need to redefine the __init__signature on all cameras. An advantage of the init script route is that it puts all of the camera config in one place and forces you to think about it when you add a camera.

A `PYME.config` option would offload much of the verbosity, but makes it hard to deal with the fact that different cameras need different things. It's unclear if it would reduce the possibility of parameter rot or exacerbate it as the config is machine local and and somewhat invisible. This could be somewhat offset by putting the global config dir (/etc/PYME or <anaconda dir>/etc/PYME) under version control, but that also incurs overhead and has some risks.

All told, I like the idea of having things in an external .yaml file (or similar - it should be human readable/editable whatever we use).

My proposal:

- we modify the base class to specify that all cameras should implement a `.noise_properties` property that returns a dictionary of the current properties
- we modify the `__init__()` signature to accept a `noise_info` parameter which can either be a path to a .yaml file (preferred), or a dictionary. The resulting `noise_info` dictionary **must** be keyed on the camera serial number (this should help avoid problems when swapping cameras), but what sits under that is up to the individual camera.
- we make the above parameter optional, with cameras using either a `PYME.config` option to specify the location of the .yaml file, or a fixed location within the pyme config tree by default (this allows, amongst other things, a smooth upgrade path by transferring the current hard-coded properties to a .yaml that we distribute with PYME).

The use of a path to a .yaml allows a master file to be held on shared storage, if desired, making camera swaps etc ... easy.

David-Baddeley:
As an upgrade path, we could also keep the hard coded values in the files for now, with the data coming from `noise_info` being used to `update()` the hard-coded dictionaries.

David-Baddeley:
The spooler stuff should now be consistent across cameras, using a `.noise_properties` property, modulo a bit of cleanup, largely invalidating this PR. Still need to follow up on part 2 (improving how the noise properties are specified).

PR #145, opened 2020-02-07
Spatial-sorting ActionPanel and associated traveling salesperson tweaks

barentine:
- Adds `PathMinimizingActionPanel`, which overrides the `_add_ROIs` method to optimize the imaging route
- modifies the standard `ActionPanel._add_ROIs` method to ensure timeout is long enough to image the ROIs being queued
- adds a greedy / nearest-neighbor path optimizer, which is an important performance bootstrap for two-opt sorting
- adds kmeans point splitting for multiprocessed two-opt chunk function
- updates deprecated names in recipes/test_measurement


PYME/Acquire/ui/actionUI.py
```diff
@@ -175,29 +177,30 @@ def OnAddSequence(self, event):
...
 
-        for x, y in rois:
-            args = {'state': {'Positioning.x': float(x) - roi_offset_x, 'Positioning.y': float(y) - roi_offset_y}}
+        # do a greedy sort relative to our stage position
+        positions = np.reshape(rois, (len(rois), 2)).astype(float) - np.array([roi_offset_x, roi_offset_y])[None, :]
+
+        n_frames = int(self.tNumFrames.GetValue())
+        nice = float(self.tNice.GetValue())
+        time_per_task = n_frames * n_frames / self.scope.cam.GetFPS()
+        timeout = max(float(self.tTimeout.GetValue()), 1.25 * time_per_task)  # allow enough time for what we queue
```
David-Baddeley: I'm pretty sure this timeout calculation is not doing what you want - needs to be > ` n_sequences_queued*n_frames/FPS` (but even this is not particularly safe if there are any higher priority tasks in there)

PYME/Acquire/ui/actionUI.py
```diff
@@ -175,29 +177,30 @@ def OnAddSequence(self, event):
...
         roi_offset_x, roi_offset_y = self.scope.get_roi_offset()
 
-        for x, y in rois:
-            args = {'state': {'Positioning.x': float(x) - roi_offset_x, 'Positioning.y': float(y) - roi_offset_y}}
+        # do a greedy sort relative to our stage position
+        positions = np.reshape(rois, (len(rois), 2)).astype(float) - np.array([roi_offset_x, roi_offset_y])[None, :]
+
+        n_frames = int(self.tNumFrames.GetValue())
+        nice = float(self.tNice.GetValue())
+        time_per_task = n_frames * n_frames / self.scope.cam.GetFPS()
```
David-Baddeley: why `n_frames*n_frames` ???

PYME/Acquire/ui/actionUI.py
```diff
@@ -222,5 +225,57 @@ def OnROIsFromTileviewer(self, event):
...
+        """
+        Parameters
+        ----------
+        sort_epsilon: float
+            Relative per-index exit tolerance for two-opt sorting. See traveling_salesperson.two_opt
+        """
+        ActionPanel.__init__(self, parent, actionManager, scope)
+        self._sort_epsilon = sort_epsilon
+
+    def _add_ROIs(self, rois):
```
David-Baddeley: Several points:

- I don't like the idea of over-riding `_add_rois()` as it duplicates logic. Much better to call `_add_rois(sort_rois(rois))` or similar.
- There is far too much logic here in what should be GUI code (what happened to model view separation?). The original `_add_ROIs()` function is already pushing it.
- I don't like the idea of subclassing a GUI class to change the backend logic (this is a really gross design pattern)

PYME/localization/traveling_salesperson.py
```diff
@@ -84,7 +126,7 @@ def two_opt_test(route, i, k, distances, k_max):
 
 
 
-def two_opt_(distances, epsilon, initial_route=None):
+def two_opt(distances, epsilon, initial_route=None, fixed_endpoint=False):
```
David-Baddeley: Gits merge seems to have done something gross when merging our alterations to this file - it ends up having two competing definitions of `two_opt()` with the same name (and possibly other issues).

PYME/Acquire/ui/actionUI.py
David-Baddeley: Medium term I could envisage letting ActionUI have a user selectable sort function. As a short term solution I'll suggest the following:

- give `actionUI` an `roi_sort_fcn` property, which defaults to `roi_sort_fcn = lambda rois : rois`. Might need to optimise the signature here depending on what the sort needs to know, e.g. current position? 

- modify `_add_rois()` so it looks like:
```python
def _add_rois(self, rois):
    rois = self.roi_sort_fcn(rois)
    ### rest of _add_rois here ...
```

David-Baddeley: Longer term, potentially factor some of the rest of `_add_rois()` into the ActionManager, rather than actionUI.  

Note, the travelling salesman `roi_sort_fcn` should ideally be in a separate file to the actionUI (although could potentially be a module-level function for now - this lets us write it in a way which semi-separate from the GUI, and worry about where to put it later).

PR #146, opened 2020-02-07
Offset piezo / e816T dll event logging

barentine:
- Adds PiezoOnTarget events for the e816T dll piezo so we know where we actually land and when rather than guessing / being inaccurate (given our target tolerance)
- Adds `LogFocusCorrection` call to disablelock such that we get an 'update offset' event and know how to interpret the position logged by the PiezoOnTarget events, since these are from the base piezo (with no offset included in the position)

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
@@ -23,10 +23,10 @@
 
 #import serial
 import time
-
+import numpy as np
 import logging
 import threading
-
+from PYME.Acquire.eventLog import logEvent
 from .base_piezo import PiezoBase
 from PYME.Acquire.Hardware.GCS import gcs
 
@@ -197,10 +197,9 @@ class piezo_e816(PiezoBase):
     def GetTargetPos(self,iChannel=0):
         return self.lastPos
 
-import numpy as np
 class piezo_e816T(PiezoBase):
     def __init__(self, identifier=None, maxtravel=12.00, Osen=None, hasTrigger=False, target_tol=.002,
-                 update_rate=0.005, spool_controller=None):
+                 update_rate=0.005):
         """
 
         Parameters
@@ -221,7 +220,6 @@ class piezo_e816T(PiezoBase):
         self.units = 'um'
         self._target_tol = target_tol
         self._update_rate = update_rate
-        self._spool_controller = spool_controller
 
         self.lock = threading.Lock()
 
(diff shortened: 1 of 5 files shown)


PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
```diff
@@ -200,7 +200,7 @@ def GetTargetPos(self,iChannel=0):
 import numpy as np
 class piezo_e816T(PiezoBase):
     def __init__(self, identifier=None, maxtravel=12.00, Osen=None, hasTrigger=False, target_tol=.002,
-                 update_rate=0.005):
+                 update_rate=0.005, spool_controller=None):
```
David-Baddeley: really unacceptable to require a reference to the spool controller.

PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
```diff
@@ -294,6 +295,13 @@ def _Loop(self):
                     # logging.debug('Moving piezo to target: %f' % (pos[0],))
 
                 if np.allclose(self.position, self.targetPosition, atol=self._target_tol):
+                    if not self.onTarget and self._spool_controller:
+                        try:
+                            self._spool_controller.spooler.evtLogger.logEvent('PiezoOnTarget',
```
David-Baddeley: the correct way to log an event is:

```python
from PYME.Acquire import eventLog
eventLog.logEvent(....)
```

This automatically logs the event with anything which has registered as an event consumer (ie. the spool controller will get the event - you don't need to hardcode the routing.)

David-Baddeley: Also, you should generally **not** supply a timestamp - leave that for the spooler. This is mostly to make sure that the simulator works (it runs faster than real time and spoofs it's timestamps), but also allows us to, e.g. use a hardware timestamp source in the future.

PYME/Acquire/Scripts/init_htsms.py
```diff
@@ -51,10 +51,9 @@ def pz(scope):
     from PYME.Acquire.Hardware.Piezos import piezo_e816_dll, offsetPiezoREST as opr
 
     # try and update the pifoc position roughly as often as the PID / camera, but a little faster if we can
-    scope._piFoc = piezo_e816_dll.piezo_e816T(maxtravel=100, target_tol=0.05, update_rate=0.002)
-    #scope.hardwareChecks.append(scope._piFoc.OnTarget)
+    scope._piFoc = piezo_e816_dll.piezo_e816T(maxtravel=100, target_tol=0.035, update_rate=0.002,
+                                              spool_controller=scope.spoolController)
```
David-Baddeley: see comments above about not passing the spool controller around

David-Baddeley:
I've updated eventLog.py to add a bit more info about how the event logging is supposed to work.

PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
barentine: > I've updated eventLog.py to add a bit more info about how the event logging is supposed to work.

Thanks for this, @David-Baddeley  - sorry I'd missed that.

barentine: For the timestamp, is it preferable to supply it here since we are in a thread (and this piezo will never be used by the simulator)?

barentine:
How written in stone is the 'update offset' event name? Looks like it's used by the pyro offset piezo. Do we want it CapsCase like the other event names -> PiezoOffsetUpdate?

PR #147, opened 2020-02-07
silence freespace not implemented for cluster warning

barentine:
It is obviously not implemented/functional from the "nan" displayed in the spool panel, and the debug logs (much less std out) are painful to look through since this warning gets displayed on each spool progress tick


David-Baddeley:
Commented out  instead of deleting and left warning in comment with FIXME, so at least if anyone looks in the code they will realise that this is broken and needs attention.

I do think this is something we should look at having working soonish (particularly for those we want to push onto a cluster of one setup), and know I'll forget if given half a chance.

PR #148, opened 2020-02-08
Threaded pairwise distances

zacsimile:
Faster pairwise distances using threading. Order-of-magnitude improvement, but not scaling with number of threads as expected. Still useful.

### DIFF
--- a/PYME/Analysis/points/DistHist/distHistThreaded.py
+++ b/PYME/Analysis/points/DistHist/distHistThreaded.py
@@ -2,13 +2,17 @@ import numpy as np
 import threading
 from .distHist import *
 
+import multiprocessing
+
+NUM_PROCS = multiprocessing.cpu_count()
+
 def _distanceHistogramThreadWrapper(x1, y1, x2, y2, nBins, binSize, hist):
     # Wrapper to deal with capturing the threading result, because we want this to work 
     # with Python 2. If we used Python 3 only, we could use concurrent.futures.ThreadPoolExecutor
     # as an executor for the pool of threads and results.
     hist += distanceHistogram(x1, y1, x2, y2, nBins, binSize)
 
-def distanceHistogramThreaded(x1, y1, x2, y2, nBins, binSize, split=12):
+def distanceHistogramThreaded(x1, y1, x2, y2, nBins, binSize, split=NUM_PROCS):
     """
     distHist.c distanceHistogram computed in # split parallel tasks
     """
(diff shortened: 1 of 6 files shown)


PYME/Analysis/points/DistHist/distHistThreaded.py
```diff
@@ -0,0 +1,67 @@
...
+import threading
+from .distHist import *
+
+def _distanceHistogramThreadWrapper(x1, y1, x2, y2, nBins, binSize, hist):
+    # Wrapper to deal with capturing the threading result, because we want this to work 
+    # with Python 2. If we used Python 3 only, we could use concurrent.futures.ThreadPoolExecutor
+    # as an executor for the pool of threads and results.
+    hist += distanceHistogram(x1, y1, x2, y2, nBins, binSize)
+
+def distanceHistogramThreaded(x1, y1, x2, y2, nBins, binSize, split=12):
```
David-Baddeley: it might be worth making the default splitting match the number of CPUs/threads

PYME/recipes/measurement.py
```diff
@@ -389,9 +389,9 @@ def execute(self, namespace):
         pos0 = namespace[self.inputPositions]
         pos1 = namespace[self.inputPositions2 if self.inputPositions2 is not '' else self.inputPositions]
         if np.count_nonzero(pos0['z']) == 0 and np.count_nonzero(pos1['z']) == 0:
-            res = DistHist.distanceHistogram(pos0['x'], pos0['y'], pos1['x'], pos1['y'], self.nbins, self.binSize)
+            res = DistHist.distanceHistogramThreaded(pos0['x'], pos0['y'], pos1['x'], pos1['y'], self.nbins, self.binSize)
```
David-Baddeley: Due to the fact that both the cluster and bakeshop will run `n_cores` recipes in parallel, recipe modules should be single threaded by default (although they may define an optional boolean parameter to run multithreaded).

David-Baddeley:
A note on the scaling - you probably don't strictly expect things to scale with the number of threads for two reasons. a) hyperthreading only brings you so much. On a 6 core 12 thread machine, 12 threads is roughly 1.5 as fast as 6 (depending on workload), and b) at some point you become limited by memory bandwidth.

zacsimile:
Will make the changes above shortly. Does it also make sense to PEP8 this section of the code (e.g. convert distHist to dist_hist) since it's fairly self-contained?

David-Baddeley:
It might not be as self contained as we'd like (used in a few places in PYME and probably also in a few notebooks), and will force everyone using the source distribution to rebuild their c code (do a new `setup.py develop` or `setup.py build_ext -i`) the next time they pull. I'd leave the PEP8ing for now.

zacsimile:
OK. Made all other changes.

Issue #140, opened 2020-02-03
Cannot create dark and variance maps from dh5view


HelenSun09:
There is no menu option to generate dark and variance maps from within dh5view. This would be preferable to running

`python .\gen_sCMOS_maps.py "C:\Users\Helen Sun\Desktop\20200113_Dark map_50fps_000_000.dcimg"`

David-Baddeley:
Absolutely, it's been on the wishlist for a while.

barentine:
@David-Baddeley or @HelenSun09  I think this can be closed now

PR #151, opened 2020-02-12
Dark and variance map generation from wx

zacsimile:
Response to Issue #140.

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -1959,3 +1959,103 @@ class StatisticsByFrame(ModuleBase):
             pass
             
         namespace[self.output_name] = res
+
+@register_module('DarkAndVarianceMap')
+class DarkAndVarianceMap(ModuleBase):
+    input = Input('input')
+    output_variance = Output('variance')
+    output_dark = Output('dark')
+    dark_threshold = Float(10000)  # this really should depend on the gain mode (12bit vs 16 bit etc)
+    variance_threshold = Float(90000)  # again this is currently picked fairly arbitrarily
+    start = Int(0)
+    end = Int(0)
+    uniform = Bool(False)  # simulate a uniform map
+
+    def execute(self, namespace):
+        from PYME.IO.MetaDataHandler import NestedClassMDHandler
+        from PYME.Analysis import gen_sCMOS_maps
+
+        logging.basicConfig() # without it got 'No handlers could be found for logger...'
+
+        default_sensor_size = (2048,2048) # we currently assume this is correct but could be chosen based
+                                # on camera model in meta data TODO - add CCD size to camera metadata
+        blemish_variance = 1e8  # clipping factor for broken pixels (super high variance)
+
+        image = namespace[self.input]
+
+        # pre-checks before calculations to minimise the pain
+        sensor_size = list(default_sensor_size)
+        try:
+            sensor_size[0] = image.mdh['Camera.SensorWidth']
+        except (KeyError, AttributeError):
+            logger.warning('no valid sensor width in metadata - using default %d' % sensor_size[0])
+        try:
+            sensor_size[1] = image.mdh['Camera.SensorHeight']
+        except (KeyError, AttributeError):
+            logger.warning('no valid sensor height in metadata - using default %d' % sensor_size[1])
+
+        self.start = np.minimum(np.maximum(self.start, 0),image.data.shape[2])
+
+        if self.end > 0:
+            self.end = np.minimum(self.end, image.data.shape[2])
+        if self.end <= 0:
+            self.end = image.data.shape[2]
+
+        if self.start >= self.end:
+            logger.warning('Messed up boundaries, defaulting to beginning and end of stack.')
+            self.start = 0
+            self.end = image.data.shape[2]
+
+        if not ((image.mdh['Camera.ROIWidth'] == sensor_size[0]) and (image.mdh['Camera.ROIHeight'] == sensor_size[1])):
+            logger.warning('Generating a map from data with ROI set. Use with EXTREME caution.\nMaps should be calculated from the whole chip.')
+
+        logger.info('Calculating mean and variance...')
+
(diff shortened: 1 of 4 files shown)


PYME/recipes/processing.py
```diff
@@ -1959,3 +1959,103 @@ def execute(self, namespace):
...
+    variance_threshold = Float(90000)  # again this is currently picked fairly arbitrarily
+    start = Int(0)
+    end = Int(0)
+    uniform = Bool(False)  # simulate a uniform map
+
+    def execute(self, namespace):
+        from PYME.IO.MetaDataHandler import NestedClassMDHandler
+        from PYME.Analysis import gen_sCMOS_maps
+
+        logging.basicConfig() # without it got 'No handlers could be found for logger...'
```
David-Baddeley, line 1978: not safe here - will override any previous logging config. This could be a bug if `dh5view` doesn't setup logging.

PYME/recipes/processing.py
```diff
@@ -1959,3 +1959,103 @@ def execute(self, namespace):
...
+    input = Input('input')
+    output_variance = Output('variance')
+    output_dark = Output('dark')
+    dark_threshold = Float(10000)  # this really should depend on the gain mode (12bit vs 16 bit etc)
+    variance_threshold = Float(90000)  # again this is currently picked fairly arbitrarily
+    start = Int(0)
+    end = Int(0)
+    uniform = Bool(False)  # simulate a uniform map
+
+    def execute(self, namespace):
```
David-Baddeley, line 1974: In general, recipe modules should be relatively light weight (i.e. the bulk of the logic should ideally reside elsewhere). As this is already in `gen_scmos_maps`, copying it here seems a bit odd.

PYME/recipes/processing.py
```diff
@@ -1959,3 +1959,103 @@ def execute(self, namespace):
...
+        from PYME.IO.MetaDataHandler import NestedClassMDHandler
+        from PYME.Analysis import gen_sCMOS_maps
+
+        logging.basicConfig() # without it got 'No handlers could be found for logger...'
+
+        default_sensor_size = (2048,2048) # we currently assume this is correct but could be chosen based
```
David-Baddeley, line 1980: If we are revisiting this, we should update this to use metadata entries (`"Camera.SensorWidth"`, `"Camera.SensorHeight"`)

David-Baddeley: Oops, see that this is done below, ordering is a bit iffy.

PYME/recipes/processing.py
```diff
@@ -1959,3 +1959,103 @@ def execute(self, namespace):
...
+        else:
+            logger.warning('Simulating uniform maps - use with care')
+            n_bad = 0
+
+        # if the uniform flag is set, then m and ve are passed as None
+        # which makes sure that just the uniform defaults from meta data are used
+        mfull, vefull, base_mdh = gen_sCMOS_maps.insert_into_full_map(m, ve, image.mdh, sensor_size=sensor_size)
+
+        common_md = NestedClassMDHandler()
+        common_md.setEntry('Analysis.name', 'mean-variance')
```
David-Baddeley, line 2035: Need to fix these metadata entries to be vaguely sensible (they should not be in "Analysis" for a start)

PYME/recipes/processing.py
David-Baddeley: Actually I see it's copied from the `main()` function in gen_scmos_maps.py, it makes sense in the main function of a standalone program, but shouldn't be copied over to a library.

David-Baddeley:
Unfortunately this isn't as easy as it might seem as `gen_sCMOS_maps.py` really needs some attention (it was a contribution from Christian which I merged in a hurry a long while back when trying to get the Exeter and main forks consistent. I'd always meant to revisit it but never got around to it.)

Potential issues:
- units need to be double checked (I think fitting might expect ADUs for variance, rather than electrons). This is now a largish problem as I think Andrew and I have been through and propagated various other units and documentation changes based on what was in gen_sCMOS_maps, rather than what the fitting actually wanted).
- some more general code style stuff.

David-Baddeley:
Just double checked units. The ORCA is currently writing it's metadata wrong, everything else, including the camera map generation seems OK. There are still some documentation issues (e.g. metadata docs).

David-Baddeley:
see #155

David-Baddeley:
@Zach - I'm doing a bit of tidying to gen_sCMOS_maps.py so hold fire on any modifications to this for now.

PYME/recipes/processing.py
```diff
@@ -1959,3 +1959,103 @@ def execute(self, namespace):
             pass
             
         namespace[self.output_name] = res
+
+@register_module('DarkAndVarianceMap')
+class DarkAndVarianceMap(ModuleBase):
```
David-Baddeley, line 1964: Recipe module should call the refactored code in gen_sCMOS_maps (otherwise we have duplication). 

Do we need a recipe module? To what extent is this something we need to automate? For the standard full frame maps you pretty much need to take them once during the lifetime of a camera. This might change with maps for very fast frame rates on very small ROIs so I'm open to the idea of a recipe module.

PYME/recipes/processing.py
```diff
@@ -1959,3 +1959,103 @@ def execute(self, namespace):
...
+    input = Input('input')
+    output_variance = Output('variance')
+    output_dark = Output('dark')
+    dark_threshold = Float(10000)  # this really should depend on the gain mode (12bit vs 16 bit etc)
+    variance_threshold = Float(90000)  # again this is currently picked fairly arbitrarily
+    start = Int(0)
+    end = Int(0)
+    uniform = Bool(False)  # simulate a uniform map
```
David-Baddeley, line 1972: If we do go with a recipe module, it should definitely not support uniform maps (the use case for these is very limited - namely testing equivalence of the map and non-map code, and I think it's reasonable that the user should have to jump through hoops to generate them).

PR #152, opened 2020-02-14
small syntax + import errors found when building conda package

barentine:


PYME/Analysis/BleachProfile/fitRecover.py
```diff
@@ -24,10 +24,9 @@
 from . import kinModels
 
 from PYME.Analysis._fithelpers import *
-
+from pylab import *
```
David-Baddeley: We can't have from `pylab import *` at the top level of any module as it screws with tests and documentation generation, and in general we want to move away from using `from pylab import *` anywhere. To fix this, we need to change `FitTrace()` so that it has `import matplotlib.pyplot as plt`, change all the plotting functions to `plt.xxx` and change anything like `array()`, `arange()` etc ... to use the `np.array()` etc equivalents.

PYME/misc/shellutils.py
```diff
@@ -456,7 +456,7 @@ def savitzky_golay(y, window_size, order, deriv=0):
     try:
         window_size = np.abs(np.int(window_size))
         order = np.abs(np.int(order))
-    except ValueError, msg:
+    except ValueError:
```
David-Baddeley, line 461: This is really odd - why are we catching a `ValueError` and then raising a `ValueError` ??

PYME/Acquire/Hardware/AndorNeo/AndorNeo.py
```diff
@@ -590,7 +590,8 @@ def GetFPS(self):
     def __getattr__(self, name):
         if name in self.noiseProps.keys():
             return self.noiseProps[name]
-        else:  raise AttributeError, name  # <<< DON'T FORGET THIS LINE !!
```
David-Baddeley: The comment here is pretty crucial to carry over for anyone modifying the class

PYME/misc/shellutils.py
barentine: Looks like to raise it with a custom message

PYME/Acquire/Hardware/AndorNeo/AndorNeo.py
barentine: I wasn't really worried someone would forget to raise an `AttributeError` inside of a `__getattr__` method. Why is there even a `__getattr__` method for the camera just to access  items of the noise_properties dict? Is it crucial to call `cam.ElectronsPerCount` instead of `cam.noise_properties['ElectronsPerCount']`? Either seems fine, it would just be nice to stick with one way of doing it.

PYME/Analysis/BleachProfile/fitRecover.py
barentine: pylab is now removed from `fitRecover`. There's only one function so I just stuffed the imports into it since you specificially mentioned importing in `FitTrace`. Worth pointing out that this code is undocumented and magic-number ridden.

PYME/Acquire/Hardware/AndorNeo/AndorNeo.py
David-Baddeley: I've done it a few times, and in this case it generates really hard to diagnose errors (hence the warning).

Ultimately the whole attribute access stuff on cameras should be removed in favour of the `noise_properties` property, modulo a check to see if it's ever actually used in the current code.

David-Baddeley: Actually the absence of `__getattr__()` in the much better maintained AndorZyla.py is good enough for me to go ahead with removal (there's no chance it'll be required in the Neo if it's not present in the Zyla as they share all UI code).

PYME/Analysis/BleachProfile/fitRecover.py
```diff
@@ -24,37 +24,37 @@
 from . import kinModels
 
 from PYME.Analysis._fithelpers import *
-
+import numpy as np
+import matplotlib.pyplot as plt
```
David-Baddeley: I still prefer deferring the pyplot import to the actual function call, as it makes things much more robust if and when you need to specify a matplotlib backend. Matplotlib chooses a backend the first time you import pyplot if you haven't already specified one, meaning that you have to be super careful about the order you import modules if you import pyplot (or pylab) at the module level.

Deferring the pyplot import until function execution time also means that you can do a lot more on a headless system (for example, a module level pyplot import can prevent test coverage from working on a headless system).

I know this is not currently consistent throughout the code, but I'm working towards it.

PYME/Analysis/BleachProfile/fitRecover.py
David-Baddeley: The undocumented and magic number ridden critique is the case for much of the stuff under BleachProfile, and some of this should really be factored out or removed. In this case most of the magic numbers match up with a specific PYMEAcquire protocol (would need to check which one). If writing from scratch, this would probably live in an IPython notebook, rather than it's own file.

For reference, it's an attempt to measure k_dark->on, based on a protocol which turns the laser on for a brief pulse (enough to push the majority of stuff into the dark state) and then lets the sample recover in the dark for varying amounts of time to assess how much has recovered. It falls into the category of stuff which:
- probably doesn't belong in PYME core
- probably worth hanging onto / revisiting at some point in the future (i.e. don't want to delete outright)
- not sufficiently important / pressing to spend much time on at present

PYME/Analysis/BleachProfile/fitRecover.py
barentine: Oops! Sorry David - I knew you'd prefer that for matplotlib and actually just moved everything to the function since there's only one function in that file, but forgot to push it.

PYME/Acquire/Hardware/AndorNeo/AndorNeo.py
barentine: Cool, thank you for checking the Zyla for that. I'll go ahead and remove the __getattr__ method then.

PR #158, opened 2020-02-21
Make visualizing LUT optional by layer

zacsimile:
Sometimes I want to turn off the LUT for a constant-color mesh, but not for the points it's approximating. I thought about adding this to the GUI, e.g add `Item('show_lut', label='Show LUT'),` on line 292 in PYME.LMVis.layers.mesh, but I think it's fine to leave as a shell option.


David-Baddeley:
As it stands, this will leave blank gaps between LUTs if the `show_lut` property is false. Should we potentially change line 76 (checking for visible layers) instead, so it checks for both `visible` and `show_lut`? As it stands, it looks like there's a typo in `visible` on this line.

zacsimile:
I agree. Changed.

PR #159, opened 2020-02-21
roi-switching speed improvements

barentine:
Effectively just removing as many file-exists calls to the cluster as possible since they're expensive if the file doesn't exist or is otherwise not in the DataServer cache. We were basically guaranteed to make two non-cached clusterIO.exists calls on every StartSpooling due to not using a flatfield map and incrementing the series name.

To this end:
- add argument to spool frame StartSpooling to not bother checking if the new file exists. This is now the default when adding a batch of ROIs from file/tileviewer. 
- made `CameraMapMixin` cache the camera maps it finds (and doesn't find).

### DIFF
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -53,6 +53,17 @@ class CameraMapMixin(object):
         self._camera_map_fields = map_fields
         self._camera_map_cache = {}
 
+    def prefill_camera_map_cache(self, integration_time):
+        """
+        Search the cluster / local camera map directories and pre-fill the camera map cache for a given integration time
+
+        Parameters
+        ----------
+        integration_time: float
+            integration time in milliseconds
+        """
+        self.fill_camera_map_metadata({'Camera.IntegrationTime': integration_time})
+
     def fill_camera_map_metadata(self, mdh):
         """
         Store camera map locations in the input metadata. The results (including camera-map does not exist) are cached.
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -66,10 +66,40 @@ def check_mapexists(mdh, type='dark'):
 
 
 class CameraMapMixin(object):
+    def __init__(self, map_fields=('dark', 'variance', 'flatfield'),
+                 cache_prefill=None):
+        self.map_fields = map_fields
+        if cache_prefill:
+            self._cache = cache_prefill
+        else:
+            self._cache = {}
+
     def fill_camera_map_metadata(self, mdh):
-        check_mapexists(mdh, type='dark')
-        check_mapexists(mdh, type='variance')
-        check_mapexists(mdh, type='flatfield')
+        from PYME.Analysis.gen_sCMOS_maps import map_filename, STEM_TO_MDH
+        import os
+        from PYME.IO import clusterIO
+        from PYME.IO.FileUtils import nameUtils
+        for map in self.map_fields:
+            map_fn = map_filename(mdh, map)
+            if map_fn in self._cache.keys():
+                map_location = self._cache[map_fn]
+                if map_location is not None:  # don't write field if we're just faking it
+                    mdh[STEM_TO_MDH[map]] = map_location
+            else:
+                local_path = os.path.join(nameUtils.getCalibrationDir(mdh['Camera.SerialNumber']), map_fn)
+                cluster_path = 'CALIBRATION/%s/%s' % (mdh['Camera.SerialNumber'], map_fn)
+                if clusterIO.exists(cluster_path):
+                    c_path = 'PYME-CLUSTER://%s/%s' % (clusterIO.local_serverfilter, cluster_path)
+                    self._cache[map_fn] = c_path
+                    mdh[STEM_TO_MDH[map]] = c_path
+                elif os.path.exists(local_path):
+                    self._cache[map_fn] = local_path
+                    mdh[STEM_TO_MDH[map]] = local_path
+                else:  # map doesn't exist, negative cache it so we don't keep checking
+                    self._cache[map_fn] = None
+
+    def reset_cache(self):
+        self._cache = {}
 
 
 class Camera(object):
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -62,7 +62,8 @@ class CameraMapMixin(object):
         integration_time: float
             integration time in milliseconds
         """
-        self.fill_camera_map_metadata({'Camera.IntegrationTime': integration_time})
+        self.fill_camera_map_metadata({'Camera.IntegrationTime': integration_time,
+                                       'Camera.SerialNumber': self.GetSerialNumber()})
 
     def fill_camera_map_metadata(self, mdh):
         """
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -33,56 +33,44 @@ import numpy as np
 import logging
 logger = logging.getLogger(__name__)
 
-def check_mapexists(mdh, type='dark'):
-    import os
-    from PYME.IO import clusterIO
-    from PYME.IO.FileUtils import nameUtils
-    from PYME.Analysis.gen_sCMOS_maps import map_filename
-
-    if type == 'dark':
-        id = 'Camera.DarkMapID'
-    elif type == 'variance':
-        id = 'Camera.VarianceMapID'
-    elif type == 'flatfield':
-        id = 'Camera.FlatfieldMapID'
-    else:
-        raise RuntimeError('unknown map type %s' % type)
-
-    mapfn = map_filename(mdh, type)
-
-    # find and record calibration paths
-    local_path = os.path.join(nameUtils.getCalibrationDir(mdh['Camera.SerialNumber']), mapfn)
-    cluster_path = 'CALIBRATION/%s/%s' % (mdh['Camera.SerialNumber'], mapfn)
-
-    if clusterIO.exists(cluster_path):
-        c_path = 'PYME-CLUSTER://%s/%s' % (clusterIO.local_serverfilter, cluster_path)
-        mdh[id] = c_path
-        return c_path
-    elif os.path.exists(local_path):
-        mdh[id] = local_path
-        return local_path
-    else:
-        return None
-
 
 class CameraMapMixin(object):
-    def __init__(self, map_fields=('dark', 'variance', 'flatfield'),
-                 cache_prefill=None):
-        self.map_fields = map_fields
-        if cache_prefill:
-            self._cache = cache_prefill
-        else:
-            self._cache = {}
+    """
+    Handles metadata propagation. Should be used with multiple inheritance with any camera class using camera maps. Maps
+    are cached as being present or not, so the `clear_cache` method should be called if camera maps are installed during
+    use (or restart PYMEAcquire). Preference is given to camera map locations on the cluster rather than local.
+    """
+    def __init__(self, map_fields=('dark', 'variance', 'flatfield')):
+        """
+        Parameters
+        ----------
+        map_fields: tuple of str
+            Camera maps in use by the inheriting camera. Due to way this class is mixed into camera classes, you will
+            likely never change this parameter of the init function, but if you are, e.g. not using a flatfield map with
+            your camera, you can change the `_camera_map_fields` attribute after camera initialization to only include
+            the maps you are using
+        """
+        self._camera_map_fields = map_fields
+        self._camera_map_cache = {}
 
     def fill_camera_map_metadata(self, mdh):
+        """
+        Store camera map locations in the input metadata. The results (including camera-map does not exist) are cached.
+
+        Parameters
+        ----------
+        mdh: PYME.IO.MetaDataHandler
+            dict-like metadata
+        """
         from PYME.Analysis.gen_sCMOS_maps import map_filename, STEM_TO_MDH
-        import os
-        from PYME.IO import clusterIO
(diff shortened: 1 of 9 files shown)


PYME/Acquire/Spooler.py
```diff
@@ -204,23 +205,18 @@ def doStartLog(self):
...
-        mdt = MetaDataHandler.NestedClassMDHandler()
-          
-        mdt.setEntry('StartTime', self.tStart)
 
-        #loop over all providers of metadata
+        # loop over all providers of start-metadata
         for mdgen in MetaDataHandler.provideStartMetadata:
-            mdgen(mdt)
-            
-        self.md.copyEntriesFrom(mdt)
```
David-Baddeley, line 216: Generating a blank metadata handler, populating, and then copying to series metadata handler (rather than populating the series handler directly) is a critically important performance optimization, resulting in a single write rather than multiple writes to storage (at least for HDF and Queue targets - the Cluster spooler caches internally and is probably immune to the performance issues)

PYME/Acquire/protocol.py
```diff
@@ -214,7 +214,7 @@ def Init(self, spooler):
 
     def OnFrame(self, frameNum):
         if frameNum > self.startFrame:
-            fn = int(floor((frameNum - self.startFrame)/self.dwellTime) % len(self.zPoss))
```
David-Baddeley, line 217: I'm pretty sure this should be floor - what is your reasoning to the contrary?

David-Baddeley: Protocol logic is that a protocol task executes strictly **after** the designated frame

PYME/Analysis/gen_sCMOS_maps.py
```diff
@@ -42,12 +49,27 @@ def _meanvards(dataSource, start=0, end=-1):
 
     return (m,v)
 
-def map_filename(mdh, type):
-    if type != 'flatfield':
+def map_filename(mdh, stem):
```
David-Baddeley, line 52: This seems like an un-necessary and confusing change in signature. I'm pretty sure that all uses of this code haven't been updated (might not strictly matter, as I suspect the arguments are passed as positional rather than keyword args, but can't be certain).

David-Baddeley:
The map->exists queries should, in theory, already be cached in clusterIO (i.e. you should incur overhead on the first call, but not on subsequent calls). If this is not the case we should look at why. In general it would be good to see if we could speed up the `.exists()` calls (although this might not be a super quick solution)

I'm more than a little hesitant about adding an additional layer of caching in the camera class (this seems like the wrong place to be worrying about caching). I'm even more uncertain about caching a failure to find something (especially in a way which would not seem to expire).

David-Baddeley:
I'm also not super comfortable with the idea of omitting the existence check prior to spooling, as the cluster integrity relies on not writing to the same filename twice. As it is we are flying by the seat of our pants a bit here already (due to the absence of locking, the result of our current `.exists()` call may be out of date by the time we receive the reply), and completely abandoning all checking feels rather reckless. It's the kind of thing that is likely to come back and bite us in the future (e.g. resulting in days of broken data).

I can see how it is attractive as a short term stop-gap, but ultimately we really need to fix the performance of `.exists()`.

David-Baddeley:
Have you actually checked whether the `.exists()` calls on the maps are a problem? Just double checked the code and they are definitely cached in clusterIO (maps which exist are cached directly, ones which don't exist drop through to the directory listing cache, although this currently expires fairly quickly). 

We could potentially look at increasing the cache sizes, although I would expect them not to be limiting in the spooler case.

David-Baddeley:
I can think of a few simple performance improvements related to `.exists()` (specifically checking to see if the parent directory exists within the cache if we fail the file location caches, before querying)

PYME/Acquire/protocol.py
barentine: Correct, it should be a floor and it *is* a floor, just don't need the floor call. Int truncates, i.e. rounds towards zero. Floor rounds towards negative inf, but since the line above ensures we only get here if we're positive, int is both our int-cast and our floor.

PYME/Analysis/gen_sCMOS_maps.py
barentine: This doesn't break anything - the map mixin was the only keyword use and this PR changes that anyway. Let me know if you want me to back this out, but I think its good practice not to override built-ins.

PYME/Acquire/Spooler.py
barentine: How much do we care? Critically important for performance optimization of the 95% deprecated localization analysis enough not to speed up the fast version?
 
image

barentine:
Yes, clusterIO caches hits and that makes it only a problem for files that don't exist like my flatfield map, as I mentioned

image

Speeding up `exists()` would be great, since I agree that it's nice to check when spooling, even though that adds about 0.25 s per series at the moment and we expect all of those to be cache-misses. 

For the camera maps, I don't think it's crazy to detach them a bit from clusterIO - it seems a bit silly to check on each series whether the camera maps have changed, though if someone adds/changes maps all they have to do for them to be discovered is call the clear cache method. That said, we could get a substantial speed-up regardless of map mixin caching if we simply toggled which maps we bothered to ask for (you'll notice this can now be done with the `_camera_maps_fields` attribute of the map mixin).

PYME/Acquire/Spooler.py
David-Baddeley: Not just queue spooling - spooling directly to disk (which is not deprecated) is also affected. 

Even if it were only the queue spooling, I'm not prepared to break that this badly at this point.  The alternative being 95% ready does not equate to the old-style being deprecated yet - there are a lot of people using it and I'm not prepared to arbitrarily break it for them. IE we care a lot

I suspect that any difference that you are seeing in metadata collection between the collect and copy vs direct collect is likely to be a profiling artifact, and in any case you need to show the comparative cost (which is unlikely to change). Even given profiling overhead, those figures seem a little on the high side though, so might be worth looking into further, but the `mdgen` cost should be the same for both new and old code.

PYME/Acquire/protocol.py
David-Baddeley: integer truncation / rounding behaviour is somewhat platform specific, and the floor makes this explicit and the code much more readable.

David-Baddeley:
Both of those `exists()==False` cases could theoretically evaluate to cache hits on the directory cache (rather than location cache) in a tweaked `exists()`. In practice this might not be as simple as I'd like due to the current directory cache lifetime of 1s (which is shorter than the series duration).

PYME/Analysis/gen_sCMOS_maps.py
David-Baddeley: I find `type` a much more intuitive name than `stem` in this context. If we are really worried about clobbering the builtin `type` (for a small function like this it's unlikely to be an issue) we could potentially use something like `map_type`.

PYME/Acquire/protocol.py
barentine: We are not rounding. Name an OS that "truncates" a positive float to the next highest int?

Do we not call this on every frame? ProtocolFocus events are fired after this line and I can run my camera at kHz. I want this to be as fast as we can make it.

David-Baddeley:
Thinking about the pre-filling a bit more, I could see an argument for pre-filling the flatfield map as this doesn't change with integration time. Pre-filling/caching the dark and variance maps is a lot riskier as there are scenarios where you'd want to change the integration time between series. There a quite a few use cases where this might be appropriate, e.g. if you had a sample labelled with both a STORM and a PAINT label - the simplest way of doing this would likely to be to do the STORM acquisition first, change the integration time, and then do the PAINT imaging. Taking a widefield stack (potentially in different colour channels) for contextual information prior to each STORM image is another.

PYME/Acquire/Spooler.py
barentine: The mdgen cost is high because the current repo head features 3x clusterIO exist calls and one of my maps isn't cached. 

Do you have time for a quick skype @David-Baddeley?

PYME/Acquire/Hardware/Camera.py
```diff
@@ -33,43 +33,73 @@
 import logging
 logger = logging.getLogger(__name__)
 
-def check_mapexists(mdh, type='dark'):
```
David-Baddeley, line 36: I like having this as a stand-alone function, rather than part of a camera mixin, as it would let us use it, e.g. from the LMAnalysis plugin for dh5view if map locations haven't been recorded at acquisition time, or if their locations have changed as we're analysing on a different computer.

Ultimately it probably belongs somewhere else (a refactored and renamed gen_scmod_maps.py?), but let's not worry about that for now.

David-Baddeley:
I'm going to have a crack at adapting the current code to do pre-filling for the flat-field map only. This should fix your performance issues related to the maps, whilst leaving things safe.

barentine:
The cache doesn't cache a single dark, or a single variance map, it caches by name following the convention in gen_sCMOS_maps, i.e. dark and var maps will be changed if you change integration time in this PR...

barentine:
Do you mind hopping on skype, @David-Baddeley ?

PYME/Acquire/Spooler.py
David-Baddeley: Using ipython to time things without profiler overhead ...

```python
>>> from PYME.IO import MetaDataHandler
>>> mdh = MetaDataHandler.load_json('/Users/david/srvtest/test1/david/2019_5_10/10_5_series_B.pcs/metadata.json')
>>> md2 = MetaDataHandler.NestedClassMDHandler()
>>> %timeit md2.copyEntriesFrom(mdh)
1000 loops, best of 3: 1.37 ms per loop
>>> %timeit md3 = MetaDataHandler.NestedClassMDHandler(mdh)
1000 loops, best of 3: 1.19 ms per loop
```

the metadata stuff is fast enough (~1ms) that we shouldn't care

David-Baddeley:
Sorry, in the open plan office at present, could do first thing tomorrow if that suits.

PYME/Acquire/protocol.py
David-Baddeley: truncation/rounding/cast behaviour depends on the processor, flags, and compiler, not platform. In any case int(xxx) truncation is not guaranteed, not clear, and with `floor()` taking around 1us (i.e. 3 orders of magnitude faster than anything we need to care about) this is not a case where we should be optimizing.

David-Baddeley: I suspect the line showed up slightly red in the profile, but this is likely a case of being caught out by profiler overhead - there are a bunch of function calls on this line which all incur significant profiling overhead.

barentine:
@David-Baddeley , for the camera map stuff I still think it's silly to query the cluster to ask if a camera map exists on each series. We can easily (as in this PR) cache maps of various integration times so we always have the right map. If you update the camera maps on the cluster, they should have the same name anyway, so even if you change midway (weird) you're fine. 

If you like I can add a timeout on the cache, though your flatfield caching doesn't do this either because I think you'll agree it is of pointless. Maybe a timeout on the negative cache in case you get unlucky the first go? If it's the case you can get unlucky then you currently stand that risk every time. In practice the clusterIO.exists works, it just takes time. I don't think we're being irresponsible for caching all map names/existance when it comes up, and I think the camera map mixin is the logical place to cache that information. What do you think?

Also your alternative flatfield-only caching is broken. `update_flatfield_map` passes an empty mdh to check mapexists.  Also, `check_mapexists` could have a more descriptive name since it's really filling-in the map metadata.
```
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\ui\HDFSpoolFrame.py", line 460, in OnBStartSpoolButton
    cluster_h5=self.cbClusterh5.GetValue())
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\SpoolController.py", line 263, in StartSpooling
    self.spooler.StartSpool()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\HTTPSpooler.py", line 224, in StartSpool
    sp.Spooler.StartSpool(self)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Spooler.py", line 129, in StartSpool
    self.doStartLog()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Spooler.py", line 216, in doStartLog
    mdgen(mdt)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\Camera.py", line 1048, in GenStartMetadata
    self.camera_class.GenStartMetadata(self, mdh)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\HamamatsuDCAM\HamamatsuORCA.py", line 417, in GenStartMetadata
    self.fill_camera_map_metadata(mdh)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\Camera.py", line 77, in fill_camera_map_metadata
    self.update_flatfield_map()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\Camera.py", line 72, in update_flatfield_map
    check_mapexists(self._prefilled, type='flatfield')
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\Camera.py", line 54, in check_mapexists
    local_path = os.path.join(nameUtils.getCalibrationDir(mdh['Camera.SerialNumber']), mapfn)
  File "c:\users\bergamot\code\python-microscopy\PYME\IO\MetaDataHandler.py", line 208, in __getitem__
    raise KeyError('Key %s not defined' % name)
KeyError: 'Key Camera.SerialNumber not defined'
```

barentine:
Should be OK to close now that camera map caching is in the head. Only thing to maybe hold over is whether we add an option to disable checking if series exists when batch-queuing (alternatively checking all at once whether the batch exists or so). Will address this later if need-be

Issue #165, opened 2020-03-02
Color-assignment error in Python 3


lukasfue:
I tried to assign color channels for salvaged fluorescence in the Colour tab in VisGUI and received the following traceback.

```
Traceback (most recent call last):                                                                                        

File "c:\users\laf62\python-microscopy\PYME\LMVis\colourPanel.py", line 298, in OnSpecAdd                                 self.pipeline.colour_mapper.species_ratios[key] = float(val)                                                          

File "C:\Users\laf62\.conda\envs\pyme\lib\site-packages\traits\trait_handlers.py", line 3318, in __setitem__              raise excp                                                                                                            

File "C:\Users\laf62\.conda\envs\pyme\lib\site-packages\traits\trait_handlers.py", line 3314, in __setitem__              key = validate(object, self.name, key)                                                                                

File "C:\Users\laf62\.conda\envs\pyme\lib\site-packages\traits\trait_handlers.py", line 1190, in validate                 self.error(object, name, value)                                                                                       

File "C:\Users\laf62\.conda\envs\pyme\lib\site-packages\traits\trait_handlers.py", line 236, in error                     object, name, self.full_info(object, name, value), value                                                            

traits.trait_errors.TraitError: Each key of the 'species_ratios' trait of a ProcessColour instance must be a value of class 'str', but a value of b'A' <class 'bytes'> was specified. 
```

David-Baddeley:
fixed in #166

PR #166, opened 2020-03-03
py3 bytes compat in colourPanel

barentine:
Fix to issue #165.

### DIFF
--- a/PYME/LMVis/colourPanel.py
+++ b/PYME/LMVis/colourPanel.py
@@ -290,12 +290,12 @@ class colourPanel(wx.Panel):
         if ret == wx.ID_OK:
             val = float(dlg.tVal.GetValue())
 
-            key = dlg.tKey.GetValue()
+            key = str(dlg.tKey.GetValue())  # unicode builds of wx 3 return a unicode str, which breaks wx.ListCtrl.InsertStringItem
 
             if key == "":
                 return
 
-            self.pipeline.colour_mapper.species_ratios[key] = float(val)
+            self.pipeline.colour_mapper.species_ratios[key] = val
 
             ind = self.lFluorSpecies.InsertStringItem(UI_MAXSIZE, key)
             self.lFluorSpecies.SetStringItem(ind,1, '%3.2f' % val)
--- a/PYME/LMVis/colourPanel.py
+++ b/PYME/LMVis/colourPanel.py
@@ -290,7 +290,7 @@ class colourPanel(wx.Panel):
         if ret == wx.ID_OK:
             val = float(dlg.tVal.GetValue())
 
-            key = dlg.tKey.GetValue().encode()
+            key = dlg.tKey.GetValue()
 
             if key == "":
                 return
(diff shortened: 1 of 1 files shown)


David-Baddeley:
this `.encode()` is strictly necessary for py2 (wx returns unicode strings on some platforms). Use `six.ensure_str()`.

David-Baddeley:
note that using `six.ensure_str()` might mean that we have to update our conda package for six (seems to be missing from earlier packages)

barentine:
Not interested in six.

Which version of wx on which platform are you concerned with unicode return from wx breaking a wx str set method?

David-Baddeley:
I can't recall which wx version  and platform, but most of the recent wx versions return `unicode` (at least on win and OSX). 

After having spent a fair bit of time hunting down bugs introduced into the cluster stuff through dropping `.encode()` calls, I'm getting a bit tired of these bullshit PRs. "Not interested in six" doesn't cut it. While it would be nice to drop py2 support, there are still a bunch of bugs on Py3 in fairly fundamental bits (e.g. rendering 2 colour images in VisGUI) and until such time as these are gone, we need to play nice.

Due to the fact that `ensure_str()` seems to be a recent addition to six, however, it might be worth considering an alternative rather than having to deal with conda packaging woes.

barentine:
I'm sorry for being a punk, David. I just pushed a fix which I think we can agree on.

The build of wxpython in pyme-depends in your conda channel is a unicode build (see [here](https://wiki.wxpython.org/UnicodeBuild)). And while the docs would suggest wx functions would then prefer unicode and convert other str to unicode automatically, you are correct that this isn't always the case.

```
>>> from PYME.LMVis.colourPanel import specAddDialog
>>> dlg = specAddDialog(colp)
>>> ret = dlg.ShowModal()
>>> val = float(dlg.tVal.GetValue())
>>> key = dlg.tKey.GetValue()
>>> key
u'test'
>>> pipeline.colour_mapper.species_ratios[key] = float(val)
  File "<input>", line 1, in <module>
  File "/Users/Andrew/miniconda3/envs/py27dev/lib/python2.7/site-packages/traits/trait_handlers.py", line 3318, in __setitem__
    raise excp
TraitError: Each key of the 'species_ratios' trait of a ProcessColour instance must be a value of type 'str', but a value of u'test' <type 'unicode'> was specified.
>>> import wx
>>> wx.version()
'3.0.2.0 osx-cocoa (classic)'
>>> 'unicode' in wx.PlatformInfo
True
```
Which can be fixed with a str cast rather than the bytes encode in the current head
```
>>> from PYME.LMVis.colourPanel import specAddDialog
>>> dlg = specAddDialog(colp)
>>> ret = dlg.ShowModal()
>>> val = float(dlg.tVal.GetValue())
>>> key = str(dlg.tKey.GetValue())
>>> pipeline.colour_mapper.species_ratios[key] = val
>>> import wx
>>> wx.version()
'3.0.2.0 osx-cocoa (classic)'
>>> 'unicode' in wx.PlatformInfo
True
```

For python 3 of course we're fine with unicode build wx4, but if we're interested we might be able to change the py2 pyme-depends version of wx(still 3? or do we want 4) to an ansi build.

David-Baddeley:
yep - `str()` cast seems to be the answer for most (but not all) `.encodes()`  - i.e. reliably converting unicode to str in a portable way.

PR #167, opened 2020-03-03
Add handling for PiezoOnTarget and PiezoOffsetUpdate events

barentine:
Two main functions added:
1 - flagging piezo movement (frames between `ProtocolFocus` and `PiezoOnTarget` events)
2 - correcting focus positions from the target position in `ProtocolFocus` events to the position logged with `PiezoOnTarget` events, including accounting for OffsetPiezo offsets (on-target events use the base piezo positions while focus events use offset-included positions)

Both have associated recipe modules. The piezo-movement flagging can be accessed through a VisGUI plug / added to the pipeline that way, while the focus events are handling during pipeline.py/_process_events.

Several DSView modules which use focus positions from events have been updated to use the correct focus positions


barentine:
OK @David-Baddeley , now updated based on our conversation yesterday

PYME/Analysis/piecewiseMapping.py
```diff
@@ -21,12 +21,21 @@
...
-from numpy import *
 import numpy as np
-import sys
-import six
 
-def timeToFrames(t, events, mdh):
+def times_to_frames(t, events, mdh):
+    """
+    Use events and metadata to convert time-stamps to frame numbers
+    :param t: ndarray
```
David-Baddeley: If you are adding docstrings, they should be numpy formatted!

PYME/LMVis/pipeline.py
```diff
@@ -114,17 +113,25 @@ def _add_eventvars_to_ds(ds, ev_mappings):
...
+        piezo_moving = ev_mappings.get('piezo_moving')
+        if piezo_moving:
+            ds.addColumn('piezo_moving', piezo_moving(ds['t']))
+
     xm = ev_mappings.get('xm', None)
     if xm:
+        # todo - comment on why the -0.01??
         scan_x = 1.e3 * xm(ds['t'] - .01)
         ds.addColumn('scanx', scan_x)
-        #ds.setMapping('x', 'x + scanx')
```
David-Baddeley, line 121: Should not be deleted (ditto below)

PYME/recipes/processing.py
```diff
@@ -257,7 +257,6 @@ def calc_flow_mp(self, data, chanNum):
         return flow_x, flow_y
         
     def execute(self, namespace):
-        import multiprocessing
```
David-Baddeley: This is not redundant! The code which uses it happens to be commented out for debugging, but it can't safely be removed.

PYME/recipes/processing.py
```diff
@@ -336,7 +335,6 @@ class WavefrontVelocity(ModuleBase):
         
     
     def execute(self, namespace):
-        from skimage.measure import profile_line
```
David-Baddeley: Same as previous comment - used in code which is currently commented out for debugging, but not safe to remove.

PYME/recipes/processing.py
```diff
@@ -1252,7 +1250,6 @@ def GetDec(self, dp, vshint):
         return self._decCache[decKey]
     
     def applyFilter(self, data, chanNum, frNum, im):
-        from PYME.Analysis import optic_flow
```
David-Baddeley: Ditto the above - not safe to remove.

barentine:
@David-Baddeley I got a little confused by a 'submit review' box and accidentally deleted my replies to your comments. Anyway, I've made the newly requested changes.

Issue #94, opened 2019-10-31
mpld3 fails to get d3.js on Linux, wx4, py3


David-Baddeley:
**[Original report](https://bitbucket.org/david_baddeley/python-microscopy/issue/94) by David Baddeley (Bitbucket: [david_baddeley](https://bitbucket.org/david_baddeley), ).**

----------------------------------------

From @{AndrewBarentine} 

I get the following message when calibrating an astigmatic PSF \(and no plots\)

‌

```
** Message: 11:58:03.849: console message:  @47: failed to load library https://mpld3.github.io/js/d3.v3.min.js
```

‌

David-Baddeley:
**Original comment by David Baddeley (Bitbucket: [david_baddeley](https://bitbucket.org/david_baddeley), ).**

----------------------------------------

I suspect this is a problem with the wx4 webview implementation, and specifically the priviledge / emulation level it is running on.

Strongly suspect that the quoted error message is coming from the webbrowser plugin itself \(i.e. not from anywhere in python\). Maybe it can’t access the internet? Do we know which browser is used? Is it Chromium? This could be an issue of a preferred browser not being available \(installed\) on the particular linux installation. Could also be some form of cross-origin protection.

Linux makes it tricky as you don’t necessarily have a default browser that is guaranteed to be present \(wx webview uses IE on windows, webkit on OSX, and these are generally guaranteed to be present\)

@{AndrewBarentine} - More info needed:

* what browsers do you have installed?
* what browser is webview trying to use \(not sure how you’d find this out, but should be in the wx docs\)
* can you have external links to anything \(e.g. images\)

Possible solutions:

* install a different webrowser
* distribute our own copy of d3 \(to avoid cross-origin issues\)

‌

David-Baddeley:
**Original comment by David Baddeley (Bitbucket: [david_baddeley](https://bitbucket.org/david_baddeley), ).**

----------------------------------------

Under the hood this looks like it’s an error with https, which is not supported \(or not properly supported\) by the version of webkit that wx is linked against \(suggestion is to install glib-networking, but this is installed\). This is compounded by the fact that github automatically promotes http requests to https.

This probably has to do with wx having been built against an old version of gtk. This may well be an anaconda packaging issue.

Potential workarounds: 

* serve d3 and mpld3 js from local files \(problematic as it would make the generated reports less portable\)
* chuck the js somewhere which doesn’t automatically promote http connections to https \(e.g. python-microscopy.org\)
[...]

PR #168, opened 2020-03-03
PSF extraction / astigmatism tweaks

barentine:
- make PSF tagging getIntCenter more robust to intensity variations in time/frame (e.g. if your mountant auto-fluoresces but bleaches quickly, you do a weird z-scan pattern, etc.. The current code will give you ~the brightest frame)
- skip webview plotting for linux/wx4. See issue #94

### DIFF
--- a/PYME/Analysis/PSFEst/extractImages.py
+++ b/PYME/Analysis/PSFEst/extractImages.py
@@ -114,15 +114,11 @@ def getIntCenter(im):
     im = im.squeeze()
     X, Y, Z = np.ogrid[0:im.shape[0], 0:im.shape[1], 0:im.shape[2]]
 
-    #from pylab import *
-    #imshow(im.max(2))
-
     X = X.astype('f') - X.mean()
     Y = Y.astype('f') - Y.mean()
     Z = Z.astype('f')
 
-    im2 = im - im.min()
-    im2 = im2 - 0.5*im2.max()
+    im2 = im - np.median(im, axis=(0, 1))
     im2 = im2*(im2 > 0)
 
     ims = im2.sum()
@@ -130,8 +126,6 @@ def getIntCenter(im):
     x = (im2*X).sum()/ims
     y = (im2*Y).sum()/ims
     z = (im2*Z).sum()/ims
-
-    #print x, y, z
 
     return x, y, z
 
--- a/PYME/DSView/modules/psfTools.py
+++ b/PYME/DSView/modules/psfTools.py
@@ -303,6 +303,7 @@ class PSFTools(HasTraits):
         from PYME.Analysis.PSFEst import extractImages
         import wx
         from PYME.Analysis.points.astigmatism import astigTools
+        import sys
 
         # query user for type of calibration
         # NB - GPU fit is not enabled here because it exits on number of iterations, which is not necessarily convergence for very bright beads!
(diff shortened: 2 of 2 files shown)


barentine:
Example motivating the `getIntCenter` change: frame-averaged intensity for a now-sorted/resampled stack which was taken with a triangle-wave z-step pattern and had a significant amount of bleaching background (should probably make a new bead sample).
image
The current head will give a z-center slice index of ~1.5, while the proposed change successfully pulls out the focus position of ~50

David-Baddeley:
In general thresholding at half max will give a MUCH more robust centroid
than the median. In a sparse image such as beads, the median is a
reasonable estimate for the background. Thresholding (and clipping) at the
background will include all the pixels where noise puts them above the
background (i.e. roughly half of all pixels). This will strongly bias the
centroid towards the centre of the ROI.

I expect the correct solution in this case is a new bead sample (or new
acquisition). If you are getting significant bleaching during a bead
acquisition the PSF is always going to be dodgy - minimal bleaching is
generally considered a pre-requisite for PSF measurements. You might want
to look at the acquisition parameters (laser power, integration time, z
step size). Invariably there are compromises with PSF measurements, and
these might require you to e.g. live with poorer SNR and average after the
fact. I think you might currently take smaller z steps than I would (I'd
use 50 nm as a reasonable compromise between bleaching and not having to
interpolate too much - in theory you could get away with 200 if you did the
interpolation right).

barentine:
I can show you the stack in question in our meeting this week. I don't think it's crazy to want to use it for astigmatism calibration. The bead SNR looks good throughout, and I don't think there is much cause for concern that the beads are bleaching so much that the fitted sigma values are ruined (and the extraction gives me a normalized 3D CRB of 1.5).

You make good point about biasing the centroid towards the... center, which I would hope in practice wouldn't be problematic, but similarly you hope that bleaching in practice won't be problematic. In the spirit of just making it easier to extract a PSF (for triage, etc.) regardless of stack: is there an alternative to median-threshold centroiding and half-max threshold centroiding which would be robust both to bleaching and to the stack not being centered on the focus?

David-Baddeley:
The extraction of sigma vs z curves is much more robust against bleaching than the extraction of full PSFs for PSF based fitting or deconvolution (what the program was originally written for). I suspect that whilst your stack might work for your case, it wouldn't for the original to applications (which I am unprepared to break). 

When averaging multiple beads, it's pretty important to get the centres of the individual beads accurate so that they can be aligned. Any bias here can be really problematic. Widefield images are generally hard for calculating axial centroids from (due to the fact that integrated intensity is equal at each slice), so you really need to only capture the central in focus section. Based on my experience it'd be hard to persuade me that there was anything better than a threshold at half max - at least in the general case. If you can dig it up, there is some theoretical analysis of this in Rainer Heintzmanns PhD thesis.

One solution to your specific case would be to apply some form of bleaching correction to the stack prior to PSF extraction. The confocal vs widefield option in the extraction already does this to some extent (if you select widefield it normalises intensities in each frame so that the sum is constant), but I believe this currently happens post alignment (i.e. after the `getIntCenter()` call).

David-Baddeley:
Changing to the median is not going to be merged - & I think logic on webview exclusion is not quite right - it forces wx > 3 on all platforms.

PR #169, opened 2020-03-03
Added series selection to BioformatsDataSource

zacsimile:
Wanted to load LIFs with multiple data series (collected with Leica SP5) in PYME. Modified BioformatsDataSource accordingly.

### DIFF
--- a/PYME/IO/DataSources/BioformatsDataSource.py
+++ b/PYME/IO/DataSources/BioformatsDataSource.py
@@ -25,7 +25,7 @@ try:
     import bioformats
     #javabridge.start_vm(class_path=bioformats.JARS, run_headless=True)
 except:
-    pass
+    raise ImportError('Please make sure python-javabridge (https://pythonhosted.org/javabridge/) and python-bioformats (https://pythonhosted.org/python-bioformats/) are working before you use this module.')
 
 numVMRefs = 0
 
@@ -60,7 +60,7 @@ from .BaseDataSource import BaseDataSource
 
 class DataSource(BaseDataSource):
     moduleName = 'BioformatsDataSource'
-    def __init__(self, filename, taskQueue=None, chanNum = 0):
+    def __init__(self, filename, taskQueue=None, chanNum = 0, series=None):
         self.filename = getFullExistingFilename(filename)#convert relative path to full path
         self.chanNum = chanNum
         
@@ -95,7 +95,14 @@ class DataSource(BaseDataSource):
         #tf = tifffile.TIFFfile(self.filename)
         ensure_VM()
         self.bff = bioformats.ImageReader(filename)
+        self._nSeries = None
+        if self.nSeries > 1:
+            import os
+            ftype = (os.path.basename(filename).split('.')[-1]).upper()
+            print('This {} file stores {} series of data. Use setSeries(series=<series number>) to switch series.'.format(ftype, self.nSeries+1))
         
+        self.setSeries(series)
+
         self.sizeX = self.bff.rdr.getSizeX()
         self.sizeY = self.bff.rdr.getSizeY()
         self.sizeZ = self.bff.rdr.getSizeZ()
@@ -109,7 +116,16 @@ class DataSource(BaseDataSource):
         
         self.additionalDims = 'TC'
                 
+    def setSeries(self, series):
+        if series is not None:
+            self.bff.rdr.setSeries(series)
 
+    
+    @property
+    def nSeries(self):
+        if self._nSeries is None:
+            self._nSeries = self.bff.rdr.getSeriesCount()
+        return self._nSeries
 
     def getSlice(self, ind):
         #self.im.seek(ind)
--- a/PYME/IO/DataSources/BioformatsDataSource.py
+++ b/PYME/IO/DataSources/BioformatsDataSource.py
@@ -65,14 +65,14 @@ from .BaseDataSource import BaseDataSource
 
 
 class BioformatsFile(bioformats.ImageReader):
-    def __init__(self, path=None, url=None, perform_init=True):
+    def __init__(self, *args, **kwargs):
         # self.path is always the file path or the url, whichever is specified
         # url overrides a file path 
         ensure_VM()
         self._md = None
         self._series_count = None
         self._series_names = None
-        super(BioformatsFile, self).__init__(path, url, perform_init)
+        super(BioformatsFile, self).__init__(self, *args, **kwargs)
 
     @property
     def series_count(self):
@@ -85,8 +85,7 @@ class BioformatsFile(bioformats.ImageReader):
         if self._md is None:
             self._md = []
             _io = self.rdr.getSeries()
-            idxs = np.arange(self.series_count)
-            for _i in idxs:
+            for _i in range(self.series_count):
                 self.rdr.setSeries(_i)
                 series_md = javabridge.jutil.jdictionary_to_string_dictionary(self.rdr.getSeriesMetadata())
                 self._md.append(series_md)
@@ -98,13 +97,14 @@ class BioformatsFile(bioformats.ImageReader):
         # return a list of series names in the file
         if self._series_names is None:
             self._series_names = []
-            idxs = np.arange(self.series_count)
-            try:
-                for _i in idxs:
+            
+            for _i in range(self.series_count):
+                try:
                     self._series_names.append(self.md[_i]['Image name'])
-            except(KeyError):
-                print('Image names not found, naming by integers.')
-                self._series_names = ['Image {}'.format(_i) for _i in idxs.astype(str)]
+                except(KeyError):
+                    print('Image names not found, naming by integers.')
+                    self._series_names.append('Image {}'.format(_i))
+                    
         return self._series_names
 
     def __del__(self):
@@ -112,46 +112,16 @@ class BioformatsFile(bioformats.ImageReader):
 
 class DataSource(BaseDataSource):
     moduleName = 'BioformatsDataSource'
-    def __init__(self, path=None, url=None, taskQueue=None, chanNum = 0, series=None):
+    def __init__(self, image_file, taskQueue=None, chanNum = 0, series=None):
         self.chanNum = chanNum
 
-        if isinstance(path, BioformatsFile):
-            self.bff = path
+        if isinstance(image_file, BioformatsFile):
+            self.bff = image_file
         else:
-            if url is not None:
-                path = url
-            self.path = getFullExistingFilename(path)#convert relative path to full path
-            
-            #self.data = readTiff.read3DTiff(self.filename)
+            self.filename = getFullExistingFilename(image_file)#convert relative path to full path
 
-            #self.im = Image.open(filename)
-
-            #self.im.seek(0)
-
-            #PIL's endedness support is subtly broken - try to fix it
-            #NB this is untested for floating point tiffs
-            #self.endedness = 'LE'
-            #if self.im.ifd.prefix =='MM':
-            #    self.endedness = 'BE'
-
-            #to find the number of images we have to loop over them all
-            #this is obviously not ideal as PIL loads the image data into memory for each
-            #slice and this is going to represent a huge performance penalty for large stacks
-            #should still let them be opened without having all images in memory at once though
-            #self.numSlices = self.im.tell()
-            
-            #try:
-            #    while True:
-            #        self.numSlices += 1
-            #        self.im.seek(self.numSlices)
-                    
-            #except EOFError:
-            #    pass
-
-            print((self.path))
-            
-            #tf = tifffile.TIFFfile(self.filename)
-            self.bff = BioformatsFile(path, url)
+            print(self.filename)
+            self.bff = BioformatsFile(self.filename)
 
         if series is not None:
             self.bff.rdr.setSeries(series)
--- a/PYME/IO/DataSources/BioformatsDataSource.py
+++ b/PYME/IO/DataSources/BioformatsDataSource.py
@@ -150,7 +150,7 @@ class DataSource(BaseDataSource):
             print((self.path))
             
             #tf = tifffile.TIFFfile(self.filename)
-            self.bff = BioformatsFile(path)
+            self.bff = BioformatsFile(path, url)
 
         if series is not None:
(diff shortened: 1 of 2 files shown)


PYME/IO/DataSources/BioformatsDataSource.py
```diff
@@ -25,7 +25,7 @@
     import bioformats
     #javabridge.start_vm(class_path=bioformats.JARS, run_headless=True)
 except:
-    pass
+    raise ImportError('Please make sure python-javabridge (https://pythonhosted.org/javabridge/) and python-bioformats (https://pythonhosted.org/python-bioformats/) are working before you use this module.')
```
David-Baddeley: It would be good to know why these are in a `try ... pass` statement in the first place. I suspect there might be a reason for it (which may or may not be any good). There is a chance that throwing an import error here will break data loading more generally when bioformats is not present. More likely is that this was a hack to, e.g., get documentation generation running on a platform where bioformats wasn't installed. 

If going ahead with these changes I'd like to see the old code commented out, and a comment saying what had happened and that it was being removed as it wasn't obvious why.

Proposed alteration is not really that different to removing the `try ... pass` completely (which might be better as you get to see which module failed and potentially why).

PYME/IO/DataSources/BioformatsDataSource.py
```diff
@@ -109,7 +110,24 @@ def __init__(self, filename, taskQueue=None, chanNum = 0):
         
         self.additionalDims = 'TC'
                 
+    def setSeries(self, series):
```
David-Baddeley: What is the use case for calling `setSeries()` on an existing data source? In general most consuming code (dh5view, analysis, etc ...) expect a `DataSource` to be immutable (or at the very least, have constant dimensionality). I suspect that it's not going to be a good idea to to change the series that a data source points to on the fly.

PYME/IO/DataSources/BioformatsDataSource.py
zacsimile: I don't think it will break loading more generally, as I haven't installed bioformats installed before today. We can drop the `try...pass`. Which old code do want commented out? I didn't delete anything besides `pass`, or is that what you're referring to?

PYME/IO/DataSources/BioformatsDataSource.py
zacsimile: I wanted to be able to loop through the different images in the series with a single load call. I figured it would be more efficient. We can drop it if you feel it's unsafe. I have no problem loading the same file 5 times to get 5 different series within it.

PYME/IO/DataSources/BioformatsDataSource.py
David-Baddeley: I was referring to the `pass` - just wanted to have a note in the code that indicated that this has changed without knowing why the old code was how it was when someone finds an error and comes looking. The circumstances where it would fail is if we import all the datasources somewhere (current code would ignore a failure to import bioformats). This might end up being in the task distribution stuff - there are places where we load data sources by module name (encoded in a string), rather than directly. These won't show up in a simple search of the source.

PYME/IO/DataSources/BioformatsDataSource.py
David-Baddeley: I think it's preferable to create 5 different DataSources. That said, we need a way of working out what is in the file, and ultimately selecting which series we want in the UI (or on the command line for recipes etc ...).

One way of addressing this would be to:
- change `BioformatsDataSource` so that the filename parameter (potentially renamed to file - would need to check how it's called) could take either a filename or a `bioformats.ImageReader` instance (or potentially our own proxy around a bioformats file so we can track the filename etc ...).
- In `image.py` (or wherever we call it) instantiate the `ImageReader`, and query the number of series. If we have a GUI and there is more than one series, prompt the user to see which series they want to open. Pass the `ImageReader` that we constructed to the data source.


That leaves us with the command line - which we could potentially defer to later - I'd be in favour of using the http URL query syntax e.g. `/path/to/somefile.lif?series=xxx` in the filename.

David-Baddeley:
PS - thanks for looking at this - it's been on my TODO list for a while.

PYME/IO/DataSources/BioformatsDataSource.py
David-Baddeley: Doing it this way I can see a clear path to having something like the UI that FIJI has for loading multi-series files - where you see a list of series with their dimensionality, series name (if we can get it) and (potentially - if there is sufficient demand) a thumbnail.

zacsimile: Yes, I agree. I'll have to look a little more into the http URL query syntax. Series name is accessible through `self.bff.rdr.getSeriesMetadata(<series number>)`, I think. Maybe the `BioformatsDataSource` should contain the whole LIF (or whatever bioformats image) and then if we want to open a series from it we just spin off an `ArrayDataSource` containing the series?

zacsimile:
No problem. I needed it to work.

PYME/IO/DataSources/BioformatsDataSource.py
David-Baddeley: You don't want to load everything and then spin out an `ArrayDataSource` as it then forces everything into memory (a problem if have large data sets). At present, we load lazily, meaning that we only have a few frames in memory at once.

zacsimile: Oh, that's right.

OK, so, do we want almost exactly what we have here then? We probably want to put the option of which series to load in the GUI, but the data source should just accept a series as an argument and just load that one and never change, right?

zacsimile: I looked into the URL format. I don't see a clear reason to go with it over just loading the file and making a getSeries() call. Seems to be just variations on a pointer. Is there more to it?

David-Baddeley: We'd still be loading a file and calling getSeries(), my idea with the URL format would just be a way of specifying which series we want on the command line (for cases where we don't want to prompt the user and/or don't have a GUI - e.g. automated analysis in bakeshop). This is why I said we could defer it. We currently use something similar with the `SupertileDataSource`.

David-Baddeley: We're a bit out of sync - replying to your comment 2nd to last comment: yep - pretty much what we want. I'd ideally:
- keep the series argument to the constructor
- back out the `getSeries()`, `setSeries()`, `setXYZTX()` calls (which are potentially confusing if we never expect them to get called)
- remove the `nSeries` property
- remove the current check on the number of series in the constructor.
- add the following to the constructor (before setting x and y sizes etc):
```python
if isinstance(filename, BioformatsFile): #TODO - change filename parameter name
    self.bff = filename.bff
else:
    self.bff = bioformats.ImageReader(filename)

if not series is None:
    self.bff.rdr.setSeries(series)
```
I'd then copy the `getSeries` and `nSeries` logic to a new class:

```python
class BioformatsFile(object):
    def __init__ (self, filename):
        self.filename = filename
        self.bff = bioformats.ImageReader(filename)

    def get_series_info(self):
        # potentially rename
        # return a list of series names in the file, potentially also with dimensionality etc ...
```

Usage in `Image._loadBioformats()` 

```python
series_num = None
#try parsing filename for series URL parameter, over-write series_num if sucessful

bioformats_file = BioformatsDataSource.BioformatsFile(filename)
series_info = bioformats_file.get_series_info()

if series_num is None and len(series_info) >1:
    print("File has multiple series, need to pick one")

    if self.haveGUI:
        #wx multichoice dialog on series_info list, set series_num to result
    else:
        logger.warning('No GUI, using 0th series')

BioformatsDataSource.DataSource(bioformats_file, series=series_num)
```

David-Baddeley: some of the ordering above is wonky - create the `BioformatsFile` class and copy logic before removing from the data source.

David-Baddeley: Actually thinking about it a bit more, the `BioformatsFile` class as such a thin proxy is a bit gross - it should probably be a subclass of `ImageReader` rather than a class which simply holds a reference to an `ImageReader` instance. We're basically just adding an easy way to get the number of series and present them in a GUI.

zacsimile: Ah, that's true. So subclass `ImageReader` and add the standard `DataSource` functions like `getSlice()`, `getSliceShape()`, etc.? Or multi-inherit `BaseDataSource` and `ImageReader`?

David-Baddeley: Neither - Subclass `ImageReader` to provide simpler access to info about the different series (no additional functions like `getSlice()`) - this is just to make the UI for series selection easier - you could in principle put all this logic in `Image` and make it call `bff.rdr.getSeriesCount()` and `[bff.rdr.getSeriesMetadata(n) for n in range(bff.rdr.getSeriesCount())]`, but it might be a little cleaner to keep it out of there. 

Make the `DataSource` own a reference to an `ImageReader` (or subclass thereof) (as it currently does).

David-Baddeley: Quick correction on the dialog in the GUI - you want a `wx.SingleChoiceDialog`

zacsimile: Oh I see. I missed `BioformatsFile`.

zacsimile:
Made the changes. Everything works. I put a note on line 898 of `image.py` that we're being a little inefficient with our metadata handling. The main reason for this is it's kind of hard to grab metadata from bioformats images through `python-bioformats` (or there's a very easy way that we're missing), but we should probably look into updating how we handle this at some point.

PYME/IO/DataSources/BioformatsDataSource.py
```diff
@@ -20,12 +20,16 @@
...
-except:
-    pass
+# try:
+#     import javabridge
+#     import bioformats
+#     #javabridge.start_vm(class_path=bioformats.JARS, run_headless=True)
+# except:
+#     pass
+
+# Try/pass loop removed to reveal loading issues
```
David-Baddeley:
```suggestion
# Try/pass removed to reveal loading issues
# unclear why try ... pass block was present in the first place TODO - check for potential regressions caused by removal
```

PYME/IO/DataSources/BioformatsDataSource.py
```diff
@@ -58,44 +62,99 @@ def release_VM():
...
+        return self._series_names
+
+    def __del__(self):
+        release_VM()
+
 class DataSource(BaseDataSource):
     moduleName = 'BioformatsDataSource'
-    def __init__(self, filename, taskQueue=None, chanNum = 0):
-        self.filename = getFullExistingFilename(filename)#convert relative path to full path
+    def __init__(self, path=None, url=None, taskQueue=None, chanNum = 0, series=None):
```
David-Baddeley: Can't change the positional argument signature (before `chanNum` - can add things after `chanNum`) here, due to the way datasources are passed around in tasks. 

What is the `url` argument supposed to do? I'm guessing this is to allow it to be passed to `bioformats.ImageReader`, but seeing as it's completely undocumented I don't think we need/want to support it. If we do want to support it, I'd suggest we infer whether our filename is a path or a url before passing it.

David-Baddeley: Bioformats *should* do this inference internally - it's a bit stink that it doesn't

zacsimile: Yeah, I just copied how bioformats did it since I assumed there was a reason. No reason we can't infer a URL, though.

zacsimile:
Looks like you're fixing things. Should I hold off then?

David-Baddeley:
May as well - I thought it'd be simpler than lots of back and forth. Think we're just about done now, modulo doing anything about potential triple loading of metadata (it should be double at worst as the rdr and file metadata should be loaded redundantly, although I wouldn't put anything past bioformats).

zacsimile:
I posted in forum.image.sc about the metadata. Still waiting for a response.

zacsimile:
It doesn't slow anything down significantly, so I think the triple metadata is fine, just ungainly. I think we can merge. As soon as I get a sufficient response or figure out a fix, I'll do another PR to clean up the metadata calls.

David-Baddeley:
it depends on what filetype - getting the metadata is horrendously slow for some bioformats files (deltavision OMX .dv files in particular). It might be that the access is somewhat lazy in which case calling `getSeriesMetadataValue('Image name')` rather than getting the entire dict might make sense (then we only load the entire metadata once, with the get ome xml call).

As it stands, however, we might not be getting the right metadata for multi-series images (we get the global metadata associated with the file)

zacsimile:
We get the per-series metadata the way it's written in `BioformatsFile`.

David-Baddeley:
The OME API documentation is garbage! It's also really odd that they map their metadata to field->value pairs in the java API whereas it's a deeply nested and highly controlled ontology in OME xml. I wonder how they handle the nesting - maybe a field maps to a java object (ick) rather than a simple value.

zacsimile:
Fields are stored in Java objects. OME metadata is stored as a Java `Hashtable` (https://docs.oracle.com/javase/8/docs/api/java/util/Hashtable.html). You can then either get it as a `Collection` (Java object) or `Enumeration` (Java object). You can spit the whole thing to string, but then you have to parse it back into a dictionary or list. Javabridge has a function for moving hashtables to python dicts that's much better (we use it).

zacsimile:
Ah, I see what you mean. Yes, I think the `Enumeration` then also stores an iterable set of Java objects.

zacsimile:
Eh, technically it stores them as Java `String` objects. But each `String` is flat. So one string will be `HardwareSetting|ScannerSettingRecord|dblZoom` and it's value, then the next `HardwareSetting|ScannerSettingRecord|csLutName07` and it's value, and so on. The nesting is represented by pipes.

David-Baddeley:
That's actually pretty close to the best case scenario - we should be able to map to our metadata pretty easily by just replacing pipes with dots.

zacsimile:
We'll have to strip spaces, too, but yes!

PR #171, opened 2020-03-04
Fix playback OnFrame conflict with displayOptions.py zp setter

zacsimile:
Playback wasn't looping on stacks in dh5view.

### DIFF
--- a/PYME/DSView/modules/playback.py
+++ b/PYME/DSView/modules/playback.py
@@ -121,7 +121,10 @@ class PlayPanel(wx.Panel):
             self.bPlay.SetBitmapLabel(self.bmPlay)
 
     def OnFrame(self):
-        self.do.zp +=1
+        if self.do.zp >= (self.do.ds.shape[2]-1):
+            self.d0.zp = 0
+        else:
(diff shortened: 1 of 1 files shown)


PYME/DSView/modules/playback.py
```diff
@@ -122,7 +122,7 @@ def OnPlay(self, event):
 
     def OnFrame(self):
         self.do.zp +=1
```
David-Baddeley: The proposed change will skip the last frame - we want to check if we are on the last frame and increment to 0 instead. This will still show the last frame

```suggestion
        if self.do.zp >= (self.do.ds.shape[2]-1):
            self.d0.zp = 0
        else:
            self.do.zp +=1
```

zacsimile: I have no strong feelings on this either way, especially since this is already merged, but just as a note in case we run into this again: it didn't skip the last frame because of the `self.OnChange()` call in `@zp.setter`. The frame would move on `self.do.zp += 1` and then move again on `self.do.zp = 0`. I tested it to be sure.

PR #173, opened 2020-03-05
fixing broken units in measurement.TilePhysicalCoords

barentine:
Current head assumes `meas['x']` is both in units of nanometers and pixels, which is of course only the case for one pixel size. Docs state it should be pixels, but `Measure2D` converts using a `pixelSize` attribute of the incoming labels, so I set this up to work for recipe-specified units.

Did we decide for certain that we want to keep `'x_um'`? Or is this a good opportunity to switch to just using `'x'`

### DIFF
--- a/PYME/recipes/measurement.py
+++ b/PYME/recipes/measurement.py
@@ -794,7 +794,7 @@ class TilePhysicalCoords(ModuleBase):
 
     input_measurements = Input('measurements')
     input_supertile = Input('input')
-    measurement_units = CStr('um')
+    measurement_units = CStr('nm')
     output_name = Output('meas_physical_coords')
     
     def execute(self, namespace):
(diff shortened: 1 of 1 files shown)


PYME/recipes/measurement.py
```diff
@@ -776,37 +776,56 @@ def execute(self, namespace):
...
+        `inputMeasurements`
     """
 
-    inputMeasurements = Input('measurements')
-    inputImage = Input('input')
-    
-    outputName = Output('meas_physical_coords')
+    input_measurements = Input('measurements')
+    input_supertile = Input('input')
+    measurement_units = CStr('um')
```
David-Baddeley:
```suggestion
    measurement_units = CStr('nm')
```

PR #185, opened 2020-03-11
Add CameraMapMixin caching

barentine:


David-Baddeley:
I know this is going to be annoying, but I still don't like it ... my specific dislikes are:

- I don't believe that the camera map filename name logic belongs in the mixin (this should be coded in one place so you can use the same logic when re-analysing files which were taken without a recorded map (ideally involving moving the `check_map_exists()` to `gen_sCMOS_maps.py`)
- caching by filename (rather than by map_type, serial number, integration time) seems a bit counter-intuitive to me
- I don't like looping over camera map fields - I feel this adds complexity and makes things hard to read for no real gain. I don't see a realistic scenario where we would be using anything other than the 3 standard maps.

My preferred alternative below:

```python
class CameraMapMixin(object):
    def _map_cache_key(self, mdh, map_type):
        if map_type == 'flatfield':
            return (map_type, mdh['Camera.SerialNumber'])
        else:
            return (map_type, mdh['Camera.SerialNumber'], mdh['Camera.IntegrationTime'])
        
    def get_camera_map_id(self, mdh, map_type):
        #create cache if not already present
        if not hasattr(self, '_map_cache'):
            self._camera_map_cache = {}
            
        cache_key = self._map_cache_key(mdh, map_type)
        try:
            return self._camera_map_cache[cache_key]
        except KeyError:
            map_fn = check_mapexists(mdh, map_type, fill=False)
            self._camera_map_cache[cache_key] = map_fn
        
    
    def fill_camera_map_metadata(self, mdh):
        dark = self.get_camera_map_id(mdh, 'dark')
        if dark is not None:
            mdh['Camera.DarkMapID'] = dark

        var = self.get_camera_map_id(mdh, 'variance')
        if var is not None:
            mdh['Camera.VarianceMapID'] = var

        flat = self.get_camera_map_id(mdh, 'flatfield')
        if flat is not None:
            mdh['Camera.FlatfieldMapID'] = flat
```

along with a modification of `check_map_exists()` to not fill metadata (this side-effect was pretty ugly to start with), and potentially a cache clearing function.

David-Baddeley:
Or ...

```python
class CameraMapMixin(object):
    def _map_cache_key(self, mdh, map_type):
        if map_type == 'flatfield':
            return (map_type, mdh['Camera.SerialNumber'])
        else:
            return (map_type, mdh['Camera.SerialNumber'], mdh['Camera.IntegrationTime'])
        
    def _fill_camera_map_id(self, mdh, mdh_key, map_type):
        #create cache if not already present
        if not hasattr(self, '_map_cache'):
            self._camera_map_cache = {}
            
        cache_key = self._map_cache_key(mdh, map_type)
        try:
            map_fn = self._camera_map_cache[cache_key]
        except KeyError:
            map_fn = check_mapexists(mdh, map_type, fill=False)
            self._camera_map_cache[cache_key] = map_fn
        
        if not map_fn is None:
            mdh[mdh_key] = map_fn
    
    def fill_camera_map_metadata(self, mdh):
        self._fill_camera_map_id(mdh, 'Camera.DarkMapID', 'dark')
        self._fill_camera_map_id(mdh, 'Camera.VarianceMapID', 'variance')
        self._fill_camera_map_id(mdh, 'Camera.FlatfieldMapID', 'flatfield')
```

barentine:
> * I don't like looping over camera map fields - I feel this adds complexity and makes things hard to read for no real gain. I don't see a realistic scenario where we would be using anything other than the 3 standard maps.

This is what I had in mind:
https://doi.org/10.1038/s41598-019-53698-x
For gen3 sCMOS I doubt it's an issue, but those of us ising gen2's might one day be interested in a 4th map?

barentine:
> * I don't believe that the camera map filename name logic belongs in the mixin (this should be coded in one place so you can use the same logic when re-analysing files which were taken without a recorded map (ideally involving moving the `check_map_exists()` to `gen_sCMOS_maps.py`)

This is what we do for the camera map filename, but I'm assuming you mean for the full path, i.e. cluster/local, which I think you're right would be nice to have factored out. 

> * caching by filename (rather than by map_type, serial number, integration time) seems a bit counter-intuitive to me

Good point

David-Baddeley:
I think the flatfield map should capture and correct RQE already...

barentine:
They agree with you for Gaussian least squares, but not Poisson MLE. I'm not trying to argue about that at the moment, I'm just pointing out a ~realistic scenario.

Another scenario is not wanting to bother with a flatmap at all.

barentine:
bad merge, marking as draft while I sort out

barentine:
Current head should be fixed with 1f7b86776f6e8d243343b3d722db6efd1565148b, will open something new if it doesn't work once we're back in lab

PR #186, opened 2020-03-12
Dumbell fit 3d

zacsimile:
Created a 3d dumbbell fitter. Also tried to consolidate code in FFBase.


tests/PYME/localization/test_localization_algorithms.py
```diff
@@ -40,6 +40,26 @@ def test_LatGaussFitFR():
     
     assert errors_over_pred_IQR < 2.5
 
+# def test_Dumbell3DFitR():
```
barentine: maybe we can remove this?

David-Baddeley:
I really, really don't like the changes to FFBase. I think it's made it much more verbose and harder to read. All the `if multiview:` stuff everywhere is really gross. There are some places where we should legitimately look at some refactoring (the noise calculation is the most pressing), as well as looking at what is shared between the split and multiview stuff. As it stands, `getMultiviewROIAtPoint()` is experimental and largely untested, existing for stalled 4Pi fitting attempts.  The long term plan here is to eventually unify on one multi-view function which handles both the old splitting case and the arbitrary quandrant number case, but having the multiview one tested and working first is a pre-requisite, followed by adapting the splitter fits to call it. 

Another reason I'm going to push back so strongly on this is that I suspect that the current unit tests won't catch some of the more subtle errors that could crop up with the splitter code (namely how ROI extraction works when there are large shifts.

David-Baddeley:
I'm leaning strongly towards rejecting all the `FFBase` changes - there are only really 2 functions, `getROIAtPoint()` and `GetSplitROIAtPoint()` and they don't share enough to warrant major refactoring, particularly before we have the multi-split code working (all the if statements in your refactoring attempt are testament to this).

`GetMultiviewROIAtPoint` is, unfortunately, a red herring in that it's a very incomplete code stub (basically a copy of GetSplitROIAtPoint at the moment, but expected to diverge significantly once we actually work on it).  I've updated the docs and factored out the sigma calculation stuff so that that is consistent. The other bit we could potentially factor out is the voxel size calculation, but that is so small it doesn't seem worth it (your proposed refactoring relies on `voxelsize.z` being present, which we can't guarantee, also, appending to a list is gross).

tests/PYME/localization/test_localization_algorithms.py
David-Baddeley: I think Zach is being nice and leaving it commented out as it currently doesn't pass / he hasn't found the right text criteria and doesn't want to add a failing test.

zacsimile:
OK. I needed almost exactly the same code for `getROIAtPoint` to get it to work in 3D. It would be good if we could keep the `dim=` refactoring part of `getROIAtPoint` at least.

I do think it's cleaner. If we delete the commented-out code, I think that becomes more apparent. I may have gone overboard. When I was working on it, it turned out the code for `getROIAtPoint` and `getROIAtPoint3D`, which got wrapped into `getROIAtPoint` in a later pull request, was almost identical to the code in  `getSplitROIAtPoint` and `getMultiviewROIAtPoint`, which I think are identical functions at the moment. Mostly I wanted to remove duplicated code in the file. Reducing the file to two functions would also work. I do think the `pixel_size` property could also still be useful.

tests/PYME/localization/test_localization_algorithms.py
zacsimile: This was my thought, yes. I do believe this test will pass in the near future.

zacsimile:
I reverted most of the changes back to the original `FFBase`. It's less of a refactoring now and more of a minor cleanup, plus adding a `dim` argument to `getROIAtPoint`. What do you think?

Issue #187, opened 2020-03-12
Memory allocation issue in pairwise distance histogram calculation with numpy=1.18


zacsimile:
I am unable to run pairwise distance histogram calculations from VisGUI on a relatively small file (4 million localizations) in Windows 10, Python 3. I can calculate a pairwise distance histogram on the same file on my Mac just fine. Traceback below.

```
Traceback (most recent call last):                                                                                        
    File "c:\users\zrc4\code\python-microscopy\PYME\LMVis\Extras\clusterAnalysis.py", line 281, in OnPairwiseDistanceHistogram                                                                                                                 
        distances = distogram.execute()                                                                                       
    File "c:\users\zrc4\code\python-microscopy\PYME\recipes\base.py", line 519, in execute                                    
        m.execute(self.namespace)                                                                                             
    File "c:\users\zrc4\code\python-microscopy\PYME\recipes\measurement.py", line 399, in execute                             
        res = DistHist.distanceHistogram(pos0['x'], pos0['y'], pos1['x'], pos1['y'], self.nbins, self.binSize)              
RuntimeError: Error allocating output array
```

zacsimile:
Downgraded from numpy=1.18.1 to numpy=1.16.5 and it works again.

zacsimile:
Upgraded to numpy=1.17.0 and that also works. So numpy=1.18.1 is the problem.

David-Baddeley:
That's pretty weird - windows is pretty bad at memory allocation (you'll sometimes get allocation issues due to the lack of contiguous blocks when allocating a large array even if it seems like there should be enough memory for it), but the array being allocated here is small. 

What version of numpy is PYME compiled against in this case?

I suspect the issue is that this code is old, and uses `PyArray_FromDims()` to allocate the array, but this is on the deprecation path. Replacing with `PyArray_SimpleNew()` would probably do the trick.

Just looked at the numpy 1.18 release notes (https://numpy.org/devdocs/release/1.18.0-notes.html), and they would seem to confirm this - look under expired deprecations.

David-Baddeley:
Looks like this will effect a few other places too ...

```bash
DB3:python-microscopy david$ grep -n -r PyArray_FromDims .
./PYME/Analysis/points/astigmatism/astiglookup.c:57:    out_z = (PyArrayObject*) PyArray_FromDims(1,outDimensions,PyArray_INT);
[...]

PR #188, opened 2020-03-12
Remove deprecated Numpy C macros

zacsimile:
Fix to #187.


zacsimile:
The only one I'm not sure will work is `fft_test.c` because I don't have control over the datatype of `f->dimensions` from this file. This needs to be an array of datatype `npy_intp`.

David-Baddeley:
I think the fft_test stuff doesn't really matter - we can fix that if it ever gets used (otherwise maybe purge it).

PR #190, opened 2020-03-13
Dialog for saving pairwise distance histograms in recipe

zacsimile:
By request


David-Baddeley:
This seems like a reasonable short-term hack, and matches what we do with shift-fields etc. Longer term we might want to think about something a bit more elegant.

My long term vision for things like this, along with photophysics, particle tracking, PSF calibrations etc ...  is for them to generate an HTML report in a new tab with, e.g., 'Save Report' and 'Save Data' options. Unfortunately the HTML viewer support is currently not as cross platform as I'd like it to be.

PR #196, opened 2020-03-25
Added easy visualization of Delaunay triangulation

zacsimile:


zacsimile:
I missed the `Tesselation` class in `PYME.recipes.pointcloud`, which does roughly the same thing. Maybe this is the wrong place for the `Tesselation` class?

David-Baddeley:
Agree that the Tessellation class should ultimately be moved, but not sure where to - it exists to expose a tessellation in a recipe friendly manner, so is mostly a shim rather than anything else - although it does contain the circumcentres logic. 

Will have a think about how best to combine both. Circumcentres etc ... might be a legit property of any triangle mesh, so just subclassing could be fine, but that might run into issues with higher dimensional tessellations (a Delaunay tesselation in 3-space will give you valence 2 faces and >2 valence edges and still be correct - it's a volume mesh rather than a surface mesh, so much of the stuff in the mesh class won't work).

Having a `from_delaunay` function in `TriangleMesh` has the same problem - it won't give you a valid (surface) mesh if applied to 3D data. You can get a valid surface mesh (convex hull unless you've, e.g., previously dropped simplices which don't contain points) by taking all valence-1 faces. This might possibly be what you want ...

David-Baddeley:
There's a bit of work towards getting surfaces from the Delaunay volume mesh in `PYME.Analysis.points.gen3DTriangs` (start at `gen3DBlobs()` and work your way down - the internal face culling is in `gen3DTriangsTF()`). That said, this code is ugly, slow, and incomplete enough that it might be better to start from scratch.

David-Baddeley:
lines 230 - 255 to be specific for the internal culling. The algorithm is about as about as naive as it gets, and as a consequence really slow, There are some pretty simple ways to improve it (e.g. below), but there might be even better ways using face hashes etc ...
```python
    if internalCull:
        fcs = {} 
        for i, triI in enumerate(triInds):
            t_t = tuple(triI)
            prev_loc = fcs.get(t_t, -1)            
            if prev_loc > -1:
                surfInds[prev_loc] = 0
                surfInds[i] = 0
            else:
                fcs[t_t] = i
```

zacsimile:
While I think this would be a cool thing to do, I wrote the `from_delaunay` method with the idea that I could render a Delaunay triangulation in VisGUI alongside a point cloud for debugging purposes. I think `Tesselation` does this just fine and so there's no need for this method.

David-Baddeley:
I thought it might have been related to our Delaunay based hole-punching idea, hence the (rather verbose) description.

zacsimile:
It is, but I was just using it for visual debugging. I could go further down the face culling route, if you think that might help.

I wonder if we could get the surface faces by just looking for triangle normals that pass through no other tetrahedra. Line-plane intersection is cheap and tetrahedara are just 4 planes.

David-Baddeley:
That would potentially work, but you would need to compare each face with
all simplices (O(N^2), with complex comparison). Looking at the face
valences is much easier.

zacsimile:
Will submit different pull request addressing this.

PR #197, opened 2020-03-27
vector tools / spherical harmonic improvements and refactoring

barentine:
This is not ready to merge - have some questions @David-Baddeley about what sort of legacy support we want/how to go about it. Is anyone using the spherical harmonic shell fitting? @zacsimile mentioned he was writing up some related docs, so I thought maybe we should have a look at this now rather than later.

Changes
---------
- During spherical harmonic shell fitting, scale points by spread along principle axes rather than by a hyperparameter along z only
- Add ability to load spherical harmonic shells which have already been fit into visgui
- refactor vector functions out of spherical_harmonics.py, add some (mostly related to shell normals/distance tests), add docs, add unit tests
- adds shell fitting/loading under the surface fitting LMVis menu

Not included: breaking/removing old code
-------------
- Deleting/deprecating PYME/Analysis/points/spherical_harmonics.py, which is replaced by vector_tools.py and spherical_harmonics/fitting.py, the latter should really just be in spherical_harmonics.py but until we decide that's OK to move I'll leave this here.
- Deleting/deprecating PYME.recipes.localisations.SphericalHarmonicShell which is replaced by a module of the same name in recipes.surface_fitting
- Deleting/deprecating PYME.LMVis.Extras.spherical_harmonics.py which is replaced by changes to Extras.surface_fitting.py
- some more docs


David-Baddeley:
A couple of quick thoughts:
- It's a hard call on whether spherical harmonics belong in surface fitting or not.  I'm not sure I see a huge need for the change in location.
- I don't think that 'vector_tools' is a particularly informative filename for what is in there.
- I'm pretty sure that `find_principle_axes()` and `distance_to_image_mask()` duplicate functionality available elsewhere.
- The convention throughout PYME is that distances to masks return negative distances for anything inside an object (see, e.g., the `processing.DistanceTransform` recipe module and `Analysis.colocalization.edtColoc`).

PYME/Analysis/points/vector_tools.py
```diff
@@ -0,0 +1,220 @@
...
+    inside_r = np.less_equal(np.sqrt(radial_component0[worth_checking] ** 2 + radial_component1[worth_checking] ** 2), radius)
+
+    # handle the indexing
+    inside = np.zeros_like(x, dtype=bool)
+    # can't slice because this copies, use where as a sort of disgusting work around
+    inside[np.where(inside_axially)[0][np.where(worth_checking)[0]][inside_r]] = True
+    return inside, axial_component[inside]
+
+
+def distance_to_image_mask(mask, points):
```
David-Baddeley: This largely duplicates `cluster_morphology.get_labels_from_image()`

zacsimile: This largely duplicates `PYME.Analysis.points.objectMeasure.get_labels_from_image()`

barentine:
Marking as draft for now.

Updates:
- changed `vector_tools.py` to `coordinate_tools.py`
- removed unused `PYME.Analysis.points.objectMeasure.get_labels_from_image()`, which duplicated `cluster_morphology.get_labels_from_image()` sans an optional feature
- refactored the image index finding code duplicated in `cluster_morphology.get_labels_from_image()` and `distance_to_image_mask` into a separate function: `pixel_index_of_points_in_image`

Still to do:

- [x] change `distance_to_mask` to return negative distances inside.
- [x] potentially move `points.objectMeasure.getPrincipleAxis` into `coordinate_tools.py`, though it could stand with some doc
- [x] integrate things such that old `z_scale` shells can still be loaded
- [x] more docstrings

barentine:
Marking some things as checked above if we've decided not to do them (e.g. support legacy zscale shells, which we have deemed not necessary).

Still need to change distance to shell to return negative distances for inside the shell. Worth thinking about how this gets integrated with the other surfaces/meshing. The distance to shell is currently approximate, since we reconstruct the shell on a (theta,phi) grid, and can be inaccurate for points close to the surface but unlucky with the solid angle sampling. I could see it being nice to instead mesh the reconstruction and use the distance to surface function - @zacsimile am I right in thinking that is fairly trivial and might additionally be a part of the new surface display anyway?

Still need to change the visualization from mayavi to a surface layer, but I think Z will get to that once it's merged

barentine:
Quick chat with @zacsimile - we think it best to mesh and use the actual surface for distance_to_surface calculations rather than our current gridded point interpolation. So once this is merged he'll change the extras plug so we get a surface layer, and we'll use the `DistanceToMesh` recipe module instead of what's currently there for the spherical harmonic shell.

barentine:
Just a note that this is ready for review. 

- I don't think I addressed in comments here the notion that `coordinate_tools.find_principle_axes` is duplicated elsewhere - it is not; `objectMeasure.getPrincipalAxis` looks like it iteratively tries to find a single 2D vector, e.g. the first 2D principle component, while `find_principle_axes` solves the 3D eigen problem and returns all 3, 3D principle axes.
- in addition to converting visualization to a PYMEVisualize layer rather than mayavi, after this is merged @zacsimile will deprecate `ScaledShell.distance_to_shell` as we'll use the already present `DistanceToMesh` recipe module for adding that column to a tabular source instead, which should have higher accuracy near the shell than computationally reasonable realizations with the two-angle grid approach of `distance_to_shell`

PYME/Analysis/points/coordinate_tools.py
```diff
@@ -0,0 +1,282 @@
...
+    """
+    from PYME.IO.MetaDataHandler import get_camera_roi_origin
+
+    x0, y0, z0 = image.origin
+
+    # account for point data ROIs
+    try:
+        roi_x0, roi_y0 = get_camera_roi_origin(points.mdh)
+
+        p_ox = roi_x0 * points.mdh['voxelsize.x'] * 1e3
```
David-Baddeley: Use `points.mdh.pixelsize_nm` instead

PYME/Analysis/points/coordinate_tools.py
```diff
@@ -0,0 +1,282 @@
...
+        p_oy = roi_y0 * points.mdh['voxelsize.y'] * 1e3
+    except AttributeError:
+        raise RuntimeError('metadata specifying ROI position and voxelsize are missing')
+
+    # Image origin is referenced to top-left corner of pixelated image.
+    # FIXME - localisations are currently referenced to centre of raw pixels
+    x_index = np.floor((points['x'] + p_ox - x0) / image.pixelSize).astype('i')
+    y_index = np.floor((points['y'] + p_oy - y0) / image.pixelSize).astype('i')
```
David-Baddeley: `image.pixelSize` is just the x axis size (which is usually OK, but not 100% safe in case you have different pixel sizes in x and y. Probably better to use `image.voxelsize_nm.y` etc ...

PYME/Analysis/points/objectMeasure.py
```diff
@@ -161,49 +161,3 @@ def calcEdgeDists(objects, objMeasures):
         minEdgeDists.append(ed)
 
     return np.array(minEdgeDists)
-
-def get_labels_from_image(inp, img):
```
David-Baddeley, line 165: Why is this being deleted?

zacsimile: I believe it duplicates `cluster_morphology.get_labels_from_image()`.

PYME/Analysis/points/coordinate_tools.py
barentine: good catch, now taken care of

PYME/Analysis/points/objectMeasure.py
barentine: Correct.

PYME/LMVis/Extras/surface_fitting.py
```diff
@@ -44,7 +47,6 @@ def OnFitQuadraticPatches(self, event):
         
         self.vis_frame.Refresh()
 
-
-def Plug(visFr):
+def Plug(vis_frame):
```
David-Baddeley:
```suggestion
def Plug(visFr):
```

The current `visFr` variable name is gross, but an existing convention. `vis_frame`  as an alternative is only marginally less gross and breaks convention. If and when we change this it should be done globally, with appropriate changes to the plugin docs, and to a name which is actually sane.

PYME/LMVis/Extras/surface_fitting.py
```diff
@@ -44,7 +47,6 @@ def OnFitQuadraticPatches(self, event):
...
     '''Plugs this module into the gui'''
-    SurfaceFitter(visFr)
\ No newline at end of file
+    SurfaceFitter(vis_frame)
```
David-Baddeley:
```suggestion
    SurfaceFitter(visFr)
```

PYME/Analysis/points/objectMeasure.py
David-Baddeley: I'd argue the other way round - the `cluster_morphology` version  duplicates this one. It may be fairly academic in any case. I'd like to refactor whatever the final version is out into somewhere more sensible, but kind of feel that this is the canonical version, rather than the other way around. Before refactoring, I'd like to know the use case for the extra `minimum_localizations` argument is in the `cluster_morphology` version. To me it seems you'd almost always be better off filtering the resulting tabular output on cluster size rather than building it in as an argument to `get_labels_from_image()`.

Issue #119, opened 2019-12-13
OctreeRenderLayer architecture inconsistent with superclass TriangleRenderLayer


zacsimile:
TriangleRenderLayer was modified and OctreeRenderLayer couldn't keep up. Function calls need to be updated accordingly. This is minor since we rarely visualize the octree.

zacsimile:
#198 was merged, I think we can close this issue.

PR #198, opened 2020-03-28
Restore octree layer rendering

zacsimile:
Response to #119

### DIFF
--- a/PYME/LMVis/layers/octree.py
+++ b/PYME/LMVis/layers/octree.py
@@ -29,7 +29,7 @@ class OctreeRenderLayer(TriangleRenderLayer):
     min_points = Int(10, desc='Number of points/node to truncate octree at')
 
     def __init__(self, pipeline, method='wireframe', dsname='', context=None, **kwargs):
-        super(OctreeRenderLayer, self).__init__(pipeline, method, dsname, context,**kwargs)
+        TriangleRenderLayer.__init__(self, pipeline, method, dsname, context, **kwargs)
 
         self.on_trait_change(self.update, 'depth')
         self.on_trait_change(self.update, 'density')
@@ -81,6 +81,36 @@ class OctreeRenderLayer(TriangleRenderLayer):
             shifts = (box_sizes[:,None]*OCT_SHIFT[None,:])*0.5
             v = (c[:,None,:] + shifts)
 
+            #
+            #     z
+            #     ^
+            #     |
+            #    v4 ----------v6
+            #    /|           /|
+            #   / |          / |
+            #  v5----------v7  |
+            #  |  |    c    |  |
+            #  | v0---------|-v2
+            #  | /          | /
+            #  v1-----------v3---> y
+            #  /
+            # x
+            #
+            # Now note that the counterclockwise triangles (when viewed straight-on) formed along the faces of the cube are:
+            #
+            # v0 v2 v1
+            # v0 v1 v5
+            # v0 v5 v4
+            # v0 v6 v2
+            # v0 v4 v6
+            # v1 v2 v3
+            # v1 v3 v7
+            # v1 v7 v5
+            # v2 v6 v7
+            # v2 v7 v3
+            # v4 v5 v6
+            # v5 v7 v6
+
             # Counterclockwise triangles (when viewed straight-on) formed along
             # the faces of an octree box
             t0 = np.vstack(v[:,[0,0,0,0,0,1,1,1,2,2,4,5],:])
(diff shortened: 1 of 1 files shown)


PYME/LMVis/layers/octree.py
```diff
@@ -11,23 +11,25 @@
...
-    #alpha = Float(1.0, desc='Face tranparency')
+    # Additional properties (the rest are inherited from TriangleRenderLayer)
     depth = Int(3, desc='Depth at which to render Octree. Set to -1 for dynamic depth rendering.')
     density = Float(0.0, desc='Minimum density of octree node to display.')
     min_points = Int(10, desc='Number of points/node to truncate octree at')
-    #method = Enum(*ENGINES.keys(), desc='Method used to display faces')
 
-    def __init__(self, pipeline, method='wireframe', dsname='', **kwargs):
-        TriangleRenderLayer.__init__(self, pipeline, method, dsname, **kwargs)
+    def __init__(self, pipeline, method='wireframe', dsname='', context=None, **kwargs):
```
David-Baddeley, line 31: Why the change here? In general specifically calling the superclass constructor is preferred (rather than using super()

PYME/LMVis/layers/octree.py
```diff
@@ -68,106 +69,77 @@ def update_from_datasource(self, ds):
...
+            clim = [0, 1]
             
             alpha = self.alpha*alpha/alpha.max()
             alpha = (alpha[None,:]*np.ones(12)[:,None])
             alpha = np.repeat(alpha.ravel(), 3)
             print('Octree scaled alpha range: %g, %g' % (alpha.min(), alpha.max()))
 
-            # Pass the restructured data to update_data
-            self.update_data(x, y, z, cmap=getattr(cm, self.cmap), clim=self.clim, alpha=alpha, xn=xn, yn=yn, zn=zn)
+            cmap = getattr(cm, self.cmap)
```
David-Baddeley, line 137: I guess this is because update_data is now missing from TriangleRenderLayer? If so, then that is likely a problem.

PYME/LMVis/layers/octree.py
```diff
@@ -68,106 +69,77 @@ def update_from_datasource(self, ds):
...
-            triangle_v1 = np.vstack((v2, v4, v3, v0, v1, v7, v6, v1, v7, v2, v4, v5))
-            triangle_v2 = np.vstack((v6, v5, v2, v1, v5, v6, v7, v0, v3, v3, v0, v4))
-
-            x, y, z = np.hstack((triangle_v0, triangle_v1, triangle_v2)).reshape(-1, 3).T
-
-            # Now we create the normal as the cross product (triangle_v2 - triangle_v1) x (triangle_v0 - triangle_v1)
-            triangle_normals = np.cross((triangle_v2 - triangle_v1), (triangle_v0 - triangle_v1))
-            # We copy the normals 3 times per triangle to get 3x(3N) normals to match the vertices shape
-            xn, yn, zn = np.repeat(triangle_normals.T, 3, axis=1)
+            shifts = (box_sizes[:,None]*OCT_SHIFT[None,:])*0.5
```
David-Baddeley, line 81: Why the condensing - I don't think this aids clarity ...

PYME/LMVis/layers/octree.py
zacsimile: It looks like `update_data` was deleted in b74fea88e6ac5236c9d526768db4d59c73b87914. Hasn't been an issue.

PYME/LMVis/layers/octree.py
zacsimile: I thought I was clarifying too much with all of the comments I made. I figured this section of code didn't need to explained in such detail and fewer lines of code would give a better visual impression of the weight of its importance.

PYME/LMVis/layers/octree.py
zacsimile: I've recently gotten in the habit of using `super()`. Why is specifically calling the superclass constructor better?

David-Baddeley: Where it really makes a difference is multi-inheritance where the use of `super()` can be ambiguous (or at the very least rather gross), but there are a number of other factors which, IMO, make the explicit form better. These include:
 - different syntax on py2 and py3 making it hard to remember which one to use (or more specifically, the python2  `super(ThisClass, self)` syntax seems needlessly verbose and gross - also needs changing if you change the class name, which seems more likely than changing the base class).
- it doesn't hide what is actually going on under the hood (you are calling the method from the other class on the current object)
- if you change out the base class, it makes you explicitly stop and think about whether the signatures of any over-ridden classes are the same.

This may just be my bias though.

PYME/LMVis/layers/octree.py
David-Baddeley: in general more comments is always better ... just because we don't usually have enough doesn't make it a great idea to change something to visually de-emphasise it.

PYME/LMVis/layers/octree.py
David-Baddeley: My rationale for calling it a problem was having essentially the same code duplicated in different places, which is never a good thing - I suspect I refactored it without realising that it was also used here.

zacsimile:
Latest commit addresses two of the three issues you've brought up. I don't think we need to put `update_data` back. I think this is ready to be merged.

Issue #199, opened 2020-03-29
DBSCAN dialog not displaying most fields on Python 3, OSX


zacsimile:
<img width="291" alt="Screen Shot 2020-03-29 at 7 05 41 PM" src="https://user-images.githubusercontent.com/1263313/77863487-dbaf2600-71f0-11ea-9c08-751a260bfd53.png">

`wxpython 4.04`
`traits 5.1.1`
`traitsui 6.1.1`

```
AttributeError: 'ImageControl' object has no attribute 'GetClientSizeTuple'
Traceback (most recent call last):
  File "/anaconda3/envs/pyme/lib/python3.6/site-packages/traitsui/wx/image_control.py", line 194, in _on_paint
    wdx, wdy = self.GetClientSizeTuple()
```

zacsimile:
This is what I get if I try to click in to the hidden fields. The same `AttributeError` just spools on the screen as I do this.

<img width="289" alt="Screen Shot 2020-03-29 at 7 13 24 PM" src="https://user-images.githubusercontent.com/1263313/77863566-62640300-71f1-11ea-9701-d99dcc0af1c7.png">

David-Baddeley:
This is a traitsui bug.

potential workarounds:

- downgrade wx to the lowest version you can on Py3
- write a custom view to force the use of a different editor for the trait (see comments on #200 )

PR #200, opened 2020-03-29
DBSCAN dialog fix

zacsimile:
Fix for #199.

### DIFF
--- a/PYME/recipes/localisations.py
+++ b/PYME/recipes/localisations.py
@@ -383,7 +383,7 @@ class DBSCANClustering(ModuleBase):
     import multiprocessing
     inputName = Input('filtered')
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Definitely not a fix (unfortunately) - an enum trait evaluates to one of the options (i.e. x, y, or, z) - we need a list of columns. Could be [x,y,z] or [x,y], or [x,y,t] etc ...

i.e. we have to stick with the `ListStr` as the trait type. Most likely workaround would be to write a custom view and to specify a standard text editor as the editor for this trait (rather than a ListEditor). I think I've done this elsewhere (not 100% sure where) as the ListEditor is buggy even on wx <4)

David-Baddeley:
proposed change does not fix problem (GUI displays, but clustering no longer works as expected)

PR #201, opened 2020-03-31
Dbscan gui fix

zacsimile:
The real fix for #199

### DIFF
--- a/PYME/recipes/localisations.py
+++ b/PYME/recipes/localisations.py
@@ -402,6 +402,10 @@ class DBSCANClustering(ModuleBase):
         inp = namespace[self.inputName]
         mapped = tabular.MappingFilter(inp)
 
+        print('How did the columns come out?')
+        print(type(self.columns))
+        print(self.columns)
+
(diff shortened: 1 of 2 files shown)


PYME/recipes/localisations.py
```diff
@@ -426,6 +426,25 @@ def execute(self, namespace):
     def hide_in_overview(self):
         return ['columns']
 
+    @property
+    def default_view(self):
+        import wx
```
David-Baddeley: What is this doing here?

David-Baddeley: IE we shouldn't need to test for wx.App

zacsimile: Can delete. Was following the example you told me to use as a template, `recipes.tablefilters`, line 71.

David-Baddeley: sorry, the ExtractTableChannel one would probably have been a better example. Have just made a few tweaks (and about to make a couple more).

David-Baddeley:
PS - should really document what the pipeline_view is about - if not immediately obvious, this is what gets shown in the pipeline view on the left in VisGUI.

PR #202, opened 2020-04-16
Ripley's K/L functions

zacsimile:


PYME/LMVis/Extras/clusterAnalysis.py
```diff
@@ -93,6 +93,58 @@ def OnClumpDBSCAN(self, event=None):
...
+        pipeline = self.visFr.pipeline
+        recipe = pipeline.recipe
+
+        # Generate a histogram rendering of the dataset as a mask for Ripley's K
+        two_d = np.count_nonzero(pipeline.dataSources[pipeline.selectedDataSourceKey]['z']) == 0
+        renderer = HistogramRenderer(self.visFr, pipeline) if two_d else Histogram3DRenderer(self.visFr, pipeline)
+        imf = renderer.GenerateGUI()
+
+        # Simulate random uniform points
+        x, y, z = locify.locify(imf.image.data, imf.image.pixelSize, 1)
```
David-Baddeley: Looking at how locify works, you almost certainly don't want to use locify with p=1 to generate your mask points. Am currently transferring the mask logic from here to the non-gui code and will fix as I do that.

Locify is likely to be good for doing the MC simulation for confidence intervals.

David-Baddeley:
I've made a bunch of changes - have a quick look. At this point the tests will be broken.

Main changes: 
- move a lot more of the logic re masks out of the GUI and into the low level functions
- change the point-set generation for the mask comparison to remove the un-needed stochasticity brought about by using locify
- make the default (no mask / region points) be to compare against a uniform rectangular region spanning the data rather than comparing against the data itself
- make the recipe module take an optional input mask as an image, rather than a set of locations (this is a much more likely usage scenario)
- change the VisGUI menu option UI:
    - offer both masked and un-masked comparisons (default use case will likely be for people to just look at their data without a mask)
    - change masked comparison to ask the user to specify a previously rendered (and hopefully thresholded / segmented) mask, as generating a mask from the data itself is unlikely to be useful, and histogram rendering is unlikely to give you a high quality mask - the data used to generate the mask should usually come from a separate channel (i.e. a total ER label in your use case) and will likely require some thought to it's segmentation if you want good results.
  - don't inject results into main VisGUI recipe / pipeline - if injected, this will run every time you change anything in the pipeline, but as results are only displayed once using the plot command, this doesn't really make sense.

David-Baddeley:
We might need to think about the units for `mask_area` - I've made them nm^2 / nm^3 but there might be an argument for using um instead

PYME/LMVis/renderers.py
```diff
@@ -327,8 +327,9 @@ class Histogram3DRenderer(HistogramRenderer):
     mode = '3Dhistogram'
 
     def genIm(self, settings, imb, mdh):
-        mdh['Origin.z'] = settings['zBounds'][0]
```
David-Baddeley, line 330: Why this change????? I think the old logic is doing the right thing

zacsimile: I think `Origin.z` is only used in `PYME.IO.image` on line `344`. It looks like this is for backward compatibility. Do we need to create new images with `Origin.z`? I am under the impression that all new images are supposed to use `Camera` metadata.

zacsimile:
I like the changes a lot. The main reason I wanted to include the mask rendering in the recipe was for ease of use. However, it's not unreasonable to expect the user to generate the mask independently before running Ripley's K, and think the code is cleaner. I'm happy with these modifications. Thanks, David!

PYME/LMVis/renderers.py
David-Baddeley: Definitely not just for backwards compatibility. image.origin is used when aligning images (e.g. making composites) either between different rendering modalities and/or different imaging modalities - e.g. a 3D reconstruction with a pre-acqisition widefield stack (potentially in a different colour channel). 

Because 3D PALM/STORM generates localizations above and below the current focal plane, and a reconstruction will naturally include extra planes for these, we can't just use the acquisition pifoc position, but need to record where our reconstruction actually starts.

zacsimile: OK, that makes sense. Why do we have to choose `Origin.x` or `Camera.ROIPosX` then in `image.origin`? Why not just swap `Origin.x`'s value into `Camera.ROIPosX` and call it good?

David-Baddeley:
Because a rendered image might not be over the full camera ROI (e.g. If you
have defined an ROI in Visgui before rendering). Camera.ROIPos is strictly
an acquisition setting, and origin.x takes precedence if defined.

Cheers,
David

PYME/LMVis/renderers.py
David-Baddeley: see reply below - replied to the notification email and it wound up in the wrong place

zacsimile: Yeah, no worries. That makes sense. So after a rendering, `Camera.ROIPosX` is essentially just historical metadata. I think it would helpful to at least add some comments to `image.origin` explaining this. Better still, I might change `Origin.x` to `Renderer.Origin.x`.

zacsimile: Ah, OK. I see it's used in cropping as well. So `Renderer.Origin.x` doesn't strictly make sense. Additional commentary in `image.origin` would suffice.

zacsimile:
> We might need to think about the units for `mask_area` - I've made them nm^2 / nm^3 but there might be an argument for using um instead

I think nm makes sense, especially since pairwise distance histogram bins are all in nm. What is the argument for um?

PR #203, opened 2020-04-16
Adding dye kinetics scripts

b3nroll1ns:
Other than the two new files, there are also minor additions to the deClump script (maximum value aggregator), as well as the corresponding additions to the mergeClumps function. It probably isn't right to add hdbscan to the dye kinetics file, but I wasn't sure where the best place for that is, maybe pyme.localisations with dbscan?


PYME/Analysis/points/DeClump/deClump.c
```diff
@@ -902,6 +902,135 @@ static PyObject * aggregateMin(PyObject *self, PyObject *args, PyObject *keywds)
     return NULL;*/
 }
 
+/*
+Aggregate data into clumps by taking the maximum. This assumes data has been sorted by clumpIndex
```
barentine, line 906: This can also be achieved with a negative sign and aggregateMin. I would suggest if we do merge in this ~as is that you delete the commented out lines which are not comments since this is only adding noise. You could also consider pep8 variable/function names.

PYME/Analysis/points/traveling_salesperson/two_opt_utils.c
```diff
@@ -0,0 +1,8019 @@
+/* Generated by Cython 0.29.13 */
```
barentine, line 1: For some reason I thought it was recommended practice to distribute the c files with cython modules, but I've obviously botched things such that we're going to get 'changes' everytime someone does a development install. With cython as a requirement, maybe we should just stick to the pyx. @David-Baddeley - any thoughts here?

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
+from PYME.recipes.base import register_module, ModuleBase, Filter, OutputModule
+from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, DictStrStr, DictStrList, Either, \
+    ListFloat, ListStr, Button, ToolbarButton, Array, Directory, File
+import numpy as np
+import pylab
```
barentine, line 5: let's try and avoid pylab

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+@register_module('DyeKineticsInit')
+class DyeKineticsInit(ModuleBase):
+    """
+    Module which sets up everything for the actual dye kinetics code.
+    A workaround for the weird way recipes are run when modules are added together without an explicitly defined UI,
+    as in the bakeshop
+
+    Without this, the only UI we see is that of the first module, which is the clump finder in this case. While this isn't
+    100% needed to run the script, user-customizability would be nil, making it harder to use.
+    """
```
barentine, line 24: Can you help me understand what you mean by this? A null execute method is a huge red flag. Also if I recall you've been developing with a fairly outdated version of PYME. David's put a lot into fixing up the pipeline (left side of VisGUI / PYMEVisualize) recipe view / linking that more firmly with the recipe tab in VisGUI. Maybe have a check that this still does what you think/want it to?

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+    min_clump_size = Int(100)
+    clump_column_name = CStr('hdbscan_id')
+    clump_prob_column_name = CStr('hdbscan_prob')
+    clump_dbscan_column_name = CStr('dbscan_id')
+    output_name = Output('hdbscan_clustered')
+
+    def execute(self, namespace):
+
+        inp = namespace[self.input_name]
+        mapped = tabular.mappingFilter(inp)
```
barentine, line 402: `mappingFilter` is deprecated in favor of `MappingFilter` -> you'll want to run test this all again merged into latest PYME head and fix any deprecation warnings

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+
+        with io.open(output_dir + '\\' + 'test_emphist' + timestr + '.json', 'w', encoding='utf8') as outfile:
+            str_ = '{"dk": ' + json.dumps(dk) + '}'
+            outfile.write(to_unicode(str_))
+
+        print('Json has been created and saved to disk')
+
+
+@register_module('HDBSCANClustering')
+class HDBSCANClustering(ModuleBase):
```
barentine, line 372: This should probably be right next to the `localisations.DBSCANClustering` module rather than another file. Perhaps even all clustering modules should be in a new file, recipes/clustering, such that they appear next to each other in the UI without naming them backwards, e.g. `localisations.ClusteringDBSCAN` and `localisations.ClusteringHDBSCAN`. If I recall, @David-Baddeley will prefer not the backwards naming, so maybe this is a good time to move them to a separate file (or maybe no one cares if they aren't next to each other in the UI).

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+    Performs HDBSCAN clustering on input dictionary
+
+    Parameters
+    ----------
+
+        minPtsForCore: The minimum size of clusters. Technically the only required parameter.
```
barentine, line 379: docstrings are only useful if they're correct ;) Maybe update these rather than using the dbscan ones.

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+        powers = np.array([res_on[0][1], res_off[0][1]])
+        times = np.array([cyctime, (max(ds[oncol]).astype(float)+1)*cyctime, (max(ds[offcol]).astype(float)+1)*cyctime])
+
+        output_list = [powers, times, n_on, n_off]
+
+        namespace[self.outputEmphistData] = output_list
+
+@register_module('save_EmpHistJSON')
+class save_EmpHistJson(ModuleBase):
+    """
```
barentine, line 323: I'm assuming you went json here since most tabular types have 1 x N columns only. We do have ragged datastructures, see `PYME.IO.ragged`. Those have a to_json method which might do what you want.

If you want to run this without a GUI, e.g. batch processed, on the cluster, etc., you'd want to use an output module in `recipes/output.py`. Unfortunately we don't have one for json at the moment, which we should maybe consider. That said, you do just have 2 length N columns and a couple constants, and those columns are nothing you'd care to look at in a human-readable uncompressed file, so I would maybe suggest instead going the hdf route and using the metadata to store scalars. There should be an example in the spherical harmonic shell stuff. There are maybe a couple ways to do this, but would be nice to use one of the existing data structures, either in ragged or tabular.

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+                   eimod(res_off[0], np.linspace(1, max(bins_off), max(bins_off) * 10) * cyctime),
+                   colours[3], lw=3, alpha=.8)
+        pylab.ylabel('Events')
+        pylab.xlabel('Event Duration [s]')
+        pylab.ylim((.9, pylab.ylim()[1]))
+        pylab.title('Blink dark times')
+
+        pylab.figtext(.6, .8 - .05, 'tau = %3.4f' % (res_off[0][1]), size=18, color=colours[1])
+
+        pylab.show()
```
barentine, line 312: You don't want to be doing any gui stuff in recipe modules. See report outputs in recipes/outputs.py, the idea being you can use jinja templating to pop plots into htmls.

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+
+        powers = np.array([res_on[0][1], res_off[0][1]])
+        times = np.array([cyctime, (max(ds[oncol]).astype(float)+1)*cyctime, (max(ds[offcol]).astype(float)+1)*cyctime])
+
+        output_list = [powers, times, n_on, n_off]
+
+        namespace[self.outputEmphistData] = output_list
```
barentine, line 319: putting a list as an output doesn't really fit with the recipes data model. Tabular (w/ metata) or ragged would be preferred.

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+        ds.addColumn(self.offTimesColName, off_times)
+
+        namespace[self.outputName] = ds
+
+
+@register_module('FitSwitchingRates')
+class FitSwitchingRates(ModuleBase):
+    """
+    Calculates switching rates based on data which has been run through a blink finding script (trackutils.findtracks2),
+    a clustering script (hdbscan), and a script which calculates all unique on and off states (findblinkstatedurations)
```
barentine, line 230: These docs could stand to be expanded quite a bit -> what's the model they're being fit to, etc. Ideally, you would include a general description of the fit, etc. up top, then type and details for each parameter. Preference is [numpy style](https://www.datacamp.com/community/tutorials/docstrings-python), and for recipe modules its nice to do a 'parameter' for each of the traits attributes.

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+    """
+    inputName = Input('kinetics')
```
barentine, line 232: pep8 naming conventions would be great throughout, like you've done with exposure_time

PYME/LMVis/Extras/kinetics_recipes.py
```diff
@@ -0,0 +1,127 @@
...
+from PYME.recipes.output import HDFOutput
+
+class DyeKinetics(object):
+    def __init__(self, visFr):
+        self.visFr = visFr
+
+        visFr.AddMenuItem('Dye Kinetics', 'Find on and off time durations for molecules', self.dye_kinetics, helpText='')
+
+    def dye_kinetics(self, event):
+        import sys
```
barentine, line 24: docstrings would be great here

PYME/LMVis/Extras/kinetics_recipes.py
```diff
@@ -0,0 +1,127 @@
...
+
+        visFr.AddMenuItem('Dye Kinetics', 'Save empirical histogram JSON file to disk', self.exp_emphist, helpText='')
+
+    def exp_emphist(self, event):
+
+        self.pipeline = self.visFr.pipeline
+
+        recipe = self.visFr.pipeline.recipe
+
+        json_exporter = kin_rcps.save_EmpHistJson(recipe, inputName='emphist_data', outputName='emphist_file', outputdir='C:\\Users\\bdr25\\Desktop\\data\\JSON files')
```
barentine, line 114: check out `PYME.IO.FileUtils.nameUtils` to default to things which will be sane on other peoples computers

PYME/LMVis/Extras/kinetics_recipes.py
```diff
@@ -0,0 +1,127 @@
...
+                                             maximumOffTimeInSecondsToFit=10000.0, offTimeFitBinWidthInSeconds=1)
+
+        if script_init.configure_traits(kind='modal'):
+
+            recipe.trait_set(execute_on_invalidation=False)
+
+            recipe.add_module(script_init)
+
+            recipe.add_module(kin_rcps.FindClumps_DK(recipe, inputName='localisations', outputName='with_clumps',
+                                                   timeWindow=script_init.blinkGapTolerance, clumpRadiusVariable=u'1.0',
```
barentine, line 46: unicode bit here is interesting. Is that necessary?

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+
+        if 'mdh' in dir(inp):
+            mdh = inp.mdh
+            cyctime = mdh.getEntry('Camera.CycleTime')
+        else:
+            cyctime = self.exposure_time
+
+        ds = tabular.mappingFilter(inp)
+
+        off_fit_limit_index = np.round(self.maxOffTimeSecondsToFit/cyctime).astype('int')
```
barentine, line 265: your bin setup / rounding, +1 and then [:-1] trimming is a bit odd and without really following through I'm worried that you might sometimes be slightly off from what you intend (e.g. when you use `bins_on[:1]` in the fit, do you actually want those bin edges, or would you rather use the bin centers?). Maybe have a look at whether this can be simplified, and if not include some comments to make it easier to follow what's going on?

PYME/recipes/kinetics_recipes.py
```diff
@@ -0,0 +1,428 @@
...
+        pylab.xlabel('Event Duration [s]')
+        pylab.ylim((.9, pylab.ylim()[1]))
+        pylab.title('Blink dark times')
+
+        pylab.figtext(.6, .8 - .05, 'tau = %3.4f' % (res_off[0][1]), size=18, color=colours[1])
+
+        pylab.show()
+
+        powers = np.array([res_on[0][1], res_off[0][1]])
+        times = np.array([cyctime, (max(ds[oncol]).astype(float)+1)*cyctime, (max(ds[offcol]).astype(float)+1)*cyctime])
```
barentine, line 315: why the +1?

barentine:
Hi @b3nroll1ns, it'll be cool to have your characterization stuff all sorted/merged. As we discussed this morning there's probably some refactoring to be done. Cleaning some of that up will hopefully help @David-Baddeley look a bit closer at, e.g. what your fit is doing, units, and some of those more critical parts of review. If you have a chance, adding some more to the docstrings especially will help get me/everyone on the same page of what you're trying to do with what incoming units, etc.

PYME/recipes/kinetics_recipes.py
b3nroll1ns: You're right - this is definitely not the right way to call a recipe. I had in it my head this should be called from the dropdown menus, and it still does what I want it to, but with the whole analysis pipeline stored as a recipe it won't be necessary any more

PYME/recipes/kinetics_recipes.py
b3nroll1ns: This was basically just for use with the save_emphistjson function, so I'll probably just get rid of the save_emphist in favor of hdfoutput. For the json output (for simulating), I'll add a traits Bool input rather than an entirely different module

PYME/LMVis/Extras/kinetics_recipes.py
b3nroll1ns: Yeah, since the module can take both integer multiples or 'error_x'

PYME/recipes/kinetics_recipes.py
b3nroll1ns: I do want those bin edges specifically for the fit, since its an integrated fit for histograms. I'll add comments to explain.

PYME/Analysis/points/traveling_salesperson/two_opt_utils.c
barentine, line 1: @b3nroll1ns I was confused earlier - I hadn't actually added this to the repository but see now that you are with this PR. I'm guessing this was an accident since you didn't change the pyx file (and to my knowledge weren't working on this part of the code). The pyx has everything to recompile and we don't want this entire file changing every-time someone does a development install. I'll submit a PR to add this to gitignore to make it easier to avoid in the future.

zacsimile: Another way to avoid this is to only add certain files to the commit. For example, if you change `kinetics_recipes.py`, you can type `git add kinetics_recipies.py`. If you do this individually for each file you modify, you'll never end up adding other files that get generated by Cython, caching, etc. to the commit. Alternatively, if there really is only one extraneous file, you can type `git add .` to add all the files in the folder to the commit and then `git reset -- two_opt_utils.c` to remove the unwanted file from the commit.

PYME/recipes/kinetics_recipes.py
barentine: I think it would be cleanest if you always output something that could be saved using an output module if you want to, but could also be loaded by the simulator. In other words, rather than adding a bool flag to allow someone to ~break the data model, would be nicer if they could just load an hdf into the simulator. I haven't looked at that code, but would that be possible?

PYME/recipes/kinetics_recipes.py
barentine: Should mention an alternative to the reports, since that definitely can take some time to get formatted as you like, etc., is just to move all of this plotting code into the vis gui plugin, so it only gets called when you are already doing gui code. Reports would be 'cooler' and maybe a bit cleaner, but if these are just for sanity checks as you run a recipe and you are unlikely to batch loads of them, visgui pyplot stuff is probably the winner of the cost-benefit.

PYME/recipes/kinetics_recipes.py
b3nroll1ns: Oh yeah that's definitely possible. Should be easy to enough to have it accept both hdf and json files, I'll include that in my next pr

PYME/Analysis/points/traveling_salesperson/two_opt_utils.c
b3nroll1ns: Thanks for the advice, I'll do this going forward!

PYME/recipes/kinetics_recipes.py
b3nroll1ns: The sanity check aspect would be nice, but in the end I definitely want to add the reports, so I'll probably end up doing both

David-Baddeley:
Could you give me a clear high level description reminding me how this differs from / builds on the dye kinetics stuff which is already in PYME (i.e. the stuff under `BleachProfile.kinModels` and `LMVis.photophysics`)? How does it deal with colour / ratiometric signals? This would help put things into context with how it should, e.g. plug into VisGUI. The differences which I can see at a glance would seem to be:

- calculating off times as well as on times
- exporting empirical histograms
- supporting different clustering methods (although this is sort of present in the current version)

From an implementation standpoint: 
- it's pretty easy to recover `tMin` and `tMax` from data coalesced using the existing coalesce functions as `tMin = t - clumpSize/2` and `tMax = t + clumpSize/2`. This would seem much easier than writing and maintaining a new coalescing function. Similarly NPhotons_total can be had as clumpSize*NPhotons.
- having a custom clump finding recipe module, and a custom coalescing module seems like overkill. Couldn't you use the existing FindClumps (which seems identical to your modified version modulo a 2-frame offset on timeWindow) and MergeClumps? If new weightings are strictly needed it might be better to think if the existing MergeClumps could not be adapted - e.g. with an optional weights parameter.
- It's unclear if `FindBlinkingRates` and `FitSwitchingRates` need to be separate modules.
- How best to save the empirical histogram is not entirely clear - as in theory it's a 2D beast and not a good map to any of the existing data types. The rule for datatypes in a recipe is to convert to one of the core data types (tabular or image) if possible, but don't force it if it would be super odd.  This probably falls into the 'super odd' exception. The best way to fix this would be probably be to define an `EmpiricalHistogram` class to be returned by `FitSwitchingRates`, which stored the data and had a`to_JSON()` method than we could call in a generic json output module (which we should probably add). The `to_JSON()` method is a convention we've used elsewhere in PYME for stuff which wants to be json serialised in a particular way (not just with json.dumps).

PR #207, opened 2020-04-23
Deprecation fixes

zacsimile:
Fixes to eliminate some of VisGUI's `DeprecationWarning`s.


David-Baddeley:
Quite a few of these break wx3 compatibility, so shelving for now.

The non-wx changes are mostly OK with a few comments which I'll add shortly

PYME/recipes/base.py
```diff
@@ -733,7 +733,7 @@ def update_from_yaml(self, data):
             with open(data) as f:
                 data = f.read()
     
-        l = yaml.load(data)
+        l = yaml.load(data, Loader=yaml.FullLoader)
```
David-Baddeley: Better to change to `load_safe` - if we are doing an unsafe load we want to be warned about it.

PYME/experimental/labview_spooling_hacks.py
```diff
@@ -30,7 +30,8 @@ def spoof_focus_from_metadata(mdh):
             'StackSettings.NumCycles'], mdh['StackSettings.FramesPerStep'])
         position = np.arange(mdh.getOrDefault('Protocol.PiezoStartPos', 0),
                              mdh.getOrDefault('Protocol.PiezoStartPos', 0) + mdh['StackSettings.NumSteps'] * mdh[
-                                 'StackSettings.StepSize'], mdh['StackSettings.StepSize'])
+                                 'StackSettings.StepSize'], (mdh['StackSettings.StepSize'] 
+                                                             if mdh['StackSettings.StepSize'] else 1))
```
David-Baddeley: What is the rationale here? - it looks like you are just replacing one nonsense value with another nonsense value. If there is really a problem with people giving you bad step sizes it would surely be better to do something like `assert(mdh['StackSettings.StepSize'] > 0)` before this line so that they get an error message that makes sense. As it is you are replacing a value you are making a logical comparison which doesn't make sense (`mdh['StackSettings.StepSize']` is not a bool) and then replacing with a value which is guaranteed to be wrong and hoping for the best.

zacsimile: This change just forces this line to throw an error that the `try` / `except` block is designed to handle (the `assert` fails). Prior to this change, this line would throw a `ZeroDivisionError` or a `ValueError` or a `RuntimeWarning` depending on how numpy felt that particular run. `assert(mdh['StackSettings.StepSize'] > 0)` is a better way of handling this. I can cheat and use `mdh['StackSettings.StepSize']` as a bool here thanks to duck typing, since it's zero or a positive number. It's not pretty, but it's technically OK.

David-Baddeley:
Converted to draft for now so we can come back to it when we are ready to drop wx3 support (tentatively the end of the year), although there would be an argument for splitting the non wx stuff out into something which can be merged straight away.

David-Baddeley:
I've cherry picked all the non-wx4 commits from this, so this may not end up being mergeable (or we might need to cherry-pick the wx commits into a separate branch at some point.

zacsimile:
Do we want to revisit these? Have been been forced onto wx4 for py2 given that wx3, with the exception of the single OSX version maintained by David, is no longer offered in any `conda` channels? Or are we maintaining compatibility for wx3 on OSX?

David-Baddeley:
New installs might have problems getting wx3, but there are a bunch of existing development installs on wx3 that are not subject to condas whims, including the instruments at Auckland, Otago, and Exeter. Don't want to force these to scrub and reinstall from scratch prior to our Jan roadmap date.

PR #210, opened 2020-04-23
Adds a Multiwell tiler and associated gui panel

barentine:
Sets things up to run many circular tilers at preset spacing. Tested with the simulator (didn't think it worth adding the init script for it though).

Currently set up to produce a tilepyramid for each well individually, and not open any tileviewers afterwards. Would be very cool to open up an entire multiwell plate in the viewer, but I think the smart way to do this would be to create a shortcut to building a large tile when you know most of it will be zero's.


David-Baddeley:
There might be an argument for refactoring some of this stuff out into it's own files, but will merge for now.

PR #212, opened 2020-04-24
more thorough OME pixelsize metadata handling and add voxelsize

barentine:
We weren't checking the units on pixelsize from OME metadata, which gave my labmate pixel sizes in micrimeters instead of micrometers
Screen Shot 2020-04-24 at 5 05 31 PM

Loading now works correctly, and uses a class which we might also want to use elsewhere to avoid pixelsize unit confusion

### DIFF
--- a/PYME/IO/MetaDataHandler.py
+++ b/PYME/IO/MetaDataHandler.py
@@ -672,6 +672,7 @@ class XMLMDHandler(MDHandlerBase):
 
 class OMEXMLMDHandler(XMLMDHandler):
     def __init__(self, XMLData = None, mdToCopy=None):
+        from PYME.IO.voxel_size import VoxelSize
         if not XMLData is None:
             #loading an existing file
             self.doc = parseString(XMLData)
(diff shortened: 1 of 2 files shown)


barentine:
@David-Baddeley I think this PR is pretty straightforward and

barentine:
Oops, sorry, started to write I had someone waiting on this and was going to ask for a review, then decided to just have them pull from my fork and somehow accidentally closed the PR. Sorry.

David-Baddeley:
I'm not a great fan of the new pixel size class, as I feel that it adds to the confusion rather than diminishes it (and does a bunch of un-necessary things like writing to metadata, along with duplicating lots of logic). Structurally, I think

- the unit mapping for OME metadata should be part of the OME handler class (although OME uses fairly sensible strings, this is part of the OME specification, and hence belongs in the OME handler, not somewhere globally - keeping file format units local to the file format helps when file formats use something screwy).

- there is no need for an n_units x n_units mapping table - simply convert to one unit (e.g. nm) and then convert back to the desired unit.

- there would be an argument for wrapping the common logic for the different functions into a utility function .e.g.

```python
class OMEXMLMDHandler(...):
   # define what we might see in the file here - this lets us easily add new units as needed for the
   # format, and is a good template, especially if files support units we might not want to support
   # globally - e.g. inches
    _OME_UNITS_TO_UM = {'m' : 1e6, 'mm' : 1e3, 'um' : 1.0, 'nm': 1e-3}
 
    @classmethod
    def _get_pixel_size_um(cls, pix, axis, default=0.1):
        axis = axis.upper()
        try:
            ps = float(pix.GetAttribute('PhysicalSize%s' % axis))
        except:
            logger.errror('No %s pixel size defined, using default' % axis)
            return default
        try:
            ps = ps*cls._OME_UNITS_TO_UM[pix.GetAttribute('PhysicalSize%sUnit' % axis]
        except:
            logger.error(('No units defined for axis %s, defaulting to um' % axis)

        return ps

    def __init__(self, ...):
        #normal stuff

        self['voxelsize.x'] = self._get_pixel_size_um(pix, 'X')
        self['voxelsize.y'] = self._get_pixel_size_um(pix, 'Y')
        self['voxelsize.z'] = self._get_pixel_size_um(pix, 'Z', 0.2) 
        self['voxelsize.units'] = 'um' #this is a courtesy - to define anything else is an error. 

```

David-Baddeley:
I could see an argument for additionally having a separate unit conversion module, e.g. `PYME.misc.units` that had exposed a single `convert_units()` function (potentially also a `convert_voxel_size_units()` function which took a voxelsize named tuple (as defined in `PYME.IO.image.VS`), although it might be possible to parse them both with the same signature. It's also possible that unit conversions are already available through, e.g. numpy or scipy.

David-Baddeley:
To elaborate further on my objections to the new `VoxelSize` object. Whilst a common, unified voxel size would probably be good, there are already a number of VoxelSize class candidates in PYME, and any common VoxelSize class should ideally be a drop-in replacement for them.

Current candidates for a VoxelSize definition are:

- `PYME.IO.image.VS`  - a NamedTuple, accessible as VS.x or VS[0], VS.y or VS[1] etc, units=nm this is what you get whenever you access `ImageStack.voxelsize` and is used in a bunch of places including all the image viewers.
- `PYME.Localisation.autopsf.roifit.VS` - the same as `PYME.IO.image.VS`
- `PYME.DSView.modules.blobMeasure.voxelsize` a 3-tuple (x, y, z) in nm
- `PYME.IO.load_psf.VoxNM` - essentially the same as `PYME.IO.image.VS`, but with a different name. Used in a lot of the deconvolution stuff
- `PYME.IO.DataSources.AlignDataSource.voxelsize` - a 3-tuple, in nm
- `PYME.localization.FitFactories.Interpolators.baseInterpolator._interpolator.setModel(..., voxelsize)` - a 3-tuple, in nm (used by all interpolators and 3D fits)

There are also a bunch of other places which take a 3-tuple in nm. The metadata is really the sole exception to the rule in that the voxelsize  is stored in um rather than nm. 

Although a global voxelsize class would be good, it must both replicate the signature of `PYME.IO.image.VS` and:

- Be lightweight (no computation on access)
- Be in units of nm
- Permit access both as a 3-tuple and using the .x, .y, and .z attributes

We could think about adding a `.units` attribute and a `convert_to_um()` function, but it might be less confusing if we were to accept that a voxelsize object always had units of nm (this means that we don't need to worry about conversion logic in any downstream functions, as we do when reading directly from metadata).

Ideally we'd just change the metadata to force nm on access, but that might be easier said than done, particularly when considering backwards compatibility.

David-Baddeley:
I'm currently adding a `.voxelsize_nm` attribute to the metadatahandler base class and converting usages - hopefully this will clean things up a little.

David-Baddeley:
closed in favour of alternate implementation (to avoid creating a new VoxelSize class). Am still open to some global unit-conversion functions (if not available as part of a 3rd party library).

PR #213, opened 2020-04-26
fix h5rfile appendToTable on python 3 / numpy >=1.14

barentine:
`isinstance(b'anything', six.string_types)` is `False` on py3 because `six.string_types` is just `(str,)`.

This would result in tables calling `numpy.rec.array(b'anything')` which fails (at least on numpy 1.14 and 1.18) due to the req that `numpy.fromstring` have a dtype or format argument.

The work-around was already in appendToTable, we just had to catch bytes on both py2 and py3. The replacement works on py2 and py3 thanks to the short-circuiting or.

Note this fixes one of the errors mentioned in issue #160.

### DIFF
--- a/PYME/IO/h5rFile.py
+++ b/PYME/IO/h5rFile.py
@@ -7,7 +7,6 @@ from PYME import config
 import numpy as np
 import traceback
 import collections
-import six
 
 #global lock across all instances of the H5RFile class as we can have problems across files
 tablesLock = threading.Lock()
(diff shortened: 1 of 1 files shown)


David-Baddeley:
hmm, we should potentially stop to think a bit here - I suspect that this should really just be `isinstance(data, bytes)` rather than including any of the unicode types, as the underlying storage in tables is a c string - i.e. if you are storing unicode you should e.g. encode it as UTF8 first. This might be a historical artefact of how it was converted for Py3 compatibility originally (i suspect all comparisons to `basestring` were replaced with `six.string_types`). If this is the case, accepting unicode types is likely to result in all sorts of nasty errors going forward.

Will have a quick search on the uses and see if I can find anywhere where it gets a unicode object.

David-Baddeley:
Accepting unicode was unsafe, as underlying storage (VLStringAtom) is bytes only. Alternative fix in 597942a

Thanks for finding the issue!

David-Baddeley:
note, ragged array was also broken on py3, and fixed in 597942a

PR #214, opened 2020-04-27
Distance to isosurface recipe

zacsimile:
This calculates the signed distance from points to a generated isosurface in PYME. If used with a tabular data source, points can be colored by signed distance to the surface.

### DIFF
--- a/PYME/LMVis/Extras/extra_layers.py
+++ b/PYME/LMVis/Extras/extra_layers.py
@@ -174,6 +174,20 @@ def save_surface(visFr):
         else:
             raise ValueError('Invalid file extension .' + str(ext))
     
+def distance_to_surface(visFr):
+    from PYME.recipes.surface_fitting import DistanceToSurface
+
+    dist_name = visFr.pipeline.new_ds_name('distance')
+    _, surf_count = visFr.pipeline.new_ds_name('surf', return_count=True)
+    surf_name = 'surf{}'.format(surf_count-1)
+
+    recipe = visFr.pipeline.recipe
+    dts = DistanceToSurface(recipe, input_surface=surf_name, input_points='Localizations', output=dist_name)
+
+    if dts.configure_traits(kind='modal'):
+        recipe.add_module(dts)
+        recipe.execute()
+        visFr.pipeline.selectDataSource(dist_name)
  
 def estimate_density(visFr):
     from PYME.recipes.pointcloud import LocalPointDensity
@@ -289,3 +303,4 @@ def Plug(visFr):
     visFr.AddMenuItem('Mesh', 'Generate Isosurface', lambda e: gen_isosurface(visFr))
     visFr.AddMenuItem('Mesh', 'Load mesh', lambda e: open_surface(visFr))
     visFr.AddMenuItem('Mesh', 'Save mesh', lambda e: save_surface(visFr))
+    visFr.AddMenuItem('Mesh>Analysis', 'Distance to surface', lambda e: distance_to_surface(visFr))
--- a/PYME/LMVis/Extras/extra_layers.py
+++ b/PYME/LMVis/Extras/extra_layers.py
@@ -175,16 +175,16 @@ def save_surface(visFr):
             raise ValueError('Invalid file extension .' + str(ext))
     
 def distance_to_surface(visFr):
-    from PYME.recipes.surface_fitting import DistanceToSurface
+    from PYME.recipes.surface_fitting import DistanceToMesh
     from PYME.experimental._triangle_mesh import TriangleMesh
 
     pipeline = visFr.pipeline
 
     dist_name = visFr.pipeline.new_ds_name('distance')
 
-    surface_names = [k for k, v in pipeline.dataSources.items() if isinstance(v, TriangleMesh)]
+    mesh_names = [k for k, v in pipeline.dataSources.items() if isinstance(v, TriangleMesh)]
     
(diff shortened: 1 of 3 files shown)


PYME/experimental/isosurface.py
```diff
@@ -25,4 +26,125 @@ def isosurface(data, isolevel, voxel_size=None, origin=None, remesh=False):
     if remesh:
         T.remesh()
     
-    return T
\ No newline at end of file
+    return T
+
+def dot(v, w):
```
David-Baddeley: why the redefinition of a bunch of functions which are available in numpy?

PYME/LMVis/Extras/extra_layers.py
```diff
@@ -174,6 +174,20 @@ def save_surface(visFr):
         else:
             raise ValueError('Invalid file extension .' + str(ext))
     
+def distance_to_surface(visFr):
+    from PYME.recipes.surface_fitting import DistanceToSurface
+
+    dist_name = visFr.pipeline.new_ds_name('distance')
+    _, surf_count = visFr.pipeline.new_ds_name('surf', return_count=True)
```
David-Baddeley: This is super gross! It will generate a new unused surface name each time it is called

David-Baddeley: Would be much less gross to look at all the surfaces which are already in the pipeline and parse their names.

zacsimile: Yeah, I'll make a new function in `pipeline` to do this. This isn't the first time I've run into this.

PYME/experimental/isosurface.py
zacsimile: We use this redefinition of `dot` all over the triangle_mesh stuff because it's faster than `np.dot` for vectors of length `(3,)`. I just formalized it here. We don't even use it, though, so I'll delete it.

PYME/LMVis/Extras/extra_layers.py
zacsimile: Wait, will this generate a new unused surface name each time it's called? If I never create the surface using the generated name, I don't think it will.

zacsimile: > Would be much less gross to look at all the surfaces which are already in the pipeline and parse their names.

I think this is what `new_ds_name` does. It just also returns the next name. However, if you don't use it, it will just return the same name/count the next time.

zacsimile: Or are you proposing that for each surface in existence I calculate the distance to that surface?

David-Baddeley: You're right - it does look to see if a surface exists before incrementing the count - I somehow thought the count was global and incremented on each call.

The best solution would probably be to do something like: 
```python
surface_names = [k for k, v in pipeline.dataSources.items() if isinstance(v, TriangleMesh)]
surf_name = wx.SingleChoiceDialog(surface_names)
```
potentially short-cutting if there is only one surface.

Relying on surface names following a `surfxx` pattern might be optimisitic.

PYME/experimental/isosurface.py
```diff
@@ -25,4 +26,71 @@ def isosurface(data, isolevel, voxel_size=None, origin=None, remesh=False):
     if remesh:
         T.remesh()
     
-    return T
\ No newline at end of file
+    return T
+
+def clamp(v, lo, hi):
```
David-Baddeley: This also has a numpy equivalent

PYME/LMVis/Extras/extra_layers.py
zacsimile: This is just an initial guess for a dataSource name, and not a bad one given how a user is expected to generate a surface (the surface name will follow the `surfxxx` pattern). The user selects the data source in `dts.configure_traits(kind='modal')` on line 187. I'm not sure creating a `wx.SingleChoiceDialog` just before the traits configuration makes sense. Is there a way to limit the list of choices for `input_surface` in `configure_traits` to instances of `TriangleMesh`?

PYME/experimental/isosurface.py
zacsimile: Ah, `np.clip`. I didn't know this existed.

PYME/LMVis/Extras/extra_layers.py
David-Baddeley: Restricting the choices might not be trivial, but is there any actual need to call `configure_traits` - the `DistanceToSurface` module doesn't have any parameters other than input and output names - we could use the current selected data source as the input points, prompt for the mesh, and generate a name for the output using `new_ds_name()`  - I suspect this would result in a better / easier UI overall.

zacsimile: OK, I'm cool with that solution. Will push a commit changing to this shortly.

PYME/recipes/surface_fitting.py
```diff
@@ -201,4 +201,31 @@ def execute(self, namespace):
             #target_length = np.mean(surf._halfedges['length'][surf._halfedges['length'] != -1])
             surf.remesh(5, l=0.5, n_relax=10)
         
-        namespace[self.output] = surf
\ No newline at end of file
+        namespace[self.output] = surf
+
+@register_module('DistanceToSurface')
+class DistanceToSurface(ModuleBase):
```
David-Baddeley: I wonder if it might not be better to call this `DistanceToMesh` to make it a bit clearer what the input should be. There are multiple ways of defining/parameterising a surface, with a mesh being only one of them.

zacsimile: Addressed

Issue #219, opened 2020-05-05
Ripley's analysis is not working


lukasfue:
I tried testing the new Ripley's analysis using the default parameter values and got this traceback:

Traceback (most recent call last):
  File "c:\users\laf62\python-microscopy\PYME\LMVis\Extras\clusterAnalysis.py", line 411, in OnRipleys
    result = recipe.apply(inputPositions=pipeline.selectedDataSource, inputMask=mask)
AttributeError: 'ModuleCollection' object has no attribute 'apply'

barentine:
Looks like that `apply()` should be an `execute()`. Thanks @lukasfue.

barentine:
oops, `recipe` should be `r`. Anyway, we'll take care of it

zacsimile:
I'm taking care of it now. Almost there, I think. There were a few other slight bugs after fixing `apply`.

David-Baddeley:
Apply is correct, not execute

zacsimile:
`apply` is correct for the recipe, not for the `ModuleCollection`. I switched to using `apply` on the recipe (it was previously on the `ModuleCollection`). Fixes are coming shortly. I'm just testing them on real data right now to make sure they work.

zacsimile:
I pushed fixes in #220 that make this run now (tested on 2D and 3D data/masks). I'm not 100% sure the graphs coming out make sense. Will need to test on some more data structures/more masks to confirm. We are passing tests for `PYME.analysis.points.ripleys.ripleys_k_from_mask_points()`, so at least that part should be correct.

zacsimile:
This was fixed with the merge of #220. I think we can close.

PR #220, opened 2020-05-05
Ripleys fix

zacsimile:
Fixes for #219

### DIFF
--- a/PYME/Analysis/points/ripleys.py
+++ b/PYME/Analysis/points/ripleys.py
@@ -76,7 +76,8 @@ def points_from_mask(mask, sampling, three_d = True, coord_origin=(0,0,0)):
     vx, vy, vz = mask.voxelsize
     x0_m, y0_m, z0_m = mask.origin
     x0_p, y0_p, z0_p = coord_origin
-    stride_x, stride_y, stride_z = [max(1, int(sampling / v)) for v in [vx, vy, vz]]
+    safe_div = lambda x, y: int(x/y) if y > 0 else 0
+    stride_x, stride_y, stride_z = [max(1, safe_div(sampling, v)) for v in [vx, vy, vz]]
     
     if three_d:
         #convert mask to boolean image
--- a/PYME/Analysis/points/ripleys.py
+++ b/PYME/Analysis/points/ripleys.py
@@ -137,17 +137,22 @@ def ripleys_k(x, y, n_bins, bin_size, mask=None, bbox=None, z=None, threaded=Fal
     else:
         if three_d:
             if not bbox:
-                bbox = [x.min(), y.min(), z.min(), x.max(), y.max(), z.max()]
+                bbox = np.array([x.min(), y.min(), z.min(), x.max(), y.max(), z.max()])
                 
             xu, yu, zu = np.mgrid[bbox[0]:bbox[3]:sampling, bbox[1]:bbox[4]:sampling, bbox[2]:bbox[5]:sampling]
             mask_area = np.prod((bbox[3:] - bbox[:3]))
         else:
             if not bbox:
-                bbox = [x.min(), y.min(), x.max(), y.max()]
+                bbox = np.array([x.min(), y.min(), x.max(), y.max()])
     
-            xu, yu = np.mgrid[bbox[0]:bbox[3]:sampling, bbox[1]:bbox[4]:sampling]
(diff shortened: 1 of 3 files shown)


PYME/LMVis/renderers.py
```diff
@@ -328,7 +328,7 @@ class Histogram3DRenderer(HistogramRenderer):
 
     def genIm(self, settings, imb, mdh):
         mdh['Origin.z'] = settings['zBounds'][0]
-        return visHelpers.rendHist3D(self.colourFilter['x'],self.colourFilter['y'], self.colourFilter['z'], imb, settings['pixelSize'], settings['zBounds'], settings['zSliceThickness'])
```
David-Baddeley, line 331: Why this change???? We should respect the z bounds that we set in the  generate3D dialog

zacsimile: Because `visHelpers.rendHist3D` doesn't accept a zbound argument. It has a max of 6 arguments.

David-Baddeley: oops - that gets handled in `generate()`, although `generate()` sets `'Origin.z'` incorrectly.

PYME/Analysis/points/ripleys.py
```diff
@@ -76,7 +76,8 @@ def points_from_mask(mask, sampling, three_d = True, coord_origin=(0,0,0)):
     vx, vy, vz = mask.voxelsize
     x0_m, y0_m, z0_m = mask.origin
     x0_p, y0_p, z0_p = coord_origin
-    stride_x, stride_y, stride_z = [max(1, int(sampling / v)) for v in [vx, vy, vz]]
+    safe_div = lambda x, y: int(x/y) if y > 0 else 0
+    stride_x, stride_y, stride_z = [max(1, safe_div(sampling, v)) for v in [vx, vy, vz]]
```
David-Baddeley: this shouldn't be needed - `v==0` is an error, and should cause a failure.  We could put an `assert(np.all(np.array(mask.voxelsize)>0)))` in before assigning vx, vy, and vz

David-Baddeley: I guess `vz` could legitimately be 0 if `three_d=False`

zacsimile: Correct. And that's a situation I ran into, hence this fix.

David-Baddeley: but suspect the correct way of handling this would be to test for the special case rather than replacing a division with a `safe_div()` which is going to cause us to silently get garbage results

zacsimile: What case would silently give us garbage results?

David-Baddeley: ie:

```python
if (vz < eps) and not three_d:
    vz = 1 #dummy value to prevent div by zero when calculating strides we don't use

assert ((vx>0) and (vy>0) and (vz > 0))
```

where `eps` is something small (because vz is a float we can't test `vz==0`).

David-Baddeley: or `if not (vz > 0):`

zacsimile: OK, that is cleaner. I'm cool with changing to that. Why would `safe_div` silently give us garbage results, though?

David-Baddeley: The garbage results case would be when someone loads an image to use as a mask and neglects to define the pixel size. Examples:

- using a different modality (e.g. widefield/confocal/STED) to define the mask
- manually manipulating the image to generate a mask in the shell / an IPython notebook and not keeping track of metadata etc ... 
- (maybe) 3D comparison to 2D mask
- any bug which causes us to drop metadata

I guess my gut feeling before thinking about the 2D case was that something else must have gone wrong to get us to this point.

zacsimile: I think if the voxelsize was undefined, `save_div` would still fail, and we would not get garbage results.

```
>>> None > 0
  File "<stdin>", line 1, in <module>
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

I do think your proposed solution is cleaner, since it directly addresses the `vz==0` case in 2d, which is what I wanted to fix. I've pushed a commit changing to this method of testing.

David-Baddeley: 'Undefined' for the voxelsize defaults to either 0 or -1 not `None` (depending on how things are loaded)

PR #221, opened 2020-05-05
Animation fixes for py3

zacsimile:
Some animation fixes for py3


zacsimile:
Hang on it pushed some unwanted commits. I must've branched incorrectly.

PR #222, opened 2020-05-06
Animation fixes for py3

zacsimile:


PYME/LMVis/views.py
```diff
@@ -153,6 +153,14 @@ def __div__(self, scalar):
                     self.clip_plane_position / scalar
                     )
 
+    def __floordiv__(self, scalar):
+        # Integer division py3
+        return self.__div__(scalar)
+
+    def __truediv__(self, scalar):
+        # Float division py3
```
barentine, line 161: pretty sure you can `//` on py2 and py3, no?

barentine: Just note that normal int-> float promotion still applies
```
(py27) Andrews-MacBook-Pro-5:~ Andrew$ python
Python 2.7.16 |Anaconda, Inc.| (default, Sep 24 2019, 16:55:38) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
>>> 2.5 // 2
1.0
>>> 3 // 2
1
>>> quit()
(py27) Andrews-MacBook-Pro-5:~ Andrew$ conda deactivate
Andrews-MacBook-Pro-5:~ Andrew$ conda activate base
(base) Andrews-MacBook-Pro-5:~ Andrew$ python
Python 3.6.9 |Anaconda, Inc.| (default, Jul 30 2019, 13:42:17) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
>>> 2.5 // 2
1.0
>>> 3 // 2
1
>>> 

```

zacsimile: Yes, but `/` is used in the current code. In py2, '/' defaults to `__div__`, in py3 it defaults to `__truediv__`. So, when we do `/` in py3, it throws an error, because it checks `__truediv__`, which is isn't overloaded.

PYME/LMVis/views.py
```diff
@@ -153,6 +153,14 @@ def __div__(self, scalar):
                     self.clip_plane_position / scalar
                     )
 
+    def __floordiv__(self, scalar):
```
David-Baddeley: what exactly is the bug this is trying to fix??? In any case I think you've got the fix wrong. Everything here should be float, including scalar, so there is no way you should be casting scalar to an int. The only potential change I could see (which is more about ensuring correct behaviour on py2 than fixing anything on py3) would be to cast scalar to a float in `__div__`.

PYME/LMVis/Extras/animation.py
```diff
@@ -57,7 +57,7 @@ def __init__(self, parent_panel, **kwargs):
         self.next_view_id = 0
 
     def create_buttons(self, vertical_sizer):
-        grid_sizer = wx.GridSizer(3, 3)
+        grid_sizer = wx.GridSizer(3, 3, 3)
```
David-Baddeley: should be `wx.GridSizer(3,3,0,0)` or `wx.GridSizer(rows=3, cols=3, vgap=0, hgap=0)`to match both wx3 and wx4 signatures (I half expect the wx4 signature to change as it's currently pretty broken).

PYME/LMVis/views.py
David-Baddeley: I see, `__div__` doesn't exist on py3. For the View object we have no need to define `__floordiv__` (and so should leave it undefined). `__truediv__` should be identical to `__div__` - ie just a line under the definition of `__div__` saying `__truediv__ = __div__`. Could potentially also add a cast to float in `__div__` to make things safer on py2 .

PYME/LMVis/Extras/animation.py
```diff
@@ -253,7 +253,7 @@ def refill(self):
         self.clear_view()
         for snapshot in self.snapshots:
             index = len(self.snapshots)
-            self.view_table.InsertStringItem(index, snapshot.view_id)
+            self.view_table.InsertItem(index, snapshot.view_id)
```
David-Baddeley: This breaks wx3 compatibility (which we are maintaining until at least Jan 2021)

PYME/LMVis/views.py
zacsimile: I added `__floordiv__` just in case. I think it is supposed to be integer division. If you look in `animation.py` on line 196, it will default to integer division in py2. Therefore I think we do want to cast to int, not float.

PYME/LMVis/Extras/animation.py
zacsimile: The wx4 signature certainly accepts both (https://wxpython.org/Phoenix/docs/html/wx.GridSizer.html). No idea about wx3.

PR #229, opened 2020-05-07
Enhanced MATLAB loader

zacsimile:
Generalized MATLAB loading, which allows arbitrary mappings of variable names.


zacsimile:
That last commit should say multicolor, not two-color

PYME/LMVis/importTextDialog.py
```diff
@@ -255,4 +273,33 @@ def GetFieldNames(self):
 
     def GetVarName(self):
         return self.cbVarNames.GetValue()
-        
\ No newline at end of file
+        
+class ImportMatlabDialog(ImportDialog):
+    # MATLAB importer with variable mapping
+    def __init__(self, parent, matFileName):
+        wx.Dialog.__init__(self, parent, title='Import data from MATLAB file')
```
David-Baddeley: This should be in `ImportDialog`

PYME/LMVis/importTextDialog.py
```diff
@@ -31,22 +31,33 @@ def isnumber(s):
...
     requiredVariables = {'x':'x position [nm]',
                         'y':'y position [nm]'}
     recommendedVariables = {'A':'amplitude of Gaussian',
                             't':'time [frames]',
                             'sig':'std. deviation of Gaussian [nm]',
                             'error_x':'fit error in x direction [nm]'}
     niceVariables = {'error_y':'fit error in y direction [nm]'}
-
-    def __init__(self, parent, textFileName):
-        wx.Dialog.__init__(self, parent, title='Import data from text file')
```
David-Baddeley, line 46: Really no excuse for not calling `wx.Dialog.__init__()` here.

could add a virtual method, e.g.

```python

def _parse_header(self, file):
    raise NotImplementedError

```

which can be over-ridden in derived classes

PYME/LMVis/importTextDialog.py
```diff
@@ -31,22 +31,33 @@ def isnumber(s):
     except:
         return False
     
-
-
-class ImportTextDialog(wx.Dialog):
+class ImportDialog(wx.Dialog):
```
David-Baddeley: Would be better to call this 'ColumnMappingDialog` or similar - `ImportDialog` is a bit too generic

PYME/IO/tabular.py
```diff
@@ -535,7 +535,7 @@ def getInfo(self):
 class MatfileColumnSource(TabularBase):
     _name = "Matlab Column Source"
     
-    def __init__(self, filename):
+    def __init__(self, filename, columnnames=None):
```
David-Baddeley: The column mapping stuff should be in a subclass of `MatfileColumnSource` or better yet, just load using the old/existing `MatfileColumnSource` and then use a `tabular.MappingFilter` to re-map column names. The rationale for the second approach is to separate VisGUI loading (which requires magic column names) from more general loading of tabular data which really doesn't care what the column names are.

PYME/IO/tabular.py
```diff
@@ -546,6 +546,29 @@ def __init__(self, filename):
         self.res = scipy.io.loadmat(filename)  # TODO: evaluate why these are cast as floats
         
         self._keys = [k for k in self.res.keys() if not k.startswith('_')]
+
+        # Check for multicolor
```
David-Baddeley: I'd be inclined to make the multicolour (cell array) stuff it's own data source.

David-Baddeley: Although not as convinced on this one as some of the other comments

PYME/LMVis/importTextDialog.py
```diff
@@ -255,4 +273,33 @@ def GetFieldNames(self):
...
+
+        self.colNames, self.dataLines = self.MatlabHeaderParse(matFileName)
+
+        ImportDialog.__init__(self, parent)
+
+    def MatlabHeaderParse(self, filename):
+        from scipy.io import loadmat
+        import numpy as np
+
+        mf = loadmat(filename)
```
David-Baddeley: this would be the 3rd time we load the .mat file. Would be sensible to change the function signature to accept either a filename or a loaded file.

```python
def _parse_header(self, file):
    if isinstance(file, six.string_types):
        mf = loadmat(file)
    else:
        #assume file already loaded
        mf = file
````

File: PYME/LMVis/visCore.py [hunk 7]
```diff
@@ -672,21 +672,40 @@ def _populate_open_args(self, filename):
             from scipy.io import loadmat
         
             mf = loadmat(filename)
-            if not 'x' in mf.keys():
```
David-Baddeley, line 675:
```python
if (len([k for k in mf.keys() if not k.startswith('_')]) < 3) and not 'x' in mf.keys():
    # all the data is probably packed in a single variable
    dlg = importTextDialog.ImportMatDialog(self, [k for k in mf.keys() if not k.startswith('__')])
    ret = dlg.ShowModal()
   # rest of single var logic ....
elif not 'x' in mf.keys() and  not 'x' in mf.keys():
    importMatlabDialog(mf)
    # rest of key mapping logic
```

PYME/LMVis/pipeline.py
```diff
@@ -674,7 +674,10 @@ def _ds_from_file(self, filename, **kwargs):
                 #old style matlab import
                 ds = tabular.MatfileSource(filename, kwargs['FieldNames'], kwargs['VarName'])
             else:
-                ds = tabular.MatfileColumnSource(filename)
```
David-Baddeley, line 701: Much of the new logic in the modified `MatfileColumnSource` could be replaced with the following.

```python
ds = tabular.MatfileColumnSource(filename)

field_names = kwargs.get('FieldNames', None)
if field_names:
    ds = tabular.MappingFilter(ds, **{new_field : old_field for new_field, old_field in zip(field_names, ds.keys())})
```

PYME/IO/tabular.py
```diff
@@ -546,6 +546,29 @@ def __init__(self, filename):
...
+            n_channels = test_shape[1]
+            if columnnames is None:
+                columnnames = self._keys
+
+        # Mapping for multicolor, channels
+        if columnnames is not None:
+            tmp_res = {}
+            for i, k in enumerate(self._keys):
+                if multicolor:
+                    tmp_res[columnnames[i]] = np.hstack([self.res[k].squeeze()[_i].squeeze() for _i in range(n_channels)])
```
David-Baddeley: how do the cell arrays actually map to python? As a list? Shouldn't it be possible to just do `np.hstack(self.res[k])`

David-Baddeley: Note that as written it will break for multicolour but with sensible variable names.

zacsimile: They map as an array of arrays. I thought this would be easier, too. Every time I tried sensible stuff, like what you suggest, it would return as an array of size `(2,)` (number of channels) instead of the size `(summed_number_of_elements)`. It crossed my mind that this might be a bug in the particular version of `numpy` I'm using, but this solution with a bunch of `squeeze()`s does work, and it's not that terrible, so I didn't push looking into it.

zacsimile: I take it back. Just calling `np.vstack(self.res[k])` seems to work. Must've been something foolish I was doing in the single-line loop.

PYME/LMVis/importTextDialog.py
zacsimile: You are right. My excuse was, "it works," but it's a sloppy thing to do.

zacsimile: Also, I couldn't think of a good way to change the dialog box title at the moment I was coding this, so I thought it'd be easier to leave in each subclassed function. It's actually super easy to change the dialog box title. Not sure where my head was.

PYME/LMVis/visCore.py
zacsimile: Adding the `len([k for k in mf.keys() if not k.startswith('_')]) < 3)` is necessary. What's the argument for changing the nested if structure to an `elif`? The nested structure allows us to do a single check for the presence of `x` instead of two checks, and I think makes a more readable decision tree for the next person to come look at this section of code. I've commented what each conditional statement is "thinking".

zacsimile: Also, I think `'x' not in mf.keys()` is clearer than `not 'x' in mf.keys()`, but you seem to have a preference for the latter. I assumed it was a legacy preference, but you reiterated here. Is there a reason for this style? Should I switch to using this style?

PYME/LMVis/pipeline.py
zacsimile: Yep. Changed.

PYME/LMVis/importTextDialog.py
zacsimile: Yep. Changed.

PYME/IO/tabular.py
zacsimile: Nope, I was right. That doesn't work. However, `np.vstack(self.res[k][0])` works.

zacsimile:
Thanks for your comments, @David-Baddeley! They were super helpful. I've made (almost) all of the suggested changes, and the loading is much cleaner.

PYME/LMVis/visCore.py
zacsimile: I guess the nested vs. flat comes down to code style. In general, I agree a flat `if` statement is better than a nested `if` statement. However, I think a nested `if` statement is better than an `elif` statement without an `else`. Dangling `elif`s bother me, but that's just a personal preference. If you feel strongly, we can change it.

PYME/LMVis/importTextDialog.py
```diff
@@ -31,21 +31,26 @@ def isnumber(s):
...
-class ImportTextDialog(wx.Dialog):
+class ColumnMappingDialog(wx.Dialog):
     requiredVariables = {'x':'x position [nm]',
                         'y':'y position [nm]'}
     recommendedVariables = {'A':'amplitude of Gaussian',
                             't':'time [frames]',
                             'sig':'std. deviation of Gaussian [nm]',
                             'error_x':'fit error in x direction [nm]'}
     niceVariables = {'error_y':'fit error in y direction [nm]'}
+    colNames = []
```
David-Baddeley: why the class level empty lists here?  Surely these should be instance variables? I'd think it should be enough to just assign to `self.colNames` and `self.dataLines` in the constructor  / `_parse_header()`. Will delete if no objections.

zacsimile: I think I left this in here just so the class functions would all run without error if there was no derived class.  `CheckColNames`, for example, has a reference to self.colNames. It's not really important for them to be there. No objections to deleting.

PYME/LMVis/importTextDialog.py
```diff
@@ -31,21 +31,26 @@ def isnumber(s):
...
+        """
+        Dialog box for importing data source with arbitrary column names.
+        """
 
-    def __init__(self, parent, textFileName):
-        wx.Dialog.__init__(self, parent, title='Import data from text file')
+        wx.Dialog.__init__(self, parent, title='Import data from {} file'.format(self.fileType))
 
-        self.colNames, self.dataLines = self.TextFileHeaderParse(textFileName)
+        self._parse_header(fileName)
```
David-Baddeley:
```suggestion
        self.colNames, self.dataLines = self._parse_header(fileName)
```

PYME/LMVis/importTextDialog.py
```diff
@@ -205,8 +173,51 @@ def GetNumberComments(self):
...
+        
+        #print commentLines
+        
+        #print commentLines[-1].split(delim), len(commentLines[-1].split(delim)), numCols
+
+        if len(commentLines) > 0 and len(commentLines[-1].split(delim)) == numCols:
+            colNames = commentLines[-1].split(delim)
+        else:
+            colNames = ['column_%d' % i for i in range(numCols)]
+
```
David-Baddeley, line 214:
```suggestion

return colNames, dataLines
```

Issue #227, opened 2020-05-06
Convert numbers in PYMEVisualize View Clipping to editable text boxes


zacsimile:
Feature request to allow users to set the min/max manually as well as with the slider. This would be in addition to double-clicking to set Clipping range, we still want to keep that.

David-Baddeley:
I'm concerned about the amount of screen real estate this would take if we were to add standard text boxes. It feels like this should potentially be something like an 'advanced clipping settings' popout or similar - not sure how to make it easy to use and discoverable. Alternative is to further develop the custom slider control so that, e.g. you could double click on the numbers (but developing custom UI widgets is a questionable use of resources at the best of times).

zacsimile:
I think it will only add a little more width to the left-hand bar, which would be fine. We can safely assume in this day and age people will use PYME full-screen, in which case a little extra screen real estate in exchange for easy clipping control will be welcome. I think the ease-of-use benefits outweigh the screen real estate loss.

David-Baddeley:
I never use it full screen :p. I also have a habit of putting a VisGUI window and a rendered image window side by side (meaning that you only have half the screen for the VisGUI window). 

If we use standard controls, we'd end up with around 50% of the width of the side bar being used by the text boxes and we'd probably double or triple the height of the view clipping stuff (the controls are particularly large on linux, which has to do with the default font sizes, but as we need 5 figures for each of the upper and lower bounds in a typical image they will be large on any platform).

zacsimile:
I think there's a way to make this work. Vutara seems to have this functionality without taking up too much screen real estate. Maybe I'll just try it and see what happens. I bet there's a way to shrink the font/box size, which would help a lot. It looks like `wx.TextCtrl` has a `wxSize` property where you can set the width and height of the box and a `style` argument, which you can use to set the font size. Both of these arguments are available in wx3 and wx4. Those should be a good start.

David-Baddeley:
It'll be very hard to do whilst maintaining some semblance of consistency / having things work across platforms. I'd strongly advise against making the vertical size anything other than -1 (auto) - changing the font size should change the vertical size to match. I doubt it'll be OK in terms of screen real estate / general look and feel with the cropping in it's current location whatever you do.

The two solutions I had thought of were:

[...]

PR #233, opened 2020-05-10
Text control of view clipping limits

zacsimile:
Re #227, implementing David's first suggested solution.

### DIFF
--- a/PYME/LMVis/view_clipping_pane.py
+++ b/PYME/LMVis/view_clipping_pane.py
@@ -344,7 +344,9 @@ class ClippingPanel(wx.Panel):
 
     def OnTextCtrlKeyPress(self, event):
         text_ctrl = event.GetEventObject()
-        if text_ctrl.HasFocus() and (event.GetKeyCode() == wx.WXK_RETURN):
+        key = event.GetKeyCode()
+        focus = text_ctrl.HasFocus()
+        if focus and (key == wx.WXK_RETURN):
             if text_ctrl.GetId() == self.ll_ctrl.GetId():
                 self.view_limits[0] = float(text_ctrl.GetValue())
             elif text_ctrl.GetId() == self.ul_ctrl.GetId():
@@ -352,22 +354,15 @@ class ClippingPanel(wx.Panel):
             text_ctrl.Hide()
             self.Refresh()
             self.Update()
+        elif focus and (key == wx.WXK_ESCAPE):
+            # We're done
+            text_ctrl.Hide()
         event.Skip() 
 
 
 class ViewClippingPanel(wx.Panel):
-    """A GUI class for determining the settings to use when displaying points
-    in VisGUI.
-
-    Constructed as follows:
-    PointSettingsPanel(parent, pipeline, pointDisplaySettings)
-
-    where:
-      parent is the parent window
-      pipeline is the pipeline object which provides the points,
-      pointDisplaySettings is an instance of PointDisplaySettings
-
-
+    """A GUI class for dynamically filtering points 
+    spatially by adjusting sliders.
     """
     
     def __init__(self, parent, glcanvas):
--- a/PYME/LMVis/view_clipping_pane.py
+++ b/PYME/LMVis/view_clipping_pane.py
@@ -354,6 +354,7 @@ class ClippingPanel(wx.Panel):
             text_ctrl.Hide()
             self.Refresh()
             self.Update()
+            self.glcanvas.refresh()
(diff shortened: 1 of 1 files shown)


zacsimile:
Everything works as expected, except I get a "boop" noise from my laptop on key press events (return/esc). It runs fine, but still makes that error noise like I've pressed a key when I'm not supposed to. Haven't been able to figure out why yet.

PYME/LMVis/view_clipping_pane.py
```diff
@@ -308,20 +329,41 @@ def OnMouseMove(self, event):
...
+        text_ctrl.SetValue(str(value))
+
+        # Display
+        text_ctrl.Show()
+        text_ctrl.SetFocus()
+
+    def HideTextCtrl(self, event):
+        event.GetEventObject().Hide()
+
+    def OnTextCtrlKeyPress(self, event):
```
David-Baddeley: Could you bind to `EVT_TEXT_ENTER` instead? (This wouldn't handle escape but would result in less boiler plate). Merely a thought at this point - you don't need to change it.

PYME/LMVis/view_clipping_pane.py
```diff
@@ -231,14 +244,22 @@ def OnMouseScrollEvent(self, evt):
...
     def OnLeftDown(self, event):
         x = event.GetX()
-        #y = event.GetY()
+        y = event.GetY()
 
         x0p, wp, x0vp, xmvp, w, xo = self._get_coords()
+        maxy = self.Size[1] - self.ll_ctrl.Size[1]/2 + 2
     
         #hit test the limits
+        # check for clipping update via text control
```
David-Baddeley, line 262: It's not very clear where you are checking for hits here ...

David-Baddeley: A comment describing where you are checking would be useful

David-Baddeley: Also, to be consistent with buttons and other clickable controls, we should be showing the text control on the mouse button release (EVT_LEFT_UP) rather than the press, but only if we were not dragging the bounds.

zacsimile: Sure thing. On the dragging front: I originally wanted to put the test for displaying the textbox after the dragging, but I'll have to make sure `y < self.Size[1] - self.ll_ctrl.Size[1]` on each of the dragging tests. Which I suppose is  about as acceptable as what I have here.

PYME/LMVis/view_clipping_pane.py
zacsimile: I thought about this. I'm under the impression that `EVT_TEXT_ENTER` would force a change every time the user typed a new number in the box. I thought this would mean handling re-drawing and re-positioning the text box every time the user updated the value. Does it only bind the enter key and/or do I just bind it and skip the re-drawing?

David-Baddeley: As far as I recall, `EVT_TEXT` fires whenever the value changes (i.e. on every character) whereas `EVT_TEXT_ENTER` fires when you press enter or the control looses focus. I might have got the names wrong, but there are definitely 2 separate events with one doing roughly what you'd want.

David-Baddeley:
This control is the bastard child of the `histLimits` control used in the filters / display settings and it would be nice to have the functionality there too. Unfortunately I think I just copied and pasted and modified so they don't share a common base. Doesn't need to be done straight away, but something to keep in the back or our minds.

PYME/LMVis/view_clipping_pane.py
zacsimile: Yes, I just found this in the `wx` docs. I must've been confused or looking at `EVT_TEXT` the last time I looked at this. Thanks for pointing it out.

zacsimile:
I see what you mean. These are fairly close. No reason we can't knock this out now. What do we call the base class? `PaintPanel`?

zacsimile:
I've addressed your first two comments. A few more thoughts below.

- I think maybe I should be firing a `LimitChangeEvent` when we enter text for consistency, although it's not necessary from a functional standpoint. 
- We could make a superclass for `histLimits` and `view_clipping_pane` and extend this functionality to histLimits.
- We could address the problem of overlapping text when the `view_limits` get too close.

barentine:
> `histLimits` control used in the filters / display settings and it would be nice to have the functionality there too

I'm glad to see this being discussed - would be happy to see this functionality everywhere, including e.g. PYMEAcquire.

PR #239, opened 2020-05-11
organize PYMEVisualize open/save into open/open extra/save

barentine:
- move extra open items into a submenu
- move all save items into a submenu
- move snapshot save from 'View' to 'File>Save'
- consistent CapsCase on 'File' menu items

Screen Shot 2020-05-11 at 12 31 20 PM
Screen Shot 2020-05-11 at 12 30 21 PM

### DIFF
--- a/PYME/LMVis/Extras/snapshot.py
+++ b/PYME/LMVis/Extras/snapshot.py
@@ -63,4 +63,4 @@ def save_snapshot(canvas):
 
 
 def Plug(vis_fr):
-    vis_fr.AddMenuItem('View', 'Save Snapshot', lambda e: save_snapshot(vis_fr.glCanvas))
+    vis_fr.AddMenuItem('File>Save', 'View Snapshot', lambda e: save_snapshot(vis_fr.glCanvas))
--- a/PYME/LMVis/visCore.py
+++ b/PYME/LMVis/visCore.py
@@ -255,12 +255,12 @@ class VisGUICore(object):
         if not subMenu:
             self.AddMenuItem('File>Open Extra', "Raw/Prebleach Data", self.OnOpenRaw)
             self.AddMenuItem('File>Open Extra', "Channel", self.OnOpenChannel)
-            
+
+        self.AddMenuItem('File', itemType='separator')
         self.AddMenuItem('File>Save', 'Filtered Localizations', self.OnSave)
         
         if not subMenu:
-            # self.AddMenuItem('File', itemType='separator')
-            self.AddMenuItem('File>Save', "Save Measurements", self.OnSaveMeasurements)
+            self.AddMenuItem('File>Save', "Measurements", self.OnSaveMeasurements)
 
             self.AddMenuItem('File', itemType='separator')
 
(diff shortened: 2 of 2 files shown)


PYME/LMVis/Extras/snapshot.py
```diff
@@ -63,4 +63,4 @@ def save_snapshot(canvas):
 
 
 def Plug(vis_fr):
-    vis_fr.AddMenuItem('View', 'Save Snapshot', lambda e: save_snapshot(vis_fr.glCanvas))
+    vis_fr.AddMenuItem('File>Save', 'View Snapshot', lambda e: save_snapshot(vis_fr.glCanvas))
```
David-Baddeley, line 66: I'm inclined to keep this on view - to me the File menu is for data / quantitative stuff and a snapshot is not data

barentine: I get that, but I think if we put ourselves in the shoes of a user it's pretty weird to open a program with visualize in the name and not be able to save your visualization in the main 'file' menu. What if you opened Word and could only save to pdf from the view menu?

barentine: I also think we get the snapshot-is-not-data point across by calling it a 'snapshot', or in this PR, 'view snapshot'.

zacsimile:
I think this gets at a bigger issue, which is what do we want under each menu? I agree that `Take Snapshot` is something I'd expect to find under the File menu. However, I understand the argument for putting it under `View`, and it's not terribly difficult to get used to. However, following this logic, if `Take Snapshot` is under `View`, then I think `Animation` should be under `View` as well and `Add Octree Layer` should not be. In general, I think we should revisit what's under each menu (another example is the qPAINT functions could be under `Analysis>qPAINT`, etc.)

I'm in favor of a flat open/save layout. I use `Open extra channel` and `Open raw/prebleached data` enough that putting them in a submenu would be more annoying than it is helpful from an organizational standpoint. I'd like `Open extra channel` to be above `Open raw/prebleached data`. I do think the horizontal bars should separate the open functions from the save functions, and it would be nice to have consistent capitalization.

David-Baddeley:
Following up on Zachs comment,

- I also prefer the flat layout, and agree that a separator between open and save would be nice (but was thinking how to say so in a constructive way so that I'm not always criticising).
- The VisGUI menus are well overdue another cleanup (we should possibly make an issue for this, although part of me says I should just do it and/or make a branch for it to get some feedback before merging)
- Animation & scale box should both be on the view menu (although scale box might move to toolbar instead)
- Tessellation and density estimation definitely don't belong in view
- In some ways layers are part of the view - at least adding a new visualisation layer for **data which is already in the pipeline** is definitely a view function. Ultimately I'd like menu options for adding all the layer types we support (point cloud, surface, tracks, image etc ...). These could be put on an "Add layer" popout from the view menu or potentially a separate "Layers" menu. Generating the data for those layers is **not** a view function (`Add octree layer` currently does both which is how it snuck it's way into view).

barentine:
This didn't gain much traction and sounds like @David-Baddeley will eventually have a go at it on a separate branch and get feedback before merging

PR #243, opened 2020-05-12
OMEXML pixel size/unit guessing fix

zacsimile:
I noticed this while trying to do some fits on 3D distributions of gold nanospheres using `Gauss3DFitR`. `scipy.optimize.leastsq` was returning `None` for `cov_x`. The following fixes this problem.

- Changed to correct case for `getAttribute`
- Check for unicode mu symbol in py3
- Return `ps` if successful

### DIFF
--- a/PYME/IO/MetaDataHandler.py
+++ b/PYME/IO/MetaDataHandler.py
@@ -746,22 +746,23 @@ class XMLMDHandler(MDHandlerBase):
 
 
 class OMEXMLMDHandler(XMLMDHandler):
-    _OME_UNITS_TO_UM = {'m': 1e6, 'mm': 1e3, 'um': 1.0, 'nm': 1e-3}
+    _OME_UNITS_TO_UM = {'m': 1e6, 'mm': 1e3, 'um': 1.0, ('%sm'%chr(181)): 1.0, 'nm': 1e-3}
     
     @classmethod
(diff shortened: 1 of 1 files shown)


PYME/IO/MetaDataHandler.py
```diff
@@ -746,22 +746,23 @@ def getEntryNames(self):
 
 
 class OMEXMLMDHandler(XMLMDHandler):
-    _OME_UNITS_TO_UM = {'m': 1e6, 'mm': 1e3, 'um': 1.0, 'nm': 1e-3}
+    _OME_UNITS_TO_UM = {'m': 1e6, 'mm': 1e3, 'um': 1.0, ('%sm'%chr(181)): 1.0, 'nm': 1e-3}
```
David-Baddeley:
```suggestion
    _OME_UNITS_TO_UM = {'m': 1e6, 'mm': 1e3, 'um': 1.0, u'\u00B5m': 1.0, 'nm': 1e-3}
```

Using the unicode literal string is a bit cleaner and safer (much more likely to work across python versions and GUI toolkits etc ...) - unicode 181 is mu wheras ascii chr(181) is ?

David-Baddeley: note that unicode literals are in hex, not decimal (hence the b5)

PR #248, opened 2020-05-15
add scikits depends back into pip req's files

barentine:
At the moment the scikit-image dep missing causes some tests to fail (e.g. spherical_harmonics and coordinate_tools)

```
==================================== ERRORS ====================================
_____ ERROR collecting tests/PYME/Analysis/points/test_coordinate_tools.py _____
ImportError while importing test module '/home/vsts/work/1/s/tests/PYME/Analysis/points/test_coordinate_tools.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
PYME/Analysis/points/test_coordinate_tools.py:6: in <module>
    from skimage import morphology
E   ModuleNotFoundError: No module named 'skimage'
___ ERROR collecting tests/PYME/Analysis/points/test_spherical_harmonics.py ____
ImportError while importing test module '/home/vsts/work/1/s/tests/PYME/Analysis/points/test_spherical_harmonics.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
PYME/Analysis/points/test_spherical_harmonics.py:5: in <module>
    from skimage import morphology
E   ModuleNotFoundError: No module named 'skimage'

```

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -22,8 +22,8 @@ pyfftw #; python_version > '3.0'
 
 #mpld3
 cherrypy
-#scikit-image
-#scikit-learn
+scikit-image
+scikit-learn
 
 #FIXME for py3k
 #shapely [osx and py2k]
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -21,8 +21,8 @@ pyfftw #; python_version > '3.0'
 
 #mpld3
 cherrypy
-#scikit-image
-#scikit-learn
+scikit-image
+scikit-learn
 
 #FIXME for py3k
 #shapely [osx and py2k]


David-Baddeley:
The omission is deliberate as skimage often causes dependency issues (I think this is especially true on the CI system). I'm not super keen on putting it back in, at least until we have resolved all the other test issues and have the tests consistently passing in CI. We hopefully should be able to do a test import of skimage and then skip the tests which depend on it.

zacsimile:
While I understand this reasoning, I think eliminating `scikits`, `django=1.11`, etc. from PYME requirements pushes the onus of getting the package right onto the user instead of pushing the work onto us to get the versioning right. I think this will make it harder to get new users. Also, I think the CI uses `pip` instead of `conda`, so the problem is a bit different.

David-Baddeley:
It's hard to find the right line such that we don't spend 90% of our time on dependency management. I think a reasonable pragmatic solution which makes things more or less manageable would be to have a set of core dependencies which enable the most commonly used functionality (ideally without any version pinning), and then a set of optional / recommended dependencies for specific tasks that are not strictly required for the core functionality (e.g. things like django, mayavi, shapely, opencv, pyvisa etc ...), leaving these for the power user. If we put everything in we pretty much guarantee unsolvable environments and make it uninstallable.  At one point we had a pip `requirements.txt` file and a seperate `recommended.txt` for the optional dependencies.

Ultimately skimage should come back in to the core dependencies (but maybe not just yet - give it a couple of weeks of passing tests), and django (which is always going to require some form of pinning) should stay out. 

In the long term some of this will likely go away when we break PYME up into smaller packages - so that users can choose to install the parts of PYME (and corresponding dependencies) that they need.

PR #250, opened 2020-05-17
Ripleys tweaks

zacsimile:
GUI and nomenclature (L vs. H)  tweaks. Fix to 2D voxel size. Fix for mask/image registration.

Re: registration fix. I have a real data example where

```
Mask origin: 33524.79315970354, 26741.4607524257, 0
Coord origin: 47520.0, 0.0, 0
```

Subtracting the coordinate origin from the mask origin resulted in a massive shift of mask points away from the real data, causing Ripley's to indicate dispersion where there is none. We didn't notice this on generated data (wormlike chain) because their coordinate origin is `(0,0,0)`.

### DIFF
--- a/PYME/recipes/pointcloud.py
+++ b/PYME/recipes/pointcloud.py
@@ -272,7 +272,7 @@ class Ripleys(ModuleBase):
             if mask.data.shape[2] < 2:
                 raise RuntimeError('Need a 3D mask to run in 3D. Generate a 3D mask or select 2D.')
             three_d = True
-        if self.dimension == '2D':
+        else:
             if mask.data.shape[2] > 1:
                 raise RuntimeError('Need a 2D mask.')
             three_d = False
--- a/PYME/recipes/pointcloud.py
+++ b/PYME/recipes/pointcloud.py
@@ -251,11 +251,12 @@ class Ripleys(ModuleBase):
     inputPositions = Input('input')
     inputMask = Input('')
     outputName = Output('ripleys')
-    normalization = Enum(['K', 'L'])
+    normalization = Enum(['K', 'L', 'H'])
     nbins = Int(50)
     binSize = Float(50.)
     sampling = Float(5.)
     threaded = Bool(False)
+    dimension = Enum(['2D', '3D'])
     
     def execute(self, namespace):
         from PYME.Analysis.points import ripleys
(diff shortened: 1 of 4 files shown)


zacsimile:
I also fixed a small bug in `QPobjectSegment.py` in this PR. It was pointed out to me while I was already working on this branch so I just committed it with everything else.

David-Baddeley:
As I see it the registration fix actually disables registration rather than fixing anything. 

The only really acceptable way to fix this is to work out why you get a screwy origin for your coords.

zacsimile:
To me, it looks like the registration is taken care of in the mask generation. The screwy origin coordinates are straight from the metadata and are coming off a LabView scope, so they could be wrong. Worth discussing during our meeting.

PYME/recipes/pointcloud.py
```diff
@@ -251,20 +251,31 @@ class Ripleys(ModuleBase):
     inputPositions = Input('input')
     inputMask = Input('')
     outputName = Output('ripleys')
-    normalization = Enum(['K', 'L'])
+    normalization = Enum(['K', 'L', 'H'])
     nbins = Int(50)
     binSize = Float(50.)
     sampling = Float(5.)
     threaded = Bool(False)
+    dimension = Enum(['2D', '3D'])
```
David-Baddeley, line 259: What should the default be? 

Very minor point, but is there a strong case for making this an `Enum` rather than just a `threeD = Bool()`? Using a bool would clean up a lot of the subsequent code as we ultimately just use `dimension == '3D'` to define a`three_d` boolean. From a UI standpoint, it will render as a drop down list or checkbox respectively - The ideal would likely be radio buttons, but for me I'd probably prefer the checkbox. If writing recipes by hand, it's also much easier to work out `True` and `False` than trying to figure out which string constants to use.

I shouldn't really care about this but the grammar nazi in me finds  'dimension' as a trait name a bit grating (suspect "dims", "dimensions" or "dimensionality" would be better).

zacsimile: Good point. I'll change it to a bool.

PYME/recipes/pointcloud.py
```diff
@@ -251,20 +251,31 @@ class Ripleys(ModuleBase):
...
-        three_d = np.count_nonzero(points_real['z']) > 0
+
+        # three_d = np.count_nonzero(points_real['z']) > 0
+        if self.dimension == '3D':
+            if np.count_nonzero(points_real['z']) == 0:
+                raise RuntimeError('Need a 3D dataset')
+            if mask.data.shape[2] < 2:
+                raise RuntimeError('Need a 3D mask to run in 3D. Generate a 3D mask or select 2D.')
+            three_d = True
+        if self.dimension == '2D':
```
David-Baddeley:
```suggestion
        else:
```

There are only two options, and we can only be one or the other. Something like
```python
if (dim=='3D'):
    ...
elif (dim=='2D'):
    ...
else:
   raise (InvalidDimChoice)
```
would also be acceptable, but overkill as the `Enum` trait does the validation for you.

Independent `if` statements implies that they are not mutually exclusive - i.e. both could evaluate true

zacsimile: Yup, strange. I thought I'd written this as an `if`/`elif`. Thanks for catching that!

PR #251, opened 2020-05-18
add helper for measuring stage level-ness using an offset piezo

barentine:
Idea is just to automate revisiting stage positions when trying to level an insert or so. No ui other than a pyplot figure at the moment.


PYME/Acquire/Hardware/Piezos/base_piezo.py
```diff
@@ -1,5 +1,17 @@
 class PiezoBase(object):
-    """ Base class for piezos"""
+    """
+    Base class for all stages
+
+    Parameters
```
David-Baddeley:
```suggestion
    Attributes
```

PR #253, opened 2020-05-19
display free disk space on cluster in spool frame

barentine:
towards launch workers feature parity, see #160 

Left things in GB, which is probably preferred for cluster of one, though might cause slight power trips for users on production microscopes w/clusters


zacsimile:
See #255.

PR #257, opened 2020-05-21
spool panel gui consistency improvements

barentine:
- makes the autostart-analysis checkbox text more readable and accurate since we also use it to toggle dh5view -g 
- changes 'Analyse' button to 'Open', deprecating OnBAnalyse but leaving it effectively alone for those currently calling it through protocols
- remove `spoolController.LaunchLocalize` calls from htsms protocols now that the checkbox takes care of it

### DIFF
--- a/PYME/Acquire/SpoolController.py
+++ b/PYME/Acquire/SpoolController.py
@@ -346,7 +346,9 @@ class SpoolController(object):
         self.spooler.StartSpool()
         
         self.onSpoolStart.send(self)
-        
+
+        if self.autostart_analysis:
+            self.LaunchAnalysis()
         #return a function which can be called to indicate if we are done
         return lambda : not self.spooler.spoolOn
 
@@ -376,27 +378,20 @@ class SpoolController(object):
         """Launch analysis
         """
         from PYME.Acquire import QueueSpooler, HTTPSpooler
-        
-        dh5view_cmd = 'dh5view'
-        if sys.platform == 'win32':
-            dh5view_cmd = 'dh5view.exe'
-            
-        if self.autostart_analysis:
-            dh5view_cmd += ' -g'
-        
-        if isinstance(self.spooler, QueueSpooler.Spooler): #queue or not
+
+        if isinstance(self.spooler, QueueSpooler.Spooler):
+            dh5view_cmd = 'dh5view -g'
+            if sys.platform == 'win32':
+                dh5view_cmd = 'dh5view.exe -g'
             subprocess.Popen('%s -q %s QUEUE://%s' % (dh5view_cmd, self.spooler.tq.URI, self.queueName), shell=True)
-        elif isinstance(self.spooler, HTTPSpooler.Spooler): #queue or not
-            if self.autostart_analysis:
-                # launch analysis in a separate thread
-                t = threading.Thread(target=self.launch_cluster_analysis)
-                t.start()
-                # keep track of a couple launching threads to make sure they have ample time to finish before joining
-                if self._analysis_launchers.full():
-                    self._analysis_launchers.get().join()
-                self._analysis_launchers.put(t)
-            else:
-                subprocess.Popen('%s %s' % (dh5view_cmd, self.spooler.getURL()), shell=True)
+        elif isinstance(self.spooler, HTTPSpooler.Spooler):
+            # launch analysis in a separate thread
+            t = threading.Thread(target=self.launch_cluster_analysis)
+            t.start()
+            # keep track of a couple launching threads to make sure they have ample time to finish before joining
+            if self._analysis_launchers.full():
+                self._analysis_launchers.get().join()
+            self._analysis_launchers.put(t)
      
     def launch_cluster_analysis(self):
         from PYME.cluster import HTTPRulePusher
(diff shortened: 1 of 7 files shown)


PYME/Acquire/ui/HDFSpoolFrame.py
```diff
@@ -259,11 +259,11 @@ def _spool_pan(self):
     
         hsizer.Add(self.bStopSpooling, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
     
-        self.bAnalyse = wx.Button(self.spoolProgPan, -1, 'Analyse', style=wx.BU_EXACTFIT)
-        self.bAnalyse.Enable(False)
-        self.bAnalyse.Bind(wx.EVT_BUTTON, self.OnBAnalyse)
+        self.open_button = wx.Button(self.spoolProgPan, -1, 'Open', style=wx.BU_EXACTFIT)
```
David-Baddeley, line 262: It probably makes sense to follow the control type prefix - e.g. `bBlah`, `stBlah` etc naming convention for wx gui controls (we should only really be striving for pep8 in stuff which could be considered part of an 'API' - i.e. called from other code).

PYME/Acquire/SpoolController.py
```diff
@@ -346,7 +346,9 @@ def StartSpooling(self, fn=None, stack=False, compLevel = 2, zDwellTime = None,
         self.spooler.StartSpool()
         
         self.onSpoolStart.send(self)
-        
+
+        if self.autostart_analysis:
```
David-Baddeley, line 350: I still think that this is the wrong place to launch analysis and that we should have the flexibility of deciding where to do it in the protocol. Doing it here will flat out not work for some use cases, and increases the risk of race conditions in others. In the cluster spooling case we need to wait until the metadata has hit the disk on the cluster and the caches have invalidated before launching analysis - this is likely to be at least 2s after we start the acquisition (1s for the timeout on the metadata put call, 1s for cache invalidation).

I do, however, see the attraction of having it toggle-able in the UI rather than having to have different protocols.

My proposed solution is that the protocols still call a `LaunchAnalysis()` method on the spool-controller (potentially renamed as, e.g. `mark_ready_for_analysis()`,  `trigger_analysis()` or similar)  when they are ready, but that the the spool controller decides what do when that happens, depending on a couple of config settings - e.g. by referencing `automatically_start_localizing` and `automatically_open_series` properties (with associated GUI controls).

I suspect you'll still object, but I don't think that there is a clean solution to this, and am pretty certain that flexibility with analysis timing is needed, particularly if we want to be flexible for other imaging modalities and types of analysis moving forward. To round things out, I'd probably call `mark_ready_for_analysis()` automatically at the end of a protocol if it hasn't already been called - i.e if you have checked 'automatically start localisation' an acquisition will launch analysis after spooling is complete, unless you flag it as being ready earlier by calling `mark_ready_for_analysis()` earlier.

David-Baddeley: It's also worth noting that your changes quite dramatically change the way cluster spooling and analysis on the HT system work (by launching analysis at the beginning of a series rather than the end) and as such risk significant breakage. This is something I'd want to test thoroughly before use on real data and if your focus once getting back in the lab is on quickly getting the required data to finish the manuscript, this might be an un-necessary risk.

PYME/Acquire/ui/AnalysisSettingsUI.py
```diff
@@ -28,22 +28,27 @@ def __init__(self, parent, analysisSettings, mdhChangedSignal=None):
         self.fitFactories = PYME.localization.FitFactories.resFitFactories
 
         hsizer = wx.BoxSizer(wx.HORIZONTAL)
+        self.cbLocalizeInRealTime = wx.CheckBox(self, -1, 'Localize in real time')
+        self.cbLocalizeInRealTime.SetValue(False)
+        self.cbLocalizeInRealTime.Bind(wx.EVT_CHECKBOX, lambda e: self.analysisSettings.SetPropagate(e.IsChecked()))
```
David-Baddeley, line 33: I'm not at all sold on the renaming of this checkbox without changing function. As it is, this decides whether to write analysis settings to the metadata (as it says in the label), nothing more, nothing less.

Later on in `LaunchAnalysis` we decide  based on whether settings are being written whether we can automatically analyse. This is a bit arse about face, and is a consequence of the autolaunching being cobbled in quickly to get something working. There really should be a dedicated option for `automatically_launch_analysis`. As we are going to the effort of revisiting this I think we should do it properly.

`automatically_launch_analysis==True` should imply `propagate=True`, but it's a little unclear to me if the converse should also apply. One of the original use cases was to reduce the number of settings you needed to change in dh5view for each file - for example if you were doing a bunch of ratiometric 3D imaging in a session, you could set the fit factory and PSF settings in PYMEAcquire, and would only need to double check the threshold in dh5view before hitting go.

PS - I think the checkbox label should be `Auto-launch localization`, `Automatically start localising` or something similar as whether it is real time or not is a function of your data rate, fit type, and whether you have a cluster or not.

PYME/Acquire/SpoolController.py
barentine: If the protocol needs to be the authority on when analysis should be run, maybe the protocol can also provide the analysis metadata, say by adding an `mdh` kwarg to `spoolController.LaunchAnalysis`?

Needing to wait 1 s after spooling metadata doesn't break 'this', with 'this' in quotes because PR #258 refactors the localization launching to have a single entry point and a simple check there on the metadata (or 1 s, 2 s sleep) wouldn't be a big deal at all.

I also thought we'd agreed we could ditch the `automatically_open_series` concept (unless you're using open to mean something else here).

PYME/Acquire/ui/HDFSpoolFrame.py
barentine: camelCase and short prefixes aren't always gfReadability. Why would we draw a line between making our API pep8 and the rest of our code (slowly, if we are already changing it, etc.)?

PYME/Acquire/ui/AnalysisSettingsUI.py
barentine: The naming is in line with the static text for that section, which is indeed a bit ambitious and should probably also change to automatic.
Screen Shot 2020-05-24 at 6 22 17 PM

LaunchAnalysis is definitely a bit arse about face. Explaining to a user the nuance of propagating metadata, and what that means for each of the spool types is something I'd like to get around at some point. re: doing it properly, I'm spreading things out over multiple PRs. Ultimately I want the auto analysis to work as a list of chained rules, which we've discussed, but I don't think you want all of in one PR.

PYME/Acquire/SpoolController.py
barentine: re: changes for high-throughput system/taking data - I can easily hack in my analysis metadata to my protocol and call the spool controller method.

barentine:
This GUI stuff can come after rule-refactoring / cluster localization takes over queues, etc. Discussing with David, also OK'd the plan of making it possible to opt-out of the standard z stack acquisition and hdfspoolframe panels.

PR #258, opened 2020-05-21
refactor HTTPRulePusher calls so they share single entry point

barentine:
additionally:
- pulls startAt/current frame number from analysis metadata rather than making it up
- removes some superfluous unused parameters of `HTTPRulePusher.__init__`
- moves copy of events from datasource to localization results file to after rule pushing is complete (such that events will be complete)


barentine:
superseded by #272

PR #260, opened 2020-05-25
Dataserver get tabular part endpoint

barentine:
Adds a nested recarray source to PYME tabular, with a to_json method such that we can read/slice part of a table and return it as json. Submitting this PR early for feedback.

This would be a precursor to a tabular.HTTPSource which would have a polling flag to update periodically as localizations are added for live preview. In the mean time I implemented a simplistic load for visgui (which does not allow slicing during retrieval or viewing part of a tabular other than 'FitResults').

Events would be nice to handle using a base class with a to_json method. At the moment I haven't done anything there.

### DIFF
--- a/PYME/cluster/HTTPDataServer.py
+++ b/PYME/cluster/HTTPDataServer.py
@@ -683,6 +683,23 @@ class PYMEHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
             return None
 
     def get_tabular_part(self, path):
+        """
+
+        Parameters
+        ----------
+        path: str
+            path to an hdf or h5r file on the dataserver computer. Specific parts of the file can be accessed by
+            appending, e.g. /Metadata, for a resulting `path` of something like folder/test.h5r/Metadata. The 'part' of
+            the file returned defaults to FitResults if unspecified, and can additionally include a slice to take before
+            returning. In other words, a `path` of 'folder/test.h5r/FitResultsslice(0,100,None)' will return the first
+            hundred elements of the FitResults table in test.h5r.
+
+        Returns
+        -------
+        f: BytesIO
+            Requested part of the file encoded as bytes
+
+        """
         from PYME.IO.tabular import NestedRecArraySource
         from PYME.IO import h5rFile
         ext = '.h5r' if '.h5r' in path else '.hdf'
@@ -696,11 +713,13 @@ class PYMEHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
                     mdh = h5f.mdh.to_JSON()
                     f, length = self._string_to_file(mdh)
                 elif part == 'Events':
-                    events = h5f.events
-                    raise RuntimeError('fixme / make events class with to_JSON method')
+                    # events = h5f.events
+                    #  fixme / make events class with to_JSON method
+                    self.send_error(400, 'serving events from tabular files not yet supported')
+                    return
                 else:
                     if 'slice' in part:
-                        # 'testing' + str(slice(None, 10, 2))
+                        # convert the slice back from str, avoiding eval for security
                         part, return_slice = part.strip(')').split('slice(')
                         return_slice = slice(*(None if s == 'None' else int(s) for s in return_slice.split(',')))
                     else:
--- a/PYME/cluster/HTTPDataServer.py
+++ b/PYME/cluster/HTTPDataServer.py
@@ -567,7 +567,7 @@ class PYMEHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
                 return self.list_h5(path)
             else:
                 return self.get_h5_part(path)
-        elif '.h5r' in self.path or '.hdf' in self.path:
+        elif '.h5r/' in self.path or '.hdf/' in self.path:
             return self.get_tabular_part(path)
 
         ctype = self.guess_type(path)
--- a/PYME/cluster/HTTPDataServer.py
+++ b/PYME/cluster/HTTPDataServer.py
@@ -691,31 +691,32 @@ class PYMEHTTPRequestHandler(http.server.SimpleHTTPRequestHandler):
             ctype = self.guess_type(filename + ext)
             part = part.lstrip('/').lstrip('\\')
 
-            if part == 'Events':
-                raise NotImplementedError
-            elif part == 'Metadata':
-                raise NotImplementedError
-
-            if 'slice' in part:
-                # 'testing' + str(slice(None, 10, 2))
-                part, return_slice = part.strip(')').split('slice(')
-                return_slice = slice(*(None if s=='None' else int(s) for s in return_slice.split(',')))
-            else:
-                return_slice = slice(None)
-
             with h5rFile.openH5R(filename + ext) as h5f:
-                table = h5f.getTableData(part, return_slice)
-                print(table.dtype)
-                table = NestedRecArraySource(table)
-                print(table.to_json())
-                f, length = self._string_to_file(table.to_json())
+                if part == 'Metadata':
+                    mdh = h5f.mdh.to_JSON()
+                    f, length = self._string_to_file(mdh)
+                elif part == 'Events':
+                    events = h5f.events
+                    raise RuntimeError('fixme / make events class with to_JSON method')
+                else:
+                    if 'slice' in part:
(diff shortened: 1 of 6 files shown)


David-Baddeley:
This really shouldn't need to touch the `tabular` or `pipeline` classes.

PYME/cluster/HTTPDataServer.py
```diff
@@ -567,6 +567,8 @@ def send_head(self):
                 return self.list_h5(path)
             else:
                 return self.get_h5_part(path)
+        elif '.h5r' in self.path or '.hdf' in self.path:
```
David-Baddeley: We need to test for extension **and** trailing slash - i.e. `'.h5r/` as we want a simple get on `somefile.h5r` to return the whole file.

David-Baddeley:
```suggestion
        elif '.h5r/' in self.path or '.hdf/' in self.path:
```

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+        """
+        from PYME.IO.tabular import NestedRecArraySource
+        from PYME.IO import h5rFile
+        ext = '.h5r' if '.h5r' in path else '.hdf'
+        try:
+            filename, part = path.split(ext)
+            ctype = self.guess_type(filename + ext)
+            part = part.lstrip('/').lstrip('\\')
+
+            with h5rFile.openH5R(filename + ext) as h5f:
```
David-Baddeley: The part logic should be in `h5rFile` (as with h5File)

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+                if part == 'Metadata':
+                    mdh = h5f.mdh.to_JSON()
+                    f, length = self._string_to_file(mdh)
+                elif part == 'Events':
```
David-Baddeley: Events is just an HDF array - can handle as such - we shouldn't need a specific endpoint.

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+                    return
+                else:
+                    if 'slice' in part:
+                        # convert the slice back from str, avoiding eval for security
+                        part, return_slice = part.strip(')').split('slice(')
+                        return_slice = slice(*(None if s == 'None' else int(s) for s in return_slice.split(',')))
+                    else:
+                        return_slice = slice(None)
+                        if part == '':
+                            part = 'FitResults'
```
David-Baddeley: I don't think we need to have any special case logic for 'FitResults' on the server side

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+                    table = h5f.getTableData(part, return_slice)
+                    table = NestedRecArraySource(table, translate_pipeline_keys=True)
```
David-Baddeley: Ditto here - key translation should not be happening on the server - we want to keep computation on the server to a minimum and we will ultimately be more flexible if we keep the server interface as close to pytables as possible - i.e. simply accessing a number of rows from a  named table.

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+                    f, length = self._string_to_file(table.to_json())
```
David-Baddeley: I suspect we'll run into performance issues serialising as json. Look at what we use to write data to files in the HTTPDataServer `__aggregate` endpoints and/or in `PYME.IO.clusterResults`

Ultimately we serialise as pickle (which has security issues). Using numpys disk format would be better from a security perspective, but significantly slower (although still not nearly as bad as json) need to watch recent numpys in case this is fixed).

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+        f: BytesIO
+            Requested part of the file encoded as bytes
+
+        """
+        from PYME.IO.tabular import NestedRecArraySource
+        from PYME.IO import h5rFile
+        ext = '.h5r' if '.h5r' in path else '.hdf'
+        try:
+            filename, part = path.split(ext)
+            ctype = self.guess_type(filename + ext)
```
David-Baddeley: this should be the content type you are returning, not the content type that the file has - i.e. `application\json` for your implementation

PYME/cluster/HTTPDataServer.py
```diff
@@ -679,6 +681,64 @@ def get_h5_part(self, path):
...
+    def get_tabular_part(self, path):
+        """
+
+        Parameters
+        ----------
+        path: str
+            path to an hdf or h5r file on the dataserver computer. Specific parts of the file can be accessed by
+            appending, e.g. /Metadata, for a resulting `path` of something like folder/test.h5r/Metadata. The 'part' of
+            the file returned defaults to FitResults if unspecified, and can additionally include a slice to take before
+            returning. In other words, a `path` of 'folder/test.h5r/FitResultsslice(0,100,None)' will return the first
```
David-Baddeley: I really don't like just appending slice without any form of separator. I'd consider either of:

`test.h5r/FitResults/start/end`

or, potentially better, using standard query syntax:

`test.h5r/FitResults?from=0&to=100`

David-Baddeley: We could potentially be really fancy and allow the caller to specify the file type they want:
e.g.:

`test.h5r/FitResults.json?from=0&to=100`
or 
`test.h5r/FitResults.npy?from=0&to=100`

This would be a good fit with keeping our web UI as language agnostic as possible whilst not forcing us to use json for everything.

PYME/cluster/HTTPDataServer.py
David-Baddeley: We should probably refactor `clusterResults.fileResults()` to first call a `clusterResults.format_results()` function which contains the bulk of `fileResults()` and  then the URI translation and `fileFormattedResults()` so that we can re-use `clusterResults.format_results()` when returning tabular parts.  I'll do that now.

barentine: Overall I think this is a really clean approach, but casting nested structured arrays to json using `format_results` gives us rubbish on nested arrays. This is essentially why I added the `NestedRecArraySource`, but maybe instead we could just add an unnest recarray function somewhere?

barentine: alternatively, maybe we don't care enough to pass table parts as json and we just stick with .npy

PYME/cluster/HTTPDataServer.py
```diff
@@ -680,6 +682,60 @@ def get_h5_part(self, path):
...
+                    # figure out if we have any slicing to do
+                    query = urlparse.parse_qs(query)
+                    start = int(query.get('from', [0])[0])
+                    end = None if 'to' not in query.keys() else int(query['to'][0])
+                    wire_data, output_format = clusterResults.format_results(h5f.getTableData(part, slice(start, end)),
+                                                                             '.' + return_type)
+
+            f, length = self._string_to_file(wire_data)
+            self.send_response(200)
+            self.send_header("Content-type", output_format)
```
David-Baddeley:
```suggestion
            self.send_header("Content-type", output_format if output_format else 'application/octet-stream')
```

barentine:
@David-Baddeley - worth noting that the current to_JSON method added to tabular only allows things to be sliced/keys, it does not help us with flat/nested tables in get_tabular_part since get_tabular_part doesn't make a FitResultsSource.

What would you think about a function to unnest a recarray, which could be a flag in `h5rFile.getTabularPart` which we could flag true in `get_tabular_part` when returning json?

barentine:
I added a `unnest_recarray_to_json` function and apply it in `clusterResults.format_results`when we save numpy arrays to json. 

Additionally fixed a bug where we were clipping the query string from the path before passing to get_tabular_par

barentine:
Screen Shot 2020-05-28 at 1 05 18 PM

PR #264, opened 2020-06-01
Compute statistics on a Ripley's analysis

zacsimile:
Added Monte-Carlo simulations to Ripley's for statistical analysis of clustering. Ripley's now returns a simulation envelope and a significance graph (p-values). Example graphs with N=3 simulations from a standard worm-like chain with an extra Generate Events call are shown below. Also added some documentation.

Screen Shot 2020-06-01 at 8 14 05 AM
Screen Shot 2020-06-01 at 8 14 07 AM

### DIFF
--- a/PYME/Analysis/points/ripleys.py
+++ b/PYME/Analysis/points/ripleys.py
@@ -61,7 +61,9 @@ def ripleys_k_from_mask_points(x, y, xu, yu, n_bins, bin_size, mask_area, area_p
             # in the mask divided by the total circumference
             # 1/(dw/w)
             dww = w*d.astype('f')/dw
-            dww[dw==0] = 0  # d[w==0]
+            # If there are no points in the mask, at this radius,
+            # don't include these distances
+            dww[dw==0] = 0
             # The histogram is a weighted count of d
             hist += dww
     else:
@@ -77,7 +79,7 @@ def ripleys_k_from_mask_points(x, y, xu, yu, n_bins, bin_size, mask_area, area_p
             dw = dh_func(x[_i], y[_i], z[_i], xu, yu, zu, n_bins, bin_size)
             d = dh_func(x[_i], y[_i], z[_i], x, y, z, n_bins, bin_size)
             dww = w*d.astype('f')/dw
-            dww[dw==0] = 0  # d[w==0]
+            dww[dw==0] = 0 
             hist += dww
 
     K = (float(mask_area) / (lx ** 2)) * np.cumsum(hist)  # Ripley's K-function
@@ -168,9 +170,9 @@ def mc_points_from_mask(mask, n_points, three_d=True, coord_origin=(0,0,0)):
         # after the Monte-Carlo rejection step
         n_sim = int((np.prod(bool_mask.shape)/mask_area + eps)*n_points)
         
-        # generate randomly sampled coordinates on mask
+        # generate randomly sampled coordinates on a 3D space
         xu, yu, zu = (np.random.rand(n_sim,3)*[bool_mask.shape[0]-1,bool_mask.shape[1]-1,bool_mask.shape[2]-1]).T
-        # Find corresponding (xu, yu, zu) in the mesj
+        # Find corresponding (xu, yu, zu) in the mask
         point_mask = bool_mask[np.round(xu).astype(int), np.round(yu).astype(int), np.round(zu).astype(int)]
         # Monte-Carlo reject points outside of the mask and shift points inside to coordinate position
         xu, yu, zu = vx * xu[point_mask] + x0_m - x0_p, vy * yu[
@@ -189,9 +191,10 @@ def mc_points_from_mask(mask, n_points, three_d=True, coord_origin=(0,0,0)):
         point_mask = bool_mask[np.round(xu).astype(int), np.round(yu).astype(int)]
         xu, yu = vx * xu[point_mask] + x0_m - x0_p, vy * yu[point_mask] + y0_m - y0_p
 
-    # Truncate
-    if (len(xu) < n_points) or (len(yu) < n_points) or (len(zu) < n_points):
+    if (len(xu) < n_points) or (len(yu) < n_points) or ((zu is not None) and (len(zu) < n_points)):
+        # This one's for the developers
         raise RuntimeError('Not enough points were generated in the Monte-Carlo simulations. Revisit calculation of n_sim.')
+    # Truncate
     xu, yu, zu = xu[:n_points], yu[:n_points], zu[:n_points] if zu is not None else None
 
     return xu, yu, zu
--- a/PYME/Analysis/points/ripleys.py
+++ b/PYME/Analysis/points/ripleys.py
@@ -181,7 +181,68 @@ def mc_points_from_mask(mask, n_points, three_d=True, coord_origin=(0,0,0)):
 
     return xu, yu, zu, mask_area
 
-def ripleys_k(x, y, n_bins, bin_size, mask=None, bbox=None, z=None, threaded=False, mc=False, n_sim=100, sampling=5.0, coord_origin=(0,0,0)):
+def mc_sampling_statistics(K, n_points, n_bins, bin_size, mask, three_d, 
+                            n_sim=100, threaded=False, sampling=5.0, 
+                            coord_origin=(0,0,0)):
+    """
+    Calculates simulation envelope and significance of clustering on a mask
+    by simulating random uniform distributions on the mask using Monte-Carlo 
+    sampling.
+
+    Parameters
+    ----------
+        K : np.array
+            Ripley's K-values for real points to be compared to simulations.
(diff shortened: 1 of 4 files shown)


PYME/Analysis/points/ripleys.py
```diff
@@ -158,10 +239,27 @@ def ripleys_k(x, y, n_bins, bin_size, mask=None, bbox=None, z=None, threaded=Fal
...
+        for _i in range(n_sim):
+            xm, ym, zm, _ = mc_points_from_mask(mask, len(x), three_d, coord_origin)
+            _, K_arr[_i,:] = ripleys_k_from_mask_points(x=xm, y=ym, z=zm,
+                                      xu=xu, yu=yu, zu=zu,
+                                      n_bins=n_bins, bin_size=bin_size, mask_area=mask_area, threaded=threaded)
+        # Envelope
+        K_min = np.min(K_arr, axis=0)
+        K_max = np.max(K_arr, axis=0)
+        # Probability our original data is clustered
+        p_clustered = ((K_arr>=K).sum(0) + 1)/(n_sim + 1)
```
David-Baddeley: What is the statistical basis of this test? Should be better documented here ...

PYME/Analysis/points/ripleys.py
```diff
@@ -158,10 +239,27 @@ def ripleys_k(x, y, n_bins, bin_size, mask=None, bbox=None, z=None, threaded=Fal
...
+        # Monte-Carlo simulations on the mask
+        K_arr = np.zeros((n_sim,len(K)))
+        for _i in range(n_sim):
+            xm, ym, zm, _ = mc_points_from_mask(mask, len(x), three_d, coord_origin)
+            _, K_arr[_i,:] = ripleys_k_from_mask_points(x=xm, y=ym, z=zm,
+                                      xu=xu, yu=yu, zu=zu,
+                                      n_bins=n_bins, bin_size=bin_size, mask_area=mask_area, threaded=threaded)
+        # Envelope
+        K_min = np.min(K_arr, axis=0)
+        K_max = np.max(K_arr, axis=0)
```
David-Baddeley: An envelope would more normally be +- 1 sigma or (probably better) +- one 95% CI, not the min and max of the simulation.

PYME/Analysis/points/ripleys.py
David-Baddeley: Based on the graphs, it's also clearly wrong as it should be impossible to get a p value of 1, and you should not even come close with 3 MC iterations.

David-Baddeley:
What are the read dotted lines on the clustering significance plot?

PYME/Analysis/points/ripleys.py
```diff
@@ -106,14 +121,72 @@ def points_from_mask(mask, sampling, three_d = True, coord_origin=(0,0,0)):
...
+        point_mask = bool_mask[xu.astype(int), yu.astype(int)]
+        xu, yu = vx * xu[point_mask] + x0_m - x0_p, vy * yu[point_mask] + y0_m - y0_p
+        mask_area *= vx * vy
+
+    # Truncate
+    xu, yu, zu = xu[:n_points], yu[:n_points], zu[:n_points] if zu is not None else None
+
+    return xu, yu, zu, mask_area
+
+def ripleys_k(x, y, n_bins, bin_size, mask=None, bbox=None, z=None, threaded=False, mc=False, n_sim=100, sampling=5.0, coord_origin=(0,0,0)):
```
David-Baddeley: I think it would be a lot clearer if the MC stuff was not part of the `ripleys_k` function (it's fundamentally not part of calculating Ripleys function), but rather a, e.g., `uniform_sampling_confidence_intervals()` or similar function - I guess the advantage of putting it here as an option is that you avoid an extra `points_from_mask()` call, but it's hard to know if the diminished clarity is worth that.

PYME/Analysis/points/ripleys.py
zacsimile: Will update the documentation! 

I think a p-value of 1 is fine from a mathematical standpoint, just maybe not meaningful for only 3 simulations. E.g.: `scipy.stats.ttest_ind([0.1,0.1,0.1],[0.1,0.1,0.1])` returns a p-value of 1.0.

PYME/Analysis/points/ripleys.py
zacsimile: Will change this.

zacsimile:
> What are the read dotted lines on the clustering significance plot?

Above the upper dotted line means statistically significant dispersion. Below the lower dotted line means statistically significant clustering. I mention this briefly on line 450 of `PYME.LMVis.Extras.clusterAnalysis.py`. Could be documented better. Should an explanation come up in the plot?

PYME/Analysis/points/ripleys.py
zacsimile: You are right. I will make a `uniform_sampling_confidence_intervals()` function. That will clean up a lot of `PYME.recipes.pointcloud.Ripleys`.

David-Baddeley:
Statistically significant would usually require p <.05, but the dotted lines is are at ~0.33 and ~0.66. This would also seem to be a single sided test. i.e. p< 0.05 implies that the null hypothesis of no clustering can be rejected, but high p values e.g. (p > 0.95) have no defined meaning. They are most certainly not an indication of anti-clustering.

PYME/Analysis/points/ripleys.py
David-Baddeley: I'm pretty sure p is defined as strictly less than 1, but suspect that it doesn't matter that much because p values > .05 (or what ever your significance cutoff is) have no statistical interpretation (a p value ~=1 is not evidence for the opposite of what is being tested).

David-Baddeley:
I think you need 2 separate plots and tests for clustering and anti-clustering.

David-Baddeley:
I also think the graph could be improved as it's confusing as anything at the moment - in general your eye get's tricked into thinking that high values are more significant, and the labels (as they are) would tend to reinforce this. There might be a reason why you don't often see p-values plotted. I'd suggest flipping the y axis so more significant values are higher, potentially even doing something like plotting -log(p) which more closely correlates with significance. What you to be really careful about is that p values ~ 1 are interpreted as non-significant rather than anything else.

David-Baddeley:
If you look at all the heat-maps for DNA stuff etc they all plot -log_2(p)

zacsimile:
> Statistically significant would usually require p <.05, but the dotted lines is are at ~0.33 and ~0.66. This would also seem to be a single sided test. i.e. p< 0.05 implies that the null hypothesis of no clustering can be rejected, but high p values e.g. (p > 0.95) have no defined meaning. They are most certainly not an indication of anti-clustering.

In the formalism I used, p(clustered) ~ 1 - p(dispersed). For this reason, we only need one graph. The lines are at 0.33 and 0.66 because significance occurs at 1/number simulations and 1-1/number simulations. The method is outlined in the PDF I emailed you. It's possible this method is not the way to go. Because we need this to be done ASAP, can we have a meeting tomorrow to discuss this?

David-Baddeley:
That's not how p-values work! You can never do p_one hypothesis =
1-p_another hypothesis. If you look at the notes you linked, they have 2
separate p-values for the clustered and dispersed cases (eqns 4.6.8 and
4.6.10) and p_dispersed is not 1-p_clustered.

For biologists, significance is at p < 0.05 (2-sigma) for physicists, p <
0.003 (3 sigma) or higher (sometimes 4 or 5 sigma). There is never a case
where p=0.33 would be considered significant. I suspect the 1/n_simulations
is something different - if 0.05 < p < 1/n_simulations it's possible that
the result might be significant, but you can't tell yet based on the number
of simulations you have done. You need to do more simulations. If  p >
1/n_simulations > 0.05 it's highly unlikely that it will turn out to be
significant regardless of how many simulations you do, you can give up.

In general with the simulations you are trying to find where the 5% CI lies
- anything below this is significant. To establish this with any accuracy
you'll probably need >=100 runs.

zacsimile:
- On page I.4-16, around the middle of the page it says, "it must be true that P_dispersed(h) ~ 1 - P_clustered(h). So there is generally no need to compute both." Neither the notes nor I claim equality (that would be silly), but practically it seems to work given how we're calculating p-values.

- The line at 1/n simulations as statistically significant is also in the notes. 0.33 is definitely not sufficient. I did a quick test run with 3 simulations to get some graphs to illustrate the concept. I think maybe this added to the confusion instead of helping and I will adjust my PRs in the future. We would absolutely need to do >= 100 simulations for this to be useful, but I didn't feel like waiting for that many simulations to finish before submitting this PR.

PYME/Analysis/points/ripleys.py
zacsimile: Thought about this more. I think you're right a p-value should be strictly less than 1 for a continuous distribution. I think for a discrete and finite distribution, it should be possible to have p=1.

PR #265, opened 2020-06-04
Quadtree rendering fix

zacsimile:
A few minor fixes to restore quadtree rendering functionality and make it work for py3.

### DIFF
--- a/PYME/Analysis/points/QuadTree/__init__.py
+++ b/PYME/Analysis/points/QuadTree/__init__.py
@@ -21,5 +21,5 @@
 #
 ##################
 
-from pointQT import * 
-from QTrend import * 
+from .pointQT import * 
+from .QTrend import * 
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Will merge, but ultimately we should fix this properly (as it stands, it won't work when called from a recipe). 

Ultimately this would involve 1 or more of the following:
- moving the `pipeline.GenQuads()` logic into the renderer 
- moving the `pipeline.GenQuads()` logic to a function in `visHelpers`
- adding a `QuadTree` recipe module with the `GenQuads()` logic and abusing the octree layer to display it (this might be a better option than moving `GenQuads()` to the renderer if we felt it was important to be able to display the quadtree in VisGUI.

Both of the first two do not pre-compute and cache the quadtree, but that might not matter for what is a pretty niche case.

PR #266, opened 2020-06-05
Fix composite display

zacsimile:
Composite display used deleted (as of 1.3) `scipy` function. Swapped for scipy-recommended `PIL` function.

### DIFF
--- a/PYME/LMVis/imageView2.py
+++ b/PYME/LMVis/imageView2.py
@@ -24,7 +24,8 @@
 import numpy
 
 import wx
-import scipy.misc
+# import scipy.misc
+from PIL import Image
 
 from PYME.DSView.displayOptions import DisplayOpts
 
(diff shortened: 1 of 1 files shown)


PYME/LMVis/imageView2.py
```diff
@@ -347,7 +348,11 @@ def DoPaint(self, dc):
             else:
                 im = (self.image.data[int(x0_ / self.image.pixelSize):int(x1_ / self.image.pixelSize):step, int(y0_ / self.image.pixelSize):int(y1_ / self.image.pixelSize):step, self.do.zp, chanNum].squeeze().astype('f').T)
 
-            im = self._map_image(scipy.misc.imresize(im, sc), chanNum)
+            # im = self._map_image(scipy.misc.imresize(im, sc), chanNum)
+            # scipy.misc.imresize was deprecated in scipy 1.0 and removed in 1.3
+            # replaced with PIL per https://docs.scipy.org/doc/scipy-1.2.1/reference/generated/scipy.misc.imresize.html
```
David-Baddeley: Use `scipy.ndimage.zoom` instead. PIL should be avoided whenever possible. I don't really trust it on "data" - as opposed to snapshots and the like, and wouldn't be at all surprised if it did nasty things with data types which aren't uint8 - it might have got better but historically it's been a last resort for saving /loading RGB images.

zacsimile: Changed.

Issue #224, opened 2020-05-06
Where do we interact with snapshot and animation scale factors?


zacsimile:
Should the option to change the scale factor (pixel size) always pop up when generating a screenshot or animation, or should this be set at a default value and changed elsewhere, either in the shell or a different GUI settings panel?

My vote is for pop up every time we generate a screenshot so users know it's an option, and have it listed as an option in the animation panel.

David-Baddeley:
The ideal in terms of UI (at least for a shapshot) would be be to to display it as a changeable option within the save dialog (think of most of the Adobe save dialogs for an example) - this would give us the best of both worlds - could normally be ignored, but is there if you need it. The downside to this is that we would need a custom save dialog (which are invariably gross - see the adobe ones for example - and not themed to match the OS) rather than the default OS ones.

For the animation we could have it as part of the animation panel - maybe under a collapsable 'export settings section' or similar. This would be better than an extra pop-up.

Part 2 of this should be how do we present the scale factor to the user. Do we present it as a pixel size in nm, or as a ppi value (with the default of 72 for screen resolution). If you want to measure from the resulting image, the nm value is potentially useful, and is also useful for getting 2 differnt images to the same scale. Measuring from images, however, is problematic for orthographic 3D projections and useless in perspective 3D so I'd like to heavily discourage measurement from snapshots or animations. If you are taking the snapshot to put on a poster or in a figure, ppi is probably the more relevant quality metric, but this isn't much use for making different images the same scale (you'd have to do this before export - which might actually be the right approach - expose setting the pixel size for the live display so you can match zooms and then have a ppi setting for export).

yongdengzhang:
I personally really want this feature. It would be great to output the current 3D snapshot as high-resolution images for making nice figures, not necessarily for measurements. As this depends on random orientations, it is not easy to do by other means. At the moment, I would mostly use PYME for generating high-resolution images for figures and making nice movies. It will be a bump to me if I can only output the snapshots limited by the resolution of the screen.

lukasfue:
I too really want this feature. Especially with 3D surfaces and such. There's a lot that can be visualized with PYME that can't be included in a high-resolution generated image. So the option of having high-resolution screenshots would be excellent.

David-Baddeley:
Snapshot aspect should be fixed in #1162, should probably revisit the animation stuff to make the UI similar / check that it works.

David-Baddeley:
Just for documentation:

## View->Save snapshot

[...]

Issue #223, opened 2020-05-06
Animation improvements


zacsimile:
#222 restores the animation panel to its former glory, but I think we can do more. Ideally we'd be able to 

- [x] set the number of frames between each view
- [x] set the output pixel size (or at least scale factor)
- [x] choose the output file type (JPG, PNG)
- [x] fix the coloring, particularly on the LUT bar in exported animations

It looks like Michael Graff started implementing some of these ideas in `PYME.LMVis.Extras.animation.EditPanel` and `EditDialog`. Did we make a decision not to use the `EditPanel`? Or was it never finished? Are there additional features we'd like to add to our animation module?

(Almost) everything I've proposed above is already implemented on the backend. What we should discuss here is how we want it implemented in the GUI.

David-Baddeley:
- I'd favour keeping 'Save' and 'Make' as separate buttons (ideally renaming 'Make' to 'Export') so that 'Save' refers to the key frames and other bits (something you can reload on other data) and Export gives you an option between JPG, PNG, and AVI (if you have openCV installed).
- setting the number of frames between views and setting the frame rate are a bit redundant when exporting individual images.
- what is wrong with the colouring?

The `EditDialog` gets displayed when you double click on one of the views. It would probably make more sense to have the duration as an additional column.

zacsimile:
- Good point, we'll drop frame rate. 
- I agree duration should be an additional column rather than a popup.
- The coloring is very...psychedelic. If you export an animation, you'll see what I mean.

zacsimile:
Can we add single TIFFs as an export option (not stacks)?

zacsimile:
There were requests to add `time` and `alpha` changes to the animation. I think we'd just need to put these in `PYME.LMVis.views.View`. Doesn't seem terrible.

David-Baddeley:
Was tidying and found the animation UI mockup I originally drew for Michael (with a duration column :p) 

![IMG_20170328_092203](https://user-images.githubusercontent.com/19475296/82623000-45691080-9c33-11ea-8556-769daa50afeb.jpg)

zacsimile:
Cool! We can build this, we're already fairly close. I wonder if a duration column makes sense. The duration is associated with the transition from one view to the next, so I think it should show up in between views (maybe as an arrow with a time next to it), rather than as a column associated with a view, since this looks ambiguous. However, I think this would require making a non-standard window. I think, in this case, it may be necessary.

[...]

PR #269, opened 2020-06-07
Animation updates

zacsimile:
Animation improvements, re: #223. Trying to combine some export ideas from #224. To make a collapsing panel for export options, I'd like to convert `PYME.LMVIs.Extras.dockedPanel` to subclass `PYME.ui.manualFoldPanel.foldingPane`, if that's OK with you @David-Baddeley. It looks like there should be no conflicts.


zacsimile:
There are some strange visibility issues with the `foldingPane` in `animation.py`. As it stands, my 'Animation Settings' `collapsingPane` doesn't display. If I replace line 52 with `self.AddNewElement(clp,foldable=False)`, it displays, but it won't fold (of course). I've been comparing my code with the code from `PYME.Acquire.ui.HDFSpoolFrame` and I can't tell what I'm doing differently. Any thoughts on why this `collapsingPane` won't display when foldable @David-Baddeley?

zacsimile:
Replaced by #282 following a failed rebase.

PR #271, opened 2020-06-11
Fix Generate>Current

zacsimile:
We are no longer using `LegacyGLCanvas`, so we can (and must, to avoid errors) eliminate the color map arguments.


David-Baddeley:
@zacsimile  - do you also need to remove Generate->Current from docs?

zacsimile:
Removed.

PR #272, opened 2020-06-11
Chained rule analysis

barentine:
draft posting early to discuss with David


barentine:
Note that if #284 is merged, we can add a lock to the `pipeline.recipe` init and then add something like the following to `pipeline.OpenFile`

```
if ds is None:
    if filename.endswith('.h5r/live'):
        from PYME.IO import http_tabular, unifiedIO
        filename = filename.rstrip('/live')
        fn, server_filter = unifiedIO.split_cluster_url(filename)
        ds = http_tabular.HTTPFitResultsSource(fn, server_filter, update_lock=self.recipe.execution_lock)
        ds.updated.connect(self.recipe.invalidate_data)
        self.mdh.copyEntriesFrom(ds.mdh)
        # FIXME - add events link here too
    else:
        from PYME.IO import unifiedIO
        ...
```

which allows us to open a fully(?) functional PYMEVisualize via subproccess when we post a rule(chain) containing a localization rule

barentine:
closing, as the only outstanding code which isn't in main (albeit sometimes in temporary form) is the live-updating tabular, which is the subject of #347

PR #276, opened 2020-06-16
arrayViewPanel Hotkey Fix

zacsimile:
Flipped i/k hotkeys to match flipped y-axis.


zacsimile:
I think we should additionally map these to the arrow keys (code 37, 38, 39, 40 for left, up, right, down)

David-Baddeley:
I think there was a reason for not mapping to arrow keys initially, but can't recall what that was. One direction might do z (depending on what has focus).

zacsimile:
Ah, that makes sense. It's not critical to map these to the arrow keys, just thought I'd bring it up since it seemed natural. `i`,`j`,`k`,`l`, is also fairly natural and I see no need to change it if it risks a conflict.

Issue #279, opened 2020-06-18
hdf5 file mode/lock issue in chaining recipe after localization


barentine:
opening an hdf5 file in read mode for a recipe immediately after it's been created/opened in append mode for localization fails 
```
DEBUG:ruleserver:adding chained rule
DEBUG:ruleserver:chained rule ID 000007-83b5df21d72a40fd9a862e0a7a85ae79
DEBUG:ruleserver:on completion present: False
DEBUG:ruleserver:000004-7e13c4b0cc864a2780a65a9ec5dd505e, n_max: 1, n_completed: 0, n_assigned: 0, n_failed: 1
DEBUG:ruleserver:000005-0623b7e4630047bb887397f4f6c52729, n_max: 1000000, n_completed: 0, n_assigned: 0, n_failed: 0
DEBUG:ruleserver:000006-0e31f41ec2c34e0d9767193b070211ef, n_max: 150, n_completed: 66, n_assigned: 84, n_failed: 0
HDF5-DIAG: Error detected in HDF5 (1.10.4) thread 123145318404096:
  #000: H5F.c line 509 in H5Fopen(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #001: H5Fint.c line 1400 in H5F__open(): unable to open file
    major: File accessibilty
    minor: Unable to open file
  #002: H5Fint.c line 1615 in H5F_open(): unable to lock the file
    major: File accessibilty
    minor: Unable to open file
  #003: H5FD.c line 1640 in H5FD_lock(): driver lock request failed
    major: Virtual File Layer
    minor: Can't update object
  #004: H5FDsec2.c line 941 in H5FD_sec2_lock(): unable to lock file, errno = 35, error message = 'Resource temporarily unavailable'
    major: File accessibilty
    minor: Bad file ID accessed
Traceback (most recent call last):
  File "/Users/Andrew/code/python-microscopy/PYME/cluster/taskWorkerHTTP.py", line 337, in computeLoop
    recipe.loadInput(url, key)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/base.py", line 897, in loadInput
    with tables.open_file(fn, mode='r') as h5f:
  File "/opt/miniconda3/lib/python3.6/site-packages/tables/file.py", line 320, in open_file
    return File(filename, mode, title, root_uep, filters, **kwargs)
  File "/opt/miniconda3/lib/python3.6/site-packages/tables/file.py", line 784, in __init__
[...]

PR #281, opened 2020-06-19
load hdf5's through cluster when process collisions stop local load

barentine:
This is a solution to #279 which a) does not require long delay after encountering the HDF5Ext error and b) gets all nodes of the hdf5 at once rather than potentially asking what's there and then asking again for each using get_tabular_part.

note that it is also possible in recipes.base to try the fallback with
```
with unifiedIO.local_or_temp_filename(filename) as fn, h5rFile.openH5R(fn, mode='r')._h5file as h5f:
    self._populate_namespace_from_hdf5(key, h5f, fn, extension)
```
but this requires adding multiple flag-checks in `unifiedIO.local_or_temp_filename`, and it seems weird for a function named `this-or-that` to have an option to only return `that`. Adding flag to force the network access in the clusterIO call (as in this PR) would need to be done anyway.

Note that h5rFile opening on the temporary file object may not be super preferable to tables (if we're accessing through the dataserver process/locks anyway) since opening that way can encounter the following error (which is currently just printed using traceback so not functionally stopping anything)
```
  File "/Users/Andrew/code/python-microscopy/PYME/IO/h5rFile.py", line 217, in _pollQueues
    self._h5file.flush()
  File "/opt/miniconda3/lib/python3.6/site-packages/tables/file.py", line 2705, in flush
    self._check_open()
  File "/opt/miniconda3/lib/python3.6/site-packages/tables/file.py", line 2159, in _check_open
    raise ClosedFileError("the file object is closed")
tables.exceptions.ClosedFileError: the file object is closed
```

### DIFF
--- a/PYME/IO/clusterIO.py
+++ b/PYME/IO/clusterIO.py
@@ -733,7 +733,7 @@ def get_local_path(filename, serverfilter):
         if os.path.exists(localpath):
             return localpath
 
-def get_file(filename, serverfilter=local_serverfilter, numRetries=3, use_file_cache=True):
+def get_file(filename, serverfilter=local_serverfilter, numRetries=3, use_file_cache=True, local_short_circuit=True):
     """
     Get a file from the cluster.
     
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Solution is a bit too hacky for my liking - have merged and added a bunch of comments, but we should probably revisit this.

PR #282, opened 2020-06-20
Animation updates 2

zacsimile:
Animation improvements, re: #223. Trying to combine some export ideas from #224. Had to re-pr since I messed up a rebase.

### DIFF
--- a/PYME/LMVis/Extras/animation.py
+++ b/PYME/LMVis/Extras/animation.py
@@ -77,8 +77,6 @@ class VideoPanel(DockedPanel):
         return pan
     
     def _settings_pan(self, clp):
-        # dpi
-        # pixel_size
         pan = wx.Panel(clp, -1)
         vsizer = wx.BoxSizer(wx.VERTICAL)
         hsizer = wx.BoxSizer(wx.HORIZONTAL)
@@ -87,10 +85,8 @@ class VideoPanel(DockedPanel):
         hsizer.Add(self.file_type, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
         vsizer.Add(hsizer, 0, wx.LEFT | wx.RIGHT | wx.EXPAND, 0)
         hsizer = wx.BoxSizer(wx.HORIZONTAL)
-        self.dpi = wx.TextCtrl(pan, -1, size=(60, -1), value='72')
         # self.pixel_size = wx.TextCtrl(pan, -1, size=(60, -1), value='5.0')
-        hsizer.Add(wx.StaticText(pan, -1, 'DPI:'), 0, wx.LEFT | wx.RIGHT | wx.ALIGN_CENTER_VERTICAL, 5)
-        hsizer.Add(self.dpi, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 2)
+        # hsizer.Add(wx.StaticText(pan, -1, 'Pixel size:'), 0, wx.LEFT | wx.RIGHT | wx.ALIGN_CENTER_VERTICAL, 5)
         # hsizer.Add(self.pixel_size, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 2)
         vsizer.Add(hsizer, 0, wx.LEFT | wx.RIGHT | wx.EXPAND, 0)
         pan.SetSizerAndFit(vsizer)
@@ -254,10 +250,8 @@ class VideoPanel(DockedPanel):
                             #    video.write(cv2.cvtColor(cv2.flip(snap.transpose(1, 0, 2), 0), cv2.COLOR_RGB2BGR))
                             #else:
                             #    video.write(cv2.flip(snap.transpose(1, 0, 2), 0))
-                            dpi = int(float(self.dpi.GetValue()))
                             im = Image.fromarray(snap).transpose(Image.FLIP_TOP_BOTTOM).save(os.path.join(dir_name, 
-                                                            'frame{:04d}.{}'.format(f_no,self.file_type.GetValue().lower())),
-                                                            dpi=(dpi,dpi))
+                                                            'frame{:04d}.{}'.format(f_no,self.file_type.GetValue().lower())))
                             f_no += 1
                         else:
                             # sleep(2.0/steps)
--- a/PYME/LMVis/Extras/animation.py
+++ b/PYME/LMVis/Extras/animation.py
@@ -34,6 +34,9 @@ from PYME.LMVis.views import VideoView
 # See PYME.LMVis.Extras.dockedPanel for a history of mfp import name
 import PYME.ui.manualFoldPanel as mfp
 
+# Export file types
+EXPORT_FILE_TYPES = ['JPG', 'PNG', 'TIFF']
+
 # noinspection PyUnusedLocal
 class VideoPanel(DockedPanel):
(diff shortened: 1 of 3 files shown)


zacsimile:
This is what it looks like now:

Screen Shot 2020-06-20 at 5 18 29 PM

PYME/LMVis/Extras/animation.py
```diff
@@ -251,9 +304,12 @@ def on_select_view(self, event):
 
     def refill(self):
         self.clear_view()
-        for snapshot in self.snapshots:
-            index = len(self.snapshots)
-            self.view_table.InsertStringItem(index, snapshot.view_id)
+        for index, snapshot in enumerate(self.snapshots):
+            # index = len(self.snapshots)
+            # NOTE: InsertStringItem and SetStringItem are deprecated in favor of InsertItem and SetItem
```
David-Baddeley:
```suggestion
            # NOTE: InsertStringItem and SetStringItem are deprecated in wx > 4.0 favour of InsertItem and SetItem. Using old methods for wx=3.x compatibility.

```

PYME/LMVis/Extras/dockedPanel.py
```diff
@@ -20,15 +20,22 @@
...
-
 import wx.lib.agw.aui as aui
 
+# This is legacy-imported as afp since much of the code was written using
+# PYME.ui.autoFoldPanel and when we switched to manual fold panel it was 
+# easiest to change import PYME.ui.autoFoldPanel as afp to 
+# import PYME.ui.manualFoldPanel as afp. This is the first time we get an
+# opportunity to write a new section of code importing manualFoldPanel,
+# and so we choose to call it its acronym mfp
+import PYME.ui.manualFoldPanel as mfp
```
David-Baddeley:
```suggestion
import PYME.ui.manualFoldPanel as afp

```

PYME/LMVis/Extras/dockedPanel.py
```diff
@@ -20,15 +20,22 @@
...
 
-class DockedPanel(wx.Panel):
+class DockedPanel(mfp.foldingPane):
```
David-Baddeley:
```suggestion
class DockedPanel(afp.foldingPane):

```

PYME/LMVis/Extras/animation.py
```diff
@@ -150,11 +189,16 @@ def run(self, event):
         self.play(False)
 
     def play(self, save):
-        width = self.get_canvas().Size[0]
-        height = self.get_canvas().Size[1]
+        # Fail gracefully if there are no snapshots
+        if len(self.snapshots) == 0:
+            dlg = wx.MessageDialog(None, 'Need views to generate animation! Please create some by pressing `Add` in the animation pane.', 'No animation views')
```
David-Baddeley: - Missing a `dlg.Destroy()`
- For future reference, `wx.MessageBox()` does this all for you (roughly equiv to  `dlg = wx.MessageDialog(); slg.ShowModal(); dlg.Destroy()`)

PYME/LMVis/Extras/animation.py
```diff
@@ -198,18 +242,27 @@ def play(self, save):
...
+                                snap = snap.transpose(1,0,2)
+                            else:
+                                snap = (255*snap).astype('uint8').transpose(1, 0, 2)
                             
                             #if snap.ndim == 3:
                             #    video.write(cv2.cvtColor(cv2.flip(snap.transpose(1, 0, 2), 0), cv2.COLOR_RGB2BGR))
                             #else:
                             #    video.write(cv2.flip(snap.transpose(1, 0, 2), 0))
-                            im = Image.fromarray(snap).save(os.path.join(dir_name, 'frame%04d.jpg' % f_no))
+                            dpi = int(float(self.dpi.GetValue()))
```
David-Baddeley: Is dpi a) useful? and b) doing what you expect? 

When passed to PIL it simply writes the value to the image metadata - it doesn't rescale the image. DPI is only really useful when printing the images (and then arguably only if it's important if the physical size of the images is important - i.e. to make sure that 12pt text is the right size etc ...). I can't really see dpi being at all relevant for video production or the majority of scientific workflows where you arrange images into subpanels with illustrator etc ... The only real time you need to worry about DPI in a scientific workflow is wen exporting image composites to figures in for papers - but that then has very little relation to the input dpi attributed to each of the panels.

You've also got the various confusions with retina displays etc (where the default ppi is 144 rather than 72), so my gut feeling is that you are adding an extra field which in practice does very little, but might not be populated with correct default values if you do for some reason want to record the physical size of the image on screen. For dpi to be at all meaningful, you'd need to interrogate the display to see what it's actual ppi is (which will be super hard to do in a platform independant way) and then re-scale the image to the desired output dpi. I'm not sure why you'd want to do this.

zacsimile: I assumed the DPI would actually change the DPI of the image for figure rendering. Since it's only metadata, this is probably useless. Thanks for pointing this out.

I'd like to be able to change the snapshot pixel size or scaling (which I'd also change by just modifying `pixel_size`). As it stands, the `pixel_size` argument in `PYME.LMVis.gl_render3D_shaders.LMGLShaderCanvas.getIm()` does not scale the image in a useful way, but instead crops the image. I propose a method of adjusting this in #283. Please let me know if you think this is the right approach.

David-Baddeley: What is the rationale for needing to change the pixel size / scaling in animation? Surely the animation is just a visual and it's unimportant what the actual pixel size is - the only reason for changing sizes is then to optimize the size on disc / fit in a powerpoint slide etc ... The easiest approach here is going to be to re-size the VisGUI window to the desired size  before capturing the movie (this has the added advantage of letting you adjust the aspect ratio at the same time).

A physical calibration of pixel size and distances is not trivial for 3D orthographic views, and not possible for 3D perspective (which we plan to make more visible in the GUI), and as a result I think that any implication in the UI that you can generate an animation scaled in anything other than pixel units is not desirable (we can have scale bars or, for 3D, scale boxes in the image).

For most animation use cases, the best way of displaying zoom settings would potentially be to offer a dropdown of preset output sizes in pixels - e.g. current (`glCanvas.Size[0]`, `glCanvas.Size[1]`), 640x480, 320x240, etc, potentially including a custom option where you could set the number of pixels in x and y. So the user experience is that you get what you see on screen, but at the pixel sizing you desire. A caveat of this is that you are going to need to deal with the aspect ratio potentially being different (I'd opt for the conservative approach where you include all of the current viewport and potentially a bit extra, rather than cropping down to the new aspect ratio), so animation no longer becomes WYSIWYG. A lot of this could alternatively be handled by documentation saying to resize to the window before capturing though. The only one case I can see that resizing the window first doesn't address is if you want the animation at full screen resolution.

zacsimile: We had discussed exposing `pixel_size` or dpi in #224. I think a `scaling` option is pretty standard for snapshot exports (e.g. Meshlab). That said, we don't need to have it here. 

What do you think about accepting this pull request (or make notes on what else needs to be changed) and we'll revisit how to handle scaling in a different PR?

David-Baddeley: sounds like a good plan. I'm not adverse to having an option for scaling of the output, just want it to make sense if and when it goes in.

PR #161, opened 2020-02-27
Unifying event name and description types

barentine:
Events ordering and format are at the moment inconsistent in that when events are passed as a structured array they're in a different order than when they're written to json as list. Additionally, because the `S` dtypes are loaded as bytes in py3, and `str(b'unfortunate') = "b'unfortunate'"`we get some unnecessary awkwardness which we at the moment carry around when we handle/display events:
image

Since the `Events` tables saved in h5r are structured, we can easily convert them from bytes to unicode and a consistent order with the list upon loading. This is handled in a new file, `PYME.IO.events`, and allows us to drop all the bytes encoding/decoding when we're really after unicode.

This PR should be Python2 compatible
```
$ python
Python 2.7.17 |Anaconda, Inc.| (default, Oct 21 2019, 19:04:46) 
[GCC 7.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import tables
>>> t = tables.open_file('/home/smeagol/Desktop/20_2_series_00002_1.h5r')
>>> t.root.Events[:]
array([('PiezoOnTarget', 1.58222555e+09, '48.262'),
       ('PiezoOnTarget', 1.58222555e+09, '48.258'),
       ('PiezoOnTarget', 1.58222555e+09, '48.264'),
       ('PiezoOnTarget', 1.58222555e+09, '48.262'),
       ('PiezoOnTarget', 1.58222555e+09, '48.260'),
       ('PiezoOnTarget', 1.58222555e+09, '48.261'),
       ('PiezoOnTarget', 1.58222555e+09, '48.258'),
       ('PiezoOnTarget', 1.58222555e+09, '48.259'),
       ('PiezoOnTarget', 1.58222555e+09, '48.257'),
       ('PiezoOnTarget', 1.58222555e+09, '48.255'),
       ('PiezoOnTarget', 1.58222555e+09, '48.264'),
       ('PiezoOnTarget', 1.58222555e+09, '48.267'),
       ('PiezoOnTarget', 1.58222555e+09, '48.268'),
       ('PiezoOnTarget', 1.58222555e+09, '48.270'),
       ('PiezoOnTarget', 1.58222555e+09, '48.270'),
       ('PiezoOnTarget', 1.58222555e+09, '48.272'),
       ('PiezoOnTarget', 1.58222555e+09, '48.276'),
       ('PiezoOnTarget', 1.58222555e+09, '48.279'),
       ('PiezoOnTarget', 1.58222555e+09, '48.282'),
[...]

PR #284, opened 2020-06-22
Update the AddPipelineDerivedVars module and use it in the pipeline

barentine:
This PR is to update the `Pipelineify` recipe module to it is up to date with mappings that curring happen in `pipeline.OpenFile` and to then add that module directly to the pipeline recipe rather than have the same things in two places. Functionally, this is a pretty small change, but is a step towards making `pipeline.Rebuild` actually rebuild the pipeline, which one might want if they e.g. have a results source which is updating live (see PR #272 ). More importantly, having the module there makes things more consistent - if you develop a recipe in PYMEVisualize and then want to save it from the recipe panel and run it on the cluster or so, at the moment you need to magically know to throw the pipelineify/addpipelinederivedvars module onto the front of your recipe. This doesn't appear to expose anything wild/unseenly to the data pipeline, and I tried not to overly clean commented out code that was left in pipeline.OpenFile perhaps as a reminder to David.
Screen Shot 2020-06-22 at 11 46 18 AM


zacsimile:
I really like this. In addition to everything Andrew mentioned, exposing pixel size is very convenient. More than once I've received user data, e.g. from the Zeiss, where someone punched in the pixel size incorrectly.

PYME/LMVis/pipeline.py
```diff
@@ -759,63 +762,46 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
         #self._get_dye_ratios_from_metadata()
 
-        self.addDataSource('Localizations', mapped_ds)
-
         # Retrieve or estimate image bounds
-        if False:  # 'imgBounds' in kwargs.keys():
-            self.imageBounds = kwargs['imgBounds']
-        elif (not (
-                'scanx' in mapped_ds.keys() or 'scany' in mapped_ds.keys())) and 'Camera.ROIWidth' in self.mdh.getEntryNames():
+        # if False:  # 'imgBounds' in kwargs.keys():
```
David-Baddeley:
```suggestion
        if False:  # 'imgBounds' in kwargs.keys(): 
            # TODO - why is this disabled? Current usage would appear to be when opening from LMAnalysis
            # during real-time localization, to force image bounds to match raw data, but also potentially useful
            # for other scenarios where metadata is not fully present.
```

PYME/LMVis/pipeline.py
```diff
@@ -759,63 +762,46 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
+        #     self.imageBounds = kwargs['imgBounds']
```
David-Baddeley:
```suggestion
             self.imageBounds = kwargs['imgBounds']
```

PYME/LMVis/pipeline.py
```diff
@@ -759,63 +762,46 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
+        if ('scanx' not in ds.keys() or 'scany' not in ds.keys()) and 'Camera.ROIWidth' in self.mdh.getEntryNames():
```
David-Baddeley:
```suggestion
        elif ('scanx' not in ds.keys() or 'scany' not in ds.keys()) and 'Camera.ROIWidth' in self.mdh.getEntryNames():
```

PYME/LMVis/pipeline.py
barentine: I think you're right that the real-time display is the only situation where that is currently passed (and ignored). Also worth noting that the kwarg is misspelled there (`imBound`) in LMAnalysis, line 725.

I'm completely fine with whatever you decide to do here, @David-Baddeley. I'm a bit new to the suggestion feature - would you like me to commit it as I'm OK with it?

David-Baddeley:
Thanks, I like the concepts of a) reducing code duplication and b) making recipes saveable from VisGUI and then runable directly from Bakeshop etc ... are both good goals.

I'm going to need to go through this in a bit more detail before merging, however, as it has the potential to break some bits of VisGUI which have not yet been updated for the recipe module approach (amongst other things I'm pretty sure that there are still a few bits of code which assume that the input datasource is a mapping filter, meaning that all inputs might still need to be wrapped in a mapping filter). This might only apply to the special 'Localisations' name, but by altering this anything which does change the mappings on 'Localisations' will potentially get erased on a pipeline rebuild. 

I'm also very iffy on exposing a 'pixel_size_nm' control in pipeline editor, suspecting that it will result in confusion.  @zacsimile - I don't think this does what you think it does. The use case is for importing data which was analysed in different localization software which outputs fluorophores positions in units of pixels, rather than nm. For that data (importing from .txt or .mat) you get an opportunity to set the localization units when you open the file, so making it editable again is redundant. You could concievably use the setting to fudge stuff if you had analysed data with the wrong pixel size settings, but here be dragons - there is significant potential for this to go wrong depending on the fit type you used (some fits may not converge properly with incorrect voxel size metadata). A better approach is usually to re-analyse with the right pixel size. At the very least, we should re-name the parameter to 'input units nm' or similar in the UI. This reminds me that renaming and sanitising `Pipelineify` is also on the TODO list

PYME/LMVis/pipeline.py
David-Baddeley: I'll commit them from this end - I think that we might ultimately want to pop this out as an option in the matlab and .txt loaders as it would make it easier to get raw images /overviews taken in other software to line up with rendered images.

barentine:
Thanks David.

Anywhere in particular you want checked for breakage re: mapping filter? I know you’ve got loads on your plate at the moment and am happy to help. 

Re: keeping `pixel_size_nm` editing out of the pipeline view I think there’s a potential compromise that would be pretty clean without killing the recipe portability. We can remove the pixelsizeNM attribute of `pipelineify` and if we check for the `PixelSize` kwarg (==1) in `pipeline.OpenFile` we can simply pop in a `tablefilters.Mapping` module into the pipeline recipe to do the pixel size stuff. Then the pixelsize mapping would only be editable from the pipeline view if the dragons were already about

PYME/LMVis/pipeline.py
```diff
@@ -759,63 +762,49 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
-        #extract information from any events
-        self.ev_mappings, self.eventCharts = _processEvents(mapped_ds, self.events, self.mdh)
+        # skip the MappingFilter wrapping, etc. in self.addDataSource and add this datasource as-is
+        ds.mdh, ds.events = self.mdh, self.events
+        self.dataSources['results_source'] = ds
 
+        add_pipeline_variables = Pipelineify(self.recipe, 
+            inputFitResults='results_source',
+            inputDriftResults='' if self.driftInputMapping is None else self.driftInputMapping,
+            inputEvents='' if self.events is None else self.events,
```
David-Baddeley: This doesn't work - `inputEvents` takes a string which is a key to the recipe namespace. We would need to inject the events into the recipe namespace (which is a bit gross).

PYME/LMVis/pipeline.py
```diff
@@ -759,63 +762,49 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
-        if 'PixelSize' in kwargs.keys():
-            mapped_ds.addVariable('pixelSize', kwargs['PixelSize'])
-            mapped_ds.setMapping('x', 'x*pixelSize')
-            mapped_ds.setMapping('y', 'y*pixelSize')
-
-        #extract information from any events
-        self.ev_mappings, self.eventCharts = _processEvents(mapped_ds, self.events, self.mdh)
+        # skip the MappingFilter wrapping, etc. in self.addDataSource and add this datasource as-is
+        ds.mdh, ds.events = self.mdh, self.events
+        self.dataSources['results_source'] = ds
```
David-Baddeley: Ideally everything should be routed through `pipeline.addDataSource()` rather than just injecting. Injecting is a bad pattern as it leads to things which won't translate to re-runable recipes. At the moment `pipeline.dataSources` is just an alias for `pipeline.recipe.namespace` but it might become a read-only proxy in the future. It's hard to completely avoid injection during file loading, but it's nice to try.

PYME/LMVis/pipeline.py
barentine: With all the events loading we do in file opening, do we actually need this here? If we need to maintain the option to pop in events from another file or something am I going really recipe crazy to suggest we write a module to load events and add them to data source? I guess events in the recipe namespace at all seems to be a bit of an api fudge

PYME/LMVis/pipeline.py
barentine: I was trying to avoid a non-recipe MappingFilter on the live-update httptabular in another PR (which I suspect will involve some discussion! Happy to use addDataSource for now and kick that conversation to later)

PYME/LMVis/pipeline.py
David-Baddeley: Events in the recipe namespace is definitely a fudge (which is one of the reasons why I have some reservations about pipelinify in VisGUI), but was the quickest way of getting things working in bakeshop etc ... It also maps pretty well to the automatic (non h5r specific) HDF loading we do in recipes where we load every table from the hdf file into the namespace.

We need to get the events from somewhere in order to generate the derived columns ... in the current VisGUI loading workflow we avoid any injection nastiness by loading them once, but that isn't a super good match to pipelinify.

David-Baddeley:
Medium / long term, I'm wondering if the following (major refactor) of VisGUI loading might make sense:

- change things to use `recipes.base.ModuleBase.loadInput()` to perform low-level data file loading. This would likely involve transferring some but not all of the loading logic from pipeline here, including .mat and .txt handling.
- create a wrapper class for events which encapusulates `_processEvents()` and other event handling functions  and load this instead of just injecting events as a table.

barentine:
We have a `PYME.IO.events` candidate in PR #161. I was being a bit of a punk about str/bytes at the time, but it'd be quick to write a class version of a file and submit another PR.

zacsimile:
@David-Baddeley You may be right that I'm confused about the `pixel_size_nm` control. My impression was it did the same thing as the pixel size setting in the `ImportDialog`. My comment was also just an offhand, hooray, I can fix some things without having to re-analyze the data. I was not advocating for this as a regularly-used tool.

Right now, the import dialog only lets you set pixel size for .txt files. Once #229 is merged this will work for .mat files as well.

zacsimile:
I think from a developer perspective refactoring everything to load from `recipes.base.ModuleBase.loadInput()` makes a lot of sense and would make it even easier to pass recipes between different parts of PYME. From a user perspective, I worry this would force us to use traits to handle the loading GUIs, supplanting some of the easier-to-use GUI loaders. I think in the best scenario we find a way to keep the GUI loaders we're currently using and/or make them more user-friendly, but move just the loading logic to `recipes.base.ModuleBase.loadInput()`. Is this what you're suggesting?

barentine:
If we wind up with a '.hdf' file in PYMEVisualize the events mappings should have already been done (when it was loaded in PYMEVisualize last time, or with `pipelineify` in bakeshop, etc.), right? Even for the 'hdf' loading where we pipe every node we find into the namespace, we have an exception for the `MetaData` table where we add a metadatahandler to each/all nodes we find. I think it would be pretty sane to do the same with Events - otherwise they shouldn't be in the same file, right? At the moment I think we do a good job of not propagating 'Events' in recipe modules, so they should really only be there upstream of the first-non `pipelineify` function for points, and upstream of any e.g. frame-averaging module on the image side.

I think it makes sense to take 'Events' out of the namespace/remove it as an input to `pipelineify`. If we add a module to add events to a namespace item and a gui menu item to 'Load Events' which pops that module in upstream of `pipelineify` I think we are golden and people can still hack in spoof'd events if they need to, which would be a cleaner solution than what we did with the labview spooling hack before where we have a (still present) special case in `pipeline._processEvents` to fake events from metadata.

David-Baddeley:
@zacsimile - you've got it on your last comment. No desire to put the loading itself into recipe modules / traits - we'd keep the current loading GUI(s) but a bunch of the `pipeline.OpenFile()` logic would end up moving into `ModuleCollection.loadInput()` (seeing as the pipeline is now mostly just a wrapper around a recipe).

barentine:
Updates following our discussions:
- `ModuleCollection.loadInput` and `pipeline._ds_from_file` now add events attribute to datasources rather than separately in the namespace. At some point would be nice if these shared some more code, but for now I hope OK as is.
  - haven't tested recipe loading yet
- pixelSizeNM is still the attribute name for `localisations.Pipelineify` so we shouldn't break any recipes but now it displays as `nanometer units` and has help text (which for some reason is always prepended with 'Specifies'.
Screen Shot 2020-06-26 at 2 56 14 PM


After I test recipe loading, is there anything else we need to do before merging? @David-Baddeley are you OK with skipping the mapping filter in `pipeline.addDataSource`? What is that used for if not downstream of `Pipelineify`? As mentioned, my goal would be to keep all mapping/filter/other tabular wrappings in the pipeline itself so `Rebuild` does a complete rebuild. I assume we keep some of the potential events refactoring/class for later.

barentine:
To summarize where I think we're at with this:
- we agree we should keep the actual loading out of the pipeline, but all of the mapping stuff should be in the pipeline
- we agree pixelSizeNM can be a confusing name, but not problematic to have in the PYMEVisualize data pipeline if it displays under a different name and has help-text (now accomplished)
- we agree events as an item in the (pipeline) recipe namespace wasn't the cleanest/hindered reproducibility, and might as well be handled as a datasource attribute which should be added during loading, as saving events for a file which they don't belong to would be obviously gross/buggy/avoided

I think we've resolved all of that.

The one thing which might be outstanding is whether we should add the most upstream datasource using the `pipeline.addDataSource` method. The thinking was that using this method enhances reproducibility/recipe portability, however dodging it and injecting directly into the namespace was actually my attempt to do just that because it can add a mapping filter which won't get rebuilt when the pipeline is rebuilt (something that would be nice for live-updating tabular results source like in #347), and won't be there if we save the PYMEVisualize recipe and then try and run it in the bakeshop. I would hate to see this hold up the PR for another month as it gets harder to come back to as it ages, and would happily fix this however you like @David-Baddeley. That said, I see three options which I'll express my views on:
1. Leave as is with the injection - obviously controversial, and not my preferred approach either
2. Use `pipeline.addDataSource` as-is, and be left with a part of the (philosophical) pipeline which doesn't rebuild and a PYMEVisualize recipe which potentially fails to port to bakeshop for some use cases (in practice though, I think we've made sure recipe modules that need a MappingFilter force a wrapping).
3. Change `pipeline.addDataSource` so that if it deems it necessary to wrap a new datasource in a `MappingFilter` it does so by adding a recipe module to the pipeline recipe so that it gets rebuilt live. I think this is pretty pointless, as the need for the mapping filter is taken care of by moving `Pipelineify` into the pipeline recipe, ie this PR. Do I miss something?
4. Change `pipeline.addDataSource` to not worry about adding a `MappingFilter`. As mentioned in (3), I don't see the point of this if we have `Pipelineify` in the pipeline recipe.

barentine:
Going to bump this again as it's getting rather old and I ran into another instance of needing to change a recipe to pass it off to a collaborator. Where are we at with it @David-Baddeley?

barentine:
I know we've got stuff going on, but going to keep bumping this each time it's a friction point for me in lab.

PYME/LMVis/pipeline.py
```diff
@@ -867,6 +858,8 @@ def OpenChannel(self, filename='', ds=None, channel_name='', **kwargs):
         if ds is None:
             #load from file
             ds = self._ds_from_file(filename, **kwargs)
+            self.mdh.copyEntriesFrom(ds.mdh)
```
David-Baddeley: `OpenChannel` is specifically for adding / opening an additional channel (after the main data is loaded). It should **not** overwrite the pipeline level metadata and/or events.

PYME/LMVis/pipeline.py
```diff
@@ -676,15 +677,17 @@ def _ds_from_file(self, filename, **kwargs):
...
             for t in h5f.list_nodes('/'):
+                if t.name == 'Events':
+                    continue
                 if isinstance(t, tables.table.Table):
                     tab = tabular.HDFSource(h5f, t.name)
                     self.addDataSource(t.name, tab)
                         
-                    if 'EventName' in t.description._v_names: #FIXME - we shouldn't have a special case here
-                        self.events = t[:]  # this does not handle multiple events tables per hdf file
+                    if 'EventName' in t.description._v_names:
```
David-Baddeley:
```suggestion
                    if 'EventName' in t.description._v_names: #FIXME - we shouldn't have a special case here
```

PYME/LMVis/pipeline.py
```diff
@@ -676,15 +677,17 @@ def _ds_from_file(self, filename, **kwargs):
             self.filesToClose.append(h5f)
 
             for t in h5f.list_nodes('/'):
+                if t.name == 'Events':
```
David-Baddeley, line 680: Presumably because the 'Events' table has already been processed???

PYME/LMVis/pipeline.py
David-Baddeley: In addition to injection worries, this has the potential to be unsafe with some legacy code which expects the pipeline input to be a mapping filter. May not be a huge issue, and probably fixable with comments for now. Suggest we use the "magic" `'input'` name here to be a 1:1 mapping onto the bakeshop.
```suggestion
        self.dataSources['input'] = ds
```

PYME/LMVis/pipeline.py
```diff
@@ -750,9 +755,10 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
         if 'zm' in dir(self):
             del self.zm
         self.filter = None
-        self.mapping = None
+        # self.mapping = None
```
David-Baddeley:
```suggestion
        self.mapping = None
```

We *should* reset the mapping when reloading data

PYME/LMVis/pipeline.py
```diff
@@ -750,9 +755,10 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
         self.colourFilter = None
         self.events = None
+        self.driftInputMapping = None
```
David-Baddeley:
```suggestion
```
I don't think this is necessary

PYME/LMVis/pipeline.py
```diff
@@ -767,64 +773,49 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
-            mapped_ds.setMapping('x', 'x*pixelSize')
-            mapped_ds.setMapping('y', 'y*pixelSize')
-
-        #extract information from any events
-        self.ev_mappings, self.eventCharts = _processEvents(mapped_ds, self.events, self.mdh)
+        # skip the MappingFilter wrapping, etc. in self.addDataSource and add this datasource as-is
+        self.dataSources['results_source'] = ds
 
+        add_pipeline_variables = Pipelineify(self.recipe, 
+            inputFitResults='results_source',
```
David-Baddeley:
```suggestion
            inputFitResults='FitResults',
```

PYME/LMVis/pipeline.py
David-Baddeley: On second thoughts, bakeshop loads data in based on table name - we should probably use `'FitResults` here to match.

```suggestion
        self.dataSources['FitResults'] = ds
```

PYME/recipes/base.py
```diff
@@ -918,37 +923,41 @@ def _inject_tables_from_hdf5(self, key, h5f, filename, extension):
...
+                # sending file over clusterIO inside of a context manager -> force it through since we already found it
+                im = ImageStack(filename=filename, mdh=mdh, events=events, haveGUI=False)
                 # assume image is the main table in the file and give it the named key
                 self.namespace[key] = im
 
     def loadInput(self, filename, key='input'):
         """
         Load input data from a file and inject into namespace
         """
+        #modify this to allow for different file types - currently only supports images
```
David-Baddeley:
```suggestion
```

patently untrue - supports tabular as either .h5r or .hdf. Raises an error on .csv and .xlsx.

PYME/recipes/base.py
```diff
@@ -918,37 +923,41 @@ def _inject_tables_from_hdf5(self, key, h5f, filename, extension):
...
+
         extension = os.path.splitext(filename)[1]
         if extension in ['.h5r', '.h5', '.hdf']:
             import tables
             from PYME.IO import h5rFile
             try:
                 with unifiedIO.local_or_temp_filename(filename) as fn, h5rFile.openH5R(fn, mode='r')._h5file as h5f:
                         self._inject_tables_from_hdf5(key, h5f, fn, extension)
-            except tables.exceptions.HDF5ExtError:  # access issue likely due to multiple processes
+            except tables.exceptions.HDF5ExtError as e:  # access issue likely due to multiple processes
```
David-Baddeley:
```suggestion
            except tables.exceptions.HDF5ExtError:  # access issue likely due to multiple processes
```

PYME/recipes/base.py
```diff
@@ -969,7 +978,7 @@ def loadInput(self, filename, key='input'):
                         self._inject_tables_from_hdf5(key, h5f, filename, extension)
                 else:
                     #not a cluster file, doesn't make sense to retry with cluster. Propagate exception to user.
-                    raise
+                    raise e
```
David-Baddeley:
```suggestion
                    raise
```

PYME/recipes/localisations.py
```diff
@@ -160,6 +190,11 @@ def execute(self, namespace):
 
             if 'LatGaussFitFR' in fitModule:
                 mapped_ds.addColumn('nPhotons', pipeline.getPhotonNums(mapped_ds, mdh))
+            
+            if 'SplitterFitFNR' in fitModule:
+                mapped_ds.addColumn('nPhotonsg', getPhotonNums({'A': mapped_ds['fitResults_Ag'], 'sig': mapped_ds['fitResults_sigma']}, self.mdh))
```
David-Baddeley:
```suggestion
                mapped_ds.addColumn('nPhotonsg', pipeline.getPhotonNums({'A': mapped_ds['fitResults_Ag'], 'sig': mapped_ds['fitResults_sigma']}, self.mdh))
```

PYME/recipes/localisations.py
```diff
@@ -160,6 +190,11 @@ def execute(self, namespace):
...
+                mapped_ds.addColumn('nPhotonsr', getPhotonNums({'A': mapped_ds['fitResults_Ar'], 'sig': mapped_ds['fitResults_sigma']}, self.mdh))
```
David-Baddeley:
```suggestion
                mapped_ds.addColumn('nPhotonsr', pipeline.getPhotonNums({'A': mapped_ds['fitResults_Ar'], 'sig': mapped_ds['fitResults_sigma']}, self.mdh))
```

PYME/LMVis/pipeline.py
David-Baddeley:
```suggestion
```

PYME/LMVis/pipeline.py
```diff
@@ -867,6 +858,8 @@ def OpenChannel(self, filename='', ds=None, channel_name='', **kwargs):
         if ds is None:
             #load from file
             ds = self._ds_from_file(filename, **kwargs)
+            self.mdh.copyEntriesFrom(ds.mdh)
+            self.events = ds.events
```
David-Baddeley:
```suggestion
```

PYME/recipes/base.py
```diff
@@ -905,6 +905,11 @@ def _inject_tables_from_hdf5(self, key, h5f, filename, extension):
             logger.warning('No metadata found, proceeding with empty metadata')
             mdh = MetaDataHandler.NestedClassMDHandler()
         
+        try:
+            events = h5f.root.Events[:]
```
David-Baddeley, line 909: This breaks our compatibility with generic HDF5 - One of the premises of our .hdf loading in recipes is that it should be able to load fairly generic HDF5 - e.g. from MATLAB etc .... and that (as opposed to the .h5r format) we should avoid 'special' table naming conventions whenever possible. It is not at all out of the question that someone could use an 'Events' table to store localisations.

It is probably sufficient to a) check for a table called 'Events' and b) check that it has the right signature / column names (which should stop things blowing up if someone has used the 'Events' name for something else).

PYME/recipes/localisations.py
```diff
@@ -147,7 +168,16 @@ def execute(self, namespace):
             mapped_ds.setMapping('y', 'y*pixelSize')
 
         #extract information from any events
-        events = namespace.get(self.inputEvents, None)
+        if self.inputEvents != '':
+            logger.warn('Having Events in the recipe namespace is deprecated, they should be associated with datasources during load, or with PYME.recipes.fudges.AddEvents')
+            events = namespace[self.inputEvents]
```
David-Baddeley:
```suggestion
            events = namespace.get(self.inputEvents, None)
```

PYME/recipes/localisations.py
```diff
@@ -147,7 +168,16 @@ def execute(self, namespace):
             mapped_ds.setMapping('y', 'y*pixelSize')
 
         #extract information from any events
-        events = namespace.get(self.inputEvents, None)
+        if self.inputEvents != '':
+            logger.warn('Having Events in the recipe namespace is deprecated, they should be associated with datasources during load, or with PYME.recipes.fudges.AddEvents')
```
David-Baddeley, line 172:
```suggestion
            # Use specified table for events if given (otherwise look for a `.events` attribute on the input data
            # TODO: resolve how best to handle events (i.e. should they be a separate table, or should they be attached to data tables)
```

I know we discussed this, but don't believe there was a final conclusion. There are a couple of competing considerations here - having events in the namespace results in clutter, but is conversely the most generic way of handling them - particularly if we want to minimise custom file formats. There is also a general ugliness with attaching attributes to tabular data (and deciding if and when to propagate them). 

As I recall, I think I was tending to not really wanting them in the VisGUI namespace (to avoid them appearing as a potential output and the confusion this would generate), but I don't think this automatically extended to recipes and the bakeshop. I definitely don't think that events should be propagated by default in recipe modules (which makes the order of modules somewhat critical - i.e. `Pipelineify` must come before any other modules - a behaviour which is a little counterintuitive). A reasonable compromise is to handle both for now.

barentine: Once events are loaded and you're ready to inject them into the namespace there is no file format, which i think fundamentally invalidates that concern. They can be loaded however you like, but what do Events mean without an associated datasource, and what happens when you have e.g. localizations and a widefield stack with different events in the same recipe? I think events floating without association are fundamentally as weird as injecting a metadata handler into the recipe namespace. Propagation wise they should be treated similarly too - if the events are invalidated by e.g. averaging frames / changing the length of an image stack, the output imagestack should either have spoofed events or no events.

PYME/recipes/base.py
barentine: seems like overkill to me, but I'm OK with it. Will try and get to implementing this soon but might be tomorrow

David-Baddeley: I'm also happy to crack on some of these - a lot of this was just notes to myself, some of which might be better transferred to comments. My next step was going to be to pull the branch down locally.

PYME/recipes/localisations.py
David-Baddeley: Putting them into the namespace has some advantages re different events in the same recipe and/or events with different file backends (the logical way of handling, e.g. a .tif stack with events or a .dcimg with events would be to put the events in an accompanying .json). If we strictly attach events to datasources on load we would need to follow some pretty prescriptive file naming patterns and have a lot of logic in the loaders to cope with that. If we can just load a json in separately as an events table (different events tables can have different names) then the custom logic for a special case config resides in the recipe that processes it. 

Not saying that this is ideal, but attaching them to data also has it's issues (one of which is how it impacts serialisation). Philosophically I'd prefer it if people didn't need to think about events when writing / using recipe modules - apart from the one or two cases where you need them. This argues against having events be an expected attribute of a tabular object.

barentine: As you said, we both know we've discussed this, and as mentioned [above in this PR discussion](https://github.com/python-microscopy/python-microscopy/pull/284#discussion_r444630231) I think it would be perfectly reasonable and sane to write an `AddEvents` module to associate events with a datasource. I don't think it's wild to ask a user to shim in their tif/dcimg acquisition events with a specific filename or manually select/indicate it. This is what we did for high-throughput dcimgspooling.

David-Baddeley:
I'm going to try and merge this on top of #472 and make a couple more tweaks before merging with master

barentine:
Cool, thanks for having a go at that @David-Baddeley

barentine:
Just a note that the `namespace.get(self.inputEvents, None)` doesn't super work because inputEvents is an `Input` type rather than `CStr` so if someone e.g. has an old recipe with `inputEvents = 'Events'` they'll now hit
```
Traceback:
----------
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/taskWorkerHTTP.py", line 340, in computeLoop
    recipe.execute()
  File "/home/ubuntu/code/python-microscopy/PYME/recipes/base.py", line 640, in execute
    m.check_inputs(self.namespace)
  File "/home/ubuntu/code/python-microscopy/PYME/recipes/base.py", line 137, in check_inputs
    raise MissingInputError('Input "%s" is missing from namespace; keys: %s' % (input, keys))
PYME.recipes.base.MissingInputError: Input "Events" is missing from namespace; keys: ['FitResults']
```
though of course they can update the recipe simply by dropping the `inputEvents` part

PR #285, opened 2020-06-23
Allow recipes to queue ActionManager acquisition tasks

barentine:
Enhancement

**Proposed changes:**
The driving goal for this PR and a series of others is to be able to start an acquisition which automatically starts analysis which itself can automatically queue more acquisition tasks.

The focus of this PR is to be able to queue acquisition tasks from recipes. In order to achieve this, we'll need a server process for the `ActionManager`. Additionally, one of the most common tasks we'll want to queue is `SpoolController.StartSpooling`, and if we are mixing different types of acquisitions we may well want to change settings on the spool controller. Adding a server process for the `SpoolController` to change settings is most definitely the wrong way to tackle this as not strictly associating any spool controller settings with the tasks allows for plenty of chaos/spooling with settings for the wrong task. A simpler alternative is to allow keyword arguments in `StartSpooling` to update the spool controller settings, which is already in the direction recent `SpoolController` refactoring were headed. 

- wrap `ActionManager` with a server process so we can queue acquisition actions from the cluster (in recipe modules to follow)
- refactor `SpoolController.StartSpooling` to accept/pass `SpoolController` settings to its `update_settings` method. 
- short-circuit `SpoolController.update_settings` method change such that we don't call _update_series_count and check_exists through an entire spooled directory each time we start spooling a new series
- allow all startspooling kwargs as well as spool controller settings to be passed in `SpoolControllerWrapper.start_spooling` endpoint.
- add standalone server process for the spoolcontroller wrapper rather than only as an `acquire_Server` endpoint. This was predominantly for testing the `UpdateSpoolerSettings` recipe module which is not really necessary, but I already wrote it and maybe it isn't crazy to have the class there.
- add recipe module to change spooler settings. As mentioned in its docstrings this is really not the preferred way of passing spooler settings from recipes, but it's there, tested, and maybe useful later
- add recipe module to queue acquisitions at specified ROI positions. Should be sufficiently flexible in terms of protocol definitions, sorting, etc.

**Checklist:**

- [ ] Tested with numpy=1.14

Well, it did, but then when I changed the signature for StartSpooling to allow the kwarg stuff I also changed `fn` to `filename`, which I don't think is terribly bold.

If I had a mailing list I would tell people who want to use it to pop it in with their ActionManager /panel init
```
@init_gui('Action manager')
def action_manager(MainFrame, scope):
    from PYME.Acquire.ui import actionUI
    from PYME.Acquire.ActionManager import ActionManagerServer

    ap = actionUI.ActionPanel(MainFrame, scope.actions, scope)
    MainFrame.AddPage(ap, caption='Queued Actions')
    ActionManagerServer(scope.actions, port=9001, bind_address='')
```


PYME/recipes/acquisition.py
```diff
@@ -0,0 +1,160 @@
+
+from .base import register_module, ModuleBase, OutputModule
+from .traits import Input, Output, CStr, Int, DictStrAny, Bool, Float, ListFloat
+import requests
+import json
+import numpy as np
+
+@register_module('UpdateSpoolerSettings')
+class UpdateSpoolerSettings(ModuleBase):
```
David-Baddeley: This is a non-starter as a recipe module as it has side effects and is not a subclass of `OutputModule`. 

The solution we discussed had us passing any settings updated to the spool controller through the `StartSpooling` method in the action manager.

PYME/Acquire/SpoolController.py
```diff
@@ -482,13 +522,43 @@ def stop_spooling(self):
...
     @webframework.register_endpoint('/start_spooling', output_is_json=False)
-    def start_spooling(self, filename=None, max_frames=sys.maxsize):
-        self.spool_controller.StartSpooling(fn=filename, maxFrames=max_frames)
+    def start_spooling(self, body='{}'):
+        settings = json.loads(body)
+        self.spool_controller.StartSpooling(**settings)
         return 'OK'
+
+
+class SpoolControllerServer(webframework.APIHTTPServer, SpoolControllerWrapper):
```
David-Baddeley: There should be no need for recipes to interact directly with the spool controller. This is super unsafe in terms of race conditions. The one and only place where recipes should interact with the microscope is through queueing actions.

PYME/Acquire/SpoolController.py
```diff
@@ -265,41 +297,49 @@ def _get_queue_name(self, fn, pcs=False):
...
+    def StartSpooling(self, filename=None, max_frames=sys.maxsize, 
+                      do_preflight_check=True, **kwargs):
         """
+
+        Parameters
+        ----------
+        filename : str, optional
+            filename can be hardcoded here, otherwise differs to the seriesName
+            property which will create one if need-be.
+        max_frames : int, optional
```
David-Baddeley: variable name changes violate contributing guidelines and change API.

PYME/Acquire/SpoolController.py
```diff
@@ -265,41 +297,49 @@ def _get_queue_name(self, fn, pcs=False):
...
-        # but default to using our internal values
-        compLevel = self.hdf_compression_level if compLevel is None else compLevel
-        pzf_compression_settings = self.pzf_compression_settings if pzf_compression_settings is None else pzf_compression_settings
-        stack = self.z_stepped if stack is None else stack
-        cluster_h5 = self.cluster_h5 if cluster_h5 is None else cluster_h5
-        fn = self.seriesName if fn in ['', None] else fn
-        zDwellTime = self.z_dwell if zDwellTime is None else zDwellTime
+        """
+        filename = self.seriesName if filename in ['', None] else filename
+        self.update_settings(kwargs)
```
David-Baddeley: I need to think a bit more about whether we should be updating the settings, or just over-riding with anything we supply. 

In principle I'd prefer not to update the spool controller settings - i.e. calling `StartSpooling()` with custom settings should over-ride the settings for this call but not change behaviour for subsequent calls which don't provide settings. 

The caveat here would relate to your use case of changing the spool directory for different wells, but this might be able to be handled separately.

PYME/Acquire/ActionManager.py
```diff
@@ -121,5 +125,83 @@ def Tick(self, **kwargs):
...
+        import json
+        params = json.loads(body)
+        function_name = params['function_name']
+        args = params.get('args', {})
+        nice = params.get('nice', 10.)
+        timeout = params.get('timeout', 1e6)
+        self.action_manager.QueueAction(function_name, args, nice, timeout)
+
+
+class ActionManagerServer(webframework.APIHTTPServer, ActionManagerWebWrapper):
```
David-Baddeley, line 213: I don't like the idea of having multiple different servers running within PYMEAcquire, both due to the overhead, and the code duplication it entails. Medium term this will be addressed by having having `acquire_server` running the whole time (even with the wx frontend).

We can possibly have this here short term, but should be flagged that it will dissapear.

PYME/Acquire/SpoolController.py
```diff
@@ -482,13 +522,43 @@ def stop_spooling(self):
         return 'OK'
 
     @webframework.register_endpoint('/start_spooling', output_is_json=False)
-    def start_spooling(self, filename=None, max_frames=sys.maxsize):
-        self.spool_controller.StartSpooling(fn=filename, maxFrames=max_frames)
+    def start_spooling(self, body='{}'):
```
David-Baddeley: this completely changes the signature. How does this effect the web front end? Is there a strong argument for getting rid of the query arguments? Using query arguments makes it much easier to debug and/or write an HTML frontend for.

PYME/recipes/acquisition.py
barentine: I figured we'd cut it, but left it in until we discussed as I'd written/debugged it before we came to the conclusion we needed to change `SpoolController.StartSpooling` signature. Will take it out

PYME/Acquire/SpoolController.py
barentine: We are changing the entire signature - while we do it is it not a good idea to match what you have in `SpoolControllerWrapper.start_spooling`?

PYME/Acquire/SpoolController.py
barentine: I think if we're in a situation where we're doing things that involve changing things between acquisitions, we are probably also in a situation where we'd better pair settings with every startspool task. Should mention I need to change the protocol too (and with chained analysis stuff hope to not need the directory change forever).

barentine: I didn't phrase that well, but I think once we have mixed acquisition types there are no 'normal' settings to return to. The overview/smlm thing is a pretty weighted case, but say I do a multicolor stack in between the overview and SMLM to decide whether a ROI is worth SMLM, which settings do we want the spoolcontroller to revert to each time?

PYME/Acquire/ActionManager.py
barentine: OK, I'll flag it for now. Running the entire acquire_server the whole time felt like a big leap but maybe I overestimated it. Worth noting that combining the offsetpiezo stuff with the acquire server might also be nice re:reducing overhead.

PYME/Acquire/SpoolController.py
barentine: I tested it with the web front end; as far as I could tell there's no change.

I googled and it sounded like once you're passing a dictionary via query you might as well pass a dictionary. It also makes the wrapper pass-through, which I think is nice. I did not have HTML frontend in mind, nor do I have experience with it, as you know

PYME/recipes/acquisition.py
barentine: This is now removed.

PYME/Acquire/ActionManager.py
barentine: added the note.

PYME/Acquire/SpoolController.py
David-Baddeley: In terms of web front-end, putting things in a json encoded body means that you can only call it via javascript (and will need to do some javascript munging to get the parameters into json). If you keep things with query arguments, you can easily make it the target of a link or button without doing anything in js. You can also test it out by simply pointing your web-browser to the endpoint with the relevant arguments in the URL. Past a certain complexity it makes sense to format things as json, but I'm not sure that applies for most start_spooling() calls.

David-Baddeley:
I think it would help to go over our blocking use cases on the spool controller again. As I understand it, you want to:

- queue actions to take overviews of each well in a multi-well plate
- have analysis on those overviews queue acquisitions (with higher priority so they execute before the next tiling run)?

and that this would involve changing between local spooling and cluster spooling.

As it stands, tiling is separate to the spool controller. Are you proposing to integrate it (how?)

PYME/Acquire/SpoolController.py
barentine: Opening a python console and importing requests is pretty minimal overhead for testing. json standing for JavaScript Object Notation I would have assumed it would be pretty easy to work with in javascript.

In practice, I'm not going to lose sleep over this and if you prefer query that's OK by me, but my opinion is that passing all 8 of our kwargs (since we don't know the state of the SpoolController when our tasks becomes active) is not very clean.

barentine:
I was planning on queueing up a `tile_triggered.py`-esque protocol for each well, which would use the spool controller, chained to a rule for detection. N + 1 would be scanning while detection is running on well N's overview, and after the overview for N + 1 (or N + 2 depending on speed / imaging / etc) the low-Nice acquisitions from N would run. As large/many overviews becomes more common manually transferring them off my local hard-drive to make room for the next isn't ideal compared to spooling with quantization to the cluster. My first use case here also shouldn't limit of our scope - I'm excited to also e.g. overview -> multicolor widefield stack -> decision -> SMLM

That said, if we keep the kwarg only rather than dict->update_settings approach, I think the only practical things I couldn't do with the current signature are changing protocol and changing spool directory. Spool directory is kind of a luxury that I'm not very concerned with - if we add a kwarg to set the protocol in StartSpooling I think I'd be set for the foreseeable future. Would that be preferable to the update_settings approach?

PYME/Acquire/SpoolController.py
barentine: I think you're right that most calls will be much less than all kwargs in StartSpooling, but if one doesn't know the state of the spoolcontroller they might fill out quite a few each time. If that's OK then I'll revert it to kwargs. Do you want me to add the rest of the StartSpooling kwargs (pep8 on the wrapper to current kwargs on the `self.spool_controller.StartSpooling` call) while I'm at it?

barentine:
@David-Baddeley I forgot to check back in on this but I backed out the update_settings inside of StartSpool and added a protocol kwarg so I can do everything I need during StartSpooling in a transient way wrt the spool controller settings. serverwrapper start_spooling is now query based again rather than body.

PR #287, opened 2020-06-24
recipes.base cleaning/docs

barentine:
Not trying to change code, but noticed there was no `OutputModule.save` method on the base class while I was making my coffee and did some cleaning. Won't be offended if it is unwelcome, it's mostly obnoxious but e.g. docstrings on the arithmetic filters could be helpful for users.

enhancement
**Proposed changes:**

cleaning without functional changes:
- add (NotImplemented) `save` method to `OutputModule` base class
- add docstrings for the arithmetic filters (could be helpful for, e.g. which is being subtracted from which, etc.)
- add typing for a couple parameters in existing docs
- pep8 one line between methods two lines between classes
- move dispatch import up top

**Checklist:**

- [ ] Tested with numpy=1.14


PYME/recipes/base.py
```diff
@@ -1134,78 +1191,131 @@ def _joinChannels(self, namespace):
...
-    """Add two images"""
+    """
+    Add two images
     
+    Parameters
+    ----------
+    inputName0: PYME.IO.image.ImageStack
+    inputName1: PYME.IO.image.ImageStack
+    outputName: PYME.IO.image.ImageStack
+    processFramesIndividually: Bool
```
David-Baddeley: We need to find a way of globally inheriting the docs for `processFramesIndividually`

Importantly, broadcasting is not officially supported and cannot be guaranteed going forwards (we should probably add an `assert(input1.shape==input2.shape`) somewhere in `ArithmaticFilter`. 

`processFramesIndividually` is a performance option which can lead to less (and lower latency) memory usage and potentially IO. For the arithmetic (sp) filters it should have no effect on the output. For other filters (e.g. Gaussian) also use it to indicate that we should apply a 2-d version of the filter (although this was not the original intent, and is a bit gross).

David-Baddeley: Medium/long term (once we get to a 5D data model and dask integration etc) `processFramesIndividually` should probably go away in favour of minimum block size hints, and an`applyToDimensions` (or similar) enum with options such as 'xy', 'xyz', 'xyzt' etc ...

barentine: For things where processFramesIndividually doesn't change the output could we deprecate the option, and either go with the lower ram/io version always or test roughly for a condition which says which will be faster?

That said I actually really like the ability to broadcast (speaking of lower ram/io usage); why would we want to disable it?

David-Baddeley: I can also see arguments for allowing broadcasting, but for now it is an unexpected bonus and it does constrain us somewhat in terms of how we change things going forward [we'd need to re-implement broadcasting for the chunked code]. If we do end up adding broadcasting support, it should not be controlled by `processFramesIndividually` (ie it should work regardless of whether this is checked or not).

barentine: People with very large data might be grateful for the attention to io here. It seems like one way to fix the `processFramesIndividually` double-meaning would be to remove it for functions where it doesn't (/shouldn't) change the output and instead make the decision for users based on the shape of the array, though I suspect that's largely out of the cards until we have a system for deprecating recipe module attributes. For now would you be satisfied with me noting in the docstrings that broadcasting functionality might not be preserved going forward? Do we have a broadcasting recipe module we should recommend calling first instead?

Alternatively, we could throw a `numpy.broadcast_to` call in the chunked code and commit to carrying it forward.

David-Baddeley: I think the best solution for now is probably to leave broadcasting undocumented (and to fix the chunking etc ... ASAP). If we do re-add broadcasting, I suspect it should only apply to arithmetic filters, so that might be the place to address it.

PR #288, opened 2020-06-24
move pr template to where it will get discovered sans query params

barentine:
Now that we have a PR template I've been copy-pasting it into my PRs since it doesn't open automatically. docs are a little [opaque](https://stackoverflow.com/questions/52139192/github-pull-requests-template-not-showing/52139815), but options are to have it named/located where this PR suggests, or do as is currently in the head but you have to somehow open the PR with query-str to specify which pr template to use, which feels a bit overkill unless we add more than one at some point.


barentine:
Note that for some reason the issue stuff plays nicely as is though.
Screen Shot 2020-06-24 at 11 23 43 AM

Alternatively we could try adding a separate PR template for enhancement/bugfix/clean and see what happens? Would also help with making the checkboxes feel more natural for a given type of PR (currently can be a bit weird)

PR #289, opened 2020-06-24
Visgui pylab farewell

zacsimile:
Together, these imports crash the Jupyter notebook kernel:

```
import matplotlib
matplotlib.use('wxagg')
import pylab
```

If we comment out `matplotlib.use('wxagg')` or `import pylab`, everything runs fine. To fix this, I've switched our only use of `pylab` in `VisGUI.py` to an equivalent `matplotlib` call.


zacsimile:
Version control did something wonky.

PR #295, opened 2020-06-28
Readme update

zacsimile:
Adresses issue # .

Enhancement

**Proposed changes:**

- Make readme look more professional by adding our logo, removing the [ ] brackets around links, providing more information, and adding the Azure badge (but I think @David-Baddeley will need to do something on the backend to make sure the badge API is on)

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,6 @@
 # <img alt="PYME" src="PYME/resources/icons/pymeLogo.png" style="float:left"> The PYthon Microscopy Environment
 
-[![Azure](https://dev.azure.com/davidbaddeleynz/pyme-ci/_apis/build/status/azure-pipeline%davidbaddeleynz.pyme-ci)](
-    https://dev.azure.com/davidbaddeleynz/pyme-ci/_build/latest?definitionId=5)
-
-The PYthon Microscopy Environment provides image acquisition and data analysis functionality for a number of  widefield microscopy applications, but with a particular emphasis on single molecule localisation microscopy (PALM/STORM). The package is multi platform, running on Windows, Linux, and OSX.
+PYME (the PYthon Microscopy Environment) is an open-source application suite for light microscopy image acquisition, data storage, visualization, and analysis. It offers a particularly comprehensive set of features for single molecule localisation based super-resolution techniques (PALM, STORM, PAINT, etc ...), but also has good support other microscopy modalities. The package is multi platform, running on Windows, Linux, and OSX.
 
 - **Website:** https://www.python-microscopy.org
 - **Installation:** https://www.python-microscopy.org/doc/Installation/InstallationWithAnaconda.html
--- a/README.md
+++ b/README.md
@@ -2,6 +2,14 @@
 
 PYME (the PYthon Microscopy Environment) is an open-source application suite for light microscopy image acquisition, data storage, visualization, and analysis. It offers a particularly comprehensive set of features for single molecule localisation based super-resolution techniques (PALM, STORM, PAINT, etc ...), but also has good support other microscopy modalities. The package is multi platform, running on Windows, Linux, and OSX.
 
+<!-- ![license](https://img.shields.io/github/license/python-microscopy/python-microscopy) -->
+![tag](https://img.shields.io/github/v/tag/python-microscopy/python-microscopy?sort=semver)
+![commits](https://img.shields.io/github/commit-activity/m/python-microscopy/python-microscopy)
+
+![conda](https://img.shields.io/conda/v/david_baddeley/python-microscopy.svg)
+![plat](https://img.shields.io/conda/pn/david_baddeley/python-microscopy.svg)
+
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,11 +1,15 @@
...
+# <img alt="PYME" src="PYME/resources/icons/pymeLogo.png" style="float:left"> The PYthon Microscopy Environment
 
-The PYthon Microscopy Environment provides image acquisition and data analysis functionality for a number of  widefield microscopy applications, but with a particular emphasis on single molecule localisation microscopy (PALM/STORM). The package is multi platform, running on Windows, Linux, and OSX.
-
-For more info see the project website at [http://www.python-microscopy.org].
+[![Azure](https://dev.azure.com/davidbaddeleynz/pyme-ci/_apis/build/status/azure-pipeline%davidbaddeleynz.pyme-ci)](
+    https://dev.azure.com/davidbaddeleynz/pyme-ci/_build/latest?definitionId=5)
 
-For installation instructions see [http://www.python-microscopy.org/doc/Installation/InstallationWithAnaconda.html]
+The PYthon Microscopy Environment provides image acquisition and data analysis functionality for a number of  widefield microscopy applications, but with a particular emphasis on single molecule localisation microscopy (PALM/STORM). The package is multi platform, running on Windows, Linux, and OSX.
```
barentine: Every time I read this I feel like we start by putting ourselves in a box. What do we think of keeping the description short and leaving details of what we currently have to subsections of e.g. PYMEAcquire where we say we currently only support camera-based systems, etc.? When I got started with PYME the only people I knew who used it were deconvolving STED images. I think our one-sentence description could stand to be a little less constrained.

PYME (PYthon Microscopy Environment) is a collection of programs which together constitute a seamless pipeline of image acquisition, data storage, visualization, and analysis for (light?) microscopy applications. PYME components can be used individually if desired, and run on Windows, Mac, and Linux.

zacsimile: How about shorter?

PYME (PYthon Microscopy Environment) is an open-source application suite for light microscopy image acquisition, data storage, visualization, and analysis.

David-Baddeley:
removed the azure badge for now, as a) it's broken and b) I don't think it's that informative. I'd add a bunch of other badges (e.g. python version badges) first.

zacsimile:
Numpy proudly displays a broken Azure badge: https://github.com/numpy/numpy/blob/master/README.md. CellProfiler displays a testing badge: https://github.com/CellProfiler/CellProfiler. I think it's the hip thing to do and will make us look more legit.

David-Baddeley:
I know it's the 'hip' thing to do, but badges are useful only in as much as they actually convey information to the user. The azure-pipelines badge is particularly unhelpful/uninformative - we currently use azure pipelines for testing, but they can also be used for scheduled builds etc ... I'm open to putting it back in at some point, but It should really link up with something on the backend, and ideally also indicate that it's tests.

Badges I'd be keen on (in order of importance/priority):

- a python versions badge (see, e.g. https://github.com/google/pybadges)
- a license badge
- a current version badge
- a badge (or matrix of badges) for the most recent version that is conda-installable
- ditto for pip
- a testing badge (sort of the azure badge, but identifying that this refers to the tests)
- a build badge (once we have automated builds rather than just automated testing)

zacsimile:
Hi David, I think that's reasonable. I'd say based on what I've seen from other repositories, we may want to stick to 3 badges or less in the README. In that case, I would prioritize a CI badge like Azure or something like codecov higher. People like to see CI highlighted, even if it's not particularly helpful to the user. I don't make the rules. If we ever get on image.sc, it would be nice to put their badge on, too (see https://github.com/CellProfiler/CellProfiler/blob/master/README.md).

README.md
barentine: I like it

barentine:
I think a CI badge, broken or not, tells me the maintainers care and are committed. I've never cared whether it was full CI or just tests; it's an icon, like the 'flex fuel' or 'hybrid blue' badges on cars - it's more about branding than informing you e.g. how much corn it takes to replace a gallon of gas, or what ~percent hybrid the car is. Honestly I judge more by which CI they use. Maybe I'm petty. Displaying a broken badge might also help motivate us to fix the testing when it breaks.

A Python version badge wouldn't be awful, but I'd strongly argue for including 3.6. If you're not comfortable with that yet @David-Baddeley I think we should hold off - a 2.7 only badge in mid 2020 is not so hip.

David-Baddeley:
python badge was going to be 2.7|3.6|3.7 (but might have to wait for pip / and/or  more readme expansion. Have the devops badge working  now (with more info than a simple pass/fail)

zacsimile:
I like the badges you settled on!

barentine:
And the description - thanks both of you!

zacsimile:
OK, I'm wrong about sticking to 3 badges. See https://github.com/napari/napari.

Issue #216, opened 2020-04-29
Naming inconsistencies in pyme modules


zacsimile:
We're introducing `pymeclusterofone` as yet another module command to launch elements of PYME. This is inconsistent with the other module names, which are all inconsistent with one another. To make entry points easier for new users, I propose the launch names be as follows.

```
PYMEAcquire -> pyme-acquire
VisGUI -> pyme-visualize
dh5view -> pyme-image
pymeclusterofone -> pyme-cluster
launchWorkers -> pyme-404
```

It would be best to change `pyme-visualize` to `pyme-points` since both `pyme-visualize` and  `pyme-image` are for visualization and analysis, just of different data types.

David-Baddeley:
It's a work in progress ... unfortunately hyphens in executable names aren't safe across platforms. 

All new entry points (including all the cluster ones) follow the pattern:
`PYMECommand`

`PYMEClusterOfOne` can't /shouldn't change to `PYMECluster` as we should be distinguishing between it and an actual cluster

`VisGUI` and `dh5view` definitely need to change (that's been on the TODO list for some time). The main reason it hasn't happened already is that I've been toying with the idea of combining them under a single entry point which would guess the data type and load the appropriate viewer. Do you have any feelings as to whether you think there is actually a need to two separate entry points here?

I'm not super interested in changing launchWorkers, fitMon, taskServerZC etc ... as I expect them to disappear.

David-Baddeley:
ps - I like `PYMEImage` for `dh5view`

zacsimile:
I agree that a single entry point makes the most sense.

I expect them to disappear, too. I was trying to make a 404 Not Found joke with that suggestion. I'll try to refrain from making deadpan jokes in issues from now on, but no promises.

zacsimile:
Addressed in #302.

PR #302, opened 2020-06-30
Alias VisGUI to PYMEVisualize

zacsimile:
Addresses issue #216.

Enhancement

**Proposed changes:**
- Entry point and import aliases to access `VisGUI` as `PYMEVisualize`
- Alias for Jupyter notebook functionality to access `ipython_visgui` as `ipython_pymevisualize`
- Entry point aliases to access `dh5view` as `PYMEImage` (there is no `import dh5view`)

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/zachcm/miniconda3/envs/pyme/bin/python"
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,0 @@
-{
-    "python.pythonPath": "/Users/zachcm/miniconda3/envs/pyme/bin/python"
-}
--- a/PYME/LMVis/__init__.py
+++ b/PYME/LMVis/__init__.py
@@ -38,4 +38,4 @@ The VisGUI can be launched using the `VisGUI` script.
 
 """
 
-from . import VisGUI as PYMEVisualize
+from .VisGUI import ipython_pymevisualize, ipython_visgui
--- a/PYME/LMVis/__init__.py
+++ b/PYME/LMVis/__init__.py
@@ -37,3 +37,5 @@ Plugins are located in :py:mod:`PYME.LMVis.Extras`.
 The VisGUI can be launched using the `VisGUI` script.
 
 """
+
+from . import VisGUI as PYMEVisualize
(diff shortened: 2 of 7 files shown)


zacsimile:
Would it make sense to allow entry points to be imported? Should we add the following lines to `PYME/__init__.py`?

```
import PYME.DSView.dsviewer as PYMEImage
import PYME.Acquire.PYMEAcquire as PYMEAcquire
import PYME.LMVis.VisGUI as PYMEVisualize
etc.
```

PYME/LMVis/__init__.py
```diff
@@ -36,4 +36,6 @@
 
 The VisGUI can be launched using the `VisGUI` script.
 
-"""
\ No newline at end of file
+"""
+
+from . import VisGUI as PYMEVisualize
```
David-Baddeley: Is this to make the notebook imports cleaner?

if so, it would probably be better/sufficient to just do the following  `from .VisGUI import ipython_pymevisualize, ipython_visgui`

Having anything import from `VisGUI` in `LMVis.__init__`  is  currently problematic however when we import modules from `LMVis` in non-gui code (prime candidates are `LMVis.pipeline`, `LMVis.visHelpers`, `LMVis.renderers` etc ...) as importing `VisGUI` imports wx (and also does a bunch of other things - e.g. calling  `logging.basicConfig()`) which have the potential to be problematic.

It's a long-standing TODO item to move some of the backend stuff out of LMVis (or conversely move the UI stuff), but has never felt like it's super high priority.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/Users/zachcm/miniconda3/envs/pyme/bin/python"
```
David-Baddeley: could you remove this file please. Can add .vscode/* to .gitignore if you like

conda-recipes/python-microscopy/menu-windows.json
```diff
@@ -15,7 +15,7 @@
                 "script": "${PREFIX}/Scripts/dh5view.exe",
                 "scriptarguments": [],
 
-                "name": "dh5view",
+                "name": "PYMEImage",
```
David-Baddeley: could we make this have the old name in brackets for a bit? At least until people have gone through one install / update cycle and our documentation is fixed.

.vscode/settings.json
zacsimile: Yep. Sorry about that.

conda-recipes/python-microscopy/menu-windows.json
David-Baddeley: Ditto for VisGUI below. Don't worry about the gnome stuff - it must be 5 years or so since anyone used that and if you are on linux you should be able to figure it out.

zacsimile: Done.

PR #303, opened 2020-06-30
include filename/url in clusterIO timeout/error logging

barentine:
Addresses issue # .

enhancement
**Proposed changes:**
- include filename/url in logging statements

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/IO/clusterIO.py
+++ b/PYME/IO/clusterIO.py
@@ -779,7 +779,7 @@ def get_file(filename, serverfilter=local_serverfilter, numRetries=3, use_file_c
     nTries = 1
     while nTries < numRetries and len(locs) == 0:
         #retry, giving a little bit of time for the data servers to come up
-        logger.debug('Could not find file, retrying ...')
+        logger.debug('Could not find %s, retrying ...' % filename)
         time.sleep(1)
         nTries += 1
         locs = locate_file(filename, serverfilter, return_first_hit=True)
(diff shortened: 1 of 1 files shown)


barentine:
as an aside, I'll update the nodes to the repo head and try again, but it seems like we may have nerf'd something on the dataservers - trying to download a 30k frame series, or even just take a slice through the stack opened in dh5view, is like pulling teeth at the moment, with many retries starting as early as frame 600 and then eventually exceeding the 3 retry count and failing on frames which in fact exist

PYME/IO/clusterIO.py
```diff
@@ -803,10 +803,10 @@ def get_file(filename, serverfilter=local_serverfilter, numRetries=3, use_file_c
         except (requests.Timeout, requests.ConnectionError) as e:
             # s.get sometimes raises ConnectionError instead of ReadTimeoutError
             # see https://github.com/requests/requests/issues/2392
-            logger.exception('Timeout on get file')
+            logger.exception('Timeout on get file %s' % url)
             logger.info('%d retries left' % (numRetries - nTries))
             if nTries == numRetries:
-                raise
+                raise e
```
David-Baddeley: is this strictly needed? I thought that `raise` on it's own in an exception block just re-raised the exception that was being handled (i.e. didn't need the explicit exception argument), and was more idiomatic python.

barentine: Not necessary, nor idiomatic. Same could be said for assigning the exception to an unused variable on 803. Would you like me to revert it?

David-Baddeley: was more a  "wait, did I have this wrong" informational question. Suspect the `except as e` bit was left over from looking at some additional info in the exception object.

barentine: Well, it's on me for changing something I didn't need to, sorry. Went ahead and took out the `as e`.

Issue #313, opened 2020-07-08
[BUG] IDS camera sometimes flakes out


barentine:
**Describe the bug**
This is possibly not a bug with PYME, but I have an IDS camera (UI-3240LE-NIR) running using the `uCam480` class which will occasionally die silently. No errors are raised, but the same frame is displayed and frameWrangler registers (correctly) that we're hitting a whopping 0 FPS. If I restart the microscope, the camera works fine. Additionally, I can try to bushwhack a reinitialization for the camera, which seems to go ~OK (returns a CCD temperature again, etc.) but restarting the frameWrangler afterwards doesn't seem to really achieve anything - the same frame is displayed. A potential solution might be to add some `restart` functionality to `uCam480` - obviously this is less clean than the thing just working, but... I'm interested in hearing any other options.

I don't know if the following shell commands I entered are useful to look at for anyone else or not, but I'm going to go ahead and include them, note these are all after the camera/wrangler froze up:
```
>>> scope.frameWrangler.stop()
>>> scope.frameWrangler.Prepare()
>>> scope.frameWrangler.start()
True
>>> scope.cam
<PYME.Acquire.Hardware.uc480.uCam480.uc480Camera object at 0x0000024069F12EF0>
>>> scope.cam.active
True
>>> scope.cam.CamReady
<bound method uc480Camera.CamReady of <PYME.Acquire.Hardware.uc480.uCam480.uc480Camera object at 0x0000024069F12EF0>>
>>> scope.cam.CamReady()
True
>>> scope.cam.GetFPS()
200.16814123864046
>>> scope.cam.DisplayError
<bound method Camera.DisplayError of <PYME.Acquire.Hardware.uc480.uCam480.uc480Camera object at 0x0000024069F12EF0>>
>>> scope.cam.errcheck
<bound method uc480Camera.errcheck of <PYME.Acquire.Hardware.uc480.uCam480.uc480Camera object at 0x0000024069F12EF0>>
>>> scope.cam.errcheck()
Traceback (most recent call last):
  File "<input>", line 1, in <module>
TypeError: errcheck() missing 2 required positional arguments: 'value' and 'msg'
>>> scope.frameWrangler.getFPS()
0.0
>>> scope.cam.active
True
>>> scope.cam.GetFPS()
200.16814123864046
>>> scope.frameWrangler.stop()
>>> scope.frameWrangler.Prepare()
>>> scope.frameWrangler.start()
True
[...]

PR #315, opened 2020-07-09
Offset piezo / focus lock performance improvements

barentine:
Addresses issue #313, maybe.

enhancement for the sake of bugfix

**Proposed changes:**
- use `request.Session.get` instead of `request.get` for socket keep-alives in `RLPIDFocusLockClient` 
- use `request.Session.get` instead of `request.get` for socket keep-alives in `OffsetPiezoClient` 
- connect focus lock to `frameWrangler.onFrameGroup` rather than `onFrame`
- make `frameWrangler` polling interval an attribute so we can change it if we like
- change `frameWrangler._polling_interval` in init_htsms_focus_lock
- move frameWrangler signal doc into docstring and clarify the part that I misunderstood earlier
- add minimal (typing) docstrings to `microscope` and `GUIInitTask` to aid linter/IDE shortcuts

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/frameWrangler.py
+++ b/PYME/Acquire/frameWrangler.py
@@ -54,6 +54,31 @@ import threading
 #sfrom PYME.ui import mytimer
 
 class FrameWrangler(object):
+    """
+    Grabs frames from the camera buffers
+
+    Notes
+    -----
+    dispatch Signals are used to allow other files to listen to key events 
+    happinging within the acquisition.
+
+    Attributes
+    ----------
+    onFrame : dispatch.Signal
+        Called once per new-frame appearing in our buffer; used to pass
+        frame data to e.g. spoolers. Note that while onFrame gets called once
+        per new frame, the new frames are only checked for once per polling
+        cycle, meaning that this event will be fired N-times for each 
+        `onFrameGroup` event.
+    onFrameGroup : dispatch.Signal
+        Called on each new frame group (once per polling interval) - use for 
+        updateing GUIs etc.
+    onStop : dispatch.Signal
+    _polling_interval : float
+        Length of time to sleep between loops in the polling thread. Defaults to
+        0.01, units of seconds.
+
+    """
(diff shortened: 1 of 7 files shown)


barentine:
converted to draft while I re-tune the htsms focus lock - performance is nice, might even be able to add in the fastGraph stuff again

barentine:
things are running very nicely, though the D was annoying enough I left it at the ZN PI servo for now. Should be all set @David-Baddeley

PYME/Acquire/frameWrangler.py
```diff
@@ -54,6 +54,31 @@
...
+        Called once per new-frame appearing in our buffer; used to pass
+        frame data to e.g. spoolers. Note that while onFrame gets called once
+        per new frame, the new frames are only checked for once per polling
+        cycle, meaning that this event will be fired N-times for each 
+        `onFrameGroup` event.
+    onFrameGroup : dispatch.Signal
+        Called on each new frame group (once per polling interval) - use for 
+        updateing GUIs etc.
+    onStop : dispatch.Signal
+    _polling_interval : float
```
David-Baddeley: It's a bit odd having a private attribute with no way of setting documented (leading underscore implies that it should not be accessed from outside the class)

David-Baddeley: This is also a bit of a 'here be dragons' type thing - I wouldn't want to encourage people to change this if it were to be made a public attribute as the results might not be super predictable.

David-Baddeley:
Should also really get this updated and into core docs (graphically illustrates frame timing)
[PYMEAcquireSchem.pdf](https://github.com/python-microscopy/python-microscopy/files/4910444/PYMEAcquireSchem.pdf)

PYME/Acquire/frameWrangler.py
David-Baddeley: Best approach might be to leave as an undocumented (or documented with a comment but not in docstring) private attribute.

PR #316, opened 2020-07-09
marginal gains when checking for series name collisions using httpspooler

barentine:
Addresses issue #314, but not really since we know that exists separate of spooling, just reduces a little dataserver overhead on each spool, which never hurts and we'd kind of talked about doing anyway

marginal enhancement

**Proposed changes:**
- add `clusterIO.series_exists` call to check for either pcs or h5 files with a single directory listing rather than 2x `clusterIO.isdir` and 2x single directory listings. Also use a set for faster membership checks if the directory is pretty full.

**Checklist:**
- [x] works
```
DEBUG:PYME.Acquire.SpoolController:Looking for Andrew/2020_7_9/000/9_7_series_00001 (.pcs or .h5) on cluster

('currSlide:', [None])
DEBUG:PYME.Acquire.HTTPSpooler:Starting spooling: Andrew/2020_7_9/000/9_7_series_00001.h5

'http://127.0.0.1:55533/__aggregate_h5/Andrew/2020_7_9/000/9_7_series_00001.h5/metadata.json'
DEBUG:PYME.Acquire.Spooler:Disconnecting from frame source

DEBUG:PYME.Acquire.Spooler:Frame source should be disconnected

DEBUG:PYME.Acquire.HTTPSpooler:Stopping spooling Andrew/2020_7_9/000/9_7_series_00001.h5

'http://127.0.0.1:55533/__aggregate_h5/Andrew/2020_7_9/000/9_7_series_00001.h5/final_metadata.json'
'http://127.0.0.1:55533/__aggregate_h5/Andrew/2020_7_9/000/9_7_series_00001.h5/events.json'
DEBUG:PYME.Acquire.SpoolController:Looking for Andrew/2020_7_9/000/9_7_series_00001 (.pcs or .h5) on cluster

ERROR:PYME.Acquire.ui.HDFSpoolFrame:IO error whilst spooling
  File "/Users/Andrew/code/python-microscopy/PYME/Acquire/ui/HDFSpoolFrame.py", line 469, in OnBStartSpoolButton
    self.spoolController.StartSpooling(fn, #stack=stack, #compLevel = compLevel,
  File "/Users/Andrew/code/python-microscopy/PYME/Acquire/SpoolController.py", line 291, in StartSpooling
    raise IOError('A series with the same name already exists')
OSError: A series with the same name already exists
  File "/Users/Andrew/code/python-microscopy/PYME/Acquire/ui/HDFSpoolFrame.py", line 469, in OnBStartSpoolButton
    self.spoolController.StartSpooling(fn, #stack=stack, #compLevel = compLevel,
  File "/Users/Andrew/code/python-microscopy/PYME/Acquire/SpoolController.py", line 291, in StartSpooling
    raise IOError('A series with the same name already exists')
OSError: A series with the same name already exists

  File "/Users/Andrew/code/python-microscopy/PYME/Acquire/ui/HDFSpoolFrame.py", line 475, in OnBStartSpoolButton
    ans = wx.MessageBox(str(e.message), 'Error', wx.OK)
AttributeError: 'OSError' object has no attribute 'message'

```

## whitespace changes
 - I removed a commented clusterIO.exists line under the one I changed in `SpoolController`. I think it's justifiable and won't overly burden your review @David-Baddeley but I'm OK with us adding it back if you like.
- accidentally committed a totally unnecessary 'fix' to a non-existent problem in `HTTPSpooler.FlushBuffer`. Sorry I didn't back it out elegantly and now there are whitespace changes. Hopefully OK (they were indented two spaces instead of four... and I had forgotten that)

### DIFF
--- a/PYME/Acquire/HTTPSpooler.py
+++ b/PYME/Acquire/HTTPSpooler.py
@@ -94,7 +94,22 @@ def getReducedFilename(filename):
     return sname
     
 def exists(seriesName):
-    return clusterIO.exists(getReducedFilename(seriesName))
+    """
+    Check for PYME Cluster Series or h5 file on the cluster so we don't end up
+    with degenerate file stubs.
+    
+    Parameters
+    ----------
+    seriesName : str
+        cluster data directory relative path and filename (without extension) to
+        look for
+
+    Returns
+    -------
+    exists : bool
+        True if series name already exists, else False
+    """
+    return clusterIO.series_exists(getReducedFilename(seriesName))
 
 #Push data to cluster from multiple threads simultaeneously to hide IO latency
 #of each individual node. Not sure what the best number is here - currenty set
@@ -292,12 +307,14 @@ class Spooler(sp.Spooler):
         
     def cleanup(self):
         self._dPoll = False
-      
+    
     def FlushBuffer(self):
-      self._postQueue.put(self._buffer)
-      self._buffer = []
-     
+        # each item in the post-queue will be pushed to a single dataserver,
+        # keep things relatively spread out
+        for chunk in chunk_buffer(self._buffer, self.buflen):
+            self._postQueue.put(chunk)
+        self._buffer = []
 
-
-
-   
+def chunk_buffer(buffer, chunk_size):
+        for ind in range(0, len(buffer), chunk_size):
+            yield buffer[ind:ind + chunk_size]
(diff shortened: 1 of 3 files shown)


David-Baddeley:
It would be good to know what, if any, performance improvement this gives. As it stands, the second `exists()` call should short-circuit through the `_dirCache` in `clusterIO` and never result in a call to the server. Ditto for the `isdir()` call within `exists()` after the `locate_file()` call. I'd be pretty strongly in favour of keeping things as simple as possible and keeping special purpose functions out of `clusterIO`. I also like the clarity of the original `if exists(A) or exists(B)` statement in the spooler. 

The double call to `exists()` should incur minimal overhead, and if that is not the case we should be trying to fix the caching to make this so, rather than adding convoluted workarounds.

barentine:
as david mentions, gains should be quite small - putting this as sufficiently low priority to close without bothering to benchmark

PR #317, opened 2020-07-09
Convert display name for mean and Gaussian curvature from H and K to …

zacsimile:
…mean_curvature and gaussian_curvature

Addresses issue # .

**Proposed changes:**

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/experimental/_triangle_mesh.pyx
+++ b/PYME/experimental/_triangle_mesh.pyx
@@ -231,7 +231,7 @@ cdef class TriangleMesh(TrianglesBase):
         self.vertex_normals
 
         # Properties we can visualize
-        self.vertex_properties = ['x', 'y', 'z', 'component', 'boundary', 'singular', 'H', 'K']
+        self.vertex_properties = ['x', 'y', 'z', 'component', 'boundary', 'singular', 'mean_curvature', 'gaussian_curvature']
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
how would you feel about renaming to `curvature_mean` and `curvature_gaussian` so they appear together?

zacsimile:
I'm torn. Logically, it makes more sense, but this was in direct response to Joerg asking us to make the options more user-friendly. I think people will get `curvature_mean`, but it's more confusing than `mean_curvature`. Currently `H` and `K` don't necessarily appear together... I think it's more user-friendly than `H` and `K`, but less user-friendly than `mean_curvature` and `gaussian_curvature`, but I'm fine switching it.

David-Baddeley:
also torn, which is why I asked ... I'm not sure there's a good answer. As to usability, you could argue either way - people will probably look for `curvature` before `mean_curvature` and might take the first curvature they find which will be the Gaussian one. With the type as a suffix they'll sort to where people would expect curvature, they'll see both, and probably either go with mean because they don't know what Gaussian curvature is, or look it up - either of which will probably give them what they want.

zacsimile:
Let's leave it as `curvature_mean` and `curvature_gaussian` for now.

PR #319, opened 2020-07-10
add image.sc to readme

barentine:
As discussed. Chicken/egg whether we're supposed to link it before we are community partners, but I would say we merge it your next chance @David-Baddeley since there's no harm in someone asking a question there 'early'

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,3 +18,6 @@ PYME (the PYthon Microscopy Environment) is an open-source application suite for
 - **Bug reports:** https://github.com/python-microscopy/python-microscopy/issues
 
 The documentation is also available in the *docs* subdirectory of the repository.
+
+**Have a question?** Get in touch! \
+We're a community partner of the image.sc forum - tag your post with `pyme` and we'll be happy to support you.
--- a/README.md
+++ b/README.md
@@ -20,4 +20,4 @@ PYME (the PYthon Microscopy Environment) is an open-source application suite for
 The documentation is also available in the *docs* subdirectory of the repository.
 
 **Have a question?** Get in touch! \
-We're a community partner of the image.sc forum - tag your post with `pyme` and we'll be happy to support you.
+You can email us on support@python-microscopy.org, or raise a [support issue](https://github.com/python-microscopy/python-microscopy/issues/new?assignees=&labels=support&template=support.md&title=%5BSUPPORT%5D). We are also active on the [image.sc forum](https://forum.image.sc/) - tag your post with `pyme` and we'll be happy to support you.


README.md
```diff
@@ -20,4 +20,4 @@ PYME (the PYthon Microscopy Environment) is an open-source application suite for
 The documentation is also available in the *docs* subdirectory of the repository.
 
 **Have a question?** Get in touch! \
-We're a community partner of the image.sc forum - tag your post with `pyme` and we'll be happy to support you.
+You can email us on support@python-microscopy.org, or raise a [support issue](https://github.com/python-microscopy/python-microscopy/issues/new?assignees=&labels=support&template=support.md&title=%5BSUPPORT%5D). We are also active on the [image.sc forum](https://forum.image.sc/) - tag your post with `pyme` and we'll be happy to support you.
```
barentine, line 23: one of image.sc req's is

> The organization does not promote additional, separate, project-specific discussion channels more prominently than this forum.

Having a contact email and git issues are fine, but maybe put image.sc first?

zacsimile: I think just removing the word "also" would fix this (i.e. "We are are active on...").

David-Baddeley: It's in the same paragraph, so I can't see them objecting. If they care that much it would be a good argument against using image.sc. Image.sc is still relatively new and unproven, and I think we have a fundamental duty to our users to put the methods we have control over first. I can guarantee that the first two will get our attention quickly - and cannot do so for image.sc. 

What is your username on imagesc? I am wondering if we haven't applied for partner status a bit prematurely - we should have just posted a release, tagged with PYME and generated a bit of traffic and experience first.

barentine: my image.sc username is `barentine`, and I've found it a friendly, helpful place. I think it unfair to say it's unproven - have you spent time on it? Filing an issue in a form template on github is not something everyone feels comfortable with, multiple people asking the same question in your(/my/Zach's) inbox is inefficient, and if we join/create a forum community of helpful users then sometimes even if Zach (I think also on image.sc), you (if you choose to join), or I aren't the first to get to it someone else might even jump in. In any event I knew what I was signing on for when I offered to be our community rep.

David-Baddeley: To clarify, I mean unproven in the fairly narrow context of providing support for PYME. I'm `DavidBaddeley` on image.sc. You don't need to convince me that being on image.sc is a good idea, the image.sc people seem like good folks. Philosophically, however, I struggle with forums as a *first* line of support - historically a lot of software projects have used forums as a means of abdicating their support responsibilities to the community, and the quality of community provided support can be very variable (particularly while the community is still comparatively small). Forums are very useful, and a valuable *part* of software support, but I think we have a responsibility to stand behind our software and make it easy for people to contact us directly first. 

I think image.sc are trying to mitigate some of the historical problems with forum based support by trying to ensure that developers and maintainers are committed to being present and active (and this is what is behind the clause we are struggling with). I can still see some technical issues, however - e.g. I can't find a setting which would notify me by email anytime `pyme` was tagged. I know that when things get busy I won't be logging in and checking image.sc several times a day - the support email and github issues however will go direct to by phone so I will see them (even if I can't reply immediately). My "would be a good argument against using image.sc" comment above was partly flippant, but also reflects a real un-willingness to be forced into prioritising forum based support over direct means.

zacsimile: I'm `zacsimile` on image.sc. 

Super minor, but you can be notified every time `pyme` is tagged once the tag exists. For example, go to https://forum.image.sc/tags/cellprofiler, click on the bell on the right, and switch your status to `Watching First Post` to be notified of every new post tagged with `cellprofiler`.

zacsimile: I'm zacsimile on image.sc.

Super minor, but you can be notified every time `pyme` is tagged once the tag exists. For example, go to https://forum.image.sc/tags/cellprofiler, click on the bell on the right, and switch your status to Watching First Post to be notified of every new post tagged with cellprofiler. Or, go to https://forum.image.sc/u/DavidBaddeley/preferences/tags. If you have `Activity Summary` checked under https://forum.image.sc/u/DavidBaddeley/preferences/emails, you'll automatically get an email digest of your notifications if you don't show up on the forum that day.

David-Baddeley: Had got the notifications bit, but as I read it you'd need to be logged in to get the notification. Didn't work see the activity summary bit.

zacsimile: I only know about it because I've been getting the emails for a few months now. I didn't opt-in, it was on by default when I signed up.

barentine:
update @David-Baddeley: we are now community partners of image.sc

https://forum.image.sc/tag/pyme

PR #320, opened 2020-07-10
Update requirements.txt so recipe panels can open without crashing on pip installs

barentine:
Addresses issue (none filed).

bugfix

**Proposed changes:**
- differ `networkx` import in PYME/misc/depGraph.py until it is called (in a function I don't think we use anywhere)
- add docutils back into requirements.txt as it is needed to display docs in any of the recipe ui's. While technically not necessary to run the program, I think it's a pretty small dependency and very helpful to users

**Checklist:**

- [ ] Tested with numpy=1.14

- [x ] Does this maintain backwards compatibility with old data?

### DIFF
--- a/PYME/misc/depGraph.py
+++ b/PYME/misc/depGraph.py
@@ -6,7 +6,6 @@ Created on Mon Mar  9 21:56:18 2015
 """
 import numpy as np
 import toposort
-import networkx as nx
 # import pylab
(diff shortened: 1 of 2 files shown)


David-Baddeley:
docutils is actually pretty big in terms of additional dependencies, but doesn't hurt to have it in.

PR #321, opened 2020-07-10
Package fix 1

zacsimile:
Addresses issue # package installs.

bugfix

**Proposed changes:**
- Update instructions based on use of latest `constructor` and `conda-build` packages. Remove use of local packages since we pull them from the `david_baddeley` channel anyway.
- Shift spec determination onto `pyme-depends`
- Disable `keep_pkgs` (no need)
- Install dependencies in order to remove errors thrown during install
- Convert `http` to `https` (for peace of mind)

**Checklist:**

- [x] Tested with numpy=1.14


zacsimile:
Next steps are to accept this, python-microscopy/pyme-conda-recipes#7 and python-microscopy/pyme-conda-recipes#8, `conda-build` the packages with the latest `conda-build`, and push the new packages to conda/david_baddely. At this point, the new install instructions with `constructor>=3.0.0` installed from the `conda` channels in a fresh environment should produce a perfectly-working installer on Windows, and likely will on OSX (although I'm still not 100% sure the framework error will be fixed, though it should be).

Getting these installers signed is a separate issue. We can sign them ourselves and attach the signature on OSX with `signing_identity_name` in `constructor.yaml`. Since this would not use a recognized signing authority, users would have to accept our signature the first time they install PYME. As long as we use the same signature file, all future PYME should be automatically accepted. I could not find clear instructions on how to sign our Windows installer. 

We should consider an internal upgrade call in the `Help` menu. Can just call `conda update` in the environment.

docs/dev/executable_installers.md
```diff
@@ -3,22 +3,14 @@
...
-4) Download the latest `conda.exe` for the platform you are building for from 
-[https://repo.anaconda.com/pkgs/misc/conda-execs/](https://repo.anaconda.com/pkgs/misc/conda-execs/) see also 
-[https://github.com/conda/constructor/issues/257](https://github.com/conda/constructor/issues/257)
-5) Change to the `python-microscopy` directory
-6) Edit `construct.yaml` to set the version to the latest conda package version
-6) Run `constructor --conda-exe PATH_TO_CONDA.EXE .`
+Instructions work for `miniconda3`.
+
+1) Create a clean conda environment (e.g. `conda create -n pyme_build python=2.7`)
+2) Activate the new environment and run `conda install instructor`. This should install `constructor>=3.0.0`.
```
David-Baddeley: instructor???

construct.yaml
```diff
@@ -35,9 +36,9 @@ exclude:
 packages:
 #  - python-2.7.9-0.tar.bz2
 
-#install_in_dependency_order: True
+install_in_dependency_order: True
 
-keep_pkgs: True
+# keep_pkgs: True
```
David-Baddeley, line 40: why?

construct.yaml
```diff
@@ -1,22 +1,23 @@
...
+  - https://repo.anaconda.com/pkgs/free/
   - https://conda.anaconda.org/david_baddeley
-  - http://repo.anaconda.com/pkgs/msys2
+  - https://repo.anaconda.com/pkgs/msys2
 
 # specifications
 specs:
-  - python
+  # - python
+  - pyme-depends
```
David-Baddeley: is a dep of python-microscopy - should not be needed.

docs/dev/executable_installers.md
```diff
@@ -3,22 +3,14 @@
...
-3) `conda install constructor` **NOTE:** At one point it was neccesary to git clone the constructor source, checkout the "v3" branch
-   and do a `python setup.py install` instead of the above. I think that the v3 branch has now been merged into main, but 
-   it might be worth trying the source install if you run into issues.
-4) Download the latest `conda.exe` for the platform you are building for from 
-[https://repo.anaconda.com/pkgs/misc/conda-execs/](https://repo.anaconda.com/pkgs/misc/conda-execs/) see also 
-[https://github.com/conda/constructor/issues/257](https://github.com/conda/constructor/issues/257)
-5) Change to the `python-microscopy` directory
-6) Edit `construct.yaml` to set the version to the latest conda package version
-6) Run `constructor --conda-exe PATH_TO_CONDA.EXE .`
+Instructions work for `miniconda3`.
```
David-Baddeley: which version??? - my stuff above was also with miniconda3

zacsimile: conda 4.8.3

construct.yaml
zacsimile: Fair enough. We'll delete that, too.

construct.yaml
zacsimile: It takes up extra space on the user's computer, and the installer works without this flag.

docs/dev/executable_installers.md
zacsimile: Typo, `constructor`.

docs/dev/executable_installers.md
zacsimile: The key is `conda install constructor` now pulls version `3.0.0`, which it should do on older versions of `conda` now as well.

zacsimile:
I think everything here is deprecated at this point.

PR #322, opened 2020-07-11
Documentation updates

zacsimile:
Addresses issue #persistent need for better documentation.

Enhancement

**Proposed changes:**
- Add documentation guidelines to contributing
- Update installation instructions for clarity
- Update doc copyright

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -19,7 +19,7 @@ Then, open a command prompt and enter
 .. code-block:: bash
 	
     conda create -n pyme python=2.7
-    source activate pyme
+    conda activate pyme
     conda config --add channels david_baddeley
     conda install python-microscopy
 
@@ -32,7 +32,7 @@ Then, open a command prompt and enter
 .. code-block:: bash
 	
     conda create -n pyme python=2.7
-    source activate pyme
+    conda activate pyme
     pip install python-microscopy
 
 conda installation for developers
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -32,7 +32,7 @@ Create a Python 3.6 virtual environment with ``conda``. Then, open a command pro
 	
     conda config --add channels david_baddeley
     conda install pyme-depends
-    git clone git@github.com:python-microscopy/python-microscopy.git
+    git clone https://github.com/python-microscopy/python-microscopy.git
     cd python-microscopy
     python setup.py develop
 
@@ -44,7 +44,7 @@ and cython are installed. Then, open a command prompt and enter
 
 .. code-block:: bash
 	
-    git clone git@github.com:python-microscopy/python-microscopy.git
+    git clone https://github.com/python-microscopy/python-microscopy.git
     cd python-microscopy
     pip install -r requirements.txt
     python setup.py develop
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -23,18 +23,6 @@ Then, open a command prompt and enter
     conda config --add channels david_baddeley
     conda install python-microscopy
 
-pip installation for users
-==========================
-
-Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_.
-Then, open a command prompt and enter
-
-.. code-block:: bash
-	
-    conda create -n pyme python=2.7
-    conda activate pyme
-    pip install python-microscopy
-
 conda installation for developers
 =================================
 
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -21,12 +21,13 @@ Then, open the *Anaconda prompt* [#anacondaprompt]_ and enter
 
 .. code-block:: bash
 	
+    conda config --append channels anaconda
     conda config --add channels david_baddeley
     conda install python-microscopy
 
 .. note::
 
-   **Which python version?** We are currently in the process of switching the default install from python 2.7 to python 3. As of 2020/8/6 the python 3 packages are not in the above conda channel, but that should change shortly. The python2.7 version is better tested, but most of the core functionality now runs on python 3 as well.
+    **Which python version?** We are currently in the process of switching the default install from python 2.7 to python 3. As of 2020/8/6 the python 3 packages are not in the above conda channel, but that should change shortly. The python2.7 version is better tested, but most of the core functionality now runs on python 3 as well.
 
 
 Updating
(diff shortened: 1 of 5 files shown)


docs/Installation/Installation.rst
```diff
@@ -1,10 +1,129 @@
...
+
+conda installation for users
+============================
+
+Download and install `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_.
+Then, open a command prompt and enter
+
+.. code-block:: bash
+	
+    conda create -n pyme python=2.7
```
barentine: I don't think we should old up Zach's doc-update for this, but maybe we should revisit this week whether there's anything left to do before we make py3 the default installation. If the answer is just packaging and docs (notably launchWorkers -> PYMEClusterOfOne) I think it'd be great if we push through on that this week - I'll have a go at the launchWorkers -> PYMEClusterOfOne.

zacsimile:
Worth noting that the `pip` install for users doesn't work.

```
(base) Mor:python-microscopy zachcm$ conda create -n pyme_pip python=2.7
(base) Mor:python-microscopy zachcm$ conda activate pyme_pip
(pyme_pip) Mor:python-microscopy zachcm$ pip install python-microscopy
DEPRECATION: Python 2.7 will reach the end of its life on January 1st, 2020. Please upgrade your Python as Python 2.7 won't be maintained after that date. A future version of pip will drop support for Python 2.7. More details about Python 2 support in pip, can be found at https://pip.pypa.io/en/latest/development/release-process/#python-2-support
ERROR: Could not find a version that satisfies the requirement python-microscopy (from versions: none)
ERROR: No matching distribution found for python-microscopy
```

I followed the link in the deprecation warning and it looks like `pip` will drop Python 2.7 support in January 2021. This prompted me to check `conda`, which has apparently already dropped Py2 support (https://docs.anaconda.com/anaconda/reference/release-notes/). Users will need to install an old version of Anaconda going forward to ensure they can use our Py2 branch. Right now, Miniconda install page still links to this old version, but it's something we'll need to be aware of.

The `conda` install for users also doesn't work on Mac because of a framework build issue. If I `conda-build python-microscopy` locally and then install it, the framework issue disappears. I'm not sure if this is because I'm compiling on my machine, or because the uploaded package is compiled incorrectly. Worth trying a rebuild using the latest version of `conda-build` and the latest packaging scripts and re-uploading to the `david_baddeley` channel.

docs/Installation/Installation.rst
David-Baddeley: Our target for making Py3 the default is September. Before then I'd like to a) have @csoeller try it out and b) update the analysis docs etc for `PYMEClusterOfOne`. I'd be open, however, to making sure that packages for py3 are built and adding an installing on py3 section / note to the installation docs.

zacsimile: These new docs tell users to install Py2 and developers to install Py3. I think this, coupled with joining image.sc, should help with beta testing Py3. I think we're far along enough to expect developers to make it through any Py3 problems. There should be none on the initial install or for general usage. Users in our lab are on Py3 with no issues. I think it's worth updating the analysis docs for the developers and see how they get on.

docs/Installation/InstallationFromSource.rst
```diff
@@ -7,7 +7,7 @@ If you want to contribute to PYMEs development, or you want to use PYME in a con
 requires modification to some of the scripts you will want to have the PYME sources somewhere you can easily find, modify
 and update them.
 
-If you just want to use the data analysis features then you are strongly advised to follow the instructions in :ref:`installationanaconda`_
```
David-Baddeley, line 10: Why delete the underscore? I'm pretty sure it's need to turn this into a link.

docs/Installation/Installation.rst
```diff
@@ -1,10 +1,129 @@
...
+.. code-block:: bash
+
+    cd /path/to/python-microscopy/
+    /path/to/mininconda/install/python.app/Contents/MacOS/python setup.py develop
+
+Additional resources
+********************
+
+- Detailed developer installation docs are located at :ref:`installationFromSource`
+- Additional PYME installation docs can be found at https://github.com/csoeller/pyme-install-docs
```
David-Baddeley: If there is something missing from our install docs, we should fix them - outsourcing our install docs is just plain weird.

David-Baddeley:
I'm not happy with the proposed changes to the install docs - for an end user it's not OK to present 6 different options. We need to present the one option which we want them to use. Ideally it would be the executable installer, but as a) that currently doesn't get built regularly b) doesn't work on OSX it's miniconda etc ....

docs/Installation/InstallationFromSource.rst
zacsimile: For external links you need the underscore, for internal ones you need to remove it. Verified by compiling the docs.

zacsimile:
Right now, if you go to our install docs you have to read several paragraphs before you can do anything. Every step is interspersed with multiple notes and warnings. This is more than a user wants to consider when first installing a program. By moving all caveats below the install instructions, you can quickly install PYME (assuming the packages work, which is addressed in #321, python-microscopy/pyme-conda-recipes#7 and python-microscopy/pyme-conda-recipes#8) without having to think about any of the steps. I’ve listed our options in a way I think makes it easy to identify relevance to the reader. Two options are for users, so users will follow one of those. Users who don’t know what to do will follow the first option, as they should. Developers should be fine following any of these instructions.

I think multiple installation options is OK/common. Numpy presents two options: https://numpy.org/install/. scipy, Cell Profiler and napari present multiple options: https://scikit-image.org/docs/stable/install.html, https://github.com/CellProfiler/CellProfiler/blob/master/README.md, https://github.com/napari/napari/blob/master/README.md. I tried to make the new install instructions I propose here similar to the install instructions from these packages since they all have a lot of users.

How do you propose presenting our multiple installation routes in a user-friendly manner?

docs/Installation/Installation.rst
zacsimile: It was a nod to Christian and his work. We can remove it.

barentine:
@zacsimile with the merge of #320 you should no longer need to specify docutils for pip install. That said, until #352 is merged sphinx needs to be specified (for bakeshop to work). 

If mpld3 is required to run PYMEVisualize I think you should submit a PR to add it into requirements.txt

docs/Installation/Installation.rst
David-Baddeley: I'm a little reluctant to tell *users* to install in an environment as (at least on windows) I've run into issues running programs from shortcuts and/or associating files with programs this way. The environment way pretty much forces people to drop into an anaconda prompt and run `conda activate pyme` before running stuff on the command line. 

Installing in the base environment avoids these issues.

docs/Installation/Installation.rst
```diff
@@ -3,8 +3,141 @@
...
+
+Installing using conda
+======================
+
+Download and install the python **2.7** version of `Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_.
+Then, open the *Anaconda prompt* [#anacondaprompt]_ and enter
+
+.. code-block:: bash
+	
+    conda config --add channels david_baddeley
```
David-Baddeley, line 25: Do we not need the `anaconda` channel?

zacsimile: The last time I nuked and paved the `anaconda` channel was there by default. Probably safer to add it.

barentine: I also thought the last time I didn't need to add it.

David-Baddeley: hmm, I though it was there for full anaconda, but not miniconda

David-Baddeley: have added for now - we can always alter after some testing.

PR #326, opened 2020-07-13
make sure ruleserver.html gets distributed with packages

barentine:
Addresses issue: Running `PYMERuleServer` on a conda installation currently fails because the ruleserver.html isn't included in package data
```
$ PYMERuleServer
Exception in thread Thread-1:
  File "/home/ubuntu/miniconda3/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/PYME/cluster/ruleserver.py", line 830, in run
    self.distributor = WFRuleServer(self.port, bind_addr=self.bind_addr)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/PYME/cluster/ruleserver.py", line 772, in __init__
    RuleServer.__init__(self)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/PYME/cluster/ruleserver.py", line 440, in __init__
    with open(os.path.splitext(__file__)[0] + '.html', 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/ubuntu/miniconda3/lib/python3.6/site-packages/PYME/cluster/ruleserver.html'
  File "/home/ubuntu/miniconda3/bin/PYMERuleServer", line 11, in <module>
    sys.exit(main())
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/PYME/cluster/PYMERuleServer.py", line 79, in main
    sa = proc.distributor.socket.getsockname()
AttributeError: 'ServerThread' object has no attribute 'distributor'
```

bugfix
**Proposed changes:**
- swap add 'cluster' directory as data directory in sepuptools rather than downstream 'cluster/clusterUI'. Alternatively we could move rulerserver.html?

**Checklist:**
- [x] tested

```
$ python setup.py sdist
$ tar tfz dist/python-microscopy-20.7.10.tar.gz | grep ruleserver
python-microscopy-20.7.10/PYME/cluster/ruleserver.py
python-microscopy-20.7.10/PYME/cluster/ruleserver.html
```
- [ ] Tested with numpy=1.14


David-Baddeley:
unfortunately we will need to move `ruleserver.html` to a directory without python code and then add that (adding a directory with code as a data dir is problematic).

barentine:
Thanks @David-Baddeley. Reverted the data dir to clusterUI, moved ruleserver.html there, tested ruleserver and setup.py sdist again and should be good to go.

David-Baddeley:
I'm a bit worried that dumping it in clusterUI will cause issues in the future (adding clusterUI as a datadir is a bit of a hack, and might well change in the future).

I'm going to move it to `PYME.resources.web` instead. It's currently still a work in progress, but the aim is to dump all our non code stuff in `PYME.resources` eventually.

David-Baddeley:
Alternative solution merged in 0e5fcc7aa5bf4d3e28cd1110c230b42288ec9481

Issue #323, opened 2020-07-12
[BUG]


csoeller:
**Describe the bug**

```
File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/recipes/pointcloud.py", line 263
SyntaxError: Non-ASCII character '\xe2' in file /Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/recipes/pointcloud.py on line 264, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details

```

**To Reproduce**
Try to execute Mesh>Generate Isosurface in VisGUI

**Expected behavior**
No error.

More seriously, not sure why this gives a bug on my system but not on the developer's machines.

**Screenshots**

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15.5
 - python-microscopy=20.07.10
 - python=2.7.17, platform=darwin
 - numpy=1.16.6, wx=3.0.2.0 osx-cocoa (classic)

**Additional context**
Quick test using

```
with open("pointcloud.py") as fp:
    for i, line in enumerate(fp):
        if "\xe2" in line:
            print i, repr(line)

```
gives

```
262 '            and A. K. Kenworthy, "On the use of Ripley\xe2\x80\x99s K-function and its \n'
264 '            pp. 1095\xe2\x80\x931103, 2009.\n'
```

Solution is presumably either

- delete offending nonascii chars

or

- `# -*- coding: utf-8 -*-` at top of file?

zacsimile:
Sorry about that @csoeller. Happened on a copy/paste from Mendeley. Not sure why this ran on my machine. Fixed in #324.

David-Baddeley:
An explanation for it running in some places but not others would be py2 vs 3 - I think py3 defaults to UTF8 if no source encoding is declared, py2 possibly defaults to ASCII

David-Baddeley:
Merged now

David-Baddeley:
We need to find a way of automatically testing more of this stuff (or un-skipping more of the import unit tests which should have picked this up - we might now have got enough wx stuff out of the way that the import tests have a better chance of succeeding.

PR #329, opened 2020-07-15
Add optional use of pre-commit hooks to PYME

zacsimile:
Addresses issue #323, others.

This came about as a result of #323. I thought there must be an automatic way to check for unicode characters to prevent this kind of error from happening again. It's actually easiest to just add `# -*- coding: utf-8 -*-` to the top of all of our files (`pre-commit` hook `fix-encoding-pragma`). I also decided to add automatic checking of variable and module names using the linter (for the times I type `import nupmy`, etc.). To make sure this wouldn't fail every time, I added a customized .pylintrc file that now ignores known but useless C-import and dynamic errors from `wx`, `OpenGL`, and `matplotlib.cm`. 

I would like to give this a try. Currently, this will do nothing if you don't want to use it. I downloaded `pre-commit` and ran `pre-commit install` in my personal copy of `python-microscopy`, so all these checks run automatically on files in staged commits coming from me. Developers who want to give this a test run are free to do so. I figure that by using this in an optional manner for a bit, we can work out the bugs (I suspect there are further tweaks to be made to `.pylintrc`). 

Long-term, I'd like to see these pre-commit hooks run in our Azure pipeline to take care of easy code review/fixes (I spelled a variable name wrong, I copied something with unicode characters into a docstring). There are additional, potentially-useful hooks at https://github.com/pre-commit/pre-commit-hooks if anyone wants to have a look.

Enhancement

**Proposed changes:**
- Create pre-commit hooks to detect undefined variables and address unicode characters in docstrings
- Add .pylintrc to address annoying `wx` and `OpenGL` c-import linting errors that aren't actually errors
- Add `# -*- coding: utf-8 -*-` to `PYME.LMVis.layers.pointcloud.py`

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,9 @@
+repos:
+- repo: local
+  hooks:
+  - id: pylint
+    name: pylint
+    entry: pylint --errors-only --extension-pkg-whitelist=wx
+    language: system
+    types: [python]
+    files: 'PYME/*'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -3,6 +3,10 @@ repos:
   hooks:
   - id: pylint
     name: pylint
-    entry: pylint --errors-only --extension-pkg-whitelist=wx
+    entry: pylint --errors-only
     language: system
     types: [python]
+- repo: https://github.com/pre-commit/pre-commit-hooks
+  rev: v3.1.0  # Use the ref you want to point at
+  hooks:
+  - id: fix-encoding-pragma  # Add # -*- coding: utf-8 -*- to the top of python files.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,4 +6,3 @@ repos:
     entry: pylint --errors-only --extension-pkg-whitelist=wx
     language: system
     types: [python]
-    files: 'PYME/*'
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Will come back to this when I get some time (would like to get lint to run on my machine and have a look at the .lintrc provided). 

Will also note that https://github.com/pre-commit/pre-commit-hooks is a 3rd party extension package to gits pre-commit hooks - the config details for something like that are pushing the boundaries for what should be in the repository rather than present on a per-user basis as an untracked file.

David-Baddeley:
I think `pylint` is prohibitively slow ... `pyflakes` of `flake8` might be a better choice.

A reasonable starting set of starting parameters would be:

`flake8 --extend-ignore=E2,E3,E4,E5,E722,E731,F405,W2,W3 --exclude=loft,Loft`

There are a few of the ignores we could drop in the future as more critical things get fixed, but I think this reduces messages to a semi-reasonable volume.

zacsimile:
Thanks for pointing this out! I think `flake8`/`pyflakes` is a good call. My understanding is that `flake8` runs `pyflakes`, as well as a PEP8 style checker and checks for cyclomatic complexity. `pyflakes` is basically the 1-1 replacement for what I wanted to do with `pylint`, which is check for unused/misnamed variables and imports. We don't need the rest of `pylint`'s capabilities, which slow things way down, on every commit. So, I think `flake8` actually does more than what I proposed here, but faster. I say go for it.

Issue #330, opened 2020-07-16
[BUG] Byte ordering in MATLAB source files causes saving issue in `tabular.TabularBase.to_hdf`


zacsimile:
**Describe the bug**
If we try to save filtered localizations from a loaded `.mat` file in PYME, we get the following traceback.

<details>
<summary>Traceback</summary>
<pre>
/Users/zachcm/Code/python-microscopy/PYME/LMVis/filterPane.py:180: wxPyDeprecationWarning: Call to deprecated item. Use InsertItem instead.
  ind = self.lFiltKeys.InsertStringItem(UI_MAXSIZE, key)
/Users/zachcm/Code/python-microscopy/PYME/LMVis/filterPane.py:181: wxPyDeprecationWarning: Call to deprecated item. Use SetItem instead.
  self.lFiltKeys.SetStringItem(ind, 1, '%3.2f' % minVal)
/Users/zachcm/Code/python-microscopy/PYME/LMVis/filterPane.py:182: wxPyDeprecationWarning: Call to deprecated item. Use SetItem instead.
  self.lFiltKeys.SetStringItem(ind, 2, '%3.2f' % maxVal)
lw update
hl update
DEBUG:root:pytables open call: /Users/zachcm/Documents/Papers/PYMEVisualize/Data/croppeddata.hdf
DEBUG:root:pytables file open: /Users/zachcm/Documents/Papers/PYMEVisualize/Data/croppeddata.hdf
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/h5rFile.py", line 118, in _appendToTable
    table = getattr(self._h5file.root, tablename)
  File "/Users/zachcm/miniconda3/envs/pyme/lib/python3.6/site-packages/tables/group.py", line 836, in __getattr__
    return self._f_get_child(name)
  File "/Users/zachcm/miniconda3/envs/pyme/lib/python3.6/site-packages/tables/group.py", line 708, in _f_get_child
    self._g_check_has_child(childname)
  File "/Users/zachcm/miniconda3/envs/pyme/lib/python3.6/site-packages/tables/group.py", line 395, in _g_check_has_child
    % (self._v_pathname, name))
tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``Localizations``

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #331, opened 2020-07-16
Cast mat columns to native byte order

zacsimile:
Addresses issue #330.

Bugfix, option 2 from #330

**Proposed changes:**
- Cast .mat columns to native byte order to match missing pipeline ds keys, which are in native byte order

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
casting to float is probably fine, there is at least a possibility that the matlab file has integer columns though.

PR #333, opened 2020-07-17
allow httpspooler buflen to be configured

barentine:
Enhancement as discussed with David this week. Configuring a longer buffer length in `HTTPSpooler` will increase data locality when spooling to multiple servers and localizing with sliding-window background estimation. Too long and you start losing out on the advantages of multiple servers.

### DIFF
--- a/PYME/Acquire/HTTPSpooler.py
+++ b/PYME/Acquire/HTTPSpooler.py
@@ -126,7 +126,7 @@ class Spooler(sp.Spooler):
         self.clusterFilter = kwargs.get('serverfilter', CLUSTERID)
         self._buffer = []
         
-        self.buflen = config.get('http-chunk-size', 50)
+        self.buflen = config.get('httpspooler-chunksize', 50)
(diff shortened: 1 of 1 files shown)


PYME/Acquire/HTTPSpooler.py
```diff
@@ -126,7 +126,7 @@ def __init__(self, filename, frameSource, frameShape, **kwargs):
         self.clusterFilter = kwargs.get('serverfilter', CLUSTERID)
         self._buffer = []
         
-        self.buflen = 50
+        self.buflen = config.get('http-chunk-size', 50)
```
David-Baddeley:
```suggestion
        self.buflen = config.get('httpspooler-chunksize', 50)
```

Make it specific to the spooler. 

Also collapsed chunksize to follow implicit config convention `<module/program>-<setting>` of using a hyphen as a separator between the program and the setting (rather than a generic word separator) - TODO document this.

Issue #246, opened 2020-05-14
Data pipeline disappears when there is more than one starting point for the pipeline


zacsimile:
<img width="1012" alt="Screen Shot 2020-05-14 at 1 05 09 PM" src="https://user-images.githubusercontent.com/1263313/81963771-93eb2f00-95e3-11ea-9b94-b92839c6d149.png">

zacsimile:
<img width="1012" alt="Screen Shot 2020-05-14 at 1 11 33 PM" src="https://user-images.githubusercontent.com/1263313/81964355-72d70e00-95e4-11ea-938b-40bfc9d211c2.png">

barentine:
or more than one output
![Screen Shot 2020-05-14 at 1 39 21 PM](https://user-images.githubusercontent.com/31105780/81967027-52a94e00-95e8-11ea-9ed2-9285482254c8.png)

David-Baddeley:
Issue is not present on py2, wx3. 

best guess, it's related to this traceback:
```
wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /opt/concourse/worker/volumes/live/b30868b5-2a7c-479c-4e29-27874bbc69a6/volume/wxpython_1547931175613/work/ext/wxWidgets/src/common/wincmn.cpp(478) in ~wxWindowBase(): any pushed event handlers must have been removed

The above exception was the direct cause of the following exception:

SystemError: <class 'wx._core.UpdateUIEvent'> returned a result with an error set

```

The implication of which is that it's probably traits not cleaning up after itself properly when we destroy the old recipe display.

Any punters to try on py2, wx4 to confirm this is a wx4 issue?

zacsimile:
Not a wx4 issue. Tried on python 2. Works with wx3 and wx4. Still receive this error with py2/wx4, but the pipeline doesn't disappear.

```
wx._core.wxAssertionError: C++ assertion "GetEventHandler() == this" failed at /opt/concourse/worker/volumes/live/80490b2f-c578-4407-562a-dbbbe365123b/volume/wxpython_1547931284446/work/ext/wxWidgets/src/common/wincmn.cpp(478) in ~wxWindowBase(): any pushed event handlers must have been removed
```

[...]

PR #334, opened 2020-07-17
Fix vertical recipe display on python 3

zacsimile:
Addresses issue #246.

Bugfix

**Proposed changes:**
-Fix sorted() call in recipe layout to work on objects for py3

**Checklist:**

- [ ] Tested with numpy=1.14
- [x] Tested on python 2.7 and 3.6
- [x] Tested with wx=3.x and wx=4.x [if UI code]

### DIFF
--- a/PYME/recipes/recipeLayout.py
+++ b/PYME/recipes/recipeLayout.py
@@ -201,7 +201,7 @@ def layout_vertical(dg, rdg):
                     _descend(child, dg, rdg)
     
     for t_i in ts:
-        for n in sorted(list(t_i)):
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Will merge, as better than status quo, but would be nice if there was an alternative to using `.__class__.__name__`. It feels like there should be a cleaner way of doing this.

zacsimile:
One cleaner way would be to ensure `ts` has sorted sublists and just use those (I think it actually does right now, they're just sorted in the reverse of the order we're interested in).

PR #338, opened 2020-07-19
fix squiggle typo

csoeller:
Addresses issue # .

Not filed as an issue yet but try plotting a profile with current PYME and you get an error about SELECTION_SQUIGGLE not being known.  This is because everywhere else the bogus name SELECTION_SQUIGLE is currently used.

Bugfix

**Proposed changes:**

Change the bogus SELECTION_SQUIGLE to SELECTION_SQUIGGLE everywhere.

**Checklist:**

- [ ] Tested with numpy=1.14
- [~ ] Does the PR avoid variable renaming in existing code, whitespace changes, and other forms of tidying? [There is a place for code tidying, but it makes reviewing 
much simpler if this is kept separate from functional changes] My emacs may have added missing newlines in a few files??


David-Baddeley:
Thanks!

Issue #342, opened 2020-07-20
[BUG] nameUtils import fails on OSX if imported from non-user processes


barentine:
**Describe the bug**
When not running from the terminal, e.g. launching from vscode debugger, os.environ dict doesn't contain `USER` key, which throws 
```
Exception has occurred: KeyError
'USER'
  File "/Users/Andrew/code/python-microscopy/PYME/IO/FileUtils/nameUtils.py", line 42, in getUsername
    return os.environ['USER']
  File "/Users/Andrew/code/python-microscopy/PYME/IO/FileUtils/nameUtils.py", line 64, in <module>
    dateDict = {'username' : getUsername(), 'day' : dtn.day, 'month' : dtn.month, 'year':dtn.year, 'sep' : os.sep, 'dataDir' : datadir, 'homeDir': homedir, 'dirSuffix': dirSuffix}
  File "/Users/Andrew/code/python-microscopy/PYME/IO/image.py", line 45, in <module>
    from PYME.IO.FileUtils.nameUtils import getRelFilename
  File "/Users/Andrew/code/pyWIF/tests/pyme_series_script.py", line 4, in <module>
    from PYME.IO.image import ImageStack
```

```
list(os.environ.keys())
['TERM_PROGRAM', 'TERM', 'SHELL', 'TMPDIR', 'CONDA_SHLVL', 'CONDA_PROMPT_MODIFIER', 'TERM_PROGRAM_VERSION', 'OLDPWD', 'CONDA_EXE', 'SSH_AUTH_SOCK', '_CE_CONDA', 'PATH', 'CONDA_PREFIX', 'PWD', 'LANG', 'XPC_FLAGS', '_CE_M', 'XPC_SERVICE_NAME', 'SHLVL', 'HOME', 'VSCODE_GIT_ASKPASS_MAIN', 'CONDA_PYTHON_EXE', 'LOGNAME', 'VSCODE_GIT_IPC_HANDLE', 'CONDA_DEFAULT_ENV', 'VSCODE_GIT_ASKPASS_NODE', 'GIT_ASKPASS', 'COLORTERM', '_', 'PYTHONIOENCODING', 'PYTHONUNBUFFERED']
```

`os.getlogin` works, however it returns `'root'` in this case. Note sure if it's a big deal for anyone else, but for now I'm locally changing to `os.getlogin` so I can use the debugger on things which import nameUtils.

**To Reproduce**
[...]

PR #343, opened 2020-07-20
allow nameUtils to be imported when running from non-user process

barentine:
Addresses issue #342 .

bugfix

**Proposed changes:**
- note that OSX can also hit no 'USER' key in `os.environ` and treat it like linux in that respect.

**Note Also**
Rather than `nobody` for user, if I call `os.getlogin()` when 'USER' key is not defined I get `root`. I see the note for linux that `os.getlogin` is broken from non-user processes, though I wonder if this has changed in recent python versions. I don't think this is terribly important in any event, and `nobody` is certainly fine by me.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/IO/FileUtils/nameUtils.py
+++ b/PYME/IO/FileUtils/nameUtils.py
@@ -39,8 +39,10 @@ def getUsername():
         import win32api
         return '_'.join(win32api.GetUserName().split(' '))
     else: # OSX / linux
+        import getpass
         #return os.getlogin() #broken when not runing from command line
-        return os.environ.get('USER', 'nobody')
+        #return os.environ.get('USER', 'nobody')
+        return getpass.getuser()
 
 
 dtn = datetime.datetime.now()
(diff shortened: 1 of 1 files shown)


David-Baddeley:
`os.getlogin()` is documented  (in a round about manner) as being broken when there is no controlling terminal in the python docs, with `getpass.getuser()` being recommended instead. Defaulting to `'root'` is a bit gross in case anyone tries to infer a home directory from it. 

Could you check what  `getpass.getuser()` returns when running in a debugger?

barentine:
`getpass.getuser()` returns Andrew, even when 'USER' is not in `os.environ.keys()`

barentine:
Sorry, should also mention `Andrew` is my user, so that's all good. Could additionally pull it from `os.environ['HOME']` if we like, which returns `/Users/Andrew/`

David-Baddeley:
I might change it to use`getpass` on both linux and OSX. `os.environ['HOME']` is what I first tried for the home directory, but didn't work in some circumstances - tilde expansion seems to be more reliable on *nix.

PR #346, opened 2020-07-21
Add (optional) offset display slider to focus lock gui

barentine:
Addresses issue having to open the shell to check what the offset is / how close we are to running out of piezo range.

enhancement
**Proposed changes:**
- add a slider to display the focus lock offset. Sliding it doesn't set the offset, and it snaps back to where it belongs each time the gui updates.
-use it in htsms. leave out of htsms_focus_lock for now

snazzy examples, with target position of 50 (and basePiezo range of [0, 100]):
image
and negative offsets
image

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Scripts/init_htsms.py
+++ b/PYME/Acquire/Scripts/init_htsms.py
@@ -178,7 +178,7 @@ def focus_keys(MainFrame, scope):
     from PYME.Acquire.Hardware import focusKeys
     from PYME.Acquire.ui.focus_lock_gui import FocusLockPanel
     fk = focusKeys.FocusKeys(MainFrame, scope.piFoc)
-    panel = FocusLockPanel(MainFrame, scope.focus_lock, scope.piFoc)
+    panel = FocusLockPanel(MainFrame, scope.focus_lock, offset_piezo=scope.piFoc)
     MainFrame.camPanels.append((panel, 'Focus Lock'))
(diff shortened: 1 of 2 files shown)


David-Baddeley:
This is htsms specific at this point, so you get a lot of discretion. At some point we should probably do a bit of re-naming however to separate this from the default  / other focus lock implementations. 

My one comment is that you might want to use a `wx.Gauge` rather than a slider - having a slider which you can't slide is not ideal in terms of an intuitive UI.

barentine:
I agree it's not ideal that someone can try to slide it and be disappointed, but most of the Gauge examples look to have a side shaded which I think would be pretty weird for what it's trying to display. I don't suspect we'll run into any issues with it on the htsms, so maybe we leave it as-is for now and pop in a replacement if we come across a less progress-bar looking Gauge at some point?

Fair point re: naming. I'm happy to rename/move it if you like. This GUI corresponds to PYME.Acquire.Hardware.focus_locks.reflection_focus_lock.py and the e.g. ReflectedLinePIDFocusLock there. Do you want to mirror that with a focus_locks folder in PYME.Acquire.ui too, with the same filenames?

David-Baddeley:
no rush on the renaming. I'd personally go for a gauge, even if the default position is half full, but will merge as is. Ultimately this is something for which a custom control might be nicest, but it's hard to justify the effort at this point.

PR #347, opened 2020-07-22
HTTP tabular datasource for live-updating visualization in PYMEVisualize

barentine:
Allows us to open full PYMEVisualize process when launching cluster analysis from dh5view (or directly from PYMEAcquire when automatically launching analysis). Another chunk of #272 but with some improvements since then.

enhancement

**Proposed changes:**
- add a tabular results source which pulls results from the cluster, updating periodically
- replace localization view in dh5view with a full PYMEVisualize (hitting 'Go' in dh5view LMAnalyser will launch analysis and open PYMEVisualize instance with the live-updating FitResultsSource)
- add fit progress graph in PYMEVisualize as a docked panel on the right if we're looking at a live tabular source

Screen Shot 2020-07-21 at 10 05 16 PM

**Checklist:**

- [ ] Tested with numpy=1.14

1.16 at the moment

just 3.6 at the moment

just 4.0.4 at the moment

Yes, but only for cluster(ofone) users. I think PYMEVisualize popping up instead of a new page in dh5view will be pretty intuitive/hope people will be OK with it.

can also load non-live tabulars as live just fine - just loads a touch slower and gives you the (completed) progress panel on the right

### DIFF
--- a/PYME/DSView/modules/LMAnalysis.py
+++ b/PYME/DSView/modules/LMAnalysis.py
@@ -680,7 +680,12 @@ class LMAnalyser2(Plugin):
         #auto load VisGUI display
         #from PYME.DSView import modules
         #modules.loadModule('LMDisplay', self.dsviewer)
-        self.dsviewer.LoadModule('LMDisplay')
+        if self.use_cluster:
+            import subprocess
+            uri = 'pyme-cluster:///' + self.analysisController.results_filename + '/live'
+            subprocess.Popen('visgui %s' % uri, shell=True)
+        else:
+            self.dsviewer.LoadModule('LMDisplay')
 
 
     def GenFitStatusPanel(self, _pnl):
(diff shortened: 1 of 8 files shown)


barentine:
Following up from yesterday - whether this is used for dh5view live-view or not, I think it's quite nice for PYMEAcquire-launched localization and potentially for folks spooling from labview, etc. That said, also nice if you 'Go' twice from the same dh5view, with different fitfactory/settings.

There was some concern over a race condition with updating - this original issue was due to exiting the datasource __init__ once fitResults were added but before the polling thread had set the transkeys (recent refactor) which is now fixed.

There is also a small opening for a race condition if the datasource fitResults is being extended in an update while its being used in the recipe (small because this only applies to the first part of a recipe, and only if that recipe actually rebuilds from the farthest upstream datasource, which is currently doesn't. See #284). We've discussed adding an execution lock to `ModuleCollection` as not ideal, but there should also be an easy fix internal to the live datasource by removing the 'update lock' and adding a 'fitResults' lock instead

PR #354, opened 2020-07-24
Bring back mpld3 in requirements.txt since it's needed to launch PYME…

zacsimile:
…Visualize

Addresses issue #322.

Bugfix

**Proposed changes:**
- Add `mpld3` back in as a `requirements.txt` requirement since it's needed to launch PYMEVisualize.

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
addressed in #353

Issue #350, opened 2020-07-23
[BUG] trigger-based tiling ignores return_to_start due to callNum % imsize when callNum >= imsize


barentine:
**Describe the bug**
Trigger-based tiling with return_to_start set to False will effectively be ignored because a couple extra Ticks get added (during stop? Haven't looked yet), and `Tiler._position_for_index` uses % operator on callNum, so if we get any extras we go back to the start.

I printed the index in `CircularTiler._position_for_index`:
```
ind: 1324
467.2 -161.70000000000002 8362 2893 -27.753685 328.2463
ind: 1325
474.7 -138.1 8427 3098 -43.753685 386.2463
ind: 1326
481.0 -114.2 8482 3306 -59.753685 323.2463
ind: 1327
486.09999999999997 -90.1 8526 3515 -59.753685 69.246315
ind: 1328
490.0 -65.69999999999999 8560 3728 -56.753685 76.246315
ind: 1329
492.70000000000005 -41.099999999999994 8583 3941 -54.753685 256.2463
ind: 0
494.09999999999997 -16.400000000000002 8596 4156 -54.753685 230.2463
ind: 1
0.0 -0.0 4299 4299 -47.753685 58.246315
ind: 2
24.7 -0.0 4514 4299 -64.753685 60.246315
Warning: Camera buffer overflowing - purging buffer
5858 -3984 877
```

**To Reproduce**
1. run a triggered tile (with short integration times - don't know if this matters or not) with return to start unchecked/False
2. watch it effectively return to start, but right next to the start

**Expected behavior**
start at current position, stop at the last position and not spill over and restart again

**Additional context**
At one point I thought the CircularTiler was going outside-in instead of inside-out from start pos - I think that must have been from a scan where I forgot to turn the laser on at the beginning because things are definitely going the right direction until it folds over on itself with the modulo.

PR #355, opened 2020-07-24
pointScanner stop_on_complete, return_to_start, and dwellTime+trigger fix

barentine:
Addresses issue #350.

bugfix
**Proposed changes:**
- when we're finished,if stop_on_complete is true then  call stop() before moving to the 'next' position.
- if we're running in trigger mode, fire the trigger on each call, not just on calls that move the stage (i.e. don't ignore dwellTime)
- don't log StartAq's unless we're logging events

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4

### DIFF
--- a/PYME/Acquire/Utils/pointScanner.py
+++ b/PYME/Acquire/Utils/pointScanner.py
@@ -213,9 +213,9 @@ class PointScanner(object):
                     eventLog.logEvent('ScannerXPos', '%3.6f' % self.scope.state['Positioning.x'])
                     eventLog.logEvent('ScannerYPos', '%3.6f' % self.scope.state['Positioning.y'])
 
-                if cam_trigger:
-                    #logger.debug('Firing camera trigger')
-                    self.scope.cam.FireSoftwareTrigger()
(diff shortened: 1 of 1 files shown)


barentine:
I did kind of a poor job explaining what the error was, so I'll spell it out here:
The issue in #350 was that rather than stopping at `imsize` number of moves (the 0th move being in the `start`), we were going 3 over (inds 0, 1, and 2). One of those is due to a `>` which should have been an `==` (or a `>=` if we're feeling superstitious). One was due to a 0 vs 1 indexing error (we start callNum at zero, so `callNum == imsize` is actually one call too many). Finally, one was due to putting the if clause to check if we should stop after the position update for the next call, so we'd move to be ready for an acquisition on the next tick that we actually didn't want.

worth noting that:
- Moving the if clause checking if we should stop up doesn't change behavior for any `_stop_on_complete=False` stuff.
- removing a tab on the if software trigger clause should allow the thing to work for `dwellTime > 1` (not an interest of mine, but maybe helpful for someone)

PR #357, opened 2020-07-25
Add sphinx back as a pip requirement

zacsimile:
Addresses issue: From a fresh `pip` install, if you try to use the bakeshop in PYMEVisualize you can't add any recipe modules. Traceback:

```
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipeGui.py", line 266, in OnSelect
    from sphinx.util.docstrings import prepare_docstring
ModuleNotFoundError: No module named 'sphinx'
```

An alternative fix would be to wrap the docstring display stuff in a `try` block.

Bugfix

**Proposed changes:**
- Add sphinx back as a `requirements.txt` dependency.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -36,7 +36,7 @@ pyyaml
 psutil
 
 docutils
-#sphinx
+sphinx
 ujson
 jinja2
 django==1.11.29


barentine:
#353 / upvote / I agree

zacsimile:
Ah, sorry. Thanks for fixing it already!

PR #358, opened 2020-07-25
Measure clusters 3d fix

zacsimile:
If you try to run `Analysis>Clustering>Measure Clusters` you get the following traceback.

```
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 641, in execute
    m.execute(self.namespace)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/localisations.py", line 648, in execute
    measurements = np.zeros(maxLabel, cmorph.measurement_dtype)
TypeError: 'numpy.float32' object cannot be interpreted as an integer
```

If you fix this and then try to run `Analysis>Clustering>Measure Clusters` and any of the clusters are of size <3, the context manager stops the run and you get the following traceback.

```
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 641, in execute
    m.execute(self.namespace)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/localisations.py", line 663, in execute
    cmorph.measure_3d(x, y, z, output=measurements[cluster_index])
  File "/Users/zachcm/Code/python-microscopy/PYME/Analysis/points/cluster_morphology.py", line 147, in measure_3d
    raise UserWarning('measure_3D can only be used on clusters of size 3 or larger')
UserWarning: measure_3D can only be used on clusters of size 3 or larger
```

Unrelated: how do we display the results of "Measure Clusters"?

Bugfix

**Proposed changes:**
- Explicitly cast labels to integer
- Create a warning and then exit `measure_3d` function if cluster of size <3 is passed. This lets us keep looping through clusters in `OnMeasureClusters`. An alternative would be to wrap `measure_3d` calls in a `try` block.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Analysis/points/cluster_morphology.py
+++ b/PYME/Analysis/points/cluster_morphology.py
@@ -171,9 +171,6 @@ def measure_3d(x, y, z, output=None):
     #principle axes
     u, s, v = np.linalg.svd(np.vstack([x_c, y_c, z_c]).T)
 
-    print(x_c.shape, y_c.shape, z_c.shape)
-    print(s.shape)
-
     standard_deviations = s / np.sqrt(N - 1)  # with bessel's correction
     for i in range(3):
         output['axis%d' % i] = v[i]
--- a/PYME/Analysis/points/cluster_morphology.py
+++ b/PYME/Analysis/points/cluster_morphology.py
@@ -144,7 +144,9 @@ def measure_3d(x, y, z, output=None):
     N = len(x)
     output['count'] = N
     if N < 3:
-        raise UserWarning('measure_3D can only be used on clusters of size 3 or larger')
+        import warnings
+        warnings.warn('measure_3D can only be used on clusters of size 3 or larger', UserWarning)
+        return
     
     #centroid
     xc, yc, zc = x.mean(), y.mean(), z.mean()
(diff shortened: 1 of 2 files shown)


barentine:
Thanks for the py3 fix!

I'm slightly weary of changing the size relative to the input on a measure module like `localisations.MeasureClusters3D`. We currently output the measures without the localization information, meaning if you wanted to link that back to the tabular (maybe this is related to your display question) and hadn't already filtered on cluster size before-hand you would hit an `IndexError`. If memory serves the <3 points error hits in the SVD to get the cluster's principle axes. An alternative would be to try-catch that, find the two principle axes, and assign the third to zeros. Clusters of size 1 are a little weird to measure... maybe that one just gets zeros for everything.

Another option, which maybe solves your display question would be to have two outputs, one of which was the surviving input localizations(stand-alone or already merged with the measurements).

zacsimile:
I could be wrong, but I don't think this changes the size relative to the input in `localisations.MeasureClusters3D`, I think it will just skip measuring clusters of size <3 and leave the measurement at the default value of 0 for all entries in `cluster_morphology.measurement_dtype`. That said, I'm not opposed to fixing this a different way.

My display question was more along the lines of how do I look at the output of "Measure Clusters". Running it only updates `clusterAnalyser.clusterMeasures`. I assume there's another command we need to run to get a plot.

barentine:
Good point! 

Writing a report is definitely an option, but simplest way to allow result display would probably be to extend the recipe module to also return a MappingFilter on the input localizations with all the 1D metrics added as columns, maybe with an optional prepended CStr on the new column names so you could add the module to the data pipeline multiple times without clobbering if you want.

barentine:
would actually be pretty easy to write a report output which plotted all the 1D metrics, similar to fitted parameter plotting in nep-fitting (and could list the 3-component vectors in a table, also examples in nep-fitting reports).

Issue #356, opened 2020-07-24
[BUG] dh5view localization live-view broken by new dsview plugin stuff


barentine:
**Describe the bug**
After hitting 'go' in dh5view

```
Traceback (most recent call last):
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/LMAnalysis.py", line 266, in <lambda>
    self.bGo.Bind(wx.EVT_BUTTON, lambda e : self.lmanal.OnGo(e))
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/LMAnalysis.py", line 660, in OnGo
    self.analysisController.pushImagesCluster(self.image)
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/LMAnalysis.py", line 348, in pushImagesCluster
    self.onImagesPushed.send(self)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/dispatch/dispatcher.py", line 162, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/LMAnalysis.py", line 683, in OnImagesPushed
    self.dsviewer.LoadModule('LMDisplay')
  File "/home/smeagol/code/python-microscopy/PYME/DSView/dsviewer.py", line 253, in LoadModule
    modules.loadModule(moduleName, self)
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/__init__.py", line 118, in loadModule
    ret = mod.Plug(dsviewer)
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/LMDisplay.py", line 324, in Plug
    return LMDisplay(dsviewer)
  File "/home/smeagol/code/python-microscopy/PYME/DSView/modules/LMDisplay.py", line 76, in __init__
    visCore.VisGUICore.__init__(self)
  File "/home/smeagol/code/python-microscopy/PYME/LMVis/visCore.py", line 88, in __init__
    gl_pan = wx.Panel(win)
TypeError: Panel(): arguments did not match any overloaded call:
  overload 1: too many arguments
  overload 2: argument 1 has unexpected type 'weakproxy'
```

**To Reproduce**
1. click 'Go' on a series in dh5view at the latest head

**Expected behavior**
open a live view

**Screenshots**
![image](https://user-images.githubusercontent.com/31105780/88433084-700a4d80-cdcb-11ea-9f55-4d63c157dd35.png)

[...]

PR #363, opened 2020-07-28
Covert weakref.proxy to weakref.ref

zacsimile:
Addresses issue #356.

Bugfix

**Proposed changes:**
Convert `weakref.proxy` to `weakref.ref`. My understanding is `weakref.proxy` prevents us from having to explicitly dereference when we want to use the `weakref`ed object. However, since we access this object strictly through the `dsviewer` property of plugin, we have the opportunity to deference every time we need the object. For some reason, this works better than proxy.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/DSView/modules/_base.py
+++ b/PYME/DSView/modules/_base.py
@@ -16,11 +16,11 @@ class Plugin(object):
     """
     
     def __init__(self, dsviewer):
-        self._dsviewer = weakref.proxy(dsviewer)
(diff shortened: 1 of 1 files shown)


zacsimile:
OK, I see `weakref.proxy` doesn't increment the `weakref` count, whereas `weakref.ref` does. That should be OK, though, I think. It should still eliminate circular reference problems.

David-Baddeley:
Had been thinking of doing exactly the same for #356

PR #364, opened 2020-07-28
Add title to FRC plot window

zacsimile:
Addresses issue # .

Enhacement

**Proposed changes:**
Add a title to the FRC plot window (instead of just calling it `Figure 1`). It would be good to title the second window as well, but I don't know what to call it. Best title I've come up with so far is `Fit Intercept`. Thoughts?

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
I wouldn't worry too much about the 2nd window - It's basically a debug plot and will probably go away. I think you might also be able to pass the title to the `figure()` call, although this might be better as that would re-use the figure for subsequent FRC plots.

PR #365, opened 2020-07-28
Photo fix its

zacsimile:
Addresses issue: display alignment of photophysical variables in plots was off. Error was thrown every time we tried to run photophysics even though 2/3 photophysics measurements work just fine.

Bugfix/enhancement

**Proposed changes:**
- Fix alignment in display of photophysical constants
- Don't throw an Error/Traceback if you run Analysis>Photophysics without clumping

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/LMVis/Extras/photophysics.py
+++ b/PYME/LMVis/Extras/photophysics.py
@@ -38,7 +38,11 @@ class DecayAnalyser:
         kinModels.fitDecay(pipeline)
         kinModels.fitFluorBrightness(pipeline)
         #kinModels.fitFluorBrightnessT(pipeline)
-        kinModels.fitOnTimes(pipeline)
+        if 'clumpSize' in pipeline.keys():
+            kinModels.fitOnTimes(pipeline)
+        else:
+            import warnings
+            warnings.warn('To fit on times, first run Corrections>Chaining>Find consecutive appearances and Corrections>Chaining>Clump consecutive appearances.', UserWarning)
         
 
     def OnRetrieveIntensitySteps(self, event):
--- a/PYME/LMVis/Extras/photophysics.py
+++ b/PYME/LMVis/Extras/photophysics.py
@@ -35,15 +35,14 @@ class DecayAnalyser:
         
         pipeline = self.visFr.pipeline
 
+        if 'clumpSize' not in pipeline.keys():
+            # Clump 
+            self.visFr.particleTracker.OnFindClumps()
+
         kinModels.fitDecay(pipeline)
         kinModels.fitFluorBrightness(pipeline)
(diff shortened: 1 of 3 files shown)


PYME/LMVis/Extras/photophysics.py
```diff
@@ -38,7 +38,11 @@ def OnCalcDecays(self, event):
         kinModels.fitDecay(pipeline)
         kinModels.fitFluorBrightness(pipeline)
         #kinModels.fitFluorBrightnessT(pipeline)
-        kinModels.fitOnTimes(pipeline)
+        if 'clumpSize' in pipeline.keys():
+            kinModels.fitOnTimes(pipeline)
+        else:
+            import warnings
+            warnings.warn('To fit on times, first run Corrections>Chaining>Find consecutive appearances and Corrections>Chaining>Clump consecutive appearances.', UserWarning)
```
David-Baddeley: I'm not sure this is 100% correct - I have a suspicion it expects to be run with chaining but without clumping (I seem to recall the model including a term to correct for the over-counting that you would get if you didn't clump)

David-Baddeley: Also, do we want this to be a visible warning / message box?

This stuff is very non-critical, although the entire photophysics stuff could use a bit of a re-write at some point.

zacsimile: Re: first point. You're right, you only need to run `Corrections>Chaining>Find consecutive appearances`.

I thought about this. I made this edit because I got tired of seeing the context manager every time I ran this when I was testing it out. It would probably confuse the user if we threw a traceback. It could be a dialog box, although that could also be kind of annoying. Mostly we could use clear documentation on this and then just expect users to follow that and throw the subtle warning.

David-Baddeley: Longer term it might make sense just to do the clumping if the user hasn't already, or always do our own clumping. The existing clump but don't coalesce stuff is pretty error prone. Another option would be to pop up an empty figure for the on-time histogram with a `fig_text` error message.

zacsimile: All of this is a good idea/fairly straightforward to implement. Any preference on what we do now? I'm leaning toward just clumping if the user hasn't already.

barentine: That's what we do in `multiviewMapping` if one 'merges' and hasn't already 'clumped'. In practice, knowing that, I never use the `clump` call except for in recipes

zacsimile:
Converted to automatically run `OnFindClumps` if a user runs `Photophysics>Estimate decay lifetimes` before declumping.

Issue #360, opened 2020-07-27
[PROPOSAL] TilePyramid construction compatible with chained analysis


barentine:
**Is your feature request related to a problem? Please describe.**
Problem is how run a tile_trigger.py type protocol spooling to cluster and assemble a `tile_pyramid.ImagePyramid` automatically, and load that as a `SuperTileDataSource` into a chained recipe (#272).

We can build `ImagePyramid`s locally from a (non-local) pcs, or build one locally live if we don't use an acquisition protocol and back-door things through `tiler`.

I don't need to build the pyramid live necessarily, but do need to be able to acquire it using an acquisition protocol and somehow have it built so I can run a recipe on the cluster.

**Describe the solution you'd like to implement**
I would advocate for the short/sweet/potentially not performance optimized route to get things up and running before plowing into a major rewrite of `ImagePyramid` construction, or special-casing an exception to the write-once cluster philosophy. The simplest solution in my eyes is to
- write an `ImagePyramidOutput` recipe module which builds/saves a pyramid locally using `tile_pyramid.create_pyramid_from_dataset`.

This would allow us to simply chain a recipe to the `tile_triggered`-type protocol which saves the pyramid, and chained to that would be a recipe which runs ___ (e.g. detection). It wouldn't build the pyramid live with acquisition, but for the use case I'm currently up against currently the latency will be mostly masked.

**Describe alternatives you've considered**
- write a recipe module to build a single pyramid level from a datasource, and use the `ImageOutput` module to hdf it. This would probably be pretty painful wrt RAM unless we looked through the positions first to see which datasource frames contribute to a given SuperTile frame and built it through that way. Even then, for large tilings we'd definitely be going into swap, and would probably be more work than the proposed solution.
- build the pyramid live on the instrument computer as we currently do through the `tile_panel`, load it into a `SuperTileDataSource` and then save that to the cluster as hdf or pzf and then launch a chained analysis task on it. Would be a little hacked in how we would accomplish that chained analysis launching _after_ the pyramid is complete. Could be done with a `TilePyramidSpooler` but I think that's getting pretty gross.

**What potential side effects do the proposed changes have?**
- Is it new functionality, or a change to existing functionality?

new
- Will it change how users interact with the software?

don't think anyone else is using SuperTiles as of yet, but would extend their ability to use them in recipes/write custom tiling acquisitions if they have some special needs.
- Will it change the dependency landscape?

nope

barentine:
taken care of by #379

PR #368, opened 2020-07-28
add output module to create an ImagePyramid

barentine:
Addresses issue #360 .

enhancement
**Proposed changes:**
-add recipe output module which builds a local `tile_pyramid.ImagePyramid` out of it.

**NOTE**
We don't currently support loading a non-local SupertileDatasource. Any thoughts on the cleanest way to do that @David-Baddeley? Alternatively we could convert this module into a non-output module which just happens to build the pyramid locally then wrap it as an ImageStack via SupertileDatasource (i.e just one level) which could then be used in the rest of the recipe. That might also allow for saving in a clusterIO-compatible format via ImageOutput so you could load a remote one and use the same pyramid in more than one recipe (without clobbering, etc.). I actually think that may be the cleanest.

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

**test script**
I didn't write a unit test for this, but used the following
```
#!/usr/bin/python

import os
from PYME.recipes.base import ModuleCollection
from PYME.recipes.output import ImagePyramidOutput

recipe = ModuleCollection()
recipe.loadInput('pyme-cluster:///Bergamot/2020_7_28/tile_trigger_dispatch_stop.pcs')

recipe.add_module(ImagePyramidOutput(recipe, input_stack='input'))

recipe.execute()

recipe.save(context={'file_stub': 'tile_trigger_dispatch_stop',
                     'output_dir': 'F:\\test_pyramid_recipe'})

```


barentine:
It's a bit gross to create a directory outside of an output module, and it's also gross to create something which clusterIO can't really deal with, so maybe the cleanest thing to do without adding a SUPERTILE endpoint to the dataservers is to build the pyramid in a tmp directory and save SupertileDatasource(s) of specified layer(s) as HDF(s). Is that crazy?

barentine:
see #379

PR #369, opened 2020-07-29
unpin numpy from 1.14 in requirements.txt

barentine:
We have tested numpy 1.16 on python 3 much more than 1.14, and the pin makes it harder for people to help us test py3

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 #PIP requirements
 #NB - this might be incomplete / buggy, the conda route is currently more reliable
 
-numpy==1.14.*
+numpy>=1.14.*
 scipy
 matplotlib
 wxpython
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,8 @@
-#PIP requirements
-#NB - this might be incomplete / buggy, the conda route is currently more reliable
+# PIP requirements
+# NB - this might be incomplete / buggy, the conda route is currently more reliable
+# 
 
-numpy>=1.14.*
+numpy>=1.14.* # we build and test against 1.14, but should work against any higher version. Try numpy==1.14.* if you run into issues.
 scipy
 matplotlib
 wxpython


David-Baddeley:
As long as we keep a hard pin for testing (which we currently do with the test_requirements) and for building packages (might need to double check our pip build scripts) so that we can guarantee that we keep testing and building against 1.14 this is probably OK.

David-Baddeley:
I find the logic a bit odd though - is there a concrete reason not to recommend numpy 1.14?

zacsimile:
Andrew and I moved to 1.16 a while ago because of a pytables conflict that I think has since been resolved. Because of this, all of our py3 testing is in 1.16, so it's arguably the "supported" version of numpy for py3 at this point. My understanding is this is the exact same logic we use for pinning to 1.14 (i.e. it's the one we've tested on and therefore support).

barentine:
x2 for the main reason not to recommend 1.14 (which I view as not even the question, since we're discussing pinning) being that we've tested more on py3 with 1.16 (and at one point, somewhere, I thought a numpy >=1.15 kwarg snuck into the main branch). I don't have huge beef with 1.14, but early on we were boxed in by some versioning/packaging issues (pytables as Zach mentioned, and also scipy at one point). I don't know if all of that still holds - some of it might have been from when I was trying to keep the cluster py2 even, or more ~legacy-type packages may have been built more recently. Still, the later numpy versions really ease up packaging concerns, particularly if you want to e.g. install keras/tensorflow and you have a new-release GPU which doesn't go back too far in compatible CUDA versions... (I did, for the record, try to set things up on the instrment computer with numpy 1.14 as recently as March, but despite using the oldest CUDA I could it got bumped to 1.16 by pip when I got around to tensorflow [not even 2.x])

I also think we probably all have our personal tastes when it comes to e.g. updating our laptop OS, etc. Personally, I hate the idea of missing out on fixes/performance improvements (current numpy release being 1.19, and even by 1.16 including performance improvements for things we use like views rather than slices for structured arrays and using half the RAM when pickling), but even more I get nervous about e.g. doing our own hdf5 locking for so long its entrenched and then one day pytables starts doing it on its own (not that they did a great job with that on any versions we've tested so far), or we create extensive nameserver support to avoid performance issues on windows w/ zeroconf only to see that 11 versions of zeroconf past our pin they've also addressed a couple windows concerns. I know you've been burned by some upgrades in the past David, but I'm sure we've burned people on PYME versions too and maybe that's just a part of the game.

Anyway, I'm really digressing, but in the course of trying to get things done I had to bump it to 1.16 at one point (or several points), and as more people start using PYME in their maybe already python, maybe already python 3.9 / numpy 1.19 workflows, I think it'd be nice if the pro-move developer pip install was more flexible, and I think it's really nice that we've got the testing set up with a separate requirements.txt.

David-Baddeley:
There's at least one remaining conflict with numpy > 1.14 that I'm aware of (in shmarray, which will largely effect jittered triangulation as well as some of the other parallel stuff). Fixed with more recent versions of ctypes, but would require a ctypes pin (along with potentially others). Probably more of an issue for people upgrading old installations than anyone installing from scratch.

PR #372, opened 2020-07-29
Add recipe/localization chaining to PYMEAcquire

barentine:
rule refactor/gui extracted from the too-big-to-live PR #272 

enhancement

**Proposed changes:**
- add localization and recipe rule classes which have their own post methods
- add rule chain which links inputs/outputs and nests all subsequent rules into the follow-on section of the first
- add optional recipe page/chained analysis GUI to PYMEAcquire

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.x

related to #350 #360, highly relevant to htsms project, and potentially even [this recent image.sc conversation](https://forum.image.sc/t/pyme-camera-roi-changes-via-protocol-and-camera-maps/40994)
will need to add some rst doc at some point I suspect


barentine:
Note that the protocol-rule association we've discussed @David-Baddeley does not work unless protocols, even 'temporary' ones [I still think the distinction is pointless bordering on erroneous] set in startspooling change the state of the spool controller (unless we want to e.g. send the current protocol as a kwarg in the dispatch signal send onStopSpooling or onStartSpooling. Even then we'd need to add a property or so

barentine:
correction - looks like we can handle it by not looking at spoolcontroller but rather spoolcontroller.spooler, modulo fixing protocol.filename to actually be set to something other than None

PYME/cluster/rules.py
```diff
@@ -0,0 +1,469 @@
...
+        self._rule_id = None
+        self._max_tasks = 1
+
+    @property
+    def template(self):
+        return self._template
+
+    @property
+    def rule(self):
+        rule = {'template': json.dumps(self.template)}  # todo - change rulenodeserver so we don't have to dumps template first
```
David-Baddeley: The template should be a template string. This is not going to change on the ruleserver.

David-Baddeley:
Just a heads up, didn't see the rule class stuff when going over the PR initially (was collapsed in the diff view because the diff was too large). I think this aspect still needs a little more review. 

For future reference, this should really have been 2 separate PRs - one for the pyramid stuff, which is now ready to merge, and one for the rule stuff.

David-Baddeley:
Wait a minute - this is super confusing - it seems that the changes from 379 are appearing in here as new changes ...

barentine:
Yeah github can be a little funky if you merge things at different times and I pull from master, etc.. Let me try rebasing it and we'll see if that clears up what's in this PR/not (hopefully your supertile doc change will stick)

barentine:
There @David-Baddeley, should be sorted. Kept your update on the supertile comment too! Thanks for having a look at this

PYME/cluster/rules.py
barentine: Works for me, let's take out the comment then

David-Baddeley:
Not super keen on your `Rule` class implementation (details to follow), but want to make sure that this doesn't hold you up too much. As I understand it, you currently only need to have this merged on your instrument computer, correct?

As I recall the rationale for a rule class was to have something to maintain a list / linked list of in the UI, and what you've come up with is quite different to what I'd originally envisaged. What really bugs me about this implementation is using the same object and re-writing the various inputs and outputs for each series. Because we are re-using the list of rules for each series (or each series of a given type) I'd envisaged the "rule" objects which got stored in the list as being something which knew how to create a rule for a given input name (ie. a form of rule factory), rather than being a full embodiment of a rule complete with a specific series name, ID, etc .... The factory concept would also us avoid the need for a separate rule list (the rule chain could easily just be implemented as a linked list, with each rule delegating the on_completion section of the json to the next rule in the linked list).

barentine:
I'm not entirely sure how your rule-factory vision works out with rules needing to generate/nest themselves in reverse order. It might be simple, but past 23 on a Sunday that's drawing a question mark on my end. Also, is there an advantage to this other than that's how you envision it?

As far as how much this holds me up: yes it's only on my instrument computer. However, as we've seen with getting it to this state, it's involved touching up several other areas to get this working, and that means overhead on my end not to just lump it into a #272 style mega PR (which itself wasn't everything I was doing in PYME at the time, but just the chaining-related changes). We discussed this refactor back in April, I posted my implementation for discussion in June more than 2 months ago, and while I know we've been busy with loads of other things, this has been on my priority list that whole time, and would be a bit of a bummer in my eyes to hold me up again now. I'm totally OK with us reworking the rule further, but what is the rational for not merging this as is and reworking from there? That would, I think, give me the most flexibility to keep making progress, and doesn't preclude a further rewrite (we can talk about where that falls into our current priorities tomorrow afternoon).

David-Baddeley:
I'm going to try and explain my concerns with the proposed rule refactor (separate to is this an acceptable stop-gap) as follows:

- Rule pushing is a fairly core part of our distributed analysis infrastructure. As such, it is something that other people are likely to need to read, understand, or potentially adapt and/or use as a reference implementation when writing, e.g. a umanager rule pusher. 
- If I create a `Rule` object with a given `analysis_metadata`, `series_uri`, `_rule_id` etc ... I'd kind of expect these to stay constant and not change on me. I'd especially not expect them to be re-written by a method called `chain_rule()`. 
- If a `Rule` object is going to be re-used for multiple different series/rules, then the series specific parameters should be passed to a method of the rule object, rather than being properties of that object itself.


The implementation I was imagining looks something like this:

``` python
class LocalisationRuleFactory(object):
    on_completion = None #RuleFactory instance for rule to execute when this rule completes

    def generate_template(self, results_metadata_uri, series_uri, ...): 
        #similar to HTTPRulePusher._TaskTemplate
        tt = {'id': '{{ruleID}}~{{taskID}}',
                      'type':'localization',
                      'taskdef': {'frameIndex': '{{taskID}}', 'metadata':results_md_uri},
                      'inputs' : {'frames': series_uri},
                      'outputs' : {'fitResults': resultsURI+'/FitResults',
                                   'driftResults': resultsURI+'/DriftResults'}
                      }
        return json.dumps(tt)

    def generate_rule(self, dataSourceID, metadata, ...):
        # do stuff in pusher __init__ **or** create a pusher instance
      
        rule = {'template' : self.generate_template(...)}
        if self.on_completion is not None:
            rule['on_completion'] = self.on_completion.generate_rule(...)
         

```

PR #373, opened 2020-07-29
make standard acquisitiontasks panels optional

barentine:
Addresses issue artistic differences / different use cases

enhancement
**Proposed changes:**
- allow users to specify if they want the standard z-stack and Time / Blinking series panels rather than always adding them

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.x

### DIFF
--- a/PYME/Acquire/acquiremainframe.py
+++ b/PYME/Acquire/acquiremainframe.py
@@ -51,7 +51,7 @@ from PYME.Acquire.ui import HDFSpoolFrame
 
 from PYME.Acquire import microscope
 from PYME.Acquire import protocol
-
+from PYME import config
 from PYME.IO import MetaDataHandler
 #from PYME.IO.FileUtils import nameUtils
 import six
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Will address in more detail later, but don't like having it as a config option. Also this addresses your pain point but not necessarily that of others ...

long term plan might be something along these lines:

- express layout as a dict with some default values e.g. (might want to think a bit more about how best to structure this)

``` python
{ 
    'cam_panels' : ['display_scaling', 'integration_time', 'camera'],
    'acquisition_panels' : ['z_stack', 'sequence'],
    ....
}
```

- have a dictionary of panel constructor functions ( means that in `doPostInit()` we have something like 
    ``` python
    for p in self.layout['cam_panels']:
        self.AddCamTool(self.panel_constructors['p'](self, self.scope))
   ```
- replace most of the `@init_gui` decorated sections in the init file with a) registering a panel constructor function and b)  appending it's name to the appropriate layout entry
- take some of the more commonly used stuff (e.g. laser control panels etc) out of the init scripts and make them autodetect the hardware.
- this puts the builtin panels on a similar footing to ones defined in the init script

barentine:
I'm OK with overhauling the panel initialization, but will hold off investing any time in it until we're on the same page. Meanwhile, I don't see what's untenable about a config-option to hold us over.

Issue #340, opened 2020-07-20
[BUG] ROIExtractNR fit results dtype incompatible with tabular dtype unnesting


barentine:
**Describe the bug**
ROIExtractNR fit result dtype is kind of weird, and breaks saving results (at least on py3)

printing the dtype, followed by each `n` in `tabular.unNestDtype` to show what we're hitting the error on (i.e. `<`)
```
[('tIndex', '<i4'), ('x', '<f4'), ('y', '<f4'), ('A', '<f4'), ('data', 'f4', (15, 15, 1)), ('sigma', 'f4', (15, 15, 1)), ('sp', [('Xg0', '<f4'), ('Yg0', '<f4'), ('Xr0', '<f4'), ('Yr0', '<f4')])]
pixel size: 115.0
('tIndex', '<i4')
('x', '<f4')
('y', '<f4')
('A', '<f4')
('data', '<f4', (15, 15, 1))
<
ERROR:PYME.recipes.base:Error in recipe module: <__main__.ExtractROIs object at 0x7fd191c729e8>
Traceback (most recent call last):
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/base.py", line 641, in execute
    m.execute(self.namespace)
  File "secret/pyme_series_script.py", line 237, in execute
    inputPositions=positions,
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/base.py", line 206, in apply_simple
    return self.apply(*args, **kwargs)[next(iter(self.outputs))]
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/base.py", line 185, in apply
    self.execute(namespace)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/measurement.py", line 233, in execute
    res = tabular.FitResultsSource(r, sort=False)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/tabular.py", line 235, in __init__
    self.setResults(fitResults, sort=sort)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/tabular.py", line 256, in setResults
    self._keys = unNestDtype(self.fitResults.dtype.descr)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/tabular.py", line 209, in unNestDtype
    unList += unNestDtype(n[1], parent + n[0] + '_')
  File "/Users/Andrew/code/python-microscopy/PYME/IO/tabular.py", line 209, in unNestDtype
[...]

PR #374, opened 2020-07-30
Fixes to make Dumbell3DFitR work

zacsimile:
Addresses three issues:

* `FFBase.get3DROIAtPoint` slices `z` using `yslice`
* `self._keys` from `Dumbell3DFitR` returns a `tuple` (I expect other fitters do this too)

```
TypeError                                 Traceback (most recent call last)
<ipython-input-10-3dee89596212> in <module>
      2 from PYME.IO import tabular
      3 res_t = tabular.FitResultsSource(res)
----> 4 res_t.toDataFrame().to_csv('my_data.csv')

~/Code/python-microscopy/PYME/IO/tabular.py in toDataFrame(self, keys)
     59         import pandas as pd
     60         if keys is None:
---> 61             keys = self.keys()
     62 
     63         d = {}

~/Code/python-microscopy/PYME/IO/tabular.py in keys(self)
    268 
    269     def keys(self):
--> 270         return self._keys + list(self.transkeys.keys())
    271 
    272     def __getitem__(self, keys):

TypeError: can only concatenate tuple (not "list") to tuple
```
* Underscores are implictly used for nesting in `FitResults`, so my `sigma_xy` and `sigma_z` variables cause problems
```
ValueError                                Traceback (most recent call last)
<ipython-input-10-3dee89596212> in <module>
      2 from PYME.IO import tabular
      3 res_t = tabular.FitResultsSource(res)
----> 4 res_t.toDataFrame().to_csv('my_data.csv')

~/Code/python-microscopy/PYME/IO/tabular.py in toDataFrame(self, keys)
     64 
     65         for k in keys:
---> 66             v = self.__getitem__(k)
     67             if np.ndim(v) == 1:
     68                 d[k] = v

~/Code/python-microscopy/PYME/IO/tabular.py in __getitem__(self, keys)
    288             return self.fitResults[k[0]][k[1]][sl]
    289         elif len(k) == 3:
--> 290             return self.fitResults[k[0]][k[1]][k[2]][sl]
    291         else:
    292             raise KeyError("Don't know about deeper nesting yet")

ValueError: no field of name sigma
```

**Proposed changes:**
* Make `FFBase.get3DROIAtPoint` slice `z` using `zslice`
* Cast `self._keys` explicitly to a list. There are probably better ways to fix this (e.g. ensure `unnest_dtype` returns a list, but I didn't want to touch that function since I'm not sure it's supposed to return a list. It would be nice if this function had a docstring indicating what it's supposed to return. In general, it would be nice if all functions at least indicated input/output types), but this works.
* Change `sigma_xy` and `sigma_z` to `wxy`, `wz`, as in `Gauss3DFitR`

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -267,7 +267,7 @@ class FitResultsSource(TabularBase):
 
 
     def keys(self):
-        return self._keys + list(self.transkeys.keys())
+        return list(self._keys) + list(self.transkeys.keys())
 
     def __getitem__(self, keys):
         key, sl = self._getKeySlice(keys)
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -259,7 +259,7 @@ class FitResultsSource(TabularBase):
         #allow access using unnested original names
         # TODO???? - replace key translation with a np.view call?
         #self._keys = unNestDtype(self.fitResults.dtype.descr)
-        self._keys = unnest_dtype(self.fitResults.dtype).names
+        self._keys = list(unnest_dtype(self.fitResults.dtype).names)
         
         #or shorter aliases
         self._set_transkeys()
@@ -267,7 +267,7 @@ class FitResultsSource(TabularBase):
 
 
     def keys(self):
-        return list(self._keys) + list(self.transkeys.keys())
+        return self._keys + list(self.transkeys.keys())
 
     def __getitem__(self, keys):
         key, sl = self._getKeySlice(keys)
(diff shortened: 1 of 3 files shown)


barentine:
The recent changes were David helping with another error #340. The keys should definitely be list not tuple, and this is resulting from the recent change. numpy.dtype.names returns a tuple, while in `FitResultsSource.setResults` we were previously setting `_keys` with a list coming out of `unNestDtype`. 

Zach you'll want to move the list wrap from `keys` property to to `self._keys = unnest_dtype(self.fitResults.dtype).names` in `FitResultsSource.setResults`

David-Baddeley:
I'm not 100% sure that the keys should be a list rather than a tuple. My gut tells me that they should possibly be a tuple (immutable). In practice it probably doesn't matter, will merge as a pragmatic step (something downstream evidently expects them to be a list), but we should consider revisiting at some point.

PR #377, opened 2020-07-31
str->bytes error in dataserver

barentine:
image


David-Baddeley:
Will merge, as most likely safe. Just wanted to flag the potential for py2 vs py3 issue here (py2 `.dumps()` takes an encoding argument implying that the output is already bytes). In practice it's unlikely to matter as py2 is permissive with `.encode()`. Possibly complicated by using ujson instead of builtin json (haven't checked if this is the case here).

PR #379, opened 2020-08-02
Allow Supertile datasources to be created directly from raw tile series in recipes

barentine:
Addresses issue #368.

enhancement

**Proposed changes:**
- change `PYME.IO.Datasources.SupertileDatasource` to accept an already constructed `ImagePyramid` in it's `__init__`, adding a staticmethod to replace the old-interface where we create an ImagePyramid from a raw tile series first.
- add `recipes.supertile.py`
- move `TilePhysicalCoords` recipe module from its awkward home in `recipes.measurement`
- add `recipes.supertile.Supertile` module to create an `ImageStack` wrapping a `SupertileDatasource` using a temporary directory for creating the underlying `tile_pyramid.ImagePyramid`, which gets removed during execution. Consequently, the resulting datasource must be completely cached in RAM
- allow cache size to be set when creating an `ImagePyramid`, using PZF IO for cached, and sql for cache size of zero.
- add `recipes.supertile.SupertileOutput` to create an `ImagePyramid` from a raw series in a temporary directory, but rather than creating a cache-only `SupertileDatasource`, create one and save it to disk.

**Checklist:**

- [ ] Tested with numpy=1.14

only whitespace changes were when I was already changing the signature and we were quite over 80 characters on a line. I did move tilephysicalcoords into the new supertile.py recipe file, addressing a TODO issue there.


barentine:
@David-Baddeley - as suggested, I've moved the tempdir stuff out of the recipe module and into ImagePyramid and reverted the cache size setting stuff. I think you're right that this is much cleaner.

```
In [2]: recipe.loadInput('SUPERTILE:pyme-cluster:///Bergamot/2020_7_28/tile_trigger_dispatch_stop.pcs?level=2')                                                                                                            
filename == SUPERTILE:pyme-cluster:///Bergamot/2020_7_28/tile_trigger_dispatch_stop.pcs?level=2
filename == pyme-cluster:///Bergamot/2020_7_28/tile_trigger_dispatch_stop.pcs
DEBUG:PYME.Analysis.tile_pyramid:Adding base tiles ...
DEBUG:PYME.Analysis.tile_pyramid:Added base tiles in 2.838041s
DEBUG:PYME.Analysis.tile_pyramid:9.965896606445312e-05
DEBUG:PYME.Analysis.tile_pyramid:Updating pyramid ...
DEBUG:PYME.Analysis.tile_pyramid:0.7826032638549805
DEBUG:PYME.Analysis.tile_pyramid:Done
In [3]: recipe.namespace['input'].data._pyr.base_dir                                                                                                                                                                       
Out[3]: '/tmp/tmplz9xfq8_'
In [4]: import glob                                                                                                                                                                                                        
In [5]: glob.glob('/tmp/tmplz9xfq8_')                                                                                                                                                                                      
Out[5]: ['/tmp/tmplz9xfq8_']
In [6]: glob.glob('/tmp/tmplz9xfq8_/*')                                                                                                                                                                                    
Out[6]: 
['/tmp/tmplz9xfq8_/4',
 '/tmp/tmplz9xfq8_/2',
 '/tmp/tmplz9xfq8_/1',
 '/tmp/tmplz9xfq8_/3',
 '/tmp/tmplz9xfq8_/metadata.json',
 '/tmp/tmplz9xfq8_/0']
In [7]: del recipe.namespace['input']                                                                                                                                                                                      
In [8]: glob.glob('/tmp/tmplz9xfq8_/*')                                                                                                                                                                                    
Out[8]: []
```

PYME/IO/DataSources/SupertileDatasource.py
```diff
@@ -38,24 +38,36 @@
...
 class DataSource(BaseDataSource):
     moduleName = 'SupertileDataSource'
-    def __init__(self, filename, taskQueue=None):
-        from PYME.Analysis.tile_pyramid import ImagePyramid
-        self.tile_base, query = filename.split('?')
-        qp = parse_qs(query)
-        self.level = int(qp.get('level', [0])[0])
-        self.stride = int(qp.get('stride', [3])[0])
-        self.overlap = int(qp.get('overlap', [1])[0])
+    def __init__(self, pyramid, level=0, stride=3, overlap=1):
```
David-Baddeley, line 41: I'll try and re-work this. 

For future reference, a `DataSourceModule.DataSource` **must**  accept a filename in the constructor (this is, amongst other things, related to how localisation analysis can work for file types other than .h5 and .pcs). The solution in this case is probably to use inheritance rather than a static method - i.e. have `SupertileDatasource.SupertileDataSource` with your new constructor, and `SupertileDatasource.DataSource` as an over-ridden version (or potentially even just a function which calls the static method).

David-Baddeley: Actually, just realised that the `from_tile_series(...)` static method does something completely different to what the initial constructor does - namely it creates a pyramid from scratch, rather than loading an existing pyramid.

David-Baddeley:
@barentine - this is now in a state where I should be happy to merge, at least in the interim (with a few caveats in place as commented). Kicking back to you to double check it won't break your workflow as I've changed a few things quite a lot. Specifically if you are creating SupertileDataSources directly from raw image series rather than already constructed pyramids (which you probably shouldn't be :p).

barentine:
Hi David, I'll have a look. Would note that the entire point of this PR was to create SupertileDataSources directly from raw image series rather than already constructed pyramids... Otherwise where do we construct the pyramid?

barentine:
Thanks for kicking it back to me, David. I see that your point is to put the pyramid directly into the recipe namespace rather than only ImageStacks. Frankly, I think you would have rejected it as breaking the data model if I had submitted it that way. 

The use case I see you mention is that if you load from filename and have to construct the pyramid, as you would in localization analysis, it will take time. That is clearly true, but is not solved by creating a pyramid in memory in a `ModuleBase`, where it would then only be accessible to the single process anyway.

Note that I originally wrote, discussed, and was convinced to remove an output module version, which is what one would actually need if they wanted to pyramid once and then distribute frames over multiple processes/computers in a subsequent chained rule. The difficulty here is that the output module in the case of a pyramid would save a directory, which would require a change to the dataservers/clusterIO to pickup, while pyramiding and creating an ImageStack in the same module allows you to save it as normal using ImageOutput and then load again, etc. in subsequent recipes if you like.

Do I miss something? Sorry if being brief comes off as gruff - I appreciate you looking through this before I'm back at the scope tomorrow

David-Baddeley:
As it stands, this should work as you describe in the recipes - the main change is on the command line `dh5view SUPERTILE://...` (or `ImageStack(filename='SUPERTILE:\\...')` has reverted to it's original behaviour which is to create a supertile datasource from an existing pyramid. This is somewhat non-negotiable for a merge (I don't want to be recreating the pyramid when I load supertiles from an existing pyramid). I also think the changes would have broken the old single coverslip high throughput approach.

Where I think we could use an additional change is if the recipe module (which currently creates both pyramid and supertiles) was split.

David-Baddeley:
I think we can leave the `Supertile` recipe module for now. The main argument for separating them is looking ahead to when we fix the cluster creation problem, allowing us to super-tile different multi-resolution inputs such as zarr, and allowing multiple workflows at different resolutions to run off a single pyramid. Longer term it might actually make sense to make a pyramid (or similar multi-resolution image format) a first-class object. It could then be appropriate for this to have a supertiles attribute ... Some of this stuff also wants to support lightsheet and a line-scan confocal extended volume imaging system I'm involved with here.

David-Baddeley:
The multiple workflows at different resolutions on the same pyramid bit is something which might hit us earlier than some of the others.

David-Baddeley:
Visualisation and super-tiled analysis is also something where you want to do more than one thing with the same pyramid. I see the current temporary directory stuff as a short term fix for generating pyramids on the cluster, but fully intend that we will solve this properly at some point.

barentine:
Thanks for the clarification. I'm OK with having to e.g. load a tiled series in dh5view and run a recipe module on it to get a supertile if I haven't already made a pyramid. I think your comment that I shouldn't be creating supertiledatasources directly from image series threw me for a loop, but I see what you mean now.

I'm all for a first-class datatype supporting multiple resolutions, but think that the timeline we'll achieve that makes a `WARNING` and a `FIXME` in the supertile module a little excessive, but that's probably just my ego talking.

Please feel free to merge as is.

Issue #376, opened 2020-07-31
[BUG] PYME.recipes.skfilters argspecing chooses some things incorrectly


zacsimile:
**Describe the bug**
Here's a good example: the following is the definition for `skimage.filters.threshold_local` in `sckikit-image=0.17.2`.

```
def threshold_local(image, block_size, method='gaussian', offset=0,
                    mode='reflect', param=None, cval=0):
```

The inspector argspec returns

```
FullArgSpec(args=['image', 'block_size', 'method', 'offset', 'mode', 'param', 'cval'], varargs=None, varkw=None, defaults=('gaussian', 0, 'reflect', None, 0), kwonlyargs=[], kwonlydefaults=None, annotations={})
```

We default unknown arguments to type `float`. This makes `block_size` a float. This is incorrect, `block_size` needs to be an integer. We also set `int` defaults to type `int`. Unfortunately, `offset` is of type float and has `0` as the default, which is of type `int`. We therefore force `offset` to be `int`. Then, when we try to load `PYME.recipes.Recipes.STED_COP3.yaml`, we get an error at the line `offset: -0.5`. (P.S. we also get an error in `PYME.recipes.Recipes.STED_COP3.yaml` because `threshold_adaptive` no longer exists and has been replaced by `threshold_local`).

I don't see a great way around this. `0` is an ambiguous default. I do think it might be safer to default to `int` type instead of `float` for unknowns. `int` can almost always be cast up to a `float`, but a `float` will throw an error if passed to `np.ones((my_misplaced_float,))`, slicing, etc.

Any thoughts on how to fix this?

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15.6
 - Python version: 3.6
 - numpy version: 1.16
 - wxpython version [if a UI bug]:

zacsimile:
I thought of a very gross possible solution. We could set all unknown args to the following type.

```
class IntFloat(BaseFloat):
    default_value = 0.0
    info_text = 'Frankenstein to deal with automatic introspection'

    def get(self, obj, name):
        val = self.get_value(obj, name)
        if val is None:
            val = self.default_value
        if val == np.round(val):
            val = int(val)
        return val
```

zacsimile:
Another option would be to add something like https://github.com/rr-/docstring_parser after the initial inspection to try to fill in the blanks from the docstrings.

zacsimile:
[...]

PR #380, opened 2020-08-02
Get PYMEImage SPT workflow working again

zacsimile:
Everything I needed to do to get the steps outlined in PYME_tracking_workflow.pdf working again. `movieplot2` still isn't plotting contours correctly. 

Bugfix

**Proposed changes:**
- Fix `skfilter` imports by shimming py2 int/float behavior into py3 (see #376)
- Update wx calls in particeTracking.py
- Create a new version of `STED_COP3.yaml` that uses `skimage.filters.threshold_local` instead of the removed `threshold_adaptive`, and add a `FractionalThreshold` to convert its output to a mask 
- Adjust how the clump information is injected into the pipeline
- Fix `ListEvent` index grabbing
- Float/int fix for slicing in py3

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Analysis/Tracking/trackUtils.py
+++ b/PYME/Analysis/Tracking/trackUtils.py
@@ -41,10 +41,13 @@ class FeaturePlot(object):
             return ''
             
         data = self.clump[key]
+
+        if len(data.shape) > 2:
+            return ''
             
         plt.ioff()
         f = plt.figure(figsize=(6,2))
-        
+
         if 't' in self.clump.keys():        
             plt.plot(self.clump['t'], data)
         else:
--- a/PYME/Analysis/Tracking/trackUtils.py
+++ b/PYME/Analysis/Tracking/trackUtils.py
@@ -42,7 +42,9 @@ class FeaturePlot(object):
             
         data = self.clump[key]
 
-        if len(data.shape) > 2:
+        if len(data.shape) >= 2:
+            # FIXME - handle this better. Potentially a kymograph for ndim==2?
+            # FIXME - warn???
             return ''
             
         plt.ioff()
--- a/PYME/recipes/tracking.py
+++ b/PYME/recipes/tracking.py
@@ -135,6 +135,8 @@ class TrackFeatures(ModuleBase):
         #self.clumps = clumps        
         # pipeline.clumps = clumps
 
+        # FIXME - is this the right thing to be adding (and under the right name?) Does the pipeline (As used in dsviewer)
+        # actually support multiple data sources in a meaningful way? Consider fixing DSView/modules/particleTracking and removing this function completely.
         pipeline.addDataSource('clumps', clumpInfo)
         
         return clumps
--- a/PYME/recipes/tracking.py
+++ b/PYME/recipes/tracking.py
@@ -92,7 +92,9 @@ class TrackFeatures(ModuleBase):
         pipe = {}
         pipe.update(clumpInfo)
         pipe.update(objects)
-        pipe = pd.DataFrame(pipe)
+        # pipe = pd.DataFrame(pipe)
+        from PYME.IO.tabular import DictSource
+        pipe = DictSource(pipe)
         
         if 'mdh' in dir(objects):
             #propagate metadata
@@ -105,7 +107,8 @@ class TrackFeatures(ModuleBase):
         else:
             clumps = [c for c in clumps.all if (c.nEvents > self.minTrackLength) ]
 
-        clumpInfo = pd.DataFrame(clumpInfo)
+        # clumpInfo = pd.DataFrame(clumpInfo)
+        clumpInfo = DictSource(clumpInfo)
         
         if 'mdh' in dir(objects):
             #propagate metadata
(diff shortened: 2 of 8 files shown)


PYME/recipes/traits.py
```diff
@@ -30,4 +31,20 @@ def __init__(self, *args, **kwargs):
...
-        return value
\ No newline at end of file
+        return value
+
+class IntFloat(BaseFloat):
+    default_value = 0.0
+    info_text = 'Frankenstein to deal with automatic introspection'
+
+    def set(self, obj, name, value):
+        setattr(self, name, value)
```
David-Baddeley, line 44: is `setattr()` call  needed? it seems a little weird. Ditto for the whole `set()` method - surely we can just let the base class handle it?

zacsimile: Yeah, it doesn't work without the `set` method. I tried it. And `traits` docs actually say you need to explicitly define it.

PYME/recipes/tracking.py
```diff
@@ -125,12 +128,14 @@ def TrackWithPipeline(self, pipeline):
...
-        pipeline.addColumn('trackVelocity', clumpInfo['trackVelocity'])
+        # pipeline.addColumn('clumpIndex', clumpInfo['clumpIndex'])
+        # pipeline.addColumn('clumpSize', clumpInfo['clumpSize'])
+        # pipeline.addColumn('trackVelocity', clumpInfo['trackVelocity'])
         
         #self.clumps = clumps        
-        pipeline.clumps = clumps
+        # pipeline.clumps = clumps
+
+        pipeline.addDataSource('clumps', clumpInfo)
```
David-Baddeley, line 140: This won't work. The legacy mode should either remain as is, or (potentially): 

- create a new mapping filter from the `selectedDataSource`
- add the required columns
- select the new data mapping as the `selectedDataSource`

This might also not work, as this is not the full VisGUI pipeline. My preference would be to leave as is until we fix `DSView.modules.particleTracking`.

zacsimile: It does work. I can't leave it because a postdoc in the Toomre lab wants to use this on Wednesday.

zacsimile: I did the minimal changes needed to get the exact workflow you outlined in that PDF working again. I've run through the whole pipeline several times with the code in this PR, so I'm confident it does work.

PYME/Analysis/Tracking/trackUtils.py
```diff
@@ -41,10 +41,13 @@ def __getitem__(self, key):
             return ''
             
         data = self.clump[key]
+
+        if len(data.shape) > 2:
```
David-Baddeley: Is this the right behaviour / check to be performing? Should it be `>=2`? Should we warn? 
Should we (eventually) plot a heatmap/kymograph for dim=2?

PYME/recipes/tracking.py
zacsimile: Ah, sorry, I think I equated the pipeline working with this specific line working. It does add the dataSource, but we never actually use `pipeline.clumps` downstream in `particleTracking`, only the `clumps` `ClumpManager` that gets returned. So it's possible this specific line is not doing the right thing. It does not throw an error, though, whereas the `pipeline` injection did. Could just comment it out entirely.

PYME/Analysis/Tracking/trackUtils.py
zacsimile: I'm not 100% sure how this is supposed to work, but it does work, looking like the graphs in the PDF, with `>2`. However, no variables with `dim==2` were shown in the example images. `>=2` or `>1` would also be fine. kymograph for `dim==2` seems like the right move, based on what I've understood.

PYME/recipes/tracking.py
zacsimile: Confirmed that commenting it out entirely does still work.

David-Baddeley: ahh, I was forgetting that the tracking module handles it's own overlays. 

I think it still breaks the (undocumented) option of opening a visgui pane in dsviewer and viewing the tracks there.

zacsimile: I think this should be OK still, because I rewrote the tracks layer to automatically figure out if we're passing it a `ClumpManager` or a `tabular` tracks source. Switching up the data source type on PYMEVisualize shouldn't confuse it.

David-Baddeley: But that might be broken anyway with recent VisGUI tracking changes.

zacsimile: I could fix it if it is. I could be wrong about this, but I think the possibility that something else is already broken shouldn't preclude us from merging something that fixes a different part of the program that we know is broken. What needs to be changed here to get this merged so a postdoc here can use it?

David-Baddeley: The might already be broken bit was me saying that it probably didn't matter. 

When I first read the code I thought that the track display within dsviewer was going to be broken (it used to be part of a very bloated arrayviewpanel). I'd forgotten I'd already factored it out into something saner. I'll try to get this in asap.

zacsimile: Ah, sorry, I misunderstood.

Thanks, David!

zacsimile: Long-term, I think it would be nice to have `PYMEImage` use layers like `VisGUI`. We could update the annotation tools to live in an `Annotation` layer that sits on top of the image and tracks layers.

zacsimile: You could actually pass the layer between `PYMEImage` and `PYMEVisualize`.

David-Baddeley: Long term that is 100% the plan, and extends a bit further to making `visgui` the viewer for everything (using the ImageRenderLayer for image data). As it stands, `arrayviewpanel` does have lightweight layers of a sort which it loops through to draw some (but not all) of the overlays. The `particleTracking` module is using these at present, but these layers are completely separate from the VisGUI layers (use wx.DC drawing commands rather than OpenGL, and not as well organised as in VisGUI).

David-Baddeley: I think it's even in the roadmap :p

PR #383, opened 2020-08-03
slowly working on scipy deprecation warnings

barentine:
Addresses issue our tests failing on linux, maybe. There are some reports of pytest segfaulting if there's too much stderr to buffer. We currently have quite a lot of deprecation warnings coming up from every scipy.sin, etc. call we hit in the tests. We'll need to do this one day anyway if we ever make it to scipy 2.0

enhancement
**Proposed changes:**
- convert standard numpy array and math functions from scipy imports to numpy


David-Baddeley:
Segfaulting / test failure issue was due to a PYMECompress bug. I dont think we've had that issue for a couple of months. Current test failures relate to dataserver name resolution in the spooling tests, and may either be a result of a race condition with the server coming up, or the CI system being a bit cautious about letting us open sockets.

This is not to say that addressing some of the scipy deprecations is bad ... just that it won't fix the test failures.

barentine:
Cool, I'll keep at it as a low priority back-burner task.

PYME/Acquire/Hardware/Simulator/fluor.py
```diff
@@ -80,17 +79,17 @@ def __init__(self, x, y, z, transitionProbablilities, excitationCrossections, th
...
     def illuminate(self, laserPowers, expTime):
-        dose = concatenate(([1],laserPowers),0)*expTime
+        dose = np.concatenate(([1],laserPowers),0)*expTime
         #grab transition matrix
         transVec = (self.transitionProbabilities[self.state,:,:]*dose).sum(1)
         transVec[self.state]= 1 - transVec.sum()
         transCs = transVec.cumsum()
         
-        r = rand()
+        r = np.random.RandomState().rand()  # replace with np.random.default_rng() on later numpy versions
```
David-Baddeley, line 92: `np.random.rand()` should work fine (alternatively, `np.random.uniform()`)
```suggestion
        r = np.random.rand() 
```

PYME/Acquire/Hardware/Simulator/fluor.py
```diff
@@ -149,7 +148,7 @@ def illuminate(self, laserPowers, expTime, position=[0,0,0], illuminationFunctio
                 transVec[m, i]= 1 - tvs[m]
             transCs = transVec.cumsum(1)
             
-            r = rand(len(self.fl))
+            r = np.random.RandomState().rand(len(self.fl))  # replace with np.random.default_rng() on later numpy versions
```
David-Baddeley, line 151:
```suggestion
            r = np.random.rand()
```

PYME/Acquire/Hardware/Simulator/rend_im.py
```diff
@@ -303,9 +302,9 @@ def interp3(X, Y, Z):
 @fluor.registerIllumFcn
 def PSFIllumFunction(fluors, position):
     im = interpModel()
-    xi = maximum(minimum(round_((fluors['x'] - position[0])/dx + im.shape[0]/2).astype('i'), im.shape[0]-1), 0)
```
David-Baddeley, line 306: why `round_()` here (and is this the same as `np.round_()` and not some overridden round that I wrote somewhere)

PYME/Acquire/Hardware/Simulator/wormlike2.py
```diff
@@ -42,45 +41,45 @@ class wormlikeChain:
...
+        zs = steplength*zs*np.ones(theta.shape)
 
         for i in range(2,numsteps):
-            sh = cross(so, so + array([0,0,2]))
+            sh = np.cross(so, so + np.array([0,0,2]))
             #sh = sh./sqrt(dot(sh, sh));
             #sh = sh./sqrt(sh*sh.');
-            sh = sh/dot(sh, sh.T)
-            sk = cross(so, sh)
+            sh = sh/np.dot(sh, sh.T)
```
David-Baddeley, line 71: I recall there being a difference between the `scipy.dot`, `numpy.dot`, and `numpy.linalg.dot` (likewise with `cross`). Are these really a 1:1 replacement and have we checked that this still works (don't think it's covered by any of the unit tests).

PYME/Acquire/Hardware/Simulator/wormlike_curled.py
```diff
@@ -37,42 +37,42 @@ class wormlikeChain:
...
+        zs = steplength*zs*np.ones(theta.shape)
 
         for i in range(2,numsteps):
-            sh = cross(so, so + array([0,0,2]))
+            sh = np.cross(so, so + np.array([0,0,2]))
             #sh = sh./sqrt(dot(sh, sh));
             #sh = sh./sqrt(sh*sh.');
-            sh = sh/dot(sh, sh.T)
-            sk = cross(so, sh)
+            sh = sh/np.dot(sh, sh.T)
```
David-Baddeley, line 67: See above comments on `dot` and `cross`

PYME/Acquire/Hardware/Simulator/fluor.py
barentine: Correct, but in the numpy docs, there is a note that `random.rand` is a convenience wrapper for people coming from matlab, and `random_sample` takes a tuple size more consistent with the rest of numpy. On the docs for `random_sample` there's a see also for `Generator.random` which "should be used for new code", and would be `np.random.default_rng().random()`. However, that is not available until numpy 1.17, hence the `random.RandomState`. 

All that's to say - I'm fine using `random.rand`. More info [here](https://numpy.org/doc/stable/reference/random/new-or-different.html#new-or-different) when we get to it.

PYME/Acquire/Hardware/Simulator/rend_im.py
barentine: Not some overridden one you wrote unless you wrote - only potential would be in `PYME/Analysis/PSFGen/ps_app.c` and I checked there.

This falls under me trying not to change style things/whitespace - they're the same:
```
>>> np.round
<function round_ at 0x7fce619d92f0>
```

PYME/Acquire/Hardware/Simulator/fluor.py
barentine: note also, the [scipy deprecation notes](https://docs.scipy.org/doc/scipy/reference/release.1.4.0.html#scipy-deprecations) specifically say to use `default_rng`

PYME/Acquire/Hardware/Simulator/wormlike2.py
barentine: `scipy.dot` is pretty tough to find documentation for... But opening in ipython:
```
Help on function dot in module numpy:

dot(...)
    scipy.dot is deprecated and will be removed in SciPy 2.0.0, use numpy.dot instead
```

and cross
```
In [4]: help(scipy.cross)                                                                                                                    

Help on function cross in module numpy:

cross(a, b, axisa=-1, axisb=-1, axisc=-1, axis=None)
    scipy.cross is deprecated and will be removed in SciPy 2.0.0, use numpy.cross instead
```
which would at least has the same signature as the numpy version:
```
Help on function cross in module numpy:

cross(a, b, axisa=-1, axisb=-1, axisc=-1, axis=None)
    Return the cross product of two (arrays of) vectors.
    
    .....
```

Note also that this particular wormlike, `wormlike2`, gets used in `tests/PYME/localization/test_MultifitSR.py`, so we do actually have coverage for it, kind of.

PR #387, opened 2020-08-04
Better contour plotting

zacsimile:
Addresses issue all the contours plot in the upper-left plot only (see screenshot).

Before this fix:
before_spt

After this fix:
after_spt

For a lot of SPT plots, this works. For some, the plots are still wonky, and I'm not sure why. Way more work this fix then worked before though (zero).

Bugfix

**Proposed changes:**
- Center contours on their respective images with `x0`, `y0`

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Analysis/graphing_filters.py
+++ b/PYME/Analysis/graphing_filters.py
@@ -282,7 +282,7 @@ def movieplot2(clump, image):
             
             if not contours is None:
                 xc, yc = contours[i].T
-                plt.plot(xc - xp + 20, yc - yp + 20, c='b')#plt.cm.hsv(clump.clumpID/16.))
+                plt.plot(xc - xp + roiHalfSize + y_0, yc - yp + roiHalfSize + x_0, c='b')#plt.cm.hsv(clump.clumpID/16.))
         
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Looks reasonable, but I think you'll want to adjust the recipe quite a bit for that data (objects look too small/dim for threshold based detection and Measure2D to work well. Very quick flowchart which I should really format nicely for the docs below.

image

David-Baddeley:
**Note:** if you use the feature based tracking on the output of a Gaussian fit, you can use both amplitude and width as features to help resolve ambiguous linkages.

zacsimile:
Thanks for the flow chart! This is definitely not how I'd like to track data, but I just ran with some default settings for contour testing 'cause it takes less time. The contour plotting does work much better with this fix.

David-Baddeley:
Just as a note, axes also look flipped in y (the top row should be a complete row, the bottom row should be a partial row). I wonder if a bunch of defaults changed in a recent matplotlib update (would also explain how things ended up in the wrong place in the photophysics stuff)

Issue #367, opened 2020-07-28
[BUG] scalebar unit in dh5view looks like lm


barentine:
**Describe the bug**
![image](https://user-images.githubusercontent.com/31105780/88730183-871aa980-d103-11ea-9550-964f28ce0b83.png)
liter-meters

**To Reproduce**
open a series in dh5view on windows, py3

**Expected behavior**
um

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] Win10
 - Python version: 3.6
 - numpy version: 1.16
 - wxpython version [if a UI bug]: 4.1.0

David-Baddeley:
works fine on OSX, wx3, py2.7.

expected behaviour is to show a greek mu.

Likely a unicode literal problem (or windows font problem - do the std windows fonts support all unicode chars?).

See `PYME.DSViewer.LMAnalysis` line 354

David-Baddeley:
If a windows font problem, could try substituting `'\u00B5'`

see https://www.fileformat.info/info/unicode/char/00b5/index.htm

David-Baddeley:
`'\u00B5'` confirmed to work on OSX (so safe substitution if it works on windows)

barentine:
Yep, that works.
![image](https://user-images.githubusercontent.com/31105780/88739622-f3e47280-d108-11ea-8a82-0242c0093c59.png)

Pretty sure this has been around for ages and is windows rather than py3/wx4, but I hardly use dh5view on windows, and I think people've just dealt with it.

Thanks David.

barentine:
fixed in ce54ff2

PR #388, opened 2020-08-04
Display units in voxSizeDialog

zacsimile:
Addresses issue the [um] part of x, y, and z wasn't showing up (py3/wx4).

Bugfix 

**Proposed changes:**
- Convert space to space character (fixed display)
- Convert mu character to Windows-accepted mu character (re: #367)

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/DSView/voxSizeDialog.py
+++ b/PYME/DSView/voxSizeDialog.py
@@ -31,7 +31,7 @@ class VoxSizeDialog(wx.Dialog):
         #sizer2 = wx.BoxSizer(wx.HORIZONTAL)
 
         sizer2 = wx.BoxSizer(wx.HORIZONTAL)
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Just out of interest, why does a regular space not work? What are the symptoms?

zacsimile:
Not sure why it doesn't work. If I leave a space, everything after the space is cut off. I.e. only 'x' gets printed in the dialog box, instead of 'x [um]'. I think this is a problem with the size of the text. On other `wx.StaticText` entries, for example the PALM dialog box in the simulator (see attached), spaces are no problem, and the text isn't cut off until near the end of the string. However, swapping in `\u00A0` doesn't fix this problem. My best guess as of this moment is

* For a fixed-size window, such as a dialog, sizer cuts off text that extends into another element after the space separating the short enough portion of the text from the too long portion of the text
* Since the font is not monospace, `\u00A0` is a slightly shorter character than a regular space

Screen Shot 2020-08-05 at 8 25 46 AM

Issue #392, opened 2020-08-04
[BUG] failing to remove intermediate directories for clusterofone io


barentine:
**Describe the bug**
If you specify a file on the command line, e.g. `dh5view PYMEData/4DN/4DNFISDEV89F.h5`, then `image.filename` gets set to that, and your nodeworkers can fail on io:
```
Traceback (most recent call last):
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/taskWorkerHTTP.py", line 307, in computeLoop
    res = task()
  File "/home/ubuntu/code/python-microscopy/PYME/localization/remFitBuf.py", line 416, in __call__
    bufferManager.updateBuffers(md, self.dataSourceModule, self.bufferLen)
  File "/home/ubuntu/code/python-microscopy/PYME/localization/remFitBuf.py", line 94, in updateBuffers
    self.dBuffer = buffers.dataBuffer(DataSource(md.dataSourceID, md.taskQueue), bufferLen)
  File "/home/ubuntu/code/python-microscopy/PYME/IO/DataSources/HDFDataSource.py", line 38, in __init__
    self.h5File = tables.open_file(self.h5Filename)
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/tables/file.py", line 315, in open_file
    return File(filename, mode, title, root_uep, filters, **kwargs)
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/tables/file.py", line 778, in __init__
    self._g_new(filename, mode, **params)
  File "tables/hdf5extension.pyx", line 374, in tables.hdf5extension.File._g_new
  File "/home/ubuntu/miniconda3/lib/python3.7/site-packages/tables/utils.py", line 154, in check_file_access
    raise IOError("``%s`` does not exist" % (filename,))
OSError: ``/home/ubuntu/PYMEData/PYMEData/4DN/4DNFISDEV89F.h5`` does not exist
```

**Expected behavior**
don't just add the entire PYMEData path if we're actually below it

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10]
ubuntu 18.04
 - Python version:
3.7
 - numpy version:
1.19
 - wxpython version [if a UI bug]:
4.0.4

barentine:
issue is directly using `image.filename` in `LMAnalysis`

```
self.pusher = HTTPRulePusher.HTTPRulePusher(dataSourceID=image.filename,
[...]

PR #393, opened 2020-08-04
ensure we have an absolute path for non-cluster filenames in PYMEImage

barentine:
Addresses issue #392 .

bugfix
**Proposed changes:**
us os.path.abspath to be an absolute path for non-cluster filenames

**Checklist:**

- [ ] Tested with numpy=1.14

1.16, 1.19

3.6 and 3.7

4.1 and 4.0

### DIFF
--- a/PYME/IO/image.py
+++ b/PYME/IO/image.py
@@ -212,6 +212,8 @@ class ImageStack(object):
         self.events = events  #events
 
         self.queueURI = queueURI
+        if filename is not None and not filename.lower().startswith('pyme-cluster'):
+            filename = os.path.abspath(filename)
         self.filename = filename
 
         self.haveGUI = haveGUI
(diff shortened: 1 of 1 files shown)


David-Baddeley:
- I think there's already functionality for this in `nameUtils` which could be used (might need modification)
- this breaks compatibility with old style analysis, with super-tile data sources, and with any other non-cluster schemes we might want to support in the future e.g. http://, omero://, bioformats:// (to replace current .tiff extension special casing for forcing the use of bioformats rather than the internal TIFF loader).
- ultimately something like the following should be safe:

``` python

if os.path.exists(filename): # is a real file on disk (lets any scheme stuff through without clobbering)
    filename = nameUtils.getRelFilename(os.path.abspath(filename))

```

This should probably all go in nameUtils (but maybe we should look at that a bit more to make sure relative paths etc are still useful). A reasonable interim solution would be:

``` python

if os.path.exists(filename): # is a real file on disk (lets any scheme stuff through without clobbering and without explicitly testing each of a list of schemes)
    filename = os.path.abspath(filename)

```

barentine:
Good point re: supertile / potential future uses, David, and thanks for finding a reasonable interim.

PR #396, opened 2020-08-06
Installation tweaks

zacsimile:
The javabridge/bioformats install instructions don't work as-is.

Bugfix

**Proposed changes:**
- Fix the conda install calls for bioformats/javabridge
- Replace with pip as the recommended installation route since that's the only route I've ever gotten to work

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -69,7 +69,7 @@ environment and enter
     conda install javabridge
     conda install python-bioformats
 
-**Caveat:** This feature is currently in development. If ``conda`` packages for ``javabridge`` and ``bioformats`` don't work, try pip. 
+**Caveat:** This feature is currently in development. If ``conda`` packages for ``javabridge`` and ``bioformats`` don't work, try ``pip``. 
 
 Verify installation
 *******************
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -61,8 +61,18 @@ On OSX, use ``/path/to/conda/environment/python.app/Contents/MacOS/python setup.
 Enable bioformats data importers
 ================================
 
-Install a JAVA JDK or JRE. Open a command prompt in the installation ``conda`` 
-environment and enter
+Download and install `JAVA JDK 1.8 <https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html>`_. 
+Note that you may need to create an account to access the installer. Open a command prompt in the installation ``conda`` 
+environment.
+
+If you are on Windows, enter
+
+.. code-block:: bash
+
+    set MSSdk=1
+    set DISTUTILS_USE_SDK=1
+
+Otherwise (or subsequently, if on Windows) enter
 
 .. code-block:: bash
 
@@ -73,12 +83,11 @@ If you'd prefer to install via ``conda``, instead enter
 
 .. code-block:: bash
 
-    conda install -c conda-forge javabridge
+    conda install -c free javabridge
     conda install -c bioconda bioformats
 
-**Caveat:** pip is the only installation route confirmed to work. 
-
-
+**Caveat:** ``pip`` is the only installation route confirmed to work. Installation issues on ``pip``
+are likely caused by ``javabridge`` and addressed at `https://pythonhosted.org/javabridge/installation.html <https://pythonhosted.org/javabridge/installation.html>`_.
 
 Verify installation
 *******************
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -57,37 +57,19 @@ This assumes a basic familiarity with python and conda. We maintain a conda meta
 
 On OSX, use ``/path/to/conda/environment/python.app/Contents/MacOS/python setup.py develop`` instead  of ``python setup.py develop`` so that the PYME programs can access the screen. 
 
-
 Enable bioformats data importers
 ================================
 
 Download and install `JAVA JDK 1.8 <https://www.oracle.com/java/technologies/javase/javase-jdk8-downloads.html>`_. 
 Note that you may need to create an account to access the installer. Open a command prompt in the installation ``conda`` 
-environment.
-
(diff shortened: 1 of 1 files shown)


David-Baddeley:
We already have a note/hint to try pip here if conda doesn't work. Ultimately we should fix the conda packages (which could be just as simple as recompiling them), but in the meantime I'm happy to leave bioformats installation slightly broken / in an extras for experts category where people need to work things out for themselves. The need to install a JVM pretty much puts this in the extras for experts box anyway. 

The problem with having a mixed pip-conda enviroment is that you are pretty much guaranteed a broken environment when you go back to trying to install (or update) anything with conda, leaving the only update path being to completely reinstall from scratch. I have (at least when I was keeping up with packaging) successfully been getting people to update by simply doing a `conda update python-microscopy`, and I'd like to keep that avenue open.

zacsimile:
I don't quite follow the idea of purposefully putting out broken install instructions. What is the advantage to this? I think opening .lif, etc. files in PYMEImage is a useful feature for most users, if we want them to adopt PYME as a primary tool in their workflow. We could fix `javabridge` install instructions for users by including a link to the JDK 1.8 single-click installer. I think the main issue with the installation of Java for `javabridge`/`bioformats` is folks either 1) install the JRE, but not the JDK, or 2) install OpenJDK, which places `JAVA_HOME` at an irregular location. A link to the correct single-click installer JDK would fix both of these issues for users and developers alike.

I'm not sure what you mean by fix the conda packages. As it stands, neither `bioformats` nor `javabridge` exist on the `anaconda` channel, so we have to change the install instructions so they pull from `bioconda` and `conda-forge`, respectively. It's true that once these are both pulled from the correct location, they do not play nicely together. It's possible that recompiling the packages from these channels could fix things, is that what you meant? I'd rather create a wrapper for the `pip` packages, as we do with `pyfftw3`, since those work.

`pip` installation of `python-bioformats` and `javabridge`(not they are listed under different names in `conda` vs. `pip`, there is a `python-bioformats` in the `bioconda` channel, but it's not updated regularly like `bioformats`, and also does not work with `javabridge`), by contrast, does work, and this has been confirmed on multiple machines. Are you certain `conda update python-microscopy` won't work if these two packages are installed via `pip`? Generally, `pip` packages don't solve for dependencies, and `python-bioformats` and `javabridge` aren't dependencies for any other PYME packages, so I believe the worst that would happen is `bioformats` stops working after a `conda update` call (although I suspect it will continue to work).

barentine:
I loathe wrong install instructions when I'm trying out new software. 

@zacsimile conda packaging for javabridge in conda-forge is only linux, while free [only hits 3.6](https://anaconda.org/free/javabridge/files) on osx I think and is otherwise <=3.5. Hopefully a github issue on [py38](https://github.com/CellProfiler/python-javabridge/pull/1) means we could build for py36? Maybe we should reach out on image.sc or so about packaging.

That said I don't think it should be a problem for people to `conda update python-microscopy`, followed by `pip install javabridge python-bioformats --upgrade`, and I think they'd appreciate that a whole lot more than instructions which don't actually work.

David-Baddeley:
I also dislike broken install instructions, but changing to pip install for `javabridge` and `bioformats` just makes the instructions broken in a different way, and one which is more likely to cause pain in the long term. There is only a very narrow set of circumstances where pip-installing on top of a conda environment doesn't break the environment when it comes to future upgrades [specifically the pip package has **no** dependencies and is **guaranteed** not to gain dependencies in the future - in practice this means a package which you control]. The conda documentation is pretty clear that `conda`, `pip`, `conda` is generally a *very bad idea*™. 

We have recipes for both `javabridge` and `bioformats` in our `pyme_conda_recipes` repo, and have packages in the channel for OSX (py27). The right solution is to debug any issues with these and build across all platforms. Longer term we should also look at making a conda package for a JRE (or seeing if there is one we can borrow).

barentine:
I don't think we have javabridge in our `pyme-conda-recipes` repo. Would you mind adding it if you've got the files somewhere David? Then I'm sure Zach and I are happy to help build py3s so the conda-only docs aren't broken

zacsimile:
Anaconda channel does have `openjdk` (https://anaconda.org/anaconda/openjdk). It might work?

I still think that while `conda`, `pip`, `conda` is a generally bad idea, this particular proposal doesn't fall victim to any of the potential problems listed here: https://www.anaconda.com/blog/using-pip-in-a-conda-environment. These packages are so stand-alone I just think it won't be an issue. With these install instruction fixes, we could get more users using our bioformats features. `pip` is currently the only way I know we can get `javabridge` and `bioformats` to work. 

I'm certainly happy to try to fix the recipes for `conda`.

zacsimile:
After discussing with David, I think we can leave the instructions as-is with the recommendation that we fix the `javabridge` packaging in `conda` relatively soon.

To ease development installations, I suggest adding a link to the preferred JDK version in the install instructions and a slight modification of the caveat, which I have pushed here.

zacsimile:
Closing since these tweaks are cosmetic and don't fix anything, and I accidentally submitted them from my master branch, which I'd like back.

csoeller:
I appreciate that this issue is closed but I also wanted to express my negative experience with the java-bridge install. Had to resort to pip in the end, also had to hand tweak where javabridge looks for the java tools, this looked decidedly *not robust*. I foresee fragility resulting from javabridge approaches almost necessarily leaking into anything dependent on it. With the variability and changeability of the selected java install I appreciate this is difficult but nevertheless will put of anybody but hardcore aficionados.

My experience was on a Catalina Mac (latest os updates) and pulling down an oracle java (which is probably where the problems began).

Anyway, I think it would be nice to have some semi-automated install test environments that run scripts to attempt an install in a fresh conda environment and do a few sanity tests and then wipe that environment in a cleanup operation. These could be run on different platforms and generate some reports that can be auto-uploaded somewhere suitable. This would be something worth investing effort into if installability by non-experts remains a goal.

Best, CS

PR #400, opened 2020-08-07
start `SpoolController._update_series_counter` from zero series

barentine:
Addresses issue if you image a couple series, then change sample and change the spool directory correspondingly, you get left at the series counter / name from the old one, e.g. your first series in a new folder will be 8_7_series_ZB.

small enhancement
**Proposed changes:**
- when we update the series counter to make sure we don't override any files if we left off and restarted, start at zero so we also get the right number if we start a new folder.
  - Note we only call this function internally to spoolcontroller when we a) pull the first series name, b) set spool method, and c) set spool directory

**Question for future PRs/not**
Would zero padding the dates be OK? As a config option? Maybe another for just leaving the gen series name in numbers rather than letters?

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

### DIFF
--- a/PYME/Acquire/SpoolController.py
+++ b/PYME/Acquire/SpoolController.py
@@ -272,6 +272,7 @@ class SpoolController(object):
         
     def _update_series_counter(self):
         logger.debug('Updating series counter')
+        self.seriesCounter = 0
         while self._checkOutputExists(self.seriesName):
             self.seriesCounter +=1
             self.seriesName = self._GenSeriesName()
--- a/PYME/Acquire/SpoolController.py
+++ b/PYME/Acquire/SpoolController.py
@@ -273,6 +273,7 @@ class SpoolController(object):
     def _update_series_counter(self):
         logger.debug('Updating series counter')
         self.seriesCounter = 0
+        self.seriesName = self._GenSeriesName()
         while self._checkOutputExists(self.seriesName):
             self.seriesCounter +=1
             self.seriesName = self._GenSeriesName()


David-Baddeley:
Part of me feels that the numbering should not reset to 0 when changing directories / spool methods, as it reduces the chance of name collisions (i.e. the series numbering should be strictly monotonic regardless of where you save). There's a philosophical argument here as well in that the data should be identified by the metadata rather than the filename so as long as you have unique filenames you should be fine. One of the design principles with the metadata and file naming scheme was that you don't need to preserve the directory structure for everything to work (acknowledging that people will copy individual files onto USB drives, attach them to email etc ...).

The pragmatic bit of me is not sure how much I care, and am aware that we can't guarantee this across restarts in any case, and that for a lot of people the metadata is incomplete. I'm also aware that not restarting at 0 when you change directories might be considered odd behaviour. I'm leaning towards letting the behaviour be a config option.

As to zero-padding the dates - that doesn't seem unreasonable. I like the alphanumeric numbering, but would be open to a config option on this.

barentine:
Thanks for seeing both sides here, David. 

I'll edit this PR to config the alphanumerics. and the restarting at 0 on set-directory. Do you want the zero-padded dates as a config option too or just standard?

David-Baddeley:
zero padded dates can be standard ...

David-Baddeley:
this has been sitting for an embarrassing length of time ...

Before merging, probably worth noting that this has the potential to cause a small performance hit as it forces an `exists()` check on every file in the directory for every series. Might not be enough to be a problem, so tempted to merge and fix later if it is (the caching should catch things in the cluster case, and `os.exists()` is relatively cheap - although hard to be sure on windows for large directories). Workaround might be to only re-set if called when changing spool directory.

David-Baddeley:
Another potential optimisation would be to list the directory and check within our locally held list.

Issue #404, opened 2020-08-11
[BUG] another small traits dict -> str type error in cbeditors


barentine:
**Describe the bug**
```
Traceback (most recent call last):
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\matplotlib\cbook\__init__.py", line 196, in process
    func(*args, **kwargs)
  File "c:\users\bergamot\code\python-microscopy\PYME\recipes\recipeGui.py", line 533, in OnPick
    outp = self.recipes.activeRecipe.namespace[k]
KeyError: 'overlap_identified'
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\ui\custom_traits_editors.py", line 48, in update_editor
    n = choices.index(self.value)
ValueError: {'filtered': 'roi_locations', 'overlap_identified': 'og_detected'} is not in list

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\matplotlib\cbook\__init__.py", line 196, in process
    func(*args, **kwargs)
  File "c:\users\bergamot\code\python-microscopy\PYME\recipes\recipeGui.py", line 531, in OnPick
    self.configureModule(k)
  File "c:\users\bergamot\code\python-microscopy\PYME\recipes\recipeGui.py", line 553, in configureModule
    k.edit_no_invalidate(handler=MControl())
  File "c:\users\bergamot\code\python-microscopy\PYME\recipes\base.py", line 260, in edit_no_invalidate
    self.edit_traits(*args, kind='modal', **kwargs)
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traits\has_traits.py", line 1678, in edit_traits
    args,
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traitsui\view.py", line 462, in ui
    ui.ui(parent, kind)
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traitsui\ui.py", line 246, in ui
    self.rebuild(self, parent)
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traitsui\wx\toolkit.py", line 133, in ui_modal
    ui_modal.ui_modal(ui, parent)
[...]

Issue #278, opened 2020-06-17
editor improvement for recipe modules with dict inputs


barentine:
Something like a `DictInputModule(ModuleBase)` would be helpful for things like  which takes a dict input rather than the standard str, e.g. all modules with `self.inputVariables = DictStrStr()` like `output.HDFOutput`.

While one could think of potential choices to populate a drop down, e.g. [json.dumps({k:k}) for k in namespace.keys()], the current `CBEditor` is just a convenient way to throw an error from the gui:
![Screen Shot 2020-06-17 at 1 09 17 PM](https://user-images.githubusercontent.com/31105780/84933919-963e1e80-b0a4-11ea-98d2-5cac9c0d981a.png)
![Screen Shot 2020-06-17 at 1 09 24 PM](https://user-images.githubusercontent.com/31105780/84933922-99390f00-b0a4-11ea-976f-199bc513a14d.png)

David-Baddeley:
The dictionary input on HDFOutput is a bit of an aberration, and not something I'd want to encourage as a pattern. We only get away with it because it's an output module and handled rather differently than standard modules. Rather than subclassing ModuleBase, I think the two reasonable options are:

1) re-jig HDFOutput so that it doesn't need a dictionary input
2) write a custom editor and view specifically for HDFOutput

barentine:
On windows 10, wx 4.1.0, py3.6, traits 6.1.0, traitsui 7.0.0 we currently get the following when trying to add an HDFOutput module in bakeshop. I can't remember if I'd also hit it on my laptop above and just 'fixed' it by casting self.values to str first or not.
```
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\recipes\recipeGui.py", line 520, in OnAddModule
    if c.configure_traits(kind='modal'):
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traits\has_traits.py", line 2015, in configure_traits
    args,
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traitsui\wx\toolkit.py", line 219, in view_application
    context, view, kind, handler, id, scrollable, args
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traitsui\wx\view_application.py", line 101, in view_application
    args=args,
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\traitsui\view.py", line 462, in ui
[...]

PR #405, opened 2020-08-12
a dict editor / view we need for hdfoutput

barentine:
Addresses issue #404 #278 .

yes
**Proposed changes:**
- fix type error in #404 
- add DictChoiceStrEditor and use it for HDFOutput default view, as suggested in #278 

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

1. So you click add HDFOutput Screen Shot 2020-08-12 at 12 11 33 PM
2. Then you click the table Screen Shot 2020-08-12 at 12 11 42 PM
3. Then you choose what you want to save using a normal recipe input dropdown Screen Shot 2020-08-12 at 12 11 53 PM
4.  type in what you want for the table name Screen Shot 2020-08-12 at 12 12 01 PM
5. and boom, you're a hero Screen Shot 2020-08-12 at 12 12 08 PM
6. but you're a flawed hero, so you try to save another item to the same name COLLISION! Screen Shot 2020-08-12 at 12 12 21 PM
7. but we added a flag for the editor to put up unique value guard-rails Screen Shot 2020-08-12 at 12 12 40 PM
8. and check it out the thing is still functional, so if you decide to give up and just add what you had you still can Screen Shot 2020-08-12 at 12 12 52 PM

**note**
some of the code could be nicer: some camelCaseNamesAreRippedFromFilterPanel, and we might not want this all slapped into the custom traits editors file. _BUT_ this fixes multiple bugs which break the current gui, and the fact we've left it broken so long indicates a certain lack of excitement for sprucing it up further.

### DIFF
--- a/PYME/recipes/output.py
+++ b/PYME/recipes/output.py
@@ -353,9 +353,10 @@ class HDFOutput(OutputModule):
         from PYME.ui.custom_traits_editors import DictChoiceStrEditor
 
         inputs, outputs, params = self.get_params()
-
-        return View([Item(tn, editor=DictChoiceStrEditor(choices=self._namespace_keys)) for tn in inputs], 
-                    buttons=['OK', 'Cancel'])
+        
+        return View([Item(tn, editor=DictChoiceStrEditor(choices=self._namespace_keys)) for tn in inputs] +
+                    [Item('_'),] +
+                    self._view_items(params), buttons=['OK', 'Cancel'])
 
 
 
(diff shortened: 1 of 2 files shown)


barentine:
@David-Baddeley the scheme, etc. are now back. Thanks again for catching that.
Screen Shot 2020-08-12 at 6 23 13 PM

David-Baddeley:
Merging, but would like to keep #278 open for now as a) the right clicking is not super intuitive, and b) I'm greedy and want the editor to be as fast and easy as the original py2 one was.

PR #417, opened 2020-08-16
MarzhauserTango improvements

barentine:
Addresses issue my stage has a max velocity of 120 mm/s, and while that's pretty cool, and I think the focus lock is pretty good these days, I don't even know if the oil stays on the objective on big jumps at the default (50 mm/s) speed.

There is more I will do on this class in the coming months, and might even do this week (particularly some of the calibration stuff), but it would be OK to merge in it's current state, as it exposes the functionality I am immediately bumping up against and @zacsimile will be adding a tango stage to his system this week.

enhancement

**Proposed changes:**
- add velocity related tango dll calls
- reuse the same ctypes memory allocations on subsequent calls (still a couple of the originals calls which don't do this, but I'll keep smoothing things out).
- add properties/setters related to velocity 
- set stage axes units in um rather than leaving it in default mm and converting downstream. add some unit comments
- remove threaded class which was unimplemented and copied from one of the PI classes. Note that the api allows quite a bit of flexibility re: synchronous / asynchronous movement, and a threaded class may not be necessary here (as it hasn't been to this point)
- set the XY as a subclass of the full stage API. As we go it would make sense to only expose axes which are present, but we have to implement the full thing anyway for the API so we might as well subclass. Meanwhile it hasn't been an issue to only register the axes we have with PYMEAcquire, but it would still be nice to keep shell calls, etc. to what makes sense for the hardware.

**Checklist:**

- [ ] Tested with numpy=1.14

1.18

3.6

4.1

there's a bit of tidying, but we really flew through this in the first weeks of trying to put the htsms on PYME

### DIFF
--- a/PYME/Acquire/Hardware/Tango/marzhauser_tango.py
+++ b/PYME/Acquire/Hardware/Tango/marzhauser_tango.py
@@ -3,6 +3,7 @@
 from PYME.Acquire.Hardware.Piezos.base_piezo import PiezoBase
 import threading
 import ctypes
+import numpy as np
 import logging
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
The threaded classes were largely to deal with serial (or serial over usb) latency, so may well not be an issue as you have a PCI controller. There is a some potential that it could be an issue for anyone else wanting to use a serial connected Marzhauser stage, but as the code is untested, it's probably safe to remove.

PR #421, opened 2020-08-20
Instructions to get E816 and Tango running in PYME on a new machine

zacsimile:
I had to get a Marzhauser Tango stage and a PI E816 running on my microscope computer. I added a brief set of instructions on where to find the drivers and where to place the DLLs to each file, because it took @barentine and I a minute to figure it out.

Enhancement

**Proposed changes:**
- Add instructions for setting up Tango, E816 in comments at the top of their respective files

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
@@ -23,7 +23,7 @@
 
 # To install the E816 driver, download 
 # http://update.pi-portal.ws/down.php?link=http://syst13.synserver.de/PIUpdateFinder/PI_C-990_CD1_V_1_0_0_6.zip&name=PI_C-990_CD1_V_1_0_0_6.zip&id=59
-# and run PISoftwareSuite.exe. Then, navigate to C:\Users\scope_baby\Downloads\PI_C-990_CD1_V_1_0_0_6\Development\C++\API\noGUI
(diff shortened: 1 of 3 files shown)


PYME/Acquire/Hardware/Piezos/piezo_e816_dll.py
```diff
@@ -21,6 +21,11 @@
 #
 ##################
 
+# To install the E816 driver, download 
+# http://update.pi-portal.ws/down.php?link=http://syst13.synserver.de/PIUpdateFinder/PI_C-990_CD1_V_1_0_0_6.zip&name=PI_C-990_CD1_V_1_0_0_6.zip&id=59
+# and run PISoftwareSuite.exe. Then, navigate to PI_C-990_CD1_V_1_0_0_6\Development\C++\API\noGUI
+# and copy the correct version of E816_DLL(_x64).dll to C:\Windows\System32
```
David-Baddeley, line 27: Should also work if you put it in `<anaconda dir>\DLLs`

David-Baddeley:
Merging, but we should really have a section in the docs for where to put DLLs (`C:\Windows\System32` works, but is the lazy option and not really good practice).

David-Baddeley:
In principle, anywhere that can be found using these rules:

https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications

In practice, anywhere on `%PATH%` when in the conda environment, or `C:\Windows\System32`)

PR #427, opened 2020-08-21
Create CameraControlFrame as general GUI class for camera controls

zacsimile:
Currently we need to create custom camera controls per API. With this new class, you can throw the following into any init file

```
from PYME.Acquire.Hardware.CameraControlFrame import ModePropertyControl, CameraControlFrame
ctrls = [ModePropertyControl()]
scope.camControls['PcoEdge42LT'] = CameraControlFrame(MainFrame, scope.cam, scope, ctrls)
```

and get

camera_control_snip

The more controls there are in the `ctrls` list, the more appear in the GUI. We'll obviously need to add a few more base variable options to get all the camera controls working (e.g. because of how SDK3 works we need a specific `ATBoolPropertyControl` that inherits from `BoolPropertyControl`), but they should be 1-2 lines of code apiece and then work for cameras using that particular variable type generally. This should be good for quickly adding camera controls to new scopes.

Enhancement

**Proposed changes:**
- Refactor `PYME.Acquire.Hardware.AndorNeo.ZylaControlPanel` to base classes

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/CameraControlFrame.py
+++ b/PYME/Acquire/Hardware/CameraControlFrame.py
@@ -99,6 +99,9 @@ class EnumPropertyControl(wx.Panel, PropertyControl):
 ######## Property controls based on PYME.Acquire.Hardware.Camera class    
 class ModePropertyControl(EnumPropertyControl):
     def __init__(self, target='_mode'):
+        """
+        List of available camera modes, grabbed from the Camera base class.
+        """
         from PYME.Acquire.Hardware.Camera import Camera
         modes = [x for x in dir(Camera) if x.startswith('MODE_')]
         choices = sorted(modes, key=lambda x: getattr(Camera, x))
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I worry that this is kind of re-inventing a traits-like object (with a lot of the corresponding issues). A number of other worries: 
- The current camera base class has no concept of properties, using functional setters and getters - e.g. `SetIntegTime()` (so having something which calls `setattr` or `getattr` on a camera class doesn't make a lot of sense)
- the `_mode` attribute in particular is private, not consistently implemented across cameras, and is not writable / writing will give super undesirable behaviour.
- This doesn't translate well to different UI frontends
- medium/long term plan is for all GUI interaction to take place through `scope.state` (current camera UI controls are best viewed as a stopgap solution until we get there), so not particularly keen on adding a new common camera UI class which doesn't go through `scope.state`.
- Part of the design rationale in PYMEAcquire is to show the parameters a user needs, only those parameters, and to have them arranged sensibly, avoiding the automatically generated and randomly ordered property lists that are a feature of the umanager UI.  There is a limit to how much we can abstract this, hopefully we can do a bit better than requiring a custom UI class for each camera type, but it probably needs a bit of thought.

I suspect that a proper solution will involve a refactoring of the camera base class, along with some way of specifying what the UI should look like without actually involving any UI code. We should look a little at traits views (although I don't think they are necessarily the right solution here) or `PYME.localization.MetaDataEdit` for inspiration. `PYME.localization.MetaDataEdit`, whilst not quite a drop-in solution, has most of the mechanics we'd need to implement something which worked through `scope.state` - one could imagine refactoring both that and  `scope.state` hardware control to sit on a common set of base classes.

zacsimile:
That's fair. 

- We could modify this code allow the user to pass the get/set attributes, but it sounds like this isn't the approach we want to take anyway.
- I'm not sure what you mean by, "This doesn't translate well to different UI frontends." Can you expand?
- I think the camera controls should be fully-abstractable, and I think redoing the `Camera` class will make this possible. We should absolutely choose per-camera which controls appear, but I think the same control for `SetIntegTime()` should work regardless of camera (as it does now).

Can we add the refactor camera controls to `scope.state` to the roadmap? Along with other associated planned refactoring?

zacsimile:
Closing this as I no longer think it's relevant to anything.

Issue #425, opened 2020-08-21
[ENHANCEMENT] Zyla backend and ControlPanel tweaks to support both Zyla and Sona


csoeller:
**Is your feature request related to a problem? Please describe.**

This is a quick quick pre-discussion before submitting a PR for the small amount of backend tweaks that allow both Zyla and Sona heads to run with the Zyla backend.

**Describe the solution you'd like**

I have an implementation that required only a small set of tweaks to the original code. The main difference with the Sona from a user point of view are a number of selectable Tempearture Control modes, i.e. 0, -25 and -45 deg cooling. I have a screenshot below that this can be largely accomodated by having another enum control in the ZylaControlPanel.

Note the cyan box which allows access to selecting the cooling mode. With the Zyla, this just contains 0 deg as only choice.

I found that I also wanted to see the TemperatureStatus which I have implemented as a cludge to enum controls, see red box in screenshot. This is a readonly property and as such it is not the best way to show this as an enum control. It would be really good to infrequently update the status somewhere in the GUI as I found that the -45 setting never stablised and stayed at -38 or so. So this is most likely not the right solution but it would be good to briefly discuss how to tackle this one.

The remainder of the smallish tweaks will be more easily discussed once I prepared the PR. If I get some guidance how we can best tackle the TemperatureStatus I will take that into account when preparing the PR.

![sona-view-screen](https://user-images.githubusercontent.com/3750860/90905937-cd90ba80-e3c8-11ea-9ec1-d994b6bed780.jpg)

This is what the interface looks like with the Zyla. Just 0.0 by default re temp setting.

![zyla-view-screen-3](https://user-images.githubusercontent.com/3750860/90906622-da61de00-e3c9-11ea-83eb-ef1784693d46.jpg)

**Describe alternatives you've considered**
Nothing to say really.

**Additional context**
Screenshots are applied above.

**Are you prepared to work on it?**
Yep, already did.

zacsimile:
@csoeller I'd love to see what you did. I ran into a lack of camera controls for the pco/orca and I'm currently writing (started this morning, nearly finished) a complete refactor of the camera controls that lets you do something like

```
ctrls = [BoolControl('SpuriousNoiseFilter'),
                etc.]
MyCamControl = CamControl(parent, cam, scope, ctrls)
```

for every camera. It seems like I may be duplicating efforts. Perhaps we should coordinate on this?

csoeller:
[...]

PR #429, opened 2020-08-22
tweaks to enable Sona with Zyla backend support

csoeller:
Addresses issue #425.

Enhancement of the Zyla backend to allow Sona heads to be connected.

**Proposed changes:**

- Changes to the AndorZyla code to self-detect some capabilities.

- Additional interface changes in ZylaControlPanel so that the cooling mode options of the Sona are autodetected and offered in the control panel as options for selection.

**Checklist:**

- [x] Tested with numpy=1.16
- [x] Tested on python 2.7 only
- [NA ] Tested with wx=3.x and wx=4.x [if UI code]

- [x] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)? [has been raised]
- [-] Does this change how users interact with the software? How will these changes be communicated?
- [-] Does this change the required dependencies?
- [-] Are there any other side effects of the change? [don't think so but will discuss]

### DIFF
--- a/PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
+++ b/PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
@@ -131,7 +131,7 @@ class AndorBase(SDK3Camera, CameraMapMixin):
     class SimpleGainEnum(object):
         def __init__(self, cam):
             self.cam = cam
-            self.gainmodes = cam.PixelEncodingForGain.keys()
+            self.gainmodes = cam.SimplePreAmpGainControl.getAvailableValues()
             self.propertyName = 'SimpleGainMode'
             
         def getAvailableValues(self):
@@ -256,6 +256,11 @@ class AndorBase(SDK3Camera, CameraMapMixin):
                     raise RuntimeError('PixelEncodingForGain mode "%s" unknown bit depth (neither 12 nor 16 bit)' % (mode))
         else:
             self.PixelEncodingForGain = self.DefaultPixelEncodingForGain
+
+        # this instance is compatible with use in Zylacontrolpanel
+        # note we make this only once the camera has been initialised and PixelEncodingForGain been made
+        self.SimpleGainEnumInstance = self.SimpleGainEnum(self)
+
 
         self.FrameCount.setValue(1)
         self.CycleMode.setString(u'Continuous')
@@ -790,8 +795,7 @@ class AndorZyla(AndorBase):
         self.TemperatureControl = ATEnum()
         self.TemperatureStatus = ATEnum()
         self.SimplePreAmpGainControl = ATEnum()
-        # this instance is compatible with use in Zylacontrolpanel
-        self.SimpleGainEnumInstance = self.SimpleGainEnum(self)
+
         self.BitDepth = ATEnum()
         
         self.ActualExposureTime = ATFloat()
--- a/PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
+++ b/PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
@@ -49,10 +49,10 @@ class AndorBase(SDK3Camera, CameraMapMixin):
     #MODE_CONTINUOUS = 1
     #MODE_SINGLE_SHOT = 0
 
-    PixelEncodingForGain = {'12-bit (low noise)': 'Mono12',
-                            '12-bit (high well capacity)': 'Mono12',
-                            '16-bit (low noise & high well capacity)' : 'Mono16'
-                            }
+    DefaultPixelEncodingForGain = {'12-bit (low noise)': 'Mono12',
+                                   '12-bit (high well capacity)': 'Mono12',
+                                   '16-bit (low noise & high well capacity)' : 'Mono16'
+    }
 
     _noise_properties = {
         'VSC-00954': {
@@ -73,6 +73,19 @@ class AndorBase(SDK3Camera, CameraMapMixin):
                 'ElectronsPerCount' : 0.5,
                 'ADOffset' : 100,
                 'SaturationThreshold' : (2**16 -1)
+            }},
+        'CSC-00425': { # this is info for a Sona
+            u'12-bit (low noise)': {
+                'ReadNoise' : 1.21,
+                'ElectronsPerCount' : 0.45,
+                'ADOffset' : 100, # check mean (or median) offset
+                'SaturationThreshold' : 1776  #(2**16 -1) # check this is really 11 bit
+            },
+            u'16-bit (high dynamic range)': {
+                'ReadNoise' : 1.84,
+                'ElectronsPerCount' : 1.08,
+                'ADOffset' : 100,
+                'SaturationThreshold' : 44185
             }},
         'VSC-02858': {
              '12-bit (low noise)': {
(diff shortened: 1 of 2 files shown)


PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
```diff
@@ -130,26 +143,9 @@ def noise_properties(self):
...
-        def getAvailableValues(self):
-            return self.gainmodes
-
-        def setString(self,str):
-            self.cam.SetSimpleGainMode(str)
-
-        def getString(self):
-            return self.cam.GetSimpleGainMode()
-
-
```
csoeller, line 150: The SimpleGainEnumClass seemed to be essentially overhead only once the code was investigated. I got rid of it in favour of a simple keyword for EnumControl objects in the Zyla Control Panel.

PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
```diff
@@ -653,8 +677,13 @@ def GenStartMetadata(self, mdh):
...
             mdh.setEntry('Camera.ROIWidth', x2 - x1)
             mdh.setEntry('Camera.ROIHeight', y2 - y1)
-            #mdh.setEntry('Camera.StartCCDTemp',  self.GetCCDTemp())
 
+            #mdh.setEntry('Camera.StartCCDTemp',  self.GetCCDTemp())
+            # these make sense with the Sona choices but should also be ok for the Zyla
+            mdh.setEntry('Camera.TemperatureControl', self.TemperatureControl.getString())
+            mdh.setEntry('Camera.TemperatureStatus', self.TemperatureStatus.getString())
+            mdh.setEntry('Camera.SensorTemperature', self.SensorTemperature.getValue())
+
```
csoeller, line 712: With the Sona temperature control options it becomes important to know in which cooling mode the camera was operated as I expect this to affect camera maps. It cannot hurt for the Zyla either.

PYME/Acquire/Hardware/AndorNeo/ZylaControlPanel.py
```diff
@@ -8,14 +8,24 @@
...
+        # this keyword allows us to override unwieldy property Names
+        if targetPropertyName is None:
+            self.targetPropertyName = target.propertyName
+        else:
+            self.targetPropertyName = targetPropertyName
+            
+        try:
+            self.targetReadOnly = target.isReadOnly()
+        except AttributeError:
+            self.targetReadOnly = False
```
csoeller: The ReadOnly check is a kludge to support the read only temperature status property in the ControlPanel. It should really be a status display of some form. This probably needs discussion.

PYME/Acquire/Hardware/AndorNeo/ZylaControlPanel.py
```diff
@@ -73,16 +84,20 @@ def __init__(self, parent, cam, scope):
         self.cam = cam
         self.scope = scope
         
-        self.ctrls = [EnumControl(self, cam.SimpleGainEnumInstance),
+        self.ctrls = [EnumControl(self, cam.SimplePreAmpGainControl, targetPropertyName='SimpleGainMode'),
```
csoeller: here we got rid of the SimpleGainEnumInstance which really only provided a simpler name for the property. This has insteads been done with a simpler keyword to the EnumControl constructor.

PYME/Acquire/Hardware/AndorNeo/ZylaControlPanel.py
```diff
@@ -73,16 +84,20 @@ def __init__(self, parent, cam, scope):
...
                       BoolControl(self, cam.SpuriousNoiseFilter),
                       BoolControl(self, cam.StaticBlemishCorrection),
-                      EnumControl(self, cam.CycleMode),]
-        
+                      EnumControl(self, cam.CycleMode),
+                      EnumControl(self, cam.TemperatureStatus),]
+
+        if len(cam.TemperatureControl.getAvailableValues()) > 0: # need to check that Zyla has at least one mode
+            self.ctrls.append(EnumControl(self, cam.TemperatureControl))
+
```
csoeller, line 98: Here we could check for >1 instead which would avoid the Zyla case where only one mode is available (0.0 deg). So far I opted for the verbose one that just shows the single selectable option.

PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
```diff
@@ -74,6 +74,19 @@ class AndorBase(SDK3Camera, CameraMapMixin):
...
+                'ElectronsPerCount' : 0.45,
+                'ADOffset' : 100, # check mean (or median) offset
+                'SaturationThreshold' : 1776  #(2**16 -1) # check this is really 11 bit
+            },
+            u'16-bit (high dynamic range)': {
+                'ReadNoise' : 1.84,
+                'ElectronsPerCount' : 1.08,
+                'ADOffset' : 100,
+                'SaturationThreshold' : 44185
+            }},
```
csoeller, line 89: This showcases the difference for the Sona which has a different set of gain modes, only two are supported and the 16bit version has a slightly different name to the Zyla.

As we discussed, eventually this info should go in a config file but this should be the subject of a separate PR/issue as it is essentially orthogonal to the Sona functionality tweaks and also needs a bit more discussion.

PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
David-Baddeley: I recall there being a good reason for this (which might have evaporated with a more recent SDK3 version), but can't put my finger on it. It could be related to Andor Neo support.

David-Baddeley: This is needed - and is not just an alias for `SimplePreAmpGainControl`, as `SetSimpleGainMode()` actually sets two properties (the gain mode **and** pixel encoding which is not set automatically by setting the `SimplePreAmpGainControl` property).

PYME/Acquire/Hardware/AndorNeo/ZylaControlPanel.py
```diff
@@ -73,16 +84,20 @@ def __init__(self, parent, cam, scope):
...
         
-        self.ctrls = [EnumControl(self, cam.SimpleGainEnumInstance),
+        self.ctrls = [EnumControl(self, cam.SimplePreAmpGainControl, targetPropertyName='SimpleGainMode'),
                       EnumControl(self, cam.PixelReadoutRate),
                       BoolControl(self, cam.SpuriousNoiseFilter),
                       BoolControl(self, cam.StaticBlemishCorrection),
-                      EnumControl(self, cam.CycleMode),]
-        
+                      EnumControl(self, cam.CycleMode),
+                      EnumControl(self, cam.TemperatureStatus),]
```
David-Baddeley: Temperature status would probably be better reported through the status bar - e.g. either by adding a callback to `microscope.StatusCallbacks`, or by re-working the exisiting camera temperature stuff in `microscope.genStatus()` to defer the status stuff a bit more to the camera / let it decide what to report for temperature etc ...

There might also be either performance or synchronisation issues in polling the temperature and/or temperature status through an Enum control. As it stands, the Zyla code has the temperature being queried only when`StartExposure()` is called, with the result cached and the cached value used for subsequent calls to `GetCCDTemp()`. It's been a while, but I'm pretty sure this was done this way on purpose to address a performance problem.

csoeller: Now addressed as a new method that can be added to `microscope.StatusCallbacks` in init file. The method is called `TemperatureStatusText` and I add below a screenshot that shows it in action. With the current SDK no performance issues apparently.

pyme-zyla-sona-tweaks

PYME/Acquire/Hardware/AndorNeo/AndorZyla.py
csoeller: I reinstated SimplePreampGainControl and see the reason for the pixel encoding. Interestingly, with the latest SDK there were no issues not doing it although I can't be sure it was stuck on 16bit. Anyway, seems to work fine now with latest commits.

PYME/Acquire/Hardware/AndorNeo/ZylaControlPanel.py
csoeller: Changed to >1 so only present if there is a real choice.

PR #431, opened 2020-08-23
Fix pco crashes

zacsimile:
I think was asking the `pco` camera for the same information too quickly in rapid succession. There must be some sort of internal timeout between calls. 

Bugfix

**Proposed changes:**
* Store the ROI in between setting a new ROI so we don't keep asking the camera for its ROI.
* Get rid of extra `SetDescription` calls.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/pco/pco_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_cam.py
@@ -48,6 +48,7 @@ class PcoCam(Camera):
         self.buffer_size = 0
         self.n_read = 0
         self.initalized = True
+        self.__roi = None
 
     @property
     def noise_properties(self):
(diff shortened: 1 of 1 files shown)


PYME/Acquire/Hardware/pco/pco_cam.py
```diff
@@ -242,12 +237,13 @@ def SetROI(self, x0, y0, x1, y1):
...
         # Recording state is reset, so set to 0
         self.n_read = 0
-
-        self.SetDescription() # Update the description, for safety. TODO: Is this necessary??
+        self.GetROI()
 
     def GetROI(self):
-        roi = self.cam.sdk.get_roi()
-        return roi['x0'], roi['y0'], roi['x1'], roi['y1']
+        if self.n_read == 0:
```
David-Baddeley, line 243: I worry a little about the `n_read` stuff - will merge for now as camera driver still very 'alpha', but feels like it might be fragile. On other cameras (uc480 in particular) we have just cached the ROI values that we set in `SetROI()`.

zacsimile: Fair enough. I don't have a particularly good reason for the `self.n_read==0` call. My original goal was to just confirm that the ROI we set is exactly the ROI we get back. I think I've confirmed this and I can probably just store the set ROI in `__roi`.

PR #432, opened 2020-08-25
Add label to point size using character for space

zacsimile:
"Point size" in points layers is currently displaying as "Points " on py3. Same issue as in https://github.com/python-microscopy/python-microscopy/pull/388. Feels like there should be a more general fix for this, but who knows? Might just fix itself in wx 4.1.

Bugfix

**Proposed changes:**
-Force character for space in "Point size"

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
I wish there was an alternative solution - I really don't want to have to litter the code with unicode non-breaking spaces if we can avoid it ...

zacsimile:
I agree. I'm not sure what to do, though. This started with the latest version of traits (7.0.0). As far as traits bugs go, this is very minor compared to what we've experienced in the past. I suspect newer versions of traits/wx will just fix it.

PR #433, opened 2020-08-26
int cast for shape in RGB image array allocation

barentine:
Addresses issue running on py3.

bugfix
**Proposed changes:**
cast to int, avoid
```
    def image_to_rgb(image, zoom=1.0, scaling='min-max', scaling_factor=0.99):
>       data = np.zeros([image.data.shape[0] / zoom, image.data.shape[1] / zoom, 3], dtype='uint8')
E       TypeError: 'float' object cannot be interpreted as an integer
../python-microscopy/PYME/IO/rgb_image.py:5: TypeError
```

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

### DIFF
--- a/PYME/IO/rgb_image.py
+++ b/PYME/IO/rgb_image.py
@@ -2,7 +2,7 @@ import numpy as np
 from scipy import ndimage
 
 def image_to_rgb(image, zoom=1.0, scaling='min-max', scaling_factor=0.99):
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Note: this is a numpy version issue, rather than anything else (py3 etc ...).

Issue #419, opened 2020-08-18
[BUG] ClusterOfOne doesn't pick up data on Windows outside of PYMEDATADIR


khu6:
```
ERROR:root:127.0.0.1 - - [17/Aug/2020 15:46:16] code 404, message File not found - /C:Userskh725Downloads/analysis/, [C:\Users\kh725\PYMEData\analysis/]

ERROR:root:127.0.0.1 - - [17/Aug/2020 15:46:16] code 404, message File not found - /C:Userskh725Downloads/analysis/, [C:\Users\kh725\PYMEData\analysis/]

ERROR:root:127.0.0.1 - - [17/Aug/2020 15:46:16] code 404, message File not found - /C:Userskh725Downloads/analysis/, [C:\Users\kh725\PYMEData\analysis/]

ERROR:root:127.0.0.1 - - [17/Aug/2020 15:46:16] code 404, message File not found - /C:Userskh725Downloads/analysis/, [C:\Users\kh725\PYMEData\analysis/]

ERROR:root:127.0.0.1 - - [17/Aug/2020 15:46:16] code 404, message File not found - /C:Userskh725Downloads/analysis/, [C:\Users\kh725\PYMEData\analysis/]

ERROR:root:127.0.0.1 - - [17/Aug/2020 15:46:16] code 404, message File not found - /C:Userskh725Downloads/analysis/, [C:\Users\kh725\PYMEData\analysis/]

DEBUG:root:pytables open call: C:\Users\kh725\PYMEData\analysis\0_000_000.h5r
DEBUG:root:pytables file open: C:\Users\kh725\PYMEData\analysis\0_000_000.h5r
DEBUG:root:H5RFile - closing: C:\Users\kh725\PYMEData\analysis\0_000_000.h5r
DEBUG:root:H5RFile - closed: C:\Users\kh725\PYMEData\analysis\0_000_000.h5r


ERROR:PYME.IO.unifiedIO:Name "C:\Users\kh725\Downloads/analysis/0_000_000.h5r" is invalid. Names must only include alphanumeric characters and underscore
DEBUG:PYME.IO.clusterIO:Request for b'http://127.0.0.1:49831/C:Userskh725Downloads/analysis/' failed with error: 404
DEBUG:PYME.IO.clusterIO:Request for b'http://127.0.0.1:49859/C:Userskh725Downloads/analysis/' failed with error: 404
[...]

PR #435, opened 2020-08-28
fix to allow cluster localization of paths outside PYMEData on windows

barentine:
Addresses issue #419. Issue was we replaced all separators with os separators, then joined dirname to analysis/basename with posix separator.

bugfix
**Proposed changes:**
escape drive letter and substitute posix separators rather than os separators

Tested this on windows localizing a file on another drive than PYMEData dir, worked fine. Should work for files inside and outside PYMEData, and on python 2/3:
```
In [52]: p
Out[52]: 'F:\\2020_7_23\\testing\\okwegetit'
In [53]: nameUtils.genClusterResultFileName(p)
Out[53]: '/2020_7_23/testing/analysis/okwegetit.h5r'
In [54]: pd
Out[54]: 'C:\\Users\\Bergamot\\PYMEData\\Bergamot\\2020_7_14'
In [55]: nameUtils.genClusterResultFileName(pd)
Out[55]: 'Bergamot/analysis/2020_7_14.h5r'
```

I opted not to touch `getRelFileName` since that it used elsewhere e.g. in `image.py` and I don't know if we can e.g. safely drop the drive letter there or not.

**Checklist:**

- [ ] Tested with numpy=1.14

1.19

3.6

4.1

### DIFF
--- a/PYME/IO/FileUtils/nameUtils.py
+++ b/PYME/IO/FileUtils/nameUtils.py
@@ -123,6 +123,7 @@ def genClusterResultFileName(dataFileName, create=True):
     """Generates a filename for saving fit results based on the original image
     filename"""
     from PYME import config
+    import posixpath
     fn, ext = os.path.splitext(dataFileName) #remove extension
     
     if fn.upper().startswith('PYME-CLUSTER://'):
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Merging as fixes a pretty important bug and better than status quo, but @barentine - if you could look into the `startswith('/')` case  and the possibility of this generating very odd filenames (see added comments) that would be great - I worry that this makes it a fragile fix

PR #440, opened 2020-09-04
add save recipe outputs to PYMEVis gui + backend for dict editing

barentine:
Addresses issue currently PYMEVis only uses output modules through menu items, e.g. one at a time, but does not support using them in a recipe. This is not necessarily an issue for most people - the solution could be to add a menu item for each output module, only things are we would have to actually do that, PYMEVis would not serve as a 100% complete recipe tester, and any output-ing people do through PYMEVis would require manual setting entry each time. @David-Baddeley what's the move here? dh5view has `run recipe` and `rune recipe and save outputs` menu items, which makes sense to me, so I added button-versions to PYMEVis (so a collaborator could run an output plugin module (pyme-omero).

I'm not 100% sure I understood what you would like to see differently long term for the choice-box dict editor we added for the `HDFOutput` view, but I wanted a dict editor for setting the output context (since the only thing I felt like we could reliably auto-fill was `file_stub`), so I used the `EditListCtrl` that you mentioned last time. If this is a bit closer to what you have in mind for `HDFOutput` module it would be easy use `DictCtrlDialog` (or panel version) for both, and add/edits would either open `DictItemEditDialog` or the choice box version. I didn't worry super much about refactoring the other into this, where it lives, etc. until I know whether it's remotely something we're interested in

Enhancement
**Proposed changes:**
- set a default style parameter in `EditListCtrl` init since `style=0` throws an error and is therefore not super helpful
- add a checkbox to toggle `pipeline.execute_on_invalidation`, a button to call `pipeline.recipe.execute()`, and a button to make sure the recipe has already run, gui-configure the output module context, and then call `pipeline.recipe.save()`.
Screen Shot 2020-09-04 at 2 18 10 PM
(maybe wasn't the move to put them on a different horizontal boxsizer, don't know)
- add a dict editor using the `EditListCtrl` to set the `context` dict for `ModuleCollection.save`. `file_stub` key auto-populates
Screen Shot 2020-09-04 at 2 18 23 PM
The rest of the screenshots are pretty superfluous and could have easily been left out, but I already took them so I'm including them
  - left or right clicking inside the listctrl gives you the menu
Screen Shot 2020-09-04 at 2 18 32 PM
  - add / edit dialog is str / str, rather than choice-box / str (for HDFOutput view) or str/float (for the other custom dict editor factory) Screen Shot 2020-09-04 at 2 18 39 PM
  - Screen Shot 2020-09-04 at 2 18 51 PM

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/recipes/recipeGui.py
+++ b/PYME/recipes/recipeGui.py
@@ -327,7 +327,24 @@ class RecipeView(wx.Panel):
         self.bSaveRecipe.Bind(wx.EVT_BUTTON, self.recipes.OnSaveRecipe)
         
         vsizer.Add(hsizer, 0, wx.EXPAND, 0)
-        
+
+        if hasattr(self.recipes, 'OnSaveOutputs'):
+            hsizer = wx.BoxSizer(wx.HORIZONTAL)
+
+            self.b_toggle_exec_live = wx.CheckBox(self, -1, 'Live-Update')
+            self.b_toggle_exec_live.SetValue(self.recipes.pipeline.recipe.execute_on_invalidation)
+            self.b_toggle_exec_live.Bind(wx.EVT_CHECKBOX, self.recipes.OnToggleLiveRun)
+            hsizer.Add(self.b_toggle_exec_live)
+
+            self.b_run = wx.Button(self, -1, 'Run Recipe')
+            self.b_run.Bind(wx.EVT_BUTTON, self.recipes.OnRun)
+            hsizer.Add(self.b_run)
+
+            self.b_save_outputs = wx.Button(self, -1, 'Save Outputs')
+            hsizer.Add(self.b_save_outputs)
+            self.b_save_outputs.Bind(wx.EVT_BUTTON, self.recipes.OnSaveOutputs)
+
+            vsizer.Add(hsizer, 0, wx.EXPAND, 0)
         
         hsizer1.Add(vsizer, 1, wx.EXPAND|wx.ALL, 2)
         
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Doing something to aid recipe output debugging from within PYMEVis is definitely desirable, but not super keen on having recipes save outputs within PYMEVis. Reasoning as follows:
- PYMEVis is philosophically a viewer, and should have minimal side effects (we currently violate this a little with the saving support that we do offer with generated images etc ...  although I think that is all well justified, and largely mitigated by the fact that each output is a manual choice with an interactively selected filename. Automagically generated filenames push a little too far out of my minimal side effects comfort zone.) 
- Setting the context in a useful way is hard (and super non obvious from an end user perspective)
- Recipes normally run repeatedly in PYMEVis, raising over-writing issues with automagic outputs (particularly for cluster endpoints which are strictly write once).
- I'm not sure saving outputs is actually super useful, as you'd need to re-load them anyway to see if it's worked

My preferred solution would be to pop up viewers for the outputs (or better yet, just have buttons/links for them in an "outputs" tab/panel), allowing the images / datasets to be inspected and manually saved if needed. This largely avoids any context issues, and is probably going to give a much better user experience when iterating on a recipe to debug it. Unfortunately a functional viewer for generic tabular data is a pre-requisite here.

I might possibly be persuaded to allow this in as a stopgap if the save outputs button disappeared from the UI and was replaced by a menu item - e.g. `Recipe-->Debugging-->Save outputs`.

I feel the toggle button for `_execute_on_invalidate` is dangerous, however, as it's a hard to appreciate the consequences and it could dramatically break the user experience. The option should always be on in PYMEVis - for the very few debugging situations where it makes sense to turn it off I think doing so in the shell has roughly the right level of difficulty.

I'm slightly concerned that there could be a lot of duplication between your new dict editor and the existing metadata editor. We should at least have a look to see if it wouldn't be feasible to use a common base.

PR #447, opened 2020-09-15
allow pickled shiftmaps to load on numpy > 1.14

barentine:
Addresses issue #numpy load was made more secure after 1.14.6 (I think in 1.15, but definitely in 1.16) in response to [this](https://nvd.nist.gov/vuln/detail/CVE-2019-6446), which is cool, but means that we hit

```
  File "c:\users\bergamot\code\python-microscopy\PYME\localization\MetaDataEdit.py", line 316, in <lambda>
    bSetFile.Bind(wx.EVT_BUTTON, lambda e : self._setFile(mdh))
  File "c:\users\bergamot\code\python-microscopy\PYME\localization\MetaDataEdit.py", line 342, in _setFile
    self.retrieveValue(mdh, False)
  File "c:\users\bergamot\code\python-microscopy\PYME\localization\MetaDataEdit.py", line 399, in retrieveValue
    dx, dy = np.load(self.filename)
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\numpy\lib\npyio.py", line 451, in load
    raise ValueError("Cannot load file containing pickled data "
ValueError: Cannot load file containing pickled data when allow_pickle=False
```
when loading shiftmaps on later numpy versions

i guess bugfix
**Proposed changes:**

- allow `numpy.load` to unpickle shiftmaps

**Checklist:**

- [ ] Tested with numpy=1.14
did not test, but the argument has been there forever it just defaulted to `True` on numpy 1.14

3.6

4.1

### DIFF
--- a/PYME/localization/MetaDataEdit.py
+++ b/PYME/localization/MetaDataEdit.py
@@ -396,7 +396,7 @@ class ShiftFieldParam(FilenameParam):
         FilenameParam.retrieveValue(self, mdh, *args, **kwargs)
         
         if not self.filename == oldfn and not self.filename in ['<none>', '']:
-            dx, dy = np.load(self.filename)
+            dx, dy = np.load(self.filename, allow_pickle=True)
             mdh.setEntry('chroma.dx', dx)
             mdh.setEntry('chroma.dy', dy)
         


barentine:
Looks like the default `allow_pickle` was changed from True to False in 1.16.3, but despite that I'm fielding some support for someone who's conda list is saying 1.14.6 (which appears to be wrong given the code in https://github.com/numpy/numpy/blob/maintenance/1.14.x/numpy/lib/npyio.py). Anyway, wouldn't hurt for this to work on later numpy versions too.

David-Baddeley:
Ideally we should make all shiftmap serialisation use the .json format used in the metadata to make things safe, but we probably need this for backwards compatibility in any case.

barentine:
Thanks David. We might need to double check how they're currently getting saved in the splitterest stuff - I was debugging this on a file from an 'sf' a collaborator made last week I think.

PR #451, opened 2020-09-23
Mesh fixes

zacsimile:
The recent mesh speedups cause the resulting mesh to look slightly different from how it used to look.

Bugfix, I think, unless the speedup also included some bugfixes.

**Proposed changes:**
- Update local manifold definition (not the fix, but useful)
- Consistency in use of int32_t and `halfedge` vs. `chalfedge` (not the fix, but useful)
- ...

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/experimental/_triangle_mesh.pyx
+++ b/PYME/experimental/_triangle_mesh.pyx
@@ -188,8 +188,8 @@ VERTEX_DTYPE2 = np.dtype([('position0', 'f4'),
 cdef extern from "triangle_mesh_utils.c":
     void _update_face_normals(np.int32_t *f_idxs, halfedge_t *halfedges, vertex_d *vertices, face_d *faces, signed int n_idxs)
     
-    void update_face_normal(int f_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
-    void update_single_vertex_neighbours(int v_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
+    void update_face_normal(np.int32_t f_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
+    void update_single_vertex_neighbours(np.int32_t v_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
 
 LOOP_ALPHA_FACTOR = (np.log(13)-np.log(3))/12
 
@@ -2075,8 +2075,7 @@ cdef class TriangleMesh(TrianglesBase):
             # TODO - move this to a helper function
             self._vertices['locally_manifold'] = 1
             self._vertices['locally_manifold'][self._halfedges['vertex'][self._halfedges['twin'] == -1]] = 0
-            self._vertices['locally_manifold'][-1] = (self._halfedges['vertex'][-1] != -1) and (self._halfedges['twin'][-1] != -1)
-
+            
             for i in range(n_halfedges):
                 if (self._chalfedges[i].vertex != -1) and (self._chalfedges[i].length < collapse_threshold):
                     self.edge_collapse(i)
@@ -2089,7 +2088,6 @@ cdef class TriangleMesh(TrianglesBase):
             # TODO - move this to a helper function / make collapse update this so we don't need to recompute
             self._vertices['locally_manifold'] = 1
             self._vertices['locally_manifold'][self._halfedges['vertex'][self._halfedges['twin'] == -1]] = 0
-            self._vertices['locally_manifold'][-1] = (self._halfedges['vertex'][-1] != -1) and (self._halfedges['twin'][-1] != -1)
             
             self.regularize()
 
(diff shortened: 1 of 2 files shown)


PYME/experimental/_triangle_mesh.pyx
```diff
@@ -23,8 +23,8 @@ cdef extern from 'triangle_mesh_utils.h':
         
     cdef struct face_t:
         np.int32_t halfedge
-        float normal[VECTORSIZE]
```
David-Baddeley, line 26: these definitions should match whatever is in triangle_mesh_utils.h (which uses `float`, not `float32_t`). There could be an argument for changing triangle_mesh_utils.h to use sized types for float, although a) I don't know know where and/or if sized floating point types are defined in c (and if this is consistent across compilers) and b) I don't think it's nearly critical - whilst `int` means different things across architectures, I'm pretty sure that float always implies 32 bit and double always implies 64 bit.

PYME/experimental/_triangle_mesh.pyx
```diff
@@ -188,8 +188,8 @@ VERTEX_DTYPE2 = np.dtype([('position0', 'f4'),
 cdef extern from "triangle_mesh_utils.c":
     void _update_face_normals(np.int32_t *f_idxs, halfedge_t *halfedges, vertex_d *vertices, face_d *faces, signed int n_idxs)
     
-    void update_face_normal(int f_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
-    void update_single_vertex_neighbours(int v_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
+    void update_face_normal(np.int32_t f_idx, halfedge_t *halfedges, vertex_d *vertices, face_d *faces)
```
David-Baddeley, line 191: This should be `int`, as it's used as an index into the array. Otherwise we force a cast when we access array elements. The fact that the indices are stored as `int32_t` within the structures to save memory is irrelevant.

PYME/experimental/triangle_mesh_utils.c
```diff
@@ -58,7 +58,7 @@ DB: Performance suggestions:
 
 */
 
-static void update_single_vertex_neighbours(int v_idx, halfedge_t *halfedges, void *vertices_, void *faces_)
+static void update_single_vertex_neighbours(int32_t v_idx, halfedge_t *halfedges, void *vertices_, void *faces_)
```
David-Baddeley, line 61: `v_idx` should be `int` (used here as an index to an array)

PYME/experimental/_triangle_mesh.pyx
zacsimile: I remember needing `float32_t` for `float` on Windows. I could be mis-remembering. Either way, this doesn't break anything, I did it to make sure we absolutely were not using the wrong data type. Can easily be reverted once we find whatever's got meshing behaving differently than it used to.

PYME/experimental/triangle_mesh_utils.c
```diff
@@ -58,7 +58,7 @@ DB: Performance suggestions:
...
 {
     int32_t i, k, orig_idx, curr_idx, twin_idx, tmp;
```
David-Baddeley, line 63: `i`, an `k` should definitely be `int`, rather than `int32_t`. Probably 6 of one/half a dozen of the other for the other variables.

zacsimile: They were all `int32_t` in the version of meshing that last behaved what I think is correctly, so I changed them back to make sure this wasn't what was causing the issue.

PYME/experimental/_triangle_mesh.pyx
zacsimile: Fair enough.

PYME/experimental/triangle_mesh_utils.c
zacsimile: Fair enough.

PYME/experimental/triangle_mesh_utils.c
```diff
@@ -262,9 +264,9 @@ static PyObject *update_vertex_neighbors(PyObject *self, PyObject *args)
 }
 
 
-static void update_face_normal(int f_idx, halfedge_t *halfedges, void *vertices_, void *faces_)
+static void update_face_normal(int32_t f_idx, halfedge_t *halfedges, void *vertices_, void *faces_)
```
David-Baddeley, line 267: `f_idx` and `k` should probably be int

David-Baddeley:
It's sometimes a hard call when to use `int32_t` and when to use `int`. In general,

``` c
int32_t j;
some_array[j];
```

will translate to when you compile.

``` c
int32_t j;
some_array[(int)j];
```

and 

```c
int32_t i;
for (i = 0; i < N; i++){
    // do something
}
```

will be slower than 

```c
int i;
for (i = 0; i < N; i++){
    // do something
}
```

i.e. for looping variables you should always use the native `int` type and array access with non-ints always involves a cast. Often it's a matter of deciding when that cast happens.

```c
void some_function(int32_t idx, ...)
{
    var = an_array[idx];

    ....
}

for (int32_t i = 0, i  < N; i ++)
{
    some_function(i);
}
```

might well end up generating something like

```c

for (int _i = 0, _i  < N; _i ++)
{
    i = (int32_t) _i;
    var = an_array[(int)i];
    ....
}
```

with a couple of redundant casts. This of course all depends on how smart the compiler is. In any case it's unlikely to make a real difference to anything. I can't see a scenario where it would result in errors. Mostly just wanted to point out why I'd used `int`.

zacsimile:
Thanks for typing this out! I wasn't clear on this, so it's good to know. The main reason I converted some of the `int`s back to `int32_t` was because they were `int32_t`s in earlier versions of meshing and I just wanted to eliminate the possibility that int8 vs. int32 was creating a problem. I never truly suspected it was, but it was an easy check. I will revert back to int before I convert this from a draft.

David-Baddeley:
Just out of curiosity, ran this through the compiler to see what it actually did. 

```c
#include <stdint.h>

int32_t foo[20];

int test()
{
    int j;
    int out=0;

    for (j=0; j<20; j++)
    {
        out += foo[j];
    }

    return out;
}

int32_t test2()
{
    int32_t j;
    int32_t out=0;

    for (j=0; j<20; j++)
    {
        out += foo[j];
    }

    return out;
}
```

and the compiler was too smart by half - generated identical assembly for both functions, interestingly using a 32 bit int for the looking variable.

David-Baddeley:
And that was without optimisation - with optimisation it completely removes the look and replaces it with 20 additions

David-Baddeley:
actually, with gcc/clang `int==int32_t` (I had thought the default was 64 bit ints on 64 bit platforms.

zacsimile:
I'm missing a couple of commits and my local copy is a bit messed up, so I'm going to close this and open a new one with better fixes.

PR #453, opened 2020-09-24
Meshing tweaks

zacsimile:
A few cleanups. Nothing major.

Bugfix, barely

**Proposed changes:**
- Remove extraneous argument from `_insert_new_edge`
- Extra safety checks for -1 halfedges in the single vertex/face update functions
- Make sure the last entry in `locally_manifold` isn't automatically marked as non-manifold
- Consistent use of `chalfedges` vs. `halfedges`

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/experimental/_triangle_mesh.pyx
+++ b/PYME/experimental/_triangle_mesh.pyx
@@ -1236,6 +1236,7 @@ cdef class TriangleMesh(TrianglesBase):
         cdef int idx
         
         try:
+            # TODO: This was originally set to pop(0). What was the reasoning?
             idx = el_vacancies.pop(-1)
         except IndexError:
             # no vacant slot, resize
@@ -1254,6 +1255,7 @@ cdef class TriangleMesh(TrianglesBase):
                 # el_vacancies = [int(x) for x in np.flatnonzero(el_arr[key] == -1)]
                 el_vacancies = np.flatnonzero(el_arr[key] == -1).tolist()
 
+            # TODO: This was originally set to pop(0). What was the reasoning?
             idx = el_vacancies.pop(-1)
 
         if idx == -1:
--- a/PYME/experimental/_triangle_mesh.pyx
+++ b/PYME/experimental/_triangle_mesh.pyx
@@ -1236,7 +1236,7 @@ cdef class TriangleMesh(TrianglesBase):
         cdef int idx
         
         try:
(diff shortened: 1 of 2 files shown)


PYME/experimental/_triangle_mesh.pyx
```diff
@@ -2075,6 +2077,7 @@ cdef class TriangleMesh(TrianglesBase):
             # TODO - move this to a helper function
             self._vertices['locally_manifold'] = 1
             self._vertices['locally_manifold'][self._halfedges['vertex'][self._halfedges['twin'] == -1]] = 0
+            self._vertices['locally_manifold'][-1] = (self._halfedges['vertex'][-1] != -1) and (self._halfedges['twin'][-1] == -1)
```
David-Baddeley, line 2080: ???? why do we need a special case on the last vertex here?  ... also what does the last half-edge have to do with things?

PYME/experimental/triangle_mesh_utils.c
```diff
@@ -72,6 +72,8 @@ static void update_single_vertex_neighbours(int v_idx, halfedge_t *halfedges, vo
     vertex_t *vertices = (vertex_t*) vertices_;
     face_t *faces = (face_t*) faces_;
 
+    if (v_idx == -1) return;
```
David-Baddeley, line 75: should we silently ignore here, or flag an error?

PYME/experimental/_triangle_mesh.pyx
zacsimile: I think I wrote this in case you ran into a case where `self._halfedges['vertex'][self._halfedges['twin'] == -1` returned a -1 (not impossible). In this case, `self._vertices['locally_manifold'][-1] = 0`. However, `self._vertices['locally_manifold'][-1]` might in fact be be manifold. So, you'd treat it as non-manifold unnecessarily. It's a silly little check (and I think I meant to write `self._halfedges['twin'][-1] != -1`). I'm not sure most of these meshing tweaks are relevant anymore.

PYME/experimental/triangle_mesh_utils.c
zacsimile: Flag an error.

David-Baddeley:
@zacsimile  could you have a quick look at the conflict?

zacsimile:
No longer relevant/helpful

PR #458, opened 2020-09-25
make FlatfieldAndDarkCorrect work for multiview data/without flatmap

barentine:
Addresses issue camera origin + camera roi size often won't get you the right darkmap slice for multiview data.

enhancement
**Proposed changes:**
- use camera info manager to get slices
- make flatfield map optional

**Checklist:**

- [ ] Tested with numpy=1.14
1.16
3.6
4.0.4

### DIFF
--- a/PYME/localization/remFitBuf.py
+++ b/PYME/localization/remFitBuf.py
@@ -136,6 +136,8 @@ class CameraInfoManager(object):
         """
         Extract ROI coordinates from metadata
 
+        TODO - refactor out of here as it is being used in non-fitting code
+
         Parameters
         ----------
         md: dict-like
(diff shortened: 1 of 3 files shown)


PYME/localization/remFitBuf.py
```diff
@@ -131,7 +131,8 @@ class CameraInfoManager(object):
     def __init__(self):
         self._cache = {}
 
-    def _parseROI(self, md):
+    @staticmethod
+    def _parseROI(md):
         """
         Extract ROI coordinates from metadata
```
David-Baddeley, line 138:
```suggestion

        TODO - refactor out of here as it is being used in non-fitting code

```

Issue #386, opened 2020-08-04
[BUG] tiler stop race condition


barentine:
**Describe the bug**
Recent fixes in `pointScanner` (#355 #366) to call it's stop method on time have fixed things for tile_triggered type acquisitions (which now work smoothly), however there appears to be a race condition with the pyramid between the tiler's stop method, which calls `self.P.update_pyramid()` and a `tile_panel` `wx.CallAfter` which also tries to `self.scope.tiler.P.update_pyramid()` in `_launch_viewer`

```
987.1 -57.5 17182 8098 -44.753685 236.2463
988.3 -32.800000000000004 17192 8313 -63.753685 249.2463
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\frameWrangler.py", line 235, in onExpReady
    self.onFrame.send(sender=self, frameData=d)
  File "C:\Users\Bergamot\Anaconda2\envs\pyme\lib\site-packages\dispatch\dispatcher.py", line 162, in send
    response = receiver(signal=self, sender=sender, **named)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Utils\tiler.py", line 80, in tick
    pointScanner.PointScanner.tick(self, frameData, **kwargs)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Utils\pointScanner.py", line 198, in tick
    self._stop()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Utils\tiler.py", line 103, in _stop
    self.P.update_pyramid()
  File "c:\users\bergamot\code\python-microscopy\PYME\Analysis\tile_pyramid.py", line 374, in update_pyramid
    self._rebuild_base()
  File "c:\users\bergamot\code\python-microscopy\PYME\Analysis\tile_pyramid.py", line 366, in _rebuild_base
    occ = self._occ.get_tile(0, xc, yc) + 1e-9
TypeError: unsupported operand type(s) for +: 'NoneType' and 'float'
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\Analysis\tile_pyramid.py", line 26, in load
    return self._cache[filename].data
KeyError: 'F:\\2020_8_3_A0_PD20_digodTAF647\\0\\036\\036_016_occ.pzf'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
[...]

PR #459, opened 2020-09-25
tweaks to HTSMS init/protocols (and one UOA_n!)

barentine:
**Proposed changes:**
- open focus lock in same script but different console
- pass popen flag to UOA_n init to get a second console as well (obviously take this out if you don't want it @David-Baddeley )
- open tileviewer with separate console so it can be killed/restarted e.g. if it hangs on a recipe, etc.
- remove LaunchAnalysis calls from htsms protocols 
- remove serial fiber shaker as we switched to a wall-plug always-on unit
- extend delay in callafter from 1 to 10 s for tile_panel update_pyramid call. See #386. This isn't an elegant fix, but at least on small tiles does the trick. Didn't intend to commit this change with the rest, but it is helpful for the small tiles I've been doing on test samples lately so I'd be OK leaving it in for now.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Scripts/init_UOA_n.py
+++ b/PYME/Acquire/Scripts/init_UOA_n.py
@@ -235,7 +235,7 @@ def drift_tracking(MainFrame, scope):
 
     def _drift_init():
         #scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking" -m "compact"' % (sys.executable, PYMEAcquire.__file__), shell=True)
-        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), shell=True)
+        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), creationflags=subprocess.CREATE_NEW_CONSOLE)
 
     time.sleep(15)
     _drift_init()
--- a/PYME/Acquire/Scripts/init_UOA_n.py
+++ b/PYME/Acquire/Scripts/init_UOA_n.py
@@ -235,7 +235,7 @@ def drift_tracking(MainFrame, scope):
 
     def _drift_init():
         #scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking" -m "compact"' % (sys.executable, PYMEAcquire.__file__), shell=True)
-        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), creationflags=subprocess.CREATE_NEW_CONSOLE)
+        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), shell=True)
 
     time.sleep(15)
     _drift_init()
@@ -253,5 +253,4 @@ joinBGInit() #wait for anyhting which was being done in a separate thread
 
 #time.sleep(.5)
 scope.initDone = True
-
 
--- a/PYME/Acquire/ui/tile_panel.py
+++ b/PYME/Acquire/ui/tile_panel.py
@@ -1,5 +1,7 @@
 import wx
 from PYME.Acquire.Utils import tiler
+import logging
+logger = logging.getLogger(__name__)
 
 class TilePanel(wx.Panel):
     def __init__(self, parent, scope):
@@ -93,7 +95,7 @@ class TilePanel(wx.Panel):
         self.scope.tiler.on_stop.disconnect(self._on_stop)
         self.scope.tiler.progress.disconnect(self._update)
         
-        wx.CallAfter(wx.CallLater,1e3, self._launch_viewer)
+        wx.CallAfter(wx.CallLater,1e4, self._launch_viewer)
         
         
     def _launch_viewer(self):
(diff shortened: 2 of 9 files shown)


PYME/Acquire/ui/tile_panel.py
```diff
@@ -111,7 +111,7 @@ def _launch_viewer(self):
         try:
             requests.get('http://127.0.0.1:8979/set_tile_source?tile_dir=%s' % self.scope.tiler._tiledir)
         except requests.ConnectionError:
-            self._gui_proc = subprocess.Popen('%s -m PYME.tileviewer.tileviewer %s' % (sys.executable, self.scope.tiler._tiledir), shell=True)
```
David-Baddeley, line 114: This is not a 1:1 substitution, as `shell=True` implies a bunch of other things in addition to `CREATE_NEW_CONSOLE` (we might run into issues with shell expansion of the command line amoungst other things - specifically you will no longer be able to use paths relative to, e.g. home, but I suspect there might be other issues. More concerningly, it will also causer the tileviewer to fail on non-windows platforms as `subprocess.CREATE_NEW_CONSOLE` is windows only.

PYME/Acquire/Scripts/init_UOA_n.py
```diff
@@ -235,7 +235,7 @@ def drift_tracking(MainFrame, scope):
 
     def _drift_init():
         #scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking" -m "compact"' % (sys.executable, PYMEAcquire.__file__), shell=True)
-        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), shell=True)
+        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), creationflags=subprocess.CREATE_NEW_CONSOLE)
```
David-Baddeley: the docs for the subprocess module state that `shell=True` implies `creationflags=subprocess.CREATE_NEW_CONSOLE` is that not the case? Does setting `creationflags=subprocess.CREATE_NEW_CONSOLE` actually create a new window?

barentine: yes. 
image

PYME/Acquire/ui/tile_panel.py
barentine: not sure what you mean by shell expansion of the command line, but this turns out to be a nice little feature to be able to kill/restart the tileviewer if I e.g. brick the GPU on a detection recipe, without having to restart PYMEAcquire or guess on windows which python process it is.

I'll add a try/except attribute error to catch for non-windows users. Will also check for abspath on tile_dir and add cwd if it's not. 

I don't see anything in shell=False that would be a problem for us since we are otherwise not using pipes / ~ / etc.

PYME/Acquire/Scripts/init_UOA_n.py
barentine: Also, I'm not seeing anything in the 3.6 docs on shell=True implying CREATE_NEW_CONSOLE

David-Baddeley: This is specifically an init file for a py2 system, but I suspect the docs are not correct. In any case we should run into this problem:

*"If passing a single string, either shell must be True (see below) or else the string must simply name the program to be executed without specifying any arguments"* (from 3.7 docs, but reasonably consistent across versions). Will play when I get time, but reverting back here for now.

David-Baddeley:
```suggestion
        scope.p_drift = subprocess.Popen('%s "%s" -i init_drift_tracking.py -t "Drift Tracking"' % (sys.executable, PYMEAcquire.__file__), shell=True)
```

PYME/Acquire/ui/tile_panel.py
David-Baddeley: Previously you absolutely could have used a path relative to `~`, but I'm not super attached to doing that. 

If it's not an absolute path, it's probably best to drop it under `PYMEDATADIR` rather than `cwd()` as the working directory is often weird if you are running using a windows shortcut. Will see if I can change to the relevant nameutils call.

I'm still not sure why we're not running into  *"If passing a single string, either shell must be True (see below) or else the string must simply name the program to be executed without specifying any arguments."* though (suspect windows might be doing some limited expansion/parsing under the hood - I've definitely seen breakages related to the single string stuff on *nix.

David-Baddeley: I've done most of the testing on *nix so have either killed as needed with `ps -A | grep tileviewer`, or launched the tileviewer separately in advance. One of the advantages of using `shell=True` is that you end up with the argument in the process name.

PYME/Acquire/ui/tile_panel.py
```diff
@@ -102,16 +102,27 @@ def _launch_viewer(self):
...
 
         self.scope.tiler.P.update_pyramid()
         
         #if not self._gui_proc is None:
         #    self._gui_proc.kill()
+
+        # abs path the tile dir
+        tiledir = self.scope.tiler._tiledir
+        if not os.path.isabs(tiledir):
+            tiledir = os.path.join(os.getcwd(), tiledir)
```
David-Baddeley:
```suggestion
            # TODO - should we be doing the `.isabs()` check on the parent directory instead?
            from PYME.IO.FileUtils import nameUtils
            tiledir = nameUtils.getFullFilename(tiledir)
```

PYME/Acquire/ui/tile_panel.py
```diff
@@ -93,7 +93,7 @@ def _on_stop(self, *args, **kwargs):
         self.scope.tiler.on_stop.disconnect(self._on_stop)
         self.scope.tiler.progress.disconnect(self._update)
         
-        wx.CallAfter(wx.CallLater,1e3, self._launch_viewer)
```
David-Baddeley, line 96: Why such a large delay? Do we need to make this customisable and/or heuristic?

PYME/Acquire/ui/tile_panel.py
```diff
@@ -93,7 +93,7 @@ def _on_stop(self, *args, **kwargs):
...
+        wx.CallAfter(wx.CallLater,1e4, self._launch_viewer)
```
David-Baddeley, line 98:
```suggestion
        # FIXME - previous delay was 1e3, which seems more reasonable. Do we need a config option (or heuristic) here ?
        # assume this change was due to the time it takes to build a pyramid after tiling ends. Might ultimately be fixed when we revisit live tiling. 
        wx.CallAfter(wx.CallLater,1e4, self._launch_viewer)
```

PYME/Acquire/ui/tile_panel.py
barentine: Yeah this isn't really a solution, but a long enough delay to avoid #386 on small overviews

Issue #469, opened 2020-10-04
[BUG] Can't open two files in a single PYMEVis window on Py3


zacsimile:
**Describe the bug**
PYMEVis won't let me open one file, and then let me decide to use and open different file, without first launching a new instance.

**To Reproduce**
1. Open any file in PYMEVis.
2. Try to open a different file.

**Expected behavior**
The second file replaces the first file in my PYMEVis window.

**Traceback**
```
 Error whilst running File>&Open
==============================================
python-microscopy=20.09.18
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 402, in OnOpenFile
    self.OpenFile(filename)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 741, in OpenFile
    self.recipeView._layout()
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/vertical_recipe_display.py", line 83, in _layout
    self.ordered, self.x0s = recipeLayout.layout_vertical(dg, rdg)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipeLayout.py", line 207, in layout_vertical
    _descend(n, dg, rdg)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipeLayout.py", line 197, in _descend
    for child in sorted(children):
TypeError: '<' not supported between instances of 'ProcessColour' and 'ProcessColour'
```

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15
 - Python version: 3.6.10
 - numpy version: 1.16.6
 - wxpython version [if a UI bug]: 4.0.4

**Additional context**

PR #470, opened 2020-10-04
Sort the children on class name for py3

zacsimile:
Addresses issue #469.

Bugfix

**Proposed changes:**
- Add a key to sort on for py3.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/recipes/recipeLayout.py
+++ b/PYME/recipes/recipeLayout.py
@@ -194,7 +194,7 @@ def layout_vertical(dg, rdg):
             for child in sorted(children):
                 _descend(child, dg, rdg)
         else:
-            for child in sorted(children, key=lambda x: x if isinstance(x,str) else x.__class__.__name__) :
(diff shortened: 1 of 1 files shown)


PYME/recipes/recipeLayout.py
```diff
@@ -194,7 +194,7 @@ def _descend(node, dg, rdg):
             for child in sorted(children):
                 _descend(child, dg, rdg)
         else:
-            for child in sorted(children):
+            for child in sorted(children, key=lambda x: x if isinstance(x,str) else x.__class__.__name__) :
```
David-Baddeley: would it be reasonable/possible to replace with 
```suggestion
            for child in sorted(children, key=lambda x: str(x)) :
```
(for a class, that will give you `__class__.__name__` and a memory address, which will ensure that sorting is reproducible on a given set of objects)

Using things like `__class__.__name__` explicitly always makes me nervous - it's generally a sign that here be dragons.

Issue #471, opened 2020-10-04
[BUG] Can't open two files in a row with different pipeline variables on py3


zacsimile:
**Describe the bug**
PYMEVis won't let me open one file, and then let me decide to use and open different file, without first launching a new instance, if the second file does not contain the same pipeline variables as the first.

**To Reproduce**
1. Open any file in PYMEVis.
2. Try to open a different file that has fewer pipeline variables than the first file.

**Expected behavior**
The second file replaces the first file in my PYMEVis window.

**Traceback**
```
 Error whilst running File>&Open
==============================================
python-microscopy=20.09.18
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 402, in OnOpenFile
    self.OpenFile(filename)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 736, in OpenFile
    self.pipeline.OpenFile(filename, **args)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/pipeline.py", line 828, in OpenFile
    self.recipe.execute()
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 641, in execute
    m.execute(self.namespace)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/tablefilters.py", line 38, in execute
    filtered = tabular.ResultsFilter(inp, **self.filters)
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 752, in __init__
    raise KeyError('Requested key not present: ' + k)
KeyError: 'Requested key not present: error_y'
```

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15
 - Python version: 3.6.10
 - numpy version: 1.16.6
 - wxpython version [if a UI bug]: 4.0.4

**Additional context**

barentine:
Call me crazy, but this might be solved with merge of #284. 
edit: I take that back

David-Baddeley:
[...]

PR #472, opened 2020-10-05
fix for 471

David-Baddeley:
Addresses issue #471  .

TODOs for full solution:

- use the clobber test again before adding recipe modules (deferred to avoid conflicts with #284 )
- workout what to do with filter keys when not clobbering
- ditto for ratiometric info


PYME/LMVis/pipeline.py
```diff
@@ -767,64 +780,49 @@ def OpenFile(self, filename= '', ds = None, **kwargs):
...
-        if False:  # 'imgBounds' in kwargs.keys():
-            self.imageBounds = kwargs['imgBounds']
-        elif (not (
-                'scanx' in mapped_ds.keys() or 'scany' in mapped_ds.keys())) and 'Camera.ROIWidth' in self.mdh.getEntryNames():
+        if False:  # 'imgBounds' in kwargs.keys(): 
+            # TODO - why is this disabled? Current usage would appear to be when opening from LMAnalysis
+            # during real-time localization, to force image bounds to match raw data, but also potentially useful
+            # for other scenarios where metadata is not fully present.
+             self.imageBounds = kwargs['imgBounds']
+        elif ('scanx' not in ds.keys() or 'scany' not in ds.keys()) and 'Camera.ROIWidth' in self.mdh.getEntryNames():
```
David-Baddeley: This needs to run on the **mapped** data, not the raw data (both the test for keys and `ImageBounds.estimateFromSource`.

David-Baddeley: Hacky and mostly equivalent solution would be is to estimate `self.imageBounds` based on `self.dataSources['Localisations']` *after* the call to `recipe.execute()`, need to think if this (a) makes sense, or whether we should b) use `.selectedDataSource` instead, or c) do the bounds estimation in `Pipelineify` instead.

Issue #475, opened 2020-10-06
[BUG] Can't open h5 files in PYMEVis


zacsimile:
**Describe the bug**
```
 Error whilst running File>&Open
==============================================
python-microscopy=20.10.05
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 402, in OnOpenFile
    self.OpenFile(filename)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 736, in OpenFile
    self.pipeline.OpenFile(filename, **args)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/pipeline.py", line 778, in OpenFile
    ds = self._ds_from_file(fn, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/pipeline.py", line 680, in _ds_from_file
    self.recipe._inject_tables_from_h5('', h5f, filename, '.hdf')
AttributeError: 'ModuleCollection' object has no attribute '_inject_tables_from_h5'

```

**To Reproduce**
1. Open an `h5` file in `PYMEVis`

**Expected behavior**
The file opens

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15
 - Python version: 3.6.10
 - numpy version: 1.16.6
 - wxpython version [if a UI bug]: 4.0.4

**Additional context**

zacsimile:
Following #477, I still can't open `h5` files but I get a different error.

```
 Error whilst running File>&Open
==============================================
python-microscopy=20.10.05
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 402, in OnOpenFile
    self.OpenFile(filename)
[...]

PR #477, opened 2020-10-06
Correct name to _inject_tables_from_hdf5

zacsimile:
Addresses issue #475.

Bugfix

**Proposed changes:**
- Change function call `_inject_tables_from_h5` to `_inject_tables_from_hdf5`

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -677,7 +677,7 @@ class Pipeline:
             self.filesToClose.append(h5f)
             
             #defer our IO to the recipe IO method - TODO - do this for other file types as well
-            self.recipe._inject_tables_from_h5('', h5f, filename, '.hdf')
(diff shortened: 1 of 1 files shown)


zacsimile:
Re: #329. I think this might actually be an error that `pylint` would catch but `pyflakes` wouldn't since `pyflakes` doesn't try importing. I still think `flake8` is the right call, just something to be aware of if true. Please double-check me that I've correctly understood how these two packages differ.

David-Baddeley:
I suspect you're right about pyflakes vs lint - with imports. It might be possible to run lint in a more restricted capacity (i.e. only on the files that are touched).

zacsimile:
#329 did run the linter only on the changed files in each commit, and I think that's absolutely the right call for `flake8` as well or whatever package we use.

Issue #478, opened 2020-10-06
[BUG] meshing compilation fails in some environments


barentine:
**Describe the bug**
```
PYME/experimental/_triangle_mesh.c:11004:3: warning: passing argument 4 of ‘_update_face_normals’ from incompatible pointer type [[0/327]
 by default]
   _update_face_normals((&(*((__pyx_t_5numpy_int32_t *) ( /* dim=0 */ (__pyx_v_idxs.data + __pyx_t_7 * __pyx_v_idxs.strides[0]) )))), __$
yx_v_self->_chalfedges, __pyx_v_self->_cvertices, __pyx_v_self->_cfaces, (__pyx_v_idxs.shape[0]));
   ^
In file included from PYME/experimental/_triangle_mesh.c:633:0:
PYME/experimental/triangle_mesh_utils.c:316:13: note: expected ‘struct face_t *’ but argument is of type ‘struct face_d *’
 static void _update_face_normals(int32_t *f_idxs, halfedge_t *halfedges, vertex_t *vertices, face_t *faces, signed int n_idxs)
             ^
error: Command "gcc -pthread -B /home/ubuntu/miniconda3/compiler_compat -Wl,--sysroot=/ -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -Ws$
rict-prototypes -fPIC -IPYME/experimental -I/home/ubuntu/miniconda3/lib/python3.6/site-packages/numpy/core/include -I/home/ubuntu/minicon
da3/lib/python3.6/site-packages/numpy/core/include -I/home/ubuntu/miniconda3/include/python3.6m -I/home/ubuntu/miniconda3/include/python3
.6m -c PYME/experimental/_triangle_mesh.c -o build/temp.linux-x86_64-3.6/PYME/experimental/_triangle_mesh.o -MMD -MF build/temp.linux-x86
_64-3.6/PYME/experimental/_triangle_mesh.o.d -O3 -fno-exceptions -ffast-math -march=native -mtune=native" failed with exit status 1

```
<details> <summary> on this build </summary>

```
# packages in environment at /home/ubuntu/miniconda3:
#
# Name                    Version                   Build  Channel
absl-py                   0.9.0                    pypi_0    pypi
alabaster                 0.7.12                   py36_0    anaconda
appdirs                   1.4.3                    pypi_0    pypi
asn1crypto                1.3.0                    py36_0    anaconda
astor                     0.8.1                    pypi_0    pypi
[...]

PR #479, opened 2020-10-06
Address #478 via casting of faces, vertices

zacsimile:
Addresses issue #478.

Bugfix

**Proposed changes:**
- Apply same face/vertex casting as in other c functions to deal with `face_t` vs. `face_d` structs, etc.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/experimental/triangle_mesh_utils.c
+++ b/PYME/experimental/triangle_mesh_utils.c
@@ -313,8 +313,10 @@ static void update_face_normal(int f_idx, halfedge_t *halfedges, void *vertices_
     }
 }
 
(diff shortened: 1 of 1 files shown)


barentine:
@David-Baddeley if you're up/PYMEing at the moment would you mind hitting this first?

David-Baddeley:
Hmm, I actually had going the other way and removing the current `void *` stuff and going back to typed pointers in the other c functions on my mental TODO list. The `void *` and cast is a brute force method (effectively disables all compiler type checking of the arguments) and was only needed when I didn't know how to do casts in cython. It is pretty messed up from a writing good code standpoint.

Something tells me it would be a good idea to work out why we are getting this issue, rather than just brute forcing it away.

David-Baddeley:
I think this is now defunkt as of #489

PR #476, opened 2020-10-06
bytes/str collision in clusterUI recipe view 

barentine:
Addresses issue
```
Traceback:
File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/exception.py" in inner
  41.             response = get_response(request)
File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/base.py" in _legacy_get_response
  249.             response = self._get_response(request)
File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  187.                 response = self.process_exception_by_middleware(e, request)
File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/base.py" in _get_response
  185.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/home/smeagol/code/python-microscopy/PYME/cluster/clusterUI/recipes/views.py" in run_template
  66.         recipe_text = recipe_text.replace('{'+file_input +'}', input_url)
Exception Type: TypeError at /recipes/run_template/
Exception Value: a bytes-like object is required, not 'str'
```

**Is this a bugfix or an enhancement?**
bugfix
**Proposed changes:**
decode the recipe text so it's str

PR #486, opened 2020-10-07
use RecipeRule in clusterUI rule pushing

barentine:
Addresses issue was running into issues with multiline str yaml recipe text in `HTTPRulePusher` with json dumps/loads required to replace our templated USERFILEs, and I am more invested in refactored rules working, and encountered less json headaches this way anyway.

see #476 

bugfix
**Proposed changes:**
- remove task pusher check
- user RecipeRule to push itself rather than HTTPRulePusher

**Checklist:**

- [ ] Tested with numpy=1.14
1.16
3.6


David-Baddeley:
Could you provide a clear explanation of what the problem you are trying to address is (file an issue?).

The stuff in `cluster._rules` is a temporary stopgap for use with your acquisition recipe chaining, but which is slated for removal (as discussed previously) once we've done some more refactoring. Although some refactoring of rule pushing is on the TODO list, I'm obviously not super keen on replacing code which doesn't depend on something which is going to be removed with code that does.

barentine:
see #487. 

`cluster._rules` may not be the final refactor of rules that push themselves, but looking at what you envision in your #372 comment, I don't see how whatever supersedes it wouldn't be a drop-in in this context.

I of course prefer to work with forward-progress code: `cluster._rules` is closer to an eventual `cluster.rules` than `HTTPRulePusher.RecipePusher`, so I think we should move on with it.

barentine:
@David-Baddeley updated to use `cluster.rules.RecipeRule` rather than `cluster._rules`.

PYME/cluster/clusterUI/recipes/views.py
```diff
@@ -28,31 +29,30 @@ def get_input_glob(request):
...
-
-    fileNames = request.POST.getlist('files', [])
-    pusher.fileTasksForInputs(input=fileNames)
+    from PYME.cluster.rules import RecipeRule
+    import posixpath
 
+    recipeURI = ('pyme-cluster://%s/' % server_filter) + request.POST.get('recipeURL').lstrip('/')
+    input_list = request.POST.getlist('files', [])
+    try:
+        # put outputs in the same directory as selected files, under 'analysis'
```
David-Baddeley: I don't like this as it has the potential to get super messy if you are analysing files across directories. Approach to date was to hard-code the output directory in the recipe / not rely on substitution (this is safe, if gross), a better option might be to have it as a UI input as in `run_template()`.

barentine: how do you analyze files across directories in this view?

barentine: agree ui input would be best

David-Baddeley: This is used by `recipe_standalone()` which accepts globs for inputs.

David-Baddeley: as well as the lightweight form in the files view.

David-Baddeley: Pre-populating the output directory as current directory + /analysis might not be a bad idea, but probably better done in the html view.

PR #489, opened 2020-10-08
Issue478v2

David-Baddeley:
Addresses issue #478.

Slightly safer alternative to #479 (was quicker than opening a new issue).

As to why this is preferable, first option is equivalent to saying:

"we'll accept anything that gets passed. Later we will assume that it is a `face_t *`, and will treat it as such. There is nothing in our function signature which gives you any information about what you should pass"

this option is like

"The function accepts a `face_t *`, we have something which is a `face_d *`. We know that the two structs are equivalent in memory, so we can tell the compiler to treat our `face_d *` as though it was a `face_d *`"

TODO - change the other void* calls (this is slightly harder as there are so many of them - and putting explicit casts everywhere would result in clutter). Options are to:

a) put casts everywhere
b) write a wrapper function which accepts `vertex_d *` and `face_d *`, casts them, and calls the original function
c) make the c function accept `vertex_d *` and `face_d *` and cast internally (this is much safer than using void*s
d) use a `union` to programatically assert that `vertex_d` and `face_d` are equivalent (this might be the best option, but my initial attempts didn't work)


David-Baddeley:
ugh - bit of a branching error here  (the wx stuff was in a separate branch and shouldn't have showed up here)

zacsimile:
I'm in favor of a or c. I think explicitly noting which types we're using could be quite helpful. However, d would result in the cleanest code, I agree, and since they are essentially the same type this would be the optimal route. I agree it should work. Maybe I'll have a go at it.

Issue #460, opened 2020-09-29
[BUG] PYMEVis will color layers on `z` by default even if it is not present


zacsimile:
**Describe the bug**
If we open a MATLAB file and only specify columns `x` and `y`, the data will load but the display will appear black because the default `Colour` option for `Layer 0 - output` will be set to `z`.

**Expected behavior**
The points should display. Perhaps we need to color on another variable by default in the absence of `z` and `t`.

**Screenshots**
![image](https://user-images.githubusercontent.com/1263313/94554329-ee4dfa80-0227-11eb-95ce-6436bef6c70d.png)

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15.6
 - Python version: 3.6.10
 - numpy version: 1.16.0
 - wxpython version [if a UI bug]: 4.0.4

yongdengzhang:
When I tried to open a 4Pi data for the first time, I didn't see z in the drop-down menu so it was not specified. Then after loading, no image was displayed because z was the default color option. Even if this bug is fixed, I will still forget to specify z as I have to manually type z in the box, which is not something straightforward for a beginner. It would be great to add z in the drop-down menu but not set it as a required variable.

David-Baddeley:
Hmm, this smells a little like a regression - If both z and t were absent I'm pretty sure we used to set the colour to `''` (or possibly `'<none>'`and it would display as a constant colour.

zacsimile:
I also thought we'd fixed this a while ago.

zacsimile:
I think the short answer is the `pipeline` is returning  a `z` key even when I don't define one.

David-Baddeley:
that should not be a problem - the pipeline should always have a `z` key and it should default to 0 if `z` is not in the data. Maybe it's the defaulting to 0 which is the problem (I suspect this currently relies on `t` being present, but would need to double check).

zacsimile:
You are right that `z` gets generated in the pipeline with values of 0. I forgot this was default behavior. There's no `t` in my pipeline. Here's what I think happens:

- In `PYME.LMVis.visCore._create_base_layer`, it checks for `t` and then `z`. Finding `z`, it sets this as `vertexColour`.
- In `PYME.LMVis.layers.pointcloud.update_from_datasource`, it sets `c=ds['z']`, which is all zeros.

[...]

PR #496, opened 2020-10-08
adding z, z_error to import dialog and making voxel size entry clearer

David-Baddeley:
Fixes usability issues with csv import as highlighted on twitter

UI changes should be reasonably intuitive. Need to double check docs to see if, e.g. `z` is already in the list of supported columns.

Minimal potential for side effects.


zacsimile:
Looks good! Worked well for me. Incidentally, #460 is still a very noticeable issue.

PR #497, opened 2020-10-08
pin matplotlib in test requirements

David-Baddeley:
pins matplotlib in test requirements to prevent pip installing an incompatible version.

### DIFF
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -1,8 +1,12 @@
-#NB - this is the testing requirements - it omits some packages which are not covered by the tests
+# NB - this is the testing requirements - it omits some packages which are not covered by the tests
+# Some packages are pinned to old versions to be sure that the code runs against the old package versions (assuming forwards compatibility
+# will also let us use more recent packages). One of the main drivers for testing so far back is to be able to roll bugfixes
+# out rapidly to users where completely installing from scratch is not an option.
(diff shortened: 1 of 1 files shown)


zacsimile:
This is because we're still pinning to `numpy=1.14`. Is there a good reason for this? I've been testing on `numpy=1.16` for about a year. It still supports py2 (it's the last version to do so). 1.16 has a lot of bugfixes. Seems natural to update.

David-Baddeley:
Reasons for pinning the **test** environment are 2-fold. 1) to try and make
it less of a moving target (at least until we are consistently passing all
tests) and 2) to provide a pain-free upgrade path - especially on a couple
of instrument computers which are running off source installs. If we
constantly let our minimum requirements drift it means having to
conda-install new versions of numpy etc whenever we upgrade, and in
practice completely reinstalling conda every 3-6 months when it decides
things are un-resolvable, potentially turning a 2 min git pull into a 2-3
hr ordeal.

There is currently nothing stopping us from running against a newer
version, but I think there is a strong argument for continuing to test
against older versions - it doesn't cost us anything.

David-Baddeley:
Once we drop py27 support I'd be keen to start testing against py36
numpy1.14  AND py[latest] np [latest] (or potentially just unpinned).

Issue #492, opened 2020-10-08
[BUG] pipelineify in pipeline recipe hiccup: metadata writing + file opened as read


barentine:
**Describe the bug**
New pipelineify in the pipeline recipe is something I'm pretty excited about, but we might need to wrap metadata handlers in nestedclassmetadatahandlers in recipe modules which add metadata (ideally all of them) or on load, to avoid the following
```
 Error whilst running Corrections>Multiview>Shift correct folded channels
==============================================
python-microscopy=20.10.05
python=3.6.9, platform=linux
numpy=1.16.5, wx=4.0.4 gtk2 (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/tables/file.py", line 898, in _create_path
    child = parent._f_get_child(pcomp)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/tables/group.py", line 708, in _f_get_child
    self._g_check_has_child(childname)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/tables/group.py", line 395, in _g_check_has_child
    % (self._v_pathname, name))
tables.exceptions.NoSuchNodeError: group ``/MetaData/Multiview`` does not have a child named ``shift_map``

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/smeagol/code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/home/smeagol/code/python-microscopy/PYME/LMVis/Extras/multiviewMapping.py", line 165, in OnShiftCorrectFolded
    recipe.execute()
  File "/home/smeagol/code/python-microscopy/PYME/recipes/base.py", line 641, in execute
    m.execute(self.namespace)
  File "/home/smeagol/code/python-microscopy/PYME/recipes/multiview.py", line 92, in execute
    mapped.mdh['Multiview.shift_map.location'] = loc
  File "/home/smeagol/code/python-microscopy/PYME/IO/MetaDataHandler.py", line 281, in __setitem__
    self.setEntry(name, value)
  File "/home/smeagol/code/python-microscopy/PYME/IO/MetaDataHandler.py", line 471, in setEntry
    currGroup = self.h5file._get_or_create_path('/'.join(['', 'MetaData']+ ep), True)
[...]

PR #500, opened 2020-10-09
stop modifying input metadata in recipe modules

David-Baddeley:
Addresses issue #492  

#492 unearthed a couple of recipe modules who were modifying their input (which violates the "no side effects" principle of recipe modules which is important if we want recipes to be reproducible and avoid odd behaviour). Fixed by creating a new metadatahandler each time before modifying.

### DIFF
--- a/PYME/recipes/measurement.py
+++ b/PYME/recipes/measurement.py
@@ -32,12 +32,13 @@ class MultifitBlobs(ModuleBase):
         
         img = namespace[self.inputImage]
         
-        img.mdh['Analysis.PSFSigma'] = self.blobSigma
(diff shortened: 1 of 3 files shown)


barentine (approved):

barentine:
Did you see my note @David-Baddeley? `GaussMultifitSR.FitFactory(self.scale*img.data[:,:,i], img.mdh, noiseSigma=np.ones_like(img.data[:,:,i].squeeze()))` probably wants to be `GaussMultifitSR.FitFactory(self.scale*img.data[:,:,i], mdh, noiseSigma=np.ones_like(img.data[:,:,i].squeeze()))`. Line ~42 of `PYME/recipes/measurement.py`, otherwise I don't think the blogSigma propagates to the fit

Issue #493, opened 2020-10-08
[BUG] ProcessColour ocassionally gets wildly too many color channels


barentine:
**Describe the bug**
If you throw a processcolour in front of a densitymapping module, the densitymapping module doesn't see that it's been passed a color-processed tabular type. That wouldn't really bother me, but I occasionally end up with something wild. I'm not yet if that's what's causing this catastrophe, but the print statement is in densitymapping and I wouldn't have expected us to process the color twice
![image](https://user-images.githubusercontent.com/31105780/95505104-cd9a4900-097b-11eb-91dd-28a5bd1680c3.png)

recipe:
![image](https://user-images.githubusercontent.com/31105780/95503339-3a601400-0979-11eb-83a1-1daa3b8500e0.png)

relevant modules
```
- localisations.ProcessColour:
    input: z_mapped
    output: colour_mapped
- localisations.DensityMapping:
    colours:
    - chan0
    - chan1
    inputLocalizations: colour_mapped
    manualXYBounds:
    - 0.0
    - 0.0
    - 30300.0
    - 30300.0
    outputImage: hist
    pixelSize: 25.0
    renderingModule: Histogram
    xyBoundsMode: manual
```

```
[...]

PR #501, opened 2020-10-09
coalesce_dict_sorted / deClump aggregate 0/1 indexing fixes

barentine:
Addresses issue #493 .

bugfix
**Proposed changes:**
- fix output of deClump to be the correct length (currently the 0th value in each out array is left to the whims of RAM)

DRAFT for now - need to fix the other aggregate versions / update docstrings

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

### DIFF
--- a/PYME/Analysis/points/DeClump/deClump.c
+++ b/PYME/Analysis/points/DeClump/deClump.c
@@ -489,7 +489,8 @@ static PyObject * aggregateWeightedMean(PyObject *self, PyObject *args, PyObject
 
     int nPts = 0;
     int nClumps = 0;
-    int currentClump = -1;
+    int currentClump = 0;
+    int clump_0 = 0;
 
     int i=0;
     //int j=0;
@@ -575,22 +576,24 @@ static PyObject * aggregateWeightedMean(PyObject *self, PyObject *args, PyObject
         outSig[i] = -1e4;
     }
     */
-
-
+    clump_0 = clumpIDs[0];
+    currentClump = clump_0;
+    weight_sum = 0;
+    var_sum = 0;
     //j = 0;
     for (i=0; i < nPts; i++)
     {
         if (currentClump != clumpIDs[i]){
             //We have moved on to the next clump
-            if (currentClump >= 0)
+            if (currentClump - clump_0 >= 0)  // FIXME - do we allow negative clumpIDs to get here?
             {
                 if (weight_sum == 0){
-                    outVar[currentClump] = 0;
-                    outSig[currentClump] = -1e4;
+                    outVar[currentClump - clump_0] = 0;
+                    outSig[currentClump - clump_0] = -1e4;
                 } else {
                     iws = 1.0/weight_sum;
-                    outVar[currentClump] = var_sum*iws;
-                    outSig[currentClump] = sqrtf(iws);
+                    outVar[currentClump - clump_0] = var_sum*iws;
+                    outSig[currentClump - clump_0] = sqrtf(iws);
                 }
             }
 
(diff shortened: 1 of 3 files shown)


barentine:
docstrings/test included now when you're ready to review @David-Baddeley

David-Baddeley:
The deClump stuff is used in a bunch of non-multiview places, so I'm a little nervous about this and want to make sure I fully understand it. Juggling a bunch of stuff the next couple of days, don't be afraid to prod me again later in the week.

barentine:
Thanks @David-Baddeley . Completely understand, but would like to note that this only affects aggregateMean/Min/Sum/WeightedMean, and as such appears to only be called in the multiview code (and maybe Ben's PR). 

Hopefully the unit test makes it pretty clear how broken it is - current head will fail at the length check as it'll turn the two clumps into three.

David-Baddeley:
Ok, just had a quick look after all - problem in issue #493 seems to be 2/3-fold (although I'm not sure the PR fully fixes it) in  would seem to be a combination of

a) the output is not initialised (can be fixed by uncommenting the loop over the output on lines 572-576, using `memset()`, or using `PyArray_FILLWBYTE()`)
b) we average all the clumpIndex=0 entries, which might give something weird
c) the code relies on everything being sorted by clumpIndex

The uninitialised output works if (and only if) we have a complete set of clump indices (no missing indices, incl. 0), but is not super safe.

Starting `currentClump` at -1 is deliberate - it's simply an arbitrary choice which is guaranteed to fail the test on line 585 the first time through the loop. It could be any -ve number / invalid clump ID and it would still work.  The updates on lines 585-595 write the results from the previous clump (which doesn't exist the first time we hit the loop).

Even though it would initially seem reasonable that `len(clumped) == len(np.unique(clumpIndex))`, I don't think it should necessarily be true. The way all the deClump stuff is written satisfies the following condition instead e.g. `clumped[idx] == mean(unclumped[clumpIndex==idx])`. If you have a situation where you don't have a contiguous range of clump indices (e.g. an empty clump 0, or data which has been filtered after the clump indices have been assigned) `NClumps = len(np.unique(clumpIndex))` is an unsafe choice - not only will you not be able to access clumps by their index, but more importantly you'll end up writing off the end of your allocated array into memory you don't own. Using the same reasoning I don't think that having a `clump0` offset is particularly useful. My preference would be for the c declumping routines to stay as they are, albeit with output initialisation fixed.

That's not to say that this is super intuitive, and it might be reasonable to offer the option to just keep all the non-trivial clumps and abandon index continuity in a subsequent step (this can pretty easily be accomplished with, e.g., a tabular filter).

barentine:
Good catch that having non-contiguous clump indices will break everything. 

Do you mind if I just rewrite the aggregates? Given that it's sorted could just do `np.unique(clumpIndex, return_counts=True)`, pass the counts with assigned and do a nested for. I think it'd be more readable / robust.

David-Baddeley:
It's used / was written for the standard `Localisations.MergeClumps` recipe module used for event chaining, and you are going to have a hard time convincing me to change the behaviour there - it's a good, fast, solution. I'm not super keen on duplicating it when un-necessary.

it would, however, be super simple to call it as:

```
max_clumpID = int(np.max(assigned) + 1)
unique_clumpIDs = np.unique(assigned)

var = deClump.aggregateMean(max_clumpID, ...)[unique_clumpIDs]

```

in the multi-view recipe modules.

barentine:
Good catch, sorry, saw the calls only in `multiview.coalesce_dict_sorted` and had forgotten the other was calling that.

I'd be happy is you PR/merged `deClump.aggregateMean(max_clumpID, ...)[unique_clumpIDs]` into master immediately, as you're right this is a simple fix. 

As you mention, merging everything in clumpIndex 0 is weird, and it looks like that currently happens in `localisations.MergeClumps` too  (assumption being as you mentioned the events are linked by  `deClump.findClumpsN` which starts clumpIndex at 1). I think we could do a clean rewrite that'd fix both _but_ would love the quick fix you mention, and don't expect the performance of a slice on each key to bother us in the short/medium term

David-Baddeley:
I think this is now superseded by #511 and #514

PR #506, opened 2020-10-12
Suggested lint fix

zacsimile:
Addresses issue with linting.

Potentially a bugfix

**Proposed changes:**
- Use `origin/master` to resolve ambiguity
- Use `..` diff notation (re: https://docs.github.com/en/free-pro-team@latest/github/collaborating-with-issues-and-pull-requests/about-comparing-branches-in-pull-requests) (I'm not sure we need this)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,7 +17,6 @@ jobs:
       run: pip install flake8 lintly
     - name: Lint with flake8
       run: |
-        git fetch origin master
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I've tried origin/master a couple of times already ... think it's unlikely to be the problem.

zacsimile:
I know, but I keep seeing it as the correct solution everywhere I look. Thought it might be worth revisiting now that you've got the pull request sha down. Another option might be `HEAD` instead of `origin/master`, but I know you've tried that, too. What's the output for `git remote -v` on the CI server?

David-Baddeley:
I think this is now redundant

PR #509, opened 2020-10-12
fix pipelineify/AddPipelineDerivedVars for splitter-fit data

barentine:
Addresses issue I missed the `self.` when migrating from `pipeline.py` code in #284 

bugfix
**Proposed changes:**
- don't treat mdh as an attribute of a recipe module


barentine:
proposed fix to #510

PR #512, opened 2020-10-13
Declump init

David-Baddeley:
Addresses issue # 473.

fix array initialisation in deClump c functions.

### DIFF
--- a/PYME/Analysis/points/DeClump/deClump.c
+++ b/PYME/Analysis/points/DeClump/deClump.c
@@ -565,6 +565,10 @@ static PyObject * aggregateWeightedMean(PyObject *self, PyObject *args, PyObject
         return NULL;
     }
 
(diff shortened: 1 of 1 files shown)


PYME/Analysis/points/DeClump/deClump.c
```diff
@@ -977,6 +989,10 @@ static PyObject * aggregateSum(PyObject *self, PyObject *args, PyObject *keywds)
         return NULL;
     }
 
+        //initialise memory in output arrays
+    PyArray_FILLWBYTE(outVarA, 0);
+    //PyArray_FILLWBYTE(outSigA, 0);
```
barentine, line 994: no reason to add the commented out lines (and would suggest taking them out), but otherwise I'd say go ahead and merge @David-Baddeley

PR #514, opened 2020-10-13
Polishing multiview.merge_clumps / coalesce_dict_sorted

barentine:
Addresses issue small remaining part of #493 .

bugfix
**Proposed changes:**
- crop valid out of weights too if non_trivial flag is set
- numpy docstrings
- dodge deprecation warning by using DictSource instead of initializing MappingFilter with a dict
- add test_multiview from #501 as its not completely degenerate with test_clumps, will fill more in later, and displays a current issue with the head(remaining even after this PR)
```
        merged = multiview.merge_clumps(unmerged, 0, 'cluster_id')
    
        # test len
>       assert len(merged['cluster_id']) == len(np.unique(unmerged['cluster_id']))
E       assert 1 == 2
E        +  where 1 = len(array([1.], dtype=float32))
E        +  and   2 = len(array([0, 1]))
E        +    where array([0, 1]) = <function unique at 0x7fa52a501158>(array([0, 0, 0, 0, 0, 1, 1, 1, 1, 1]))
E        +      where <function unique at 0x7fa52a501158> = np.unique

tests/PYME/Analysis/points/test_multiview.py:24: AssertionError
```

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

### DIFF
--- a/PYME/Analysis/points/multiview.py
+++ b/PYME/Analysis/points/multiview.py
@@ -38,23 +38,38 @@ def load_shiftmap(uri):
 def coalesce_dict_sorted(inD, assigned, keys, weights_by_key, discard_trivial=False):  # , notKosher=None):
     """
     Agregates clumps to a single event
-    Note that this will evaluate the lazy pipeline events and add them into the dict as an array, not a code
-    object.
-    Also note that copying a large dictionary can be rather slow, and a structured ndarray approach may be preferable.
-    DB - we should never have a 'large' dictionary (ie there will only ever be a handful of keys)
 
-    Args:
-        inD: input dictionary containing fit results
-        assigned: clump assignments to be coalesced
-        keys: list whose elements are strings corresponding to keys to be copied from the input to output dictionaries
-        weights_by_key: dictionary of weights.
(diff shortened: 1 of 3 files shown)


barentine:
Note, forgot to adjust this test for 0 cluster index being removed by [non_trivial]. will hit that after lunch, along with all these formatting violations

PYME/Analysis/points/multiview.py
```diff
@@ -35,31 +35,48 @@ def load_shiftmap(uri):
 
     return shift_map
 
-def coalesce_dict_sorted(inD, assigned, keys, weights_by_key, discard_trivial=False):  # , notKosher=None):
+
+def coalesce_dict_sorted(inD, assigned, keys, weights_by_key,
```
barentine: would be cool @David-Baddeley if we could get these to 'suggest' so we can just commit straight for github if we're OK with them

David-Baddeley: Yep, might also see if I can disable some of the more pedantic ones (e.g. line length, extra lines etc ...)

Issue #490, opened 2020-10-08
[BUG] RuleServer counting has potential to get bad if tasks fail/are reassigned


barentine:
**Describe the bug**
![image](https://user-images.githubusercontent.com/31105780/95499663-a3448d80-0973-11eb-9e59-7f37fbd77cb1.png)

**To Reproduce**
1. run a recipe or localization with a couple tasks which will fail
2. wait long enough for the failed tasks to get reassigned
3. if you're sufficiently unlucky you'll see some bad counting

**Expected behavior**
maybe not automatically requeue failed tasks - just let them fail and maybe we add an interface for quickly requeueing them from clusterUI if a user would like. That said, if we keep the requeue stuff, we should also sort out the counting - probably a locking issue.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] ubuntu 14.04/18.04
 - Python version: 3.6 / 3.7
 - numpy version: 1.14 / 1.19

David-Baddeley:
Bug is in `IntegerIDRule.mark_complete()` and is that we don't check what the current status is before updating `nCompleted`, `nFailed`, and `nAssigned`. This gets triggered by the following flow:

- send out a task
- task times out and is re-queued 
- original task is completed and returned (decrementing nAssigned, despite the fact that it has been re-assigned)
- requeued task is completed and returned (decrementing nAssigned again)

This is a symptom of the timeout being too short. With current code we at least get some indication that something is going wrong. Dropping the retries just masks this.

There are various nasty corollaries where the behaviour will result in truncated analysis (we call the rule done when `nAssigned <1`). I'll try and fix the accounting so that it reports things more reasonably - we might need to add a field or two to a) report how many timed out and b) how many returned after timeout.

barentine:
Sounds good. Thanks @David-Baddeley

barentine:
I ran the updated ruleserver last night with some templated from-cluster recipes using inputsByTask. Will note :  
- I am suspicious we are ~assigning an entire rule to a single nodeserver rather than distributing recipe tasks more like localization tasks (was suspicious of this a couple weeks ago, didn't immediately see why that was happening when i looked at the code). Maybe we make a separate issue on that.
[...]

PR #518, opened 2020-10-15
Change default retry behavior on rules -> no retries unless specified

barentine:
Addresses issue rarely is this helpful on a dedicated cluster, or clusterofone setup, and can be configured when needed on ad-hocs. Probably eases #490.

?
**Proposed changes:**
- change default number of retries to 0, i.e. try once and if a task times out mark it as failed.

**Checklist:**

- [ ] Tested 

### DIFF
--- a/PYME/cluster/ruleserver.py
+++ b/PYME/cluster/ruleserver.py
@@ -166,10 +166,10 @@ class IntegerIDRule(Rule):
         self._template = task_template
         self._task_info = np.zeros(max_task_ID, self.TASK_INFO_DTYPE)
         
-        # number of times to re-queue a task if it times out is set by the 'ruleserver-retries' config option
-        # NOTE: the new default of 0 effectively disables re-trying and makes analysis less robust. TODO - revisit whether this is a good idea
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Hmm - just rechecked the code (see new comments in code), and am now second guessing if this is a good idea to disable. Particularly if it's just to fix a cosmetic issue with the display of tasks.

The important thing is that retrying **doesn't occur on failure** but only on **timeout**. There is a reasonable expectation that retrying will work when the cause is a timeout. The fact that the retries are being triggered also indicates that they are needed. We could certainly reduce the number of retries, but that is unlikely to fix the accounting. Maybe we should fix the accounting instead.

barentine:
I can definitely appreciate that sentiment @David-Baddeley, and for sure I want to fix the accounting #490 , but I'm also concerned that tasks failing with Tracebacks aren't always being marked as failed. I've definitely hit this with recipe tasks anyway. One option could be default to 0 retries and add UI to easily re-queue all the timed out tasks from a rule?

David-Baddeley:
If failing tasks are not marked as failed, I suspect something else is at play. Leaning towards changing the default to 1 retry (stops us from piling on when things aren't keeping up) and fixing the accounting.

PR #521, opened 2020-10-16
get the right multiviewChannel if we don't have all of them present

barentine:
Addresses issue #if you only have one color active, on biplane, say channels 0 and 3, but use the current fold, you get multiviewChannels of 0 and 1, which will mess you up of course if you use a full [0, 1, 2, 3] shiftmap registration, will also mess up probe, etc.

bugfix
**Proposed changes:**
- check if we have `Multiview.ActiveViews` in our metadata, and use it to pull correct channel information.

**notes**
multiviewChannel was a mapping, which is not actually super cheap since we evaluate it later in the same function when we index to get probe. Shouldn't be a performance loss.

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

### DIFF
--- a/PYME/Analysis/points/multiview.py
+++ b/PYME/Analysis/points/multiview.py
@@ -109,7 +109,7 @@ def foldX(datasource, mdh, inject=False, chroma_mappings=False):
     if not inject:
         datasource = tabular.MappingFilter(datasource)
 
-    roiSizeNM = (mdh['Multiview.ROISize'][1]*mdh.voxelsize_nm.x)  # voxelsize is in um
+    roiSizeNM = (mdh['Multiview.ROISize'][1]*mdh.voxelsize_nm.x)
 
     numChans = mdh.getOrDefault('Multiview.NumROIs', 1)
     color_chans = np.array(mdh.getOrDefault('Multiview.ChannelColor', np.zeros(numChans, 'i'))).astype('i')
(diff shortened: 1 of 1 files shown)


PYME/Analysis/points/multiview.py
```diff
@@ -109,16 +109,21 @@ def foldX(datasource, mdh, inject=False, chroma_mappings=False):
...
 
     numChans = mdh.getOrDefault('Multiview.NumROIs', 1)
     color_chans = np.array(mdh.getOrDefault('Multiview.ChannelColor', np.zeros(numChans, 'i'))).astype('i')
 
     datasource.addVariable('roiSizeNM', roiSizeNM)
     datasource.addVariable('numChannels', numChans)
 
-    #FIXME - cast to int should probably happen when we use multiViewChannel, not here (because we might have saved and reloaded in between)
-    datasource.setMapping('multiviewChannel', 'clip(floor(x/roiSizeNM), 0, numChannels - 1).astype(int)')
+    active_rois = np.asarray(mdh.getOrDefault('Multiview.ActiveViews',
```
David-Baddeley, line 121:
```suggestion
    # NB - this assumes that 'Multiview.ActiveViews' is sorted the same way that the views are concatenated (probably a safe assumption)
    active_rois = np.asarray(mdh.getOrDefault('Multiview.ActiveViews', 
```

PR #522, opened 2020-10-16
Marching Tetrahedra Updates

### LINKED ISSUE #241: Generate isosurface from Delaunay tessellation is broken
I'm almost positive this one's entirely my fault. I need to revisit it at some point. 

We should also probably think about the pipeline from a user's perspective. If you select `View > Create Delaunay Tesselation` (sp. tessellation), and then `View > Generate Isosurface from Delaunay Tesselation`, it won't work. It expects the field `dn` (density) from `Estimate density` or `Estimate density (circumcenters)`, which also create, but do not display, a tessellation. It's just a bit confusing.

Also these should probably not be in the `View` menu. I suggest we move most of it to `Analysis > Density` and the isosurface generation from Delaunay tessellation to `Mesh`.

zacsimile:
Addresses issue #241.

Both

**Proposed changes:**
- Update calls in `DelaunayMarchingTetrahedra` (formerly `MarchingTetrahedra`, sorry, I thought the rename made things clearer, I know I'm not supposed to do this) recipe to be modern
- Make `View>Generate Isosurface from Delaunay Tesselation` auto-call `DelaunayTesselation` and `LocalPointDensity`
- Implement a raster version of MarchingTetrahedra as a double-check
- Add generation of test data at command line launch to `PYMEVis` (`pymevis -t`) for debugging purposes

All of this was to fix #241. 

Here is a Death Star (raster MT on a sphere):

image

<details>
<summary>death star code from jupyter notebook</summary>
</pre>
from PYME.LMVis import VisGUI

%gui wx

pymevis = VisGUI.ipython_pymevisualize()
pipeline = pymevis.pipeline

from PYME.experimental.marching_tetrahedra import RasterMarchingTetrahedra
from PYME.experimental import marching_cubes

S = marching_cubes.generate_sphere_image(8)

mt = RasterMarchingTetrahedra(S)

tris = mt.march(True)

import PYME.experimental._triangle_mesh as triangle_mesh
from PYME.LMVis.layers.mesh import TriangleRenderLayer

mesh = triangle_mesh.TriangleMesh.from_np_stl(tris)
pipeline.recipe.namespace['mt0'] = mesh
layer = TriangleRenderLayer(pipeline, dsname='mt0', method='flat', cmap='hot')
pymevis.add_layer(layer)
</pre>
</details>

MT on a Delaunay triangulation:

image

Generated with `pymevis -t` followed by `View>Generate Isosurface from Delaunay Tesselation`.

**Checklist:**

- [ ] Tested with numpy=1.14


PYME/LMVis/visCore.py
```diff
@@ -426,8 +426,51 @@ def add_layer(self, layer):
         self.glCanvas.refresh()
     
         self.layer_added.send(self)
-
     
+    def generate_test_object_layer(self, obj_type):
```
David-Baddeley: As mentioned, not super keen on having test code cluttering this file which is already pretty unwieldy.

PYME/LMVis/visCore.py
```diff
@@ -426,8 +426,51 @@ def add_layer(self, layer):
...
+        points = DictSource({'x':x,'y':y,'z':z,'t':t})
+        points.mdh = CachingMDHandler({'voxelsize.x': vx, 'voxelsize.y': vy, 'voxelsize.z': vz,
+                                   'voxelsize.units': 'um'})
+
+        print('Creating Pipeline')
+        self.pipeline.OpenFile(ds=points)
+        print('Pipeline Created')
+        
+        #############################
+        #now do all the gui stuff
```
David-Baddeley: Why is this needed - surely we should be able to just hook the standard open and launch code ...

PYME/LMVis/visCore.py
zacsimile: I think I wrote this before we had that conversation. Since this is not critical to...anything, I will probably just delete the ability to add test objects.

PYME/LMVis/visCore.py
zacsimile: I was having trouble hooking the standard open/launch code, which I assume is `visCore.VisGUICore.OpenFile()`. This `now do all the gui stuff` I copied from `visCore.VisGUICore.OpenFile()`. While `pipeline.Pipeline.OpenFile()` allows you to pass a data source as an argument, `visCore.VisGUICore.OpenFile()` is presently limited to files. Perhaps we should add a `ds` flag and/or `**kwargs` and associated handling to `visCore.VisGUICore.OpenFile()`. Or is there another function I missed?

PYME/LMVis/visCore.py
```diff
@@ -729,11 +728,19 @@ def _populate_open_args(self, filename):
...
+        if ds is not None:
+            if filename is None:
+                filename = str(ds)
+            else:
+                # File supercedes datasource
+                ds = None
 
         print('Creating Pipeline')
-        self.pipeline.OpenFile(filename, **args)
+        self.pipeline.OpenFile(filename=filename, ds=ds, **args)
```
```suggestion
       if filename is None and not ds is None:
            self.pipeline.OpenFile(ds=ds)
       else:
           args = self._populate_open_args(filename)
           self.pipeline.OpenFile(filename, **args)
```

PR #530, opened 2020-10-20
Rule refactor 2

David-Baddeley:
Addresses lack of rule chaining support in current rule-pusher and seeks to create a clearer interface and reduce duplication between different types of rule. Still needs a bunch of work (v. incomplete, but putting up to facilitate discussion)

### Open questions

- ~~should there be a separate pusher, or a `.push()` method on the rule?~~ - `.push()` probably makes he most sense
- how to best accomplish templating of input and output names in chained recipes
- specifically what happens on re-analysis where the results names can't be easily inferred from the original data name
  (current strategy here would be to have the first rule in the chain re-write/inject some variables in the context before passing
  passing it to chained recipes, but need to work out how to do this in a clean and intuitive way.
- should analysis metadata be a property of the `LocalisationRuleFactory`?

Example usage

```python
step1 = LocalisationRuleFactory(analysisMetadata=mdh)

step2 = RecipeRuleFactory(recipeURI='PYME-CLUSTER///RECIPES/render_image.yaml', 
                          input_patterns={'input':'{{spool_dir}}/analysis/{{series_stub}}.h5r'})
step1.chain(step2)

step3 = RecipeRuleFactory(recipeURI='PYME-CLUSTER///RECIPES/measure_blobs.yaml', 
                          input_patterns={'input':'{{spool_dir}}/analysis/{{series_stub}}.tif'})
step2.chain(step3)
```

or:

```python
step3 = RecipeRuleFactory(recipeURI='PYME-CLUSTER///RECIPES/measure_blobs.yaml', 
                          input_patterns={'input':'{{spool_dir}}/analysis/{{series_stub}}.tif'})

step2 = RecipeRuleFactory(recipeURI='PYME-CLUSTER///RECIPES/render_image.yaml', 
                          input_patterns={'input':'{{spool_dir}}/analysis/{{series_stub}}.h5r'},
                          on_completion=step3)

step1 = LocalisationRuleFactory(analysisMetadata=mdh, on_completion=step2)
```

then:
```python
def on_launch_analysis(context={'spool_dir': ..., 'series_stub': ...}):
    step1.get_rule(context=context).push()
```

### DIFF
--- a/PYME/cluster/rules.py
+++ b/PYME/cluster/rules.py
@@ -150,10 +150,10 @@ class Rule(object):
     
     def _output_files(self):
         """ Return a dictionary of output cluster URIs.
-        The principle output (if it makes sense to define one) should be under the 'results' key. Each entry should be a list
-        of outputs (or a cluster glob):
+        The principle output (if it makes sense to define one) should be under the 'results' key. Each entry should be a
+        single output as this gets substituted into the recipe before submission (not in the ruleserver):
         
-        For a single output, the dictionary should look like: {'results' : [results_URI,]}
+        For a single output, the dictionary should look like: {'results' : results_URI}
         
         Over-ride in derived rule classes
         """
@@ -356,10 +356,10 @@ class RecipeRule(Rule):
         # over-ride here because we need to add input info
         
         # try (in order of increasing precedence):
-        # - output of previous rule in the chain
+        #
         # - an `input_templates` variable in the context
         # - hard coded `inputs` in the context
-        inputs = context.get('rule_outputs', {})
+        inputs = {} # = context.get('rule_outputs', {})
         input_templates = context.get('input_templates', None)
         if input_templates is not None:
             inputs.update({k: v.format(context) for k, v in input_templates.items()})
@@ -367,23 +367,27 @@ class RecipeRule(Rule):
         inputs.update(context.get('inputs', {}))
 
         if not inputs:
-            raise RuntimeError('No inputs found, one of "inputs", "input_templates", or "rule_outputs" must be present in context')
-        
-        input_names = inputs.keys()
-        inputs = {k: inputs[k] if isinstance(inputs[k], list) else clusterIO.cglob(inputs[k], include_scheme=True) for k
-                  in input_names}
-
-        self._num_recipe_tasks = len(list(inputs.values())[0])
-
-        logger.debug('numTotalFrames = %d' % self._num_recipe_tasks)
-        logger.debug('inputs = %s' % inputs)
-
-        inputs_by_task = {frameNum: {k: inputs[k][frameNum] for k in inputs.keys()} for frameNum in
-                          range(self._num_recipe_tasks)}
-        
-        
+            if not context.get('rule_outputs', False):
+                raise RuntimeError('No inputs found, one of "inputs", "input_templates", or "rule_outputs" must be present in context')
+            inputs_by_task = None
+        else:
+            input_names = inputs.keys()
+            inputs = {k: inputs[k] if isinstance(inputs[k], list) else clusterIO.cglob(inputs[k], include_scheme=True) for k
+                      in input_names}
+    
+            self._num_recipe_tasks = len(list(inputs.values())[0])
+    
+            logger.debug('numTotalFrames = %d' % self._num_recipe_tasks)
+            logger.debug('inputs = %s' % inputs)
+    
+            inputs_by_task = {frameNum: {k: inputs[k][frameNum] for k in inputs.keys()} for frameNum in
+                              range(self._num_recipe_tasks)}
+            
+            
         rule = Rule._populate_rule(self, context, on_completion=on_completion)
-        rule['inputsByTask'] = inputs_by_task
+        
+        if inputs_by_task:
+            rule['inputsByTask'] = inputs_by_task
         
         return rule
     
@@ -407,6 +411,11 @@ class RecipeRule(Rule):
             task = task % ('"taskdefRef" : "%s"' % self.recipeURI, output_dir_n)
         else:
             task = task % ('"taskdef" : {"recipe": "%s"}' % self.recipe_text, output_dir_n)
+            
+        #if we are a chained rule, hard-code inputs
+        rule_outputs = context.get('rule_outputs', None)
+        if rule_outputs:
+            task.replace('{{taskInputs}}', json.dumps(rule_outputs))
         
         return task
 
@@ -442,7 +451,7 @@ class LocalisationRule(Rule):
         Rule.__init__(self, **kwargs)
         
     def _output_files(self):
-        return {'results' : [self.resultsURI,]}
+        return {'results' : self.resultsURI}
     
     def _setup(self, dataSourceID, metadata, resultsFilename, startAt=10, dataSourceModule=None, serverfilter=clusterIO.local_serverfilter):
         self.dataSourceID = dataSourceID
--- a/PYME/cluster/rules.py
+++ b/PYME/cluster/rules.py
@@ -278,6 +278,20 @@ class Rule(object):
         else:
             logging.error('Failed on releasing tasks with status code: %d' % r.status_code)
 
+    def _mark_complete(self):
+        """ Thin wrapper around release_rule_tasks api endpoint"""
+        from PYME.IO import clusterIO
+        s = clusterIO._getSession(self.taskQueueURI)
+        r = s.get('%s/mark_release_complete?ruleID=%s' % (
+            self.taskQueueURI, self._ruleID),
+                  data='',
+                  headers={'Content-Type': 'application/json'})
+    
+        if r.status_code == 200 and r.json()['ok']:
+            logging.debug('Successfully marked rule as complete')
+        else:
+            logging.error('Failed to mark rule complete with status code: %d' % r.status_code)
+
     def _poll_loop(self):
         logging.debug('task pusher poll loop started')
         # wait until clusterIO caches clear to avoid replicating the results file.
@@ -292,7 +306,9 @@ class Rule(object):
                 pass
         
             if self.complete:
-                logging.debug('all tasks pushed, ending loop.')
+                logging.debug('input data complete and all tasks pushed, marking rule as complete')
+                self._mark_complete()
+                logging.debug('ending polling loop.')
                 self.doPoll = False
             else:
                 time.sleep(1)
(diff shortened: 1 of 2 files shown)


David-Baddeley:
### Extra open questions

- should we add a `TemplatedRecipeRule`?

barentine:
Awesome - thanks for sinking some time into this @David-Baddeley.

As to the open questions:

I'll note that in April, one of the few design parameters we had in mind was that the rules should push themselves. That's not to say we were right in that, but worth double checking. I think it would be nice

I'd vote against the TemplatedRecipeRule. I agree that at the moment, we only need templating on recipes, and I think that's the simplest place to do it. If we do `class UserFloat(Float, UserParameter)` where 
```
class UserParameter(object):
    pass
```
add a`user_parameters` property to `ModuleBase`, and `ModuleCollection`, then we have everything we need to find parameters that should be editable/linkable by a user in a GUI/web view. I think this has the advantage that default parameters are still present, and makes recipe yamls more reusable. Disadvantage is that this is different than the USERFILE templating, which just does a str substitution, but I think this _needs_ to be different than str substitution if we want to maintain the hard-typing of traits for recipe parameters. There are flaws here in that this requires us to pick which parameters are sufficiently fluid to deserve space in views for them / templating. These are at some level separate issues. Maybe the real solution to templating is if we have a context dictionary where keys are `<module name>-<number for breaking module name degeneracy>-<parameter name>` and we add a trait to `ModuleBase` which is an integer for breaking module name degeneracy, defaulting to zero.

To me, a list of rules which takes care of the on_completion linking and posts _looks_ more intuitive than the walk-through you include in the docs at the top. I wonder if there's a way we could get the best of both worlds here and make e.g. a RuleChainFactory which has it's inputs/outputs unset, maybe context (for templating) unset, but will do the linking/nesting for you when the time comes? Maybe even have a post method which instantiates, starts, and returns a `RulePusher`? It's hard to look at some of our unanswered questions here and think that we're really cruising towards a solution which makes `_rules` look bad - if we haven't solved the problem of linking input/outputs in a chain, and the rules don't push themselves, then I feel like we're missing the two main points of this refactor.

David-Baddeley:
The proposed solution *is* nothing different from a list of rules/rule factories. It might be more intuitive coming from a c background, but under the hood, lists are implemented as

```c
struct list_item
{
    // list specific variables
    int foo;
    char[10] bar;

    //pointer to next list element
    list_item * next;
};
```

with the list variable simply being the first element. As it stands, `step1.get_rule()` constructs the whole chain, without needing to do any re-writing of parameters etc ...

David-Baddeley:
I think the templating details are best discussed elsewhere, but I'd be very, very reluctant to do anything other than string based template substitution for recipe templates.

David-Baddeley:
Because the .json representation of rule-chaining is a hierarchical linked list, I think there is a good argument for keeping that in our python mapping of it. It also puts the info directly where it is needed. A rule needs to know what it's `on_completion` is. I could potentially see a case for a utility `chain()` method which took a list of RuleFactories and chained them together, but I'm not sure if there is really a use case.

David-Baddeley:
### Current status:

- tested as working for localisation rules
- chaining a localisation and a recipe rule would seem to work on surface (generates a rule with a suitable structure), but on_completion rule never triggered. Not sure if this is a bug on the cluster side. Retry numbers look a bit off.
- recipe output file inference missing, so chaining 2 recipe rules requires either hard-coded or templated input filenames.

PYME/cluster/rules.py
```diff
@@ -0,0 +1,616 @@
...
+        Over-ridden in rules where all the data is not gauranteed to be present when the rule is created
+
+        Returns
+        -------
+
+        release_start, release_end : the indices of starting and ending tasks to release
+
+        """
+        numTotalFrames = self.ds.getNumSlices()
+        logging.debug('numTotalFrames: %s, currentFrameNum: %d' % (numTotalFrames, self.currentFrameNum))
```
barentine: @David-Baddeley have a look at _rules for a 0-indexed version of all this which ought to fix taskCompleted == (tasksPosted - 1)

startAt is zero indexed, as is the task array we use on the ruleserver, all of these +1 and -1 can (and should) go

David-Baddeley: I'm not sure I follow - most of these +1s are needed  and correct for **0-based** indexing. e.g. 

- have a total of 15 frames, all pushed => `numTotalFrames == 15`, `currentFrameNum ==14` ( ie `nFrames = frameNum +1`) which makes line 542 correct
- ditto for line 454 - 15 frames makes the frameNum for the last frame 14.
- and for L 548 - if your current frame number is 14 then you have released 15 frames.
- finally for the return - these are the values to pass to the release call, which uses slice notation - i.e. the last frame of the range is excluded

barentine: I see you got it sorted in the latest commit. Thanks David. Anything holding this up before merging?

David-Baddeley:
As of last night, the series completes and adds a chained analysis rule, but the chained rule doesn't run. Part of this might be that we're setting inputs via `inputsByTask` (not supported in chained rules) but not 100% sure. Having the chained recipe actually run is (for me) the point where this actually represents an advance over the status quo, and it feels like it's soo close.

David-Baddeley:
More on status:

```python
>>> from PYME.cluster import rules
>>> from PYME.IO import MetaDataHandler
>>> md = MetaDataHandler.NestedClassMDHandler()
>>> # NOTE: we are cheating with empty analysis metadata here as the series was spooled and analysed in real time using
>>> # the old framework, so analysis metadata is already present in the .h5
>>> r1 = rules.LocalisationRuleFactory(analysisMetadata=md)
>>> rec_text = '''- localisations.DensityMapping:
    inputLocalizations: results
    outputImage: rendered
    renderingModule: Histogram
- output.ImageOutput:
    filePattern: '{output_dir}/{file_stub}.tif'
    inputName: rendered
    scheme: pyme-cluster://'''
>>> r2 = rules.RecipeRuleFactory(recipe=rec_text)
>>> r1.chain(r2)
>>> r = r1.get_rule(dict(seriesName='PYME-CLUSTER:///david/2020_10_16/16_10_series_A.h5'))
DEBUG:PYME.cluster.rules:DataSource.__class__: <class 'PYME.IO.DataSources.ClusterPZFDataSource.DataSource'>
DEBUG:PYME.cluster.rules:numTotalFrames = 1
DEBUG:PYME.cluster.rules:inputs = {'results': ['PYME-CLUSTER://db3/david/2020_10_16/analysis/16_10_series_A_12.h5r']}
>>> r.push()
(NSInfo(name=u'PYMERuleServer: db3', address='\x7f\x00\x00\x01', port=55178, creation_time=1603772110.105406, URI=u''), '\x7f\x00\x00\x01')
DEBUG:PYME.cluster.rules:local_queues: [u'PYMERuleServer: db3']
DEBUG:root:resultsURI: http://127.0.0.1:55183/__aggregate_h5r/david/2020_10_16/analysis/16_10_series_A_12.h5r
DEBUG:root:fileFormattedResults - URI: http://127.0.0.1:55183/__aggregate_h5r/david/2020_10_16/analysis/16_10_series_A_12.h5r/MetaData
WARNING:root:No handler for data type found, using pickle
DEBUG:root:fileFormattedResults - URI: http://127.0.0.1:55183/__aggregate_h5r/david/2020_10_16/analysis/16_10_series_A_12.h5r/Events
'http://127.0.0.1:55183/david/2020_10_16/analysis/16_10_series_A_12.h5r.json'
DEBUG:PYME.cluster.rules:Successfully created rule
DEBUG:root:task pusher poll loop started
DEBUG:root:numTotalFrames: 2094, _next_release_start: 10
DEBUG:root:we have unpublished frames - push them
DEBUG:root:Successfully released tasks (10:2094)
DEBUG:root:input data complete and all tasks pushed, marking rule as complete
DEBUG:root:Successfully marked rule as complete
DEBUG:root:ending polling loop.
```
results in the series being successfully analysed **and** the follow on recipe task created, but the single task for that recipe not being assigned. This can be traced back to an error whilst rating in the rulenodeserver:
```
Exception in thread Thread-4:
  File "/Users/david/anaconda/python.app/Contents/lib/python2.7/threading.py", line 801, in __bootstrap_inner
    self.run()
  File "/Users/david/anaconda/python.app/Contents/lib/python2.7/threading.py", line 754, in run
    self.__target(*self.__args, **self.__kwargs)
  File "/Users/david/python-microscopy/PYME/cluster/rulenodeserver.py", line 344, in _poll_tasks
    self._update_tasks()
  File "/Users/david/python-microscopy/PYME/cluster/rulenodeserver.py", line 215, in _update_tasks
    taskID, cost = rater.get_local()
  File "/Users/david/python-microscopy/PYME/cluster/rulenodeserver.py", line 113, in get_local
    taskID, cost = self.next()
  File "/Users/david/python-microscopy/PYME/cluster/rulenodeserver.py", line 107, in next
    return self.__next__()
  File "/Users/david/python-microscopy/PYME/cluster/rulenodeserver.py", line 78, in __next__
    task = json.loads(filled_template)
ValueError: Expected object or value
```
The most likely reason for this is that (because this is a chained task that doesn't support `inputsByTask`) the template still has an un-substituted `{{taskInputs}}` slot. We need to either modify the `RecipeRule` so that is both usable with the standard `inputsByTask` and with a hard-coded `'input'` section if chained.

barentine:
In `_rules` I hard-coded the `input` section following the docs in `ruleserver` on rule chaining

> Follow on / chained rules are slightly more restricted than standard rules in that recipe "inputs" must be hardcoded (no ``{{taskInputs}}`` substitution).

Doesn't mean we can't or shouldn't work in the substitution, but I think the use case is much less critical (ie I'm not waiting on it ;) )

David-Baddeley:
Hard-coding the inputs directly would prevent it from being a drop-in for the current `RecipeRulePusher` and working for, e.g. queueing recipes from clusterUI. Handling both options is probably a fairly simple modification of `RecipeRule._populate_rule()` and `RecipeRule._task_template()`.

PYME/cluster/rules.py
```diff
@@ -0,0 +1,627 @@
...
+        
+        inputs.update(context.get('inputs', {}))
+
+        if not inputs:
+            if not context.get('rule_outputs', False):
+                raise RuntimeError('No inputs found, one of "inputs", "input_templates", or "rule_outputs" must be present in context')
+            inputs_by_task = None
+        else:
+            input_names = inputs.keys()
+            inputs = {k: inputs[k] if isinstance(inputs[k], list) else clusterIO.cglob(inputs[k], include_scheme=True) for k
```
barentine, line 375: annoying to have to look through multiple functions to find a required init parameter of recipe rule due to the kwarg passing. We should at least document it in the __init__ of recipe rule. `include_scheme` isn't a kwarg of clusterIO.cglob, so at the moment you are hosed if inputs[k] isn't a list.

barentine: working on a PR

Issue #525, opened 2020-10-19
[BUG] Can't re-run a recipe in PYMEVis after it crashes


zacsimile:
**Describe the bug**
If I run a recipe with a weird set of parameters and it crashes, e.g. I make a mesh at the wrong density that has no triangles, I get an error traceback in the context manager. If I then try to run the recipe (e.g. meshing) again, but change the parameters to something that works, the recipe still uses the previous, busted parameters. I have to restart `PYMEVis` to change the parameters. This appears to be consistent across recipes. 

**To Reproduce**
1. Make a recipe accessed from the PYMEVis menu crash
2. Try to re-run the recipe from the menu with new parameters

**Expected behavior**
The second run should use the updated, rather than the original, parameters. 

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: Win10, macOS 15
 - Python version: 3.6.10
 - numpy version: 1.16.6
 - wxpython version [if a UI bug]: 4.0.4

David-Baddeley:
Could you provide a bit more detail (e.g. an example traceback / specific scenario).

In general, if you call something from the menu which adds a recipe module a second time there's a good chance it won't work (it will try and add the recipe module a second time and both modules will be present in the recipe). This is something we should fix - potentially by keeping track of when we add a recipe module in the menu and simply calling `configure_traits()` on the added one rather than trying to re-add - although this is context specific.

The other thing that might be worth considering in the long term is making functions called from the menu transactional (i.e. able to be un-rolled on failure).

zacsimile:
Sure! So, I think another common use case might a user accidentally does something that makes no sense, realizes this, and then wants to keep working on their data. For example, if I open a single-channel image and then navigate to `Corrections>Multiview>Fold Channels` I get

```
 Error whilst running Corrections>Multiview>Fold Channels
==============================================
python-microscopy=20.10.05
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/Extras/multiviewMapping.py", line 128, in OnFold
    recipe.execute()
[...]

PR #531, opened 2020-10-20
add transactional recipe module addition

David-Baddeley:
Addresses issue #525 .

Bit of both. 

**Proposed changes:**

Add a backend giving a transactional way of adding recipe modules. Partial solution to #525 (which also requires plugins to be modified to use this)

### DIFF
--- a/PYME/LMVis/Extras/clusterAnalysis.py
+++ b/PYME/LMVis/Extras/clusterAnalysis.py
@@ -90,8 +90,7 @@ class ClusterAnalyser:
 
         clumper = localisations.DBSCANClustering(recipe, inputName=pipeline.selectedDataSourceKey,outputName='dbscanClumped')
         if clumper.configure_traits(kind='modal'):
-            recipe.add_module(clumper)
-            recipe.execute()
+            recipe.add_modules_and_execute([clumper,])
 
             pipeline.selectDataSource(clumper.outputName)
 
--- a/PYME/LMVis/Extras/extra_layers.py
+++ b/PYME/LMVis/Extras/extra_layers.py
@@ -84,8 +84,7 @@ def gen_isosurface(visFr):
     dmc = DualMarchingCubes(recipe, invalidate_parent=False, input=oc_name,output=surf_name)
     
     if dmc.configure_traits(kind='modal'):
-        recipe.add_module(dmc)
-        recipe.execute()
+        recipe.add_modules_and_execute([dmc,])
 
         print('Isosurface generated, adding layer')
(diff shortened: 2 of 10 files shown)


David-Baddeley:
extras now modified too ...

zacsimile:
Confirmed via user testing this works on python 3.6.10/macOS 10.15.7/numpy 1.16.6/wxpython 4.0.4.

David-Baddeley:
@barentine - this hits the multiview shiftmap calibration (specifically the results now don't get saved in the recipe namespace, only to disk). I can't see this being an issue, but would be great if you could have a quick look before I merge. 

I've also changed the calling pattern on some of the other multi-view modules (configure, then add, rather than disable invalidation, add, configure, re-enable invalidation) as a) this is a cleaner pattern and b) avoids having to manually roll-back if a user cancels the dialog. There is a chance that some of the `configure_traits()` calls should be swapped for `edit_no_invalidate()` calls as well.

barentine:
Hi @David-Baddeley thanks for pointing that out. All looks good to me - have not tested it as of yet, but will be sure to tomorrow and will PR any tweaks if necessary. Quick notes:
- I like that we're taking the shiftmap out of the recipe namespace in the calibration
- the calibrate shifts and map astigmatism were both calling `configure_traits` on the `ModuleCollection` rather than an individual modules - that was to allow editing of filters, back before we had the data pipeline view.
- related to the last point, we can almost certainly drop the visframe refresh on new line 395 in multiviewMapping
- I would appreciate us actually deleting the old style disable invalidation, configure, re-enable stuff, rather than just commenting it out - no reason to muddy the file

barentine:
Running into an issue with `pipeline_view_min` missing the 'OK' button
image

barentine:
problem exists with `pipeline_view` as well (unsurprising)

barentine:
Note that the default view, i.e.
`if not calibration_module.configure_traits(kind='modal'):`
works fine

David-Baddeley:
This is roughly how the pipeline view is supposed to work - condensed, no buttons on individual modules. Where are you running into this?

barentine:
would you mind documenting that in `recipes.base.ModuleBase` and `ModuleCollection` @David-Baddeley? As I understand it the `pipeline_view` of a module shouldn't be configured directly in modal (or you won't get an OK / cancel), but is really just there for the `ModuleCollection` configure?

I hit it in `PYME/LMVis/Extras/multiviewMapping.py` and will PR the change

David-Baddeley:
Yep - adding the appropriate docs would be a good idea. The pipeline view
is just there for when you want to configure the whole pipeline in one go
(and the vertical display in PYMEVis).

PR #485, opened 2020-10-07
fix rulenodeserver handling of inputsByTask/task_inputs

barentine:
Addresses issue current ruleserver code doesn't work for making multiple tasks off a single recipe rule.

After fixing bytes/str collision #476, and as mentioned in that PR, you hit a json error 
```
File "/home/ubuntu/miniconda3/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 344, in _poll_tasks
    self._update_tasks()
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 215, in _update_tasks
    taskID, cost = rater.get_local()
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 113, in get_local
    taskID, cost = self.next()
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 107, in next
    return self.__next__()
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 78, in __next__
    task = json.loads(filled_template)
ValueError: Unexpected character in found when decoding object value
```

Even for a nice inputs by task on the ruleserver side (matching the docs there)
```
intputs by task {0: {'input': 'pyme-cluster:///Bergamot/2020_9_28/secret/000/analysis/28_9_series_D.h5r'}, 1: {'input': 'pyme-cluster:///Bergamot/2020_9_28/secret/000/analysis/28_9_series_C.h5r'}, 2: {'input': 'pyme-cluster:///Bergamot/2020_9_28/secret/000/analysis/28_9_series_F.h5r'}, 3: {'input': 'pyme-cluster:///Bergamot/2020_9_28/secret/000/analysis/28_9_series_A.h5r'}, 4: {'input': 'pyme-cluster:///Bergamot/2020_9_28/secret/000/analysis/28_9_series_E.h5r'}, 5: {'input': 'pyme-cluster:///Bergamot/2020_9_28/secret/000/analysis/28_9_series_B.h5r'}}
```
[...]

Issue #473, opened 2020-10-05
[ENHANCEMENT] Non-filename templated recipe parameters


barentine:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when I have a templated recipe that needs a non-file str parameter changed between samples. While this is usually somewhere in the directory and could be grabbed automatically from the URI, that would take a special casing to ignore e.g. `analysis` directory, chunked `000`-type directories, etc. 

![image](https://user-images.githubusercontent.com/31105780/95114046-56617c80-0711-11eb-815a-3a07721fc249.png)

**Describe the solution you'd like**
While a web-UI for full editing would be neat, a stop-gap would be if we special-cased another e.g. `USERSTR` parameter name which was exposed in the same menu but didn't prepend `pyme-cluster:///`.

**Additional context**
Running this automatically from the scope w/chaining isn't a big deal (but requires tweaking the recipe for each sample) but I do find it a bit annoying to upload new recipes with OMERO project/dataset info for each sample.

**Are you prepared to work on it?**
definitely

David-Baddeley:
I wonder if this isn't a case where we should just push through with a webby recipe editor, before we hack things too much.

As it stands, the USERFILE parameters are a hacked work-around for the lack of a web-based editor. We could do a similar hack for, e.g. USERSTRING, but that would mean adding more cruft to the recipe classes. Templating anything other than strings gets super hard to do in an intuitive way, and I'd feel uncomfortable with with putting too much work into something which (at least in theory) is a temporary solution.

One advantage the USERFILE templates currently have over a full web recipe editor is that they effectively pull out the critical parameters of a recipe that a user might want to change, so it would be good to think about how to address that.

barentine:
Hi David, sorry to be brief I agree that a webby recipe editor would be great, and maybe not too horrific (can honestly be text based, for starters). 

One thing on my mind is that the USERFILE parameters are fine, and USERSTRING similarly easy because they are str, and we can just str comprehend them into our recipe templates. Can do the same for other parameters, but we need to be careful how in order not to lose hard typing. Linking, for completeness/later, our brief discussion on this in #530. 

In my dreams, whatever we do here will:
- allow more than just files to be templated (obviously the main point)
- allow easy rule-to-rule linking in PYMEAcquire, such that a set of chained rules producing one output can have that output used as a file input parameter for recipe modules in another set of chained rules. 
[...]

PR #534, opened 2020-10-22
html-recipe-editor

David-Baddeley:
A bit more work on a long-standing project - making an HTML recipe editor.

Still very much a work in progress, but not sure if there is any harm in merging it early.

image

### Remaining issues

- inferring trait types so we can display something more nuanced than a text input
- ~backend issues relating to setting `Float`, `Int` etc ... types with a string (hopefully go away once we use the right type of input)~
- jigna module used under the hood has not seen much recent activity
- adding/removing recipe modules
- changing wiring
- asthetics
- how well/poorly this plays with e.g. django for clusterUI


PYME/recipes/html_recipe_editor.py
```diff
@@ -4,9 +4,18 @@
...
-from jigna.template import Template
+#from jigna.template import Template
+from jigna.vue_template import VueTemplate
+
+import jigna.web_server
+# monkey-patch the js file location
+# The reason we do this is so we can use a more recent version of vuejs. By default, jigna ships with a jigna-vue.js
+# which contains a concatenation of jquery, vue=1.x, and jigna. In recipe_editor.html, we load a modified version of
+# jigna-vue.js, jigna-vue-bare WITHOUT vue which lets us separately load a recent version of vue
+jigna.web_server.JIGNA_JS_FILE='/Users/david/dev/jigna/jigna/js/dist/jigna.js'
```
barentine: was going to suggest just merging, as you mention being a bit agile on this one, but this bit looks not too portable.

What's the status of the current webby editor with the parameters separated from the svg render? Is that a 5 min pop-into clusterUI to take care of the friction points of #473 while we work on this more intuitive/flashier one?

David-Baddeley: The one in the code base at present does't work for non string parameters (sends everything back as string and gets type errors  / train validation errors on the back end). Also well off being able to run nicely with clusterUI. The one in this PR actually works, and is a lot closer to clusterUI integration. There's a reasonably simple fix to that portability issue (at least until jigna updates) which is to drop the required .js in PYME.resources.

David-Baddeley:
### Current status

- running under clusterUI (but not visually integrated nor with logic linked up)
- does right thing for int, float, string parameters
- not tested for Enum, File - UI will be a string input, which may or may not work properly on backend. Need to find a way of detecting these and displaying suitable editors
- also offers a text editor
- asthetically still not great

Plan would be to work on clusterUI integration first, then loop back to Enum/File parameters, then asthetics.

David-Baddeley:
image

Changes to parameters automatically update text, text changes propagated when focus lost.

David-Baddeley:
Getting close to being useful ... Can load, save (cluster aware), and runs under clusterUI
image

@barentine - suspect the next step to make it useful for templated scenarios is to allow you to run stuff with the modified recipe without actually saving it.

David-Baddeley:
image

Getting really close to being usable ... might almost be time to merge and then fix stuff later 

Remaining (known) issues:

- recipe in editor is global (i.e. you can only ever edit one recipe at a time / this will cause issues in a multi-user scenario)
- no enum/file support
- introduces new dependencies which are not in pyme-depends

Instructions:

- Install jigna
- launch clusterUI as `python clusterui_tornado.py` rather than `python manage.py runserver`
- port currently hard-coded at 8889

PYME/resources/web/static/recipe_editor.js
```diff
@@ -0,0 +1,119 @@
...
+           var val = event.target.value;
+           switch(this.type){
+               case 'number':
+                   return parseFloat(val);
+               case 'boolean':
+                   val = event.target.checked;
+                   console.log(val, typeof(val));
+                   return val;
+               default:
+                   return val;
```
barentine, line 37: If I understand this right @David-Baddeley we'll still be able to use File/Enum types, there just won't be any selectors for either of them, correct? (and I guess the enum's will have to be str which I think they usually are)

barentine: I haven't pulled/fiddled with this, since I think I should work on a couple other things first, but would it be crazy to check if val startswith `{USERFILE` and if so call `select_file`?

PYME/cluster/clusterUI/recipes/views.py
```diff
@@ -15,6 +15,15 @@ def recipe_standalone(request):
     """This allows file selection with globs like bakeshop"""
     return render(request, 'recipes/recipe_standalone.html', {'serverfilter' : server_filter})
 
+def recipe_editor(request):
+    """User interface for editing recipes"""
+    return render(request, 'recipes/recipe_editor.html', {'serverfilter' : server_filter})
+
+def recipe_editrun(request):
+    """User interface for editing recipes"""
```
barentine: I'd appreciate the docstrings being different between editrun and editor

barentine:
Hi David, with this barely touching PYME core stuff and it being functional/useful in its currently state I would say go ahead and merge it - I'll use it! and surely report back anything I run into. 

I think it would be great if we could get more docstrings on things, particularly for ~core PYME files like `clusterIO` and `recipes.base`, though for the simplicity of the functions you added there I wouldn't hold up a merge for it. Similarly a lot of the js stuff looks pretty straightforward, but eventually might help for community maintenance if any non-python code is documented especially well (we could do better for some of the c stuff too, but of course that's how documentation goes).

PYME/resources/web/static/recipe_editor.js
David-Baddeley: Yep - anything other than number or boolean gets a string editor. Should be fine, if inconvenient, for `File` and `Enum`. Probably won't work for any of the `List`, `Dict` etc ... traits.

I'd also thought about doing a check for `"USERFILE"`, but would want to know that we can't do things properly for all file inputs instead first. Suspect it might only be marginally more work to get something that works for all `File` inputs and, if possible that would be vastly preferable.

David-Baddeley:
If anyone else were submitting this I'd be having major issues with the functions I've added to `ModuleCollection` - they are pretty hacky (written late at night / early in the morning) just to get things working and there are some that I wouldn't really want to become a formal long term part of the recipe interface (e.g. does it really make sense to be updating a recipe from a file? Is it really the same recipe in this case?). @barentine - As this is something you're waiting on I might let pragmatism win and fix with docstrings in the interim.

barentine:
Hi David, I'm also on the side of pragmatism on this one.

As far as modifying recipes/modulecollections/etc., I think something which might go a long ways on documentation/repeatability side of things would be automatically saving a copy of the recipe yaml to the output directory and maybe even implement the same logic as rerunning localization analysis as far as `example.yaml`, `example_1.yaml`, etc.. Particularly if we defaulted the output_dir to the recipe yaml file stub when available. My point there being I don't really care if a recipe is changed somewhere as long as I know what was run on what (that said, we of course need to make sure nothing crazy happens where a module collection is modified while it's in use, which is maybe where your concern comes from?). My beef with having to edit/save recipes to tweak different parameters isn't because I hate the thought of having a hard reference somewhere, just that writing one locally and uploading it each time (or downloading it first and then editing, uploading) is surprisingly annoying. I would like to have my cluster's `RECIPES` directory have one `multiview-map.yaml` rather than one per calibration file, but I think it'd be great if the final `filled_template`/edited version was saved in the output directory whenever I use it.

David-Baddeley:
Saving the recipe used to generate an output into that output's metadata has been on the TODO list for quite a while now. This should address some of the repeatability issues. Additionally saving the recipe into the output dir could also be considered, but ultimately I think that writing the recipe into the metadata is more authoritative (relies less on getting directory structures right etc ...). 

I think there are 2 problems with the editable recipes as they stand:

a) knowing what recipe an output was generated from (this is easily solved using one or both of the strategies above)
b) allowing multiple recipes to be edited at once on the cluster - at the moment if you have two editor windows open, they are editing the same recipe (this becomes particularly problematic when you have a cluster with multiple users).

barentine:
Any version constraints on jigna and tornado? I'm getting no graphical recipe display and nothing on Launch Analysis.
```
$ conda list | grep jigna
jigna                     0.10.1                   pypi_0    pypi
(base) smeagol@smeagol:~/code/python-microscopy/PYME/cluster$ conda list | grep tornado
tornado                   6.0.4            py36h7b6447c_1    anaconda

```

David-Baddeley:
jigna most definitely doesn't work on py3.7 (just trying to chase #485 stuff and thought I'd use the new version of clusterIO). Not sure about 3.6, but there's a good chance it doesn't.

David-Baddeley:
Had my reservations about jigna as it didn't seem to have received much attention since when I first attempted to use it several years ago. Issue is due to the use of the `async` keyword (which would make the last supported version py3.4).

David-Baddeley:
I'll have a look at the recipe template stuff first, but might put forking and updating jigna on the TODO list this evening (looks like it will be a pretty simple matter of replacing all instances of `async` with something different). See how responsive the folks are if I PR the changes and think about long term strategy from there.

David-Baddeley:
@barentine - Things mostly work on python=3.6 (recipe text still syncs with recipe and can save, load, etc ...), but graphical display doesn't work. This is down to a difference in how `collections.namedtuple` is treated/implemented in py2/3, which results in the data looking different in the browser.  I think there might be a fairly simple workaround to this.

Stock jigna does not work on python>=3.7 as `async` is now a reserved keyword, and jigna uses it as a variable name resulting in syntax errors. Fork of jigna at https://github.com/David-Baddeley/jigna/tree/async-fixes works on py37 (although I think I might also have local changes to clusterui_tornado to avoid the use of async there), but with same namedtuple issues as on py36.

Issue #546, opened 2020-10-27
[BUG] PYMEDataServer name can be too long for later zeroconf versions


barentine:
**Describe the bug**
```
(base) Andrews-MacBook-Pro-5:~ Andrew$ PYMEDataServer
INFO:numexpr.utils:NumExpr defaulting to 8 threads.
INFO:root:========================================
PYMEDataServer, running on python 3.6.10 |Anaconda, Inc.| (default, Jan  7 2020, 15:01:53) 
[GCC 4.2.1 Compatible Clang 4.0.1 (tags/RELEASE_401/final)]

INFO:root:Serving from directory: /Users/Andrew/PYMEData
Traceback (most recent call last):
  File "/opt/miniconda3/bin/PYMEDataServer", line 33, in <module>
    sys.exit(load_entry_point('python-microscopy', 'console_scripts', 'PYMEDataServer')())
  File "/Users/Andrew/code/python-microscopy/PYME/cluster/HTTPDataServer.py", line 925, in main
    ns.register_service(service_name, ip_addr, sa[1])
  File "/Users/Andrew/code/python-microscopy/PYME/misc/pyme_zeroconf.py", line 166, in register_service
    desc)
  File "/opt/miniconda3/lib/python3.6/site-packages/zeroconf/__init__.py", line 1773, in __init__
    if not type_.endswith(service_type_name(name, allow_underscores=True)):
  File "/opt/miniconda3/lib/python3.6/site-packages/zeroconf/__init__.py", line 322, in service_type_name
    raise BadTypeInNameException("Too long: '%s'" % remaining[0])
zeroconf.BadTypeInNameException: Too long: 'PYMEDataServer [Andrews-MacBook-Pro-5]: Andrews-MacBook-Pro-5 - PID:883'
(base) Andrews-MacBook-Pro-5:~ Andrew$ conda list | grep zeroconf
zeroconf                  0.28.5                   pypi_0    pypi
```

**To Reproduce**
1. run PYMEClusterOfOne on a laptop with a silly name
2. notice that the dataserver isn't showing up
3. try running the dataserver
4. boom

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] osx 15
 - Python version: 3.6
 - numpy version: 1.16
 - wxpython version [if a UI bug]: 4.0.4

barentine:
[...]

PR #548, opened 2020-10-27
Change/return service_name when registering with nameservers

barentine:
Addresses issue #546 .

bugfix
**Proposed changes:**
- make a base nameserver
- munge if necessary and return the service name that was used.

Not complete. I've tested it with respect to the dataserver, but posting for feedback from @David-Baddeley. I'm not sure how compatible this is with pyro stuff. Obviously I'm not going to bemoan us if we drop that compatibility, but wanted to check in on what you think here.

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
discussion: 
- [x] Remove the BaseNS

barentine:
all set @David-Baddeley

PR #554, opened 2020-10-29
add software limit set functionality to MarzhauserTango

barentine:
Addresses issue #fearing for my objective.

enhancement


David-Baddeley:
Will merge, but might need to revisit to align with limit setting on other stages

Issue #560, opened 2020-10-30
[BUG] rules.RecipeRule._task_template and HTTPRulePusher.RecipeRule._taskTemplate sets up for ujson failure


barentine:
**Describe the bug**
one extra \n no matter what, with a second available if you don't provide output_dir
```
{"id": "{{ruleID}}~0",
              "type": "recipe",
              "inputs" : "task_inputs",
              "taskdefRef" : "pyme-cluster://villains/RECIPES/calibrate-multiview-shifts.yaml",
              
              }

```
gives 
```
'{"id": "{{ruleID}}~0",\n              "type": "recipe",\n              "inputs" : "task_inputs",\n              "taskdefRef" : "pyme-cluster://villains/RECIPES/calibrate-multiview-shifts.yaml",\n              \n              }'
Exception in thread Thread-4:
Traceback (most recent call last):
  File "/home/smeagol/miniconda3/lib/python3.6/threading.py", line 916, in _bootstrap_inner
    self.run()
  File "/home/smeagol/miniconda3/lib/python3.6/threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "/home/smeagol/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 352, in _poll_tasks
    self._update_tasks()
  File "/home/smeagol/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 221, in _update_tasks
    taskID, cost = rater.get_local()
  File "/home/smeagol/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 119, in get_local
    taskID, cost = self.next()
  File "/home/smeagol/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 113, in next
    return self.__next__()
  File "/home/smeagol/code/python-microscopy/PYME/cluster/rulenodeserver.py", line 82, in __next__
    task = json.loads(filled_template)
ValueError: Unexpected character in found when decoding object value

```
at least on later ujson versions

```
In [11]: task = '{"id": "{{ruleID}}~0",\n              "type": "recipe",\n              "inputs" : "task_inputs",\n              "taskdefRef" : "pyme-cluster://villains/RECIPES/
    ...: calibrate-multiview-shifts.yaml",\n              \n              }'

In [12]: task
[...]

PR #561, opened 2020-10-30
more strictly format the recipe task templates

barentine:
Addresses issue #560 .

bugfix
**Proposed changes:**
- more strictly format recipe templates

printed on the rulenodeserver at the end of `template_fill`
```
{"id": "{{ruleID}}~0", "type": "recipe", "inputs": "{"input":"pyme-cluster:\/\/\/Bergamot\/2020_10_28\/CalibrationStacks_real\/000\/analysis\/28_10_series_L_reg.h5r"}", "output_dir": "pyme-cluster:///Bergamot/2020_10_28/CalibrationStacks_real/000/analysis/analysis", "taskdefRef": "pyme-cluster://villains/RECIPES/calibrate-multiview-shifts.yaml"}
```
Note the issue with the inputs / inputsByTask substitution which still needs to be addressed, e.g. by #485

### DIFF
--- a/PYME/cluster/HTTPRulePusher.py
+++ b/PYME/cluster/HTTPRulePusher.py
@@ -329,26 +329,21 @@ class RecipePusher(object):
 
     @property
     def _taskTemplate(self):
-        task = '''{"id": "{{ruleID}}~{{taskID}}",
-              "type": "recipe",
-              "inputs" : {{taskInputs}},
-              %s,
-              %s
-              }'''
+        task = {
+            'id': '{{ruleID}}~{{taskID}}',
+            'type':'recipe',
+            'inputs': '{{taskInputs}}',
+        }
 
-        if self.output_dir is None:
-            output_dir_n = ''
-        else:
-            output_dir_n = '"output_dir": "%s",' % self.output_dir
-        
+        if self.output_dir != None:
+            task['output_dir'] = self.output_dir
+
         if self.recipeURI:
-            task = task % ('"taskdefRef" : "%s"' % self.recipeURI, output_dir_n)
+            task['taskdefRef'] = self.recipeURI
         else:
-            task = task % ('"taskdef" : {"recipe": "%s"}' % self.recipe_text, output_dir_n)
-            
-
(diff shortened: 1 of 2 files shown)


David-Baddeley:
As mentioned in comments on #560, I think it's a comma issue not an extra newline one (with extra newlines being ignored), but the comma issue is also coincidentally solved by this PR.

I'm more than a little bit reluctant to move away from string based templates for the tasks as, in general, task templates are not guaranteed to be valid json until substituted and there is not necessarily a valid dict representation of the template (in principle you should also be able to substitute integers and non-string stuff which couldn't be expressed with a "template as dict" approach). That said, the localisation rule uses a dict so I'm not sure I have much of a leg to stand on.

PYME/cluster/HTTPRulePusher.py
```diff
@@ -329,26 +329,21 @@ def __init__(self, recipe=None, recipeURI=None, output_dir = None):
...
-            task = task % ('"taskdef" : {"recipe": "%s"}' % self.recipe_text, output_dir_n)
-            
+        task = {
+            'id': '{{ruleID}}~{{taskID}}',
+            'type':'recipe',
+            'inputs': '{{taskInputs}}',
+        }
 
-        return task
+        if self.output_dir != None:
```
David-Baddeley:
```suggestion
        if self.output_dir is not None:
```

PYME/cluster/rules.py
```diff
@@ -395,22 +395,24 @@ def prepare(self):
...
+        # FIXME - task templates should ideally be a .json formatted string. Generating them from dictionarie
+        # as done here is not ideal as it limits templating to within string entries (and is potentially less
+        # less readable / incurs extra overhead in dumps.
+        task = {
+            'id': '{{ruleID}}~{{taskID}}',
+            'type':'recipe',
+            'inputs': '{{taskInputs}}',
+        }
+
+        if self.output_dir != None:
```
David-Baddeley:
```suggestion
        if self.output_dir is not None:
```

Issue #545, opened 2020-10-27
[BUG] cluster get freespace takes 0.45 s every status expiration (1 s)


barentine:
**Describe the bug**
Every guiUpdateCallback in `SpoolController` calls the `HDFSpoolFrame` tick which updates freespace. When spooling to cluster, this eventually calls `clusterIO.get_status`. While `clusterIO.get_status` caches, making subsequent calls within the expiration fast, it does not poll updates continuously in the background, meaning that if the cache has expired we call it again, and each time the cache has expired this takes roughly 0.44 s, which is happening in the GUI thread and presumably why I'm seeing the high-throughput system GUI hang every second or so

```
In [7]: import timeit

In [8]: timeit.timeit('get_status("villains")', setup='from PYME.IO.clusterIO import get_status', number=10)
Out[8]: 0.4622703999999942

In [9]: timeit.timeit('get_status("villains")', setup='from PYME.IO.clusterIO import get_status', number=50)
Out[9]: 0.4372077000000161

In [10]: timeit.timeit('get_status("villains")', setup='from PYME.IO.clusterIO import get_status', number=100)
Out[10]: 0.4391475999999841

In [11]:

In [11]: timeit.timeit('get_status("villains"); time.sleep(1)', setup='import time; from PYME.IO.clusterIO import get_s
    ...: tatus', number=10)
Out[11]: 14.45680680000001
```

**To Reproduce**
1. run PYMEAcquire with a cluster
2. spool
3. watch the gui stutter a bit

**Expected behavior**
not kill performance for a status update

**Screenshots**

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] calling from win10
 - Python version: 3.6
 - numpy version: 1.19
 - wxpython version [if a UI bug]: 4.0.4

**Question for @David-Baddeley **
OK if I write `clusterIO.get_polled_status` as a thread-polled version of get_status which starts a thread on its first call and updates every 5 s?

barentine:
conversation with David:
- don't put it in clusterIO, maybe config the update poll time
- potentially just skirt the issue with a config option to disable (and update config option docs + cluster_install docs)

follow on: why does it take a long time?

barentine:
Sorry if the follow-on should go in this same thread or not, but I'll start it with the `top` during localization analysis of 8x 30k frame series, the GPU usage is substantially lower than the CPU usage, and top on the ruleserver node is:
[...]

PR #562, opened 2020-11-02
consistently key using str for _dirCache

barentine:
Addresses issue #545.

potential bugfix - looking at the code I'm pretty sure we need this but will have to wait til tomorrow to test

- [ ] tested

### DIFF
--- a/PYME/IO/clusterIO.py
+++ b/PYME/IO/clusterIO.py
@@ -228,6 +228,18 @@ def _getSession(url):
 
 
 def _listSingleDir(dirurl, nRetries=1, timeout=5):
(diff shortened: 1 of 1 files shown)


barentine:
can confirm that adding
```
        if dirurl not in _dirCache.keys():
            try:
                dirurl.encode('utf-8') in _dirCache.keys()
                logger.error('we goofed')
            except TypeError:
                pass
```
in the `_listSingleDir` try statement prints 'we goofed' running the current repo head

barentine:
shoot, think I misread and we're caching on `dirurl` which _is_ a str rather than `url` which we encode

PR #563, opened 2020-11-02
actually use config'd cluster name during spool

barentine:
Addresses issue just making sure we get cache keys right everywhere, even though I don't think `clusterIO._locateCache` is going to be super important on the instrument computer.

maybe slight bugfix
**Proposed changes:**
- user config'd serverfilter rather than `''` in `HTTPSpooler`

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/HTTPSpooler.py
+++ b/PYME/Acquire/HTTPSpooler.py
@@ -76,8 +76,6 @@ class EventLogger:
         return json.dumps(self._events)
           
 
-CLUSTERID=''  
-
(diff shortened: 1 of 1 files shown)


barentine:
note that for auto analysis, this sets the series name in `HTTPRulePusher.launch_localize` which gets used to create at least one `ClusterPZFDataSource`.

Haven't found a big 'aha!' for where this would slow us down re: #545, however it wouldn't hurt to not ignore the specified server filter.

Issue #536, opened 2020-10-24
[PROPOSAL] basic sample metadata / dispatch.Signal on SpoolController.SetSpoolDir


barentine:
**Is your feature request related to a problem? Please describe.**
I want to get some basic sample information booked in the series metadata so it can be accessible to recipes, e.g. pyme-omero uploads (to set project, dataset, or 'details'). I don't need 1-to-1 mapping on any api, I would be well on my way if the spool directory was in the metadata somewhere, and I set up the omero upload recipe modules to substitute metadata keys, e.g. `Omero Project : {Sample.SlideID}` pulling from `mdh['Sample.SlideID']`.

**Describe the solution you'd like to implement**
I don't want the full sampleDB `sampleInformation.SampleInfoDialog`, in part because I'm not using sampleDB and do not have a database to connect to, and I don't want to translate the full sampleDB api to the OMERO api, and I don't want to plug in the full OMERO api. For now, I just want a couple of these in my metadata, and don't want to add anything that doesn't easily fit into either API. 

I'll write a handler to add in my init script, which will open a small dialog when a user sets the spool directory:
- ID text entry, with the spoor directory name suggested (unsure whether to go in as `Sample.SlideID` or `Sample.Specimen` - the latter feels weird, and would not be the wx static text I display. SampleDB docs are sparse)
- Notes text entry (will go as `Sample.Notes`)
- Skip / Save button

barentine:
No-go on the dispatch signal, but throwing a two text-entry SampleID + notes panel in camPanels isn't the worst thing in the world

barentine:
BUT - also flag that the sample info stuff will need quite a bit more refactoring / attention, and will eventually be nicer

barentine:
Sample.SlideID is supposed to be an int, our ticket is `'Sample.SlideRef'`

Also note `Sample.Creator` isn't a bad one to textentry

barentine:
closing as this is addressed by #573. further sample metadata dialog refactoring to happen farther down the line

PR #573, opened 2020-11-03
simple sample metadata panel

barentine:
Addresses issue #536.

enhancement
**Proposed changes:**
- add a sample sample metadata entry panel - pragmatic solution to have _something_ easily while we refactor the full sampleinfodialog
- use it in init_htsms

Screen Shot 2020-11-02 at 10 46 00 PM

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

### DIFF
--- a/PYME/Acquire/sampleInformation.py
+++ b/PYME/Acquire/sampleInformation.py
@@ -582,3 +582,4 @@ class SimpleSampleInfoPanel(wx.Panel):
             mdh['Sample.SlideRef'] = self.slide.GetValue()
             mdh['Sample.Creator'] = self.creator.GetValue()
             mdh['Sample.Notes'] = self.notes.GetValue()
+            mdh['AcquiringUser'] = self.imager.GetValue()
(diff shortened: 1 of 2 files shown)


PYME/Acquire/sampleInformation.py
```diff
@@ -520,4 +520,66 @@ def createImage(mdh, slide, comments=''):
...
-#MetaDataHandler.provideStartMetadata.append(lambda mdh: getSampleDataFailsafe)
\ No newline at end of file
+#MetaDataHandler.provideStartMetadata.append(lambda mdh: getSampleDataFailsafe)
+
+
+class SimpleSampleInfoPanel(wx.Panel):
+    def __init__(self, parent):
+        """Simple sample info panel to tide over until the SampleInfoDialog is
+        refactored to work independently of a sampleDB database.
+
```
David-Baddeley, line 530:
```suggestion

TODOS/LIMITATIONS: potentially better addressed by the full refactor rather than by modifying this class, but included here for information / to reinforce that this should ideally be replaced at some point in the future.

 - auto-populate and/or hide acquiring user using logon name on multi-user systems (preference would be to have a config option which can be set on a single user system, but which defaults to false)
 - include info on what is labelled (i.e. Sample.Labelling entries). This is super,super useful to have in the metadata
 - because there is no sanity checking / database backend etc ... you will likely get subtly differing values for acquiring user and creator which will make it hard to search on these if the data ever gets ingested into a database in the future.
 
```

Issue #575, opened 2020-11-03
[BUG] processing.AverageFramesByStep / piezo_movement_correction.correct_target_positions wrapping issue


barentine:
**Describe the bug**
should only have one frame per z at the end, but looks like we wrap on the end. Will need to look into whether we are effectively 1-indexing with starttime or something. Note that putting in a `frames` array into `correct_target_positions` which is too long will continue dumping the rest into the last z step. 

<details> <summary> printing z values and unique counts </summary>

```
DEBUG:PYME.recipes.processing:Averaged stack size: 865
DEBUG:PYME.recipes.processing:averaging by step
[48.757 48.775 48.801 48.802 48.803 48.804 48.815 48.828 48.829 48.831
 48.838 48.84  48.841 48.843 48.855 48.895 48.9   48.906 48.909 48.912
 48.913 48.919 48.92  48.921 48.922 48.935 48.961 48.975 48.989 48.991
 48.994 49.    49.001 49.015 49.055 49.062 49.063 49.064 49.066 49.068
 49.083 49.095 49.135 49.141 49.145 49.148 49.15  49.151 49.162 49.163
 49.175 49.202 49.215 49.226 49.229 49.23  49.231 49.232 49.238 49.24
 49.241 49.255 49.295 49.308 49.309 49.31  49.313 49.318 49.319 49.32
 49.323 49.335 49.375 49.388 49.389 49.392 49.401 49.402 49.403 49.415
 49.455 49.468 49.47  49.471 49.476 49.48  49.481 49.482 49.483 49.495
 49.535 49.548 49.549 49.551 49.559 49.561 49.562 49.575 49.615 49.62
 49.628 49.629 49.63  49.632 49.634 49.64  49.641 49.642 49.655 49.695
 49.704 49.709 49.71  49.711 49.717 49.72  49.721 49.722 49.735 49.775
 49.78  49.783 49.788 49.789 49.792 49.793 49.796 49.799 49.801 49.802
 49.815 49.855 49.866 49.869 49.87  49.871 49.874 49.881 49.882 49.895
 49.935 49.948 49.952 49.955 49.956 49.957 49.958 49.96  49.961 49.962
 49.963 49.975 50.001 50.015 50.023 50.024 50.025 50.026 50.039 50.042
 50.055 50.095 50.105 50.106 50.107 50.108 50.121 50.123 50.135 50.175
[...]

PR #576, opened 2020-11-03
focus correction event mapping fixes

barentine:
Addresses issue #575 and not actually testing piezo ontarget / offsetpiezo stuff.

both
**Proposed changes:**
- type integers in `piecewiseMapping.times_to_frames` as int so we don't get sporadic rounding errors when deciding which side of a piecewise point something falls on
- beef up the piezo movement focus correction test
- use start time from metadata in `processing.AverageFramesByZStep` so our fudged events match our metadata

**OLD, but leaving it here Note**
This brings up a potential issue with piecewiseMapping and rounding. The test fails on one of the ProtocolFocus events active like it's 1-indexed (or the focus takes place on the next frame) while the rest do not. Having StartTime at zero, and setting all frame numbers as cycletime * frame number, I wonder if we are rounding somewhere in a str->back converstion.

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -1762,7 +1762,7 @@ class AverageFramesByZStep(ModuleBase):
         logger.debug('Averaged stack size: %d' % n_steps)
 
         new_stack = []
-        t = time.time()
+        t = image_stack.mdh.getOrDefault('StartTime', time.time())
         fudged_events = []
         cycle_time = image_stack.mdh.getOrDefault('Camera.CycleTime', 1.0)
         for ci in range(image_stack.data.shape[3]):
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -1762,7 +1762,8 @@ class AverageFramesByZStep(ModuleBase):
         logger.debug('Averaged stack size: %d' % n_steps)
 
         new_stack = []
-        t = image_stack.mdh.getOrDefault('StartTime', time.time())
+       # TODO - should we default to zero or abort?
+        t = image_stack.mdh.getOrDefault('StartTime', 0)
         fudged_events = []
         cycle_time = image_stack.mdh.getOrDefault('Camera.CycleTime', 1.0)
         for ci in range(image_stack.data.shape[3]):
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Appears to be currently failing as precision of `np.assert_array_almost_equal()` is too high. Maybe change the assertion precision to something which makes physical sense. A tolerance of .01 or even .1 nm would likely be fine for it's real world application.

barentine:
Excellent point David, though I am also seeing a 1.227 um error which occurs on frame 1601 only
```
    def test_focus_correction():
        from PYME.Analysis.piezo_movement_correction import correct_target_positions
        corrected_focus = correct_target_positions(TEST_DATA_SOURCE['t'], TEST_EVENTS, TEST_MDH)
>       np.testing.assert_array_almost_equal(GROUND_TRUTH_Z, corrected_focus)
E       AssertionError: 
E       Arrays are not almost equal to 6 decimals
E       
E       Mismatch: 0.04%
E       Max absolute difference: 1.227
E       Max relative difference: 0.02398311
E        x: array([48.262, 48.307, 48.307, ..., 53.588, 53.588, 53.588])
E        y: array([48.262, 48.307, 48.307, ..., 53.588, 53.588, 53.588])
```


```
AssertionError: 
Arrays are not almost equal to 6 decimals

Mismatch: 0.04%
Max absolute difference: 1.227
Max relative difference: 0.02398311
 x: array([48.262, 48.307, 48.307, ..., 53.588, 53.588, 53.588])
 y: array([48.262, 48.307, 48.307, ..., 53.588, 53.588, 53.588])

In [3]: np.where(np.abs(corrected_focus - GROUND_TRUTH_Z) > 0.5)
Out[3]: (array([1601]),)
```

barentine:
a part of me wonders whether this (rounding) issue has anything to do with #575.

barentine:
Hi @David-Baddeley - keeping the frame int-typed in `piecewiseMapping.times_to_frames` fixed the test (no decimal tweaking needed).

barentine:
description now updated to reflect the two bugfixes

PYME/recipes/processing.py
```diff
@@ -1762,7 +1762,7 @@ def execute(self, namespace):
         logger.debug('Averaged stack size: %d' % n_steps)
 
         new_stack = []
-        t = time.time()
+        t = image_stack.mdh.getOrDefault('StartTime', time.time())
```
```suggestion
       # TODO - should we default to zero or abort?
        t = image_stack.mdh.getOrDefault('StartTime', 0)
```

Issue #579, opened 2020-11-04
[PROPOSAL] make ImagePyramid choose TileIO dynamically


adrianhollow:
**Is your feature request related to a problem? Please describe.**
The TileIO in ImagePyramid is currently hardcoded for PZFTileIO. If a user wants to load an ImagePyramid from another file format, the code in \_\_init\_\_ needs to be adapted accordingly.

**Describe the solution you'd like to implement**
Set the TileIO based on file extensions in base_dir to avoid changing the code manually.

**What potential side effects do the proposed changes have?**
- additional code will run in ImagePyramid.\_\_init\_\_ (looping through files in base_dir until a file extension is recognized)
- new TileIO classes will need to be registered in a dict or won't be recognized automatically
- if no file extension is recognized, an Exception will be raised (and end the program)

David-Baddeley:
Hi Adrian, it would be good to better understand the use case here. Are you trying to load image pyramids created in different software? Or do you just want to experiment with the different backends?

As its currently written `ImagePyramid` is designed, in the first instance, for creating an image pyramid (although it can also open an existing one that it has created). It's not really a general-purpose image pyramid class at this point (which is not to say that we couldn't make it one). For the creation use case there obviously won't be any files present to infer type from. Both for this reason, and for performance, it might be preferable to just specify the backend as an argument to the constructor. We could also write the backend used into the metadata and read this when re-opening a pyramid if a backend is not specified.

David-Baddeley:
More generally, `ImagePyramid` is due a bit of an overhaul (and a good bit of documentation) re the distinction between loading and creating.  It was written quickly and is very much alpha-quality code. My first thought would be to leave the constructor as the default way of creating a new pyramid and add a `load_exiting(...)` class method for the loading case.

We'd still need to change the constructor a bit as it over-writes a bunch of parameters which it should read out of the metadata when loading an existing pyramid (although we could probably handle most of this by reading those parameters in `load_exisiting()` and passing them to the constructor).

adrianhollow:
Hi David, I am trying to open different datasets, some of which have been saved as pzf-files and others as npy-files. When I run "dh5view SUPERTILE:\<path\>" at the moment, I can only open datasets whose corresponding TileIO is hardcoded in \_\_init\_\_ of ImagePyramid. I thought it would make sense to adjust the TileIO selection at runtime so a user can avoid touching the code when they want to open a different dataset.

[...]

PR #580, opened 2020-11-04
support backend switching in ImagePyramid

adrianhollow:
Addresses issue https://github.com/python-microscopy/python-microscopy/issues/579.

It's an enhancement.

**Proposed changes:**
Loop through the files in base_dir of ImagePyramid to find the correct TileIO. Set the TileIO based on the file extensions.

**Checklist:**

- [ ] Tested with numpy=1.14


PYME/Analysis/tile_pyramid.py
```diff
@@ -254,7 +254,20 @@ def flush(self):
     def __del__(self):
         self._cur.close()
         self._conn.close()
-    
+
+
+def tile_init_ref(base_dir, map_type_tile):
```
barentine:
```
TILEIO_EXT = {
    '.pzf': PZFTileIO,
    '.npy': NumpyTileIO,
    '.db': SqliteTileIO,
}

def get_tileio_backend(base_directory):
    """ find TileIO backend for a given ImagePyramid
    
    Parameters
    ----------
    base_directory : str
        root directory of an ImagePyramid instance
    
    Returns
    -------
    class
        which TileIO derived class the ImagePyramid can be
        built with.
    """
    for root, dirs, files in os.walk(base_dir):
        for f in files:
            if os.path.splitext(f)[-1] in TILEIO_EXT.keys():
                return TILEIO_EXT[file_extension]
    raise IOError("No files found for loading ImagePyramid.")
```

barentine (commented):
Thanks @adrianhollow! Any thoughts on these tweaks?

PYME/Analysis/tile_pyramid.py
adrianhollow: Makes sense to me. I've included this in the next commit.

PYME/Analysis/tile_pyramid.py
```diff
@@ -278,28 +310,30 @@ def __init__(self, storage_directory, pyramid_tile_size=256, mdh=None,
...
-            
-        self._imgs = PZFTileIO(base_dir=self.base_dir, suff='img')
-        self._acc = PZFTileIO(base_dir=self.base_dir, suff='acc')
-        self._occ = PZFTileIO(base_dir=self.base_dir, suff='occ')
 
-        # self._imgs = SqliteTileIO(base_dir=self.base_dir, suff='img')
-        # self._acc = SqliteTileIO(base_dir=self.base_dir, suff='acc')
-        # self._occ = SqliteTileIO(base_dir=self.base_dir, suff='occ')
-    
+        map_type_tile = {
```
David-Baddeley: This duplicates `TILEIO_EXT` above

PYME/Analysis/tile_pyramid.py
```diff
@@ -278,28 +310,30 @@ def __init__(self, storage_directory, pyramid_tile_size=256, mdh=None,
...
+            ".pzf": PZFTileIO,
+            ".npy": NumpyTileIO,
+            ".db": SqliteTileIO,
+        }
+        tile_init_ref = get_tile_init_ref(self.base_dir, map_type_tile)
```
David-Baddeley: `tile_init_ref` is not defined, nor is `file_type`

PYME/Analysis/tile_pyramid.py
```diff
@@ -254,7 +254,39 @@ def flush(self):
...
+    """
+    for root, dirs, files in os.walk(base_directory):
+        for file in files:
+            file_extension = os.path.splitext(file)[-1]
+            if file_extension in TILEIO_EXT.keys():
+                return TILEIO_EXT[file_extension]
+    raise IOError("No files found for loading ImagePyramid.")
 
 class ImagePyramid(object):
     def __init__(self, storage_directory, pyramid_tile_size=256, mdh=None,
```
David-Baddeley, line 293: suggest adding `backend` parameter, defaulting to `backend=PZFTileIO` (i.e. the status quo). We need to specify this when creating a new pyramid, and the current default was chosen for performance - ie is probably the right one to use in most scenarios.

PYME/Analysis/tile_pyramid.py
```diff
@@ -254,7 +254,39 @@ def flush(self):
...
-    
+
+
+TILEIO_EXT = {
+    '.pzf': PZFTileIO,
+    '.npy': NumpyTileIO,
+    '.db': SqliteTileIO,
+}
+
+def get_tileio_backend(base_directory):
```
David-Baddeley:
```suggestion
def infer_tileio_backend(base_directory):
```

Subtle change to make it clear that we are guessing the backend / rather than specifying

PYME/Analysis/tile_pyramid.py
```diff
@@ -278,28 +310,30 @@ def __init__(self, storage_directory, pyramid_tile_size=256, mdh=None,
...
         self._mdh['Pyramid.PixelSize'] = pixel_size
-        
+
         if not os.path.exists(self.base_dir):
             os.makedirs(self.base_dir)
-            
+
         #self._tilecache = TileCache()
-            
-        self._imgs = PZFTileIO(base_dir=self.base_dir, suff='img')
```
David-Baddeley, line 295: replace `PZFTileIO()` calls with `backend(...)`

Leaves current behaviour as default, but lets you over-ride it

PYME/Analysis/tile_pyramid.py
```diff
@@ -278,28 +310,30 @@ def __init__(self, storage_directory, pyramid_tile_size=256, mdh=None,
...
         # TODO - should we be re-assigning these on load, not just when we create a new pyramid?
         self._mdh['Pyramid.x0'] = x0
         self._mdh['Pyramid.y0'] = y0
         self._mdh['Pyramid.PixelSize'] = pixel_size
-        
+
         if not os.path.exists(self.base_dir):
             os.makedirs(self.base_dir)
-            
+
```
David-Baddeley, line 325: Add a test somewhere in here along the following lines:

``` python
if backend == 'infer': 
    # special case value for backend to let us load with inference
    # could alternatively implement as a test against None instead of 'infer' 
    backend = infer_tileio_backend(self.base_dir)
```

Issue #481, opened 2020-10-06
What should PYMEVis Help>About do??


zacsimile:
**Describe the bug**
```
 Error whilst running Help>&About
==============================================
python-microscopy=20.10.05
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
TypeError: 'NoneType' object is not callable
```

**To Reproduce**
1. Open PYMEVis
2. Click `Help>About`

**Expected behavior**
An about window pops up.

**Possible alternative**
Remove `Help>About`.

David-Baddeley:
Even with the fix applied, the current 'About' dialog is super out of date and useless. Might almost be better to remove the menu item completely.

a) Do we actually need/want a `help-->about` menu item? What does it actually bring us? 
b) If we do, what info should we present in `help-->about` and should we have one about dialog which is re-used across all programs.

David-Baddeley:
Seeing as the bug is now fixed, changing this to a discussion issue about what help-->about should actually show.

zacsimile:
I think it should include the PYME version number and maybe a link to our issue tracker labeled `Report an issue`. Maybe also a link to a page on our website that lists contributors and the libraries we use (acknowledgements).

zacsimile:
Do we want to continue having this discussion, or are we satisfied with the current about dialog?

csoeller:
One bit of info it should show in addition to what is there now is the supporting packages version set etc as at the top of bug reports:

```
python-microscopy=20.10.05
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

```
I am always looking for this info when I want to report a bug that did not throw open the 'bug window'. Tedious to do it manually.

barentine:
I agree with @csoeller on this one. I never learned the windows equivalent of `grep`, so when I'm troubleshooting on someone elses machine a quick 'survey of the environment' in PYME would be grand

PR #587, opened 2020-11-06
Update about dialog

zacsimile:
Addresses issue #481.

Enhancement

**Proposed changes:**

image

image

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/LMVis/VisGUI.py
+++ b/PYME/LMVis/VisGUI.py
@@ -248,13 +248,28 @@ class VisGUIFrame(AUIFrame, visCore.VisGUICore):
 
 
     def OnAbout(self, event):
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Have commented out developers for now as providing no info here is preferable to only recognising one.  Also commented out license for now as it looks weird in the drop-down on it's own. Ultimately we should probably move this to PYME.ui and populate the common fields (version, copyright, icon, license, developers etc ...) there, providing only component name and description in the individual apps.

David-Baddeley:
PS - we should also add a couple of help menu items which do a `webbrowser.open()` on either the docs, or the biorxiv paper.

Issue #571, opened 2020-11-03
[ENHANCEMENT] Update number of localizations displayed after a filter in the PYMEVis Data pipeline


zacsimile:
**Is your feature request related to a problem? Please describe.**
Right now, if I look at `Data pipeline` before and after applying a filter I have
![image](https://user-images.githubusercontent.com/1263313/97934113-b82dfa00-1d42-11eb-9b35-5f1a65ac5d18.png)
and
![image](https://user-images.githubusercontent.com/1263313/97934122-bebc7180-1d42-11eb-878e-927460338593.png)
where the number of events listed under `filtered_localizations` is the same. Visually I can see this is not the case.

**Describe the solution you'd like**
Have the data pipeline accurately reflect the number of localizations present after applying a filter.

**Are you prepared to work on it?**
Yes

David-Baddeley:
This has been on my annoyance list for ages. A bit messy to implement which is why it's not been done yet (have sort of been punting it until I can put the whole lot in html). Some options:

- completely recreate the `vertical_recipe_display` on any change to the pipeline (undesirable from a performance perspective)
- only update the event counts when the pipeline updates (need to keep track of the static text elements when creating the panel - maybe in a dict - and find some way of triggering off any recipe change. Not as bad as completely recreating the panel, but still involves doing a `len()` on everything
 - only update the label for outputs which have changed. Probably ideal, but hard to implement.

barentine:
this set @zacsimile?

zacsimile:
Yeah

PR #589, opened 2020-11-06
Update number of localizations displayed after recipe execution

zacsimile:
Addresses issue #571.

Yes

**Proposed changes:**
- Option 2 proposed by @David-Baddeley in #571: only update the event counts when the pipeline updates. Conveniently, we were already keeping track of stack text elements in `RecipeDisplayPanel.data_panes`.

image

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
That's much cleaner than I thought it was going to be, thanks!

Issue #570, opened 2020-11-03
[ENHANCEMENT] Change zoom speed on PYMEVis canvas.


zacsimile:
**Is your feature request related to a problem? Please describe.**
Even slight scrolling, especially on track pads, causes drastic changes in the zoom on some machines.

**Describe the solution you'd like**
Allow the user to adjust the zoom speed (change `ZOOM_FACTOR` in `PYME.LMVis.gl_render3D`). Ideally this setting would be saved when the user adjust the speed and be re-loaded at opening of PYMEVis.

I might put this under a new dialog, `File>Settings`. Alternatively, holding control while scrolling could adjust the zoom speed, but this might be a bit weird.

**Describe alternatives you've considered**
Leave it hardcoded and slow down zooming for everyone.

**Are you prepared to work on it?**
Yes

David-Baddeley:
This is going to be a bit hard to get right/intuitive and might be worth doing a little more research on. It would be great if we knew which hardware/os the complaint came from (or if it is observable on multiple machines). I wonder if the issues might arise from a pretty edge use case e.g. running a windows VM on a mac. As well as having `ZOOM_FACTOR` set-able, having sensible defaults would also be good. When I initially determined the current values, I tested win with wheel mouse, win with trackpad, linux, and OSX on a few different machines.  A good first step would be to make a `PYME.config` option for it which would over-ride the defaults. Whether we put a GUI round it or just document how to set the config option should probably depend on how common an issue it is.

The rolls-royce solution would be to record the time between scroll events and if it's super short, pop up a dialog prompting the user to change the zoom factor. Otherwise it's always going to be a setting that will take some effort to discover and it's going to feel wrong in the meantime.

zacsimile:
I did a bit of research. As best I can tell, our zoom speed is strictly set by the rate at which `wx` throws `EVT_MOUSEWHEEL`. I can't find any data on how this rate is set. I know that it does not change if I change the speed of my trackpad (macOS 10.15.7), so it's not tied to system mouse settings in the way I thought it would be. 

[...]

PR #590, opened 2020-11-06
Prog config

zacsimile:
Partially addresses issue #570. We need a method of storing a user's preferred zoom factor.

Enhancement

**Proposed changes:**
- Add ability to programmatically change simple values in PYME's `config.yaml`s (and similar)

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/config.py
+++ b/PYME/config.py
@@ -540,12 +540,12 @@ def update_yaml_keys(fn, d):
 
     # Update the appropriate keys
     for k, v in d.items():
-        x = re.search("\n{}\s*:\s*([^(#|\n|\s)]*)".format(k),data)
+        x = re.search('^{}\s*:\s*([^(#|\n|\s)]*)'.format(k),data,flags=re.MULTILINE)
         if x is None:
             data += '\n{}: {}'.format(k, v)
         else:
-            data = re.sub("\n{}\s*:\s*([^(#|\n|\s)]*)".format(k),
-                          "\n{}: {}".format(k,v),data)
+            data = re.sub('^{}\s*:\s*([^(#|\n|\s)]*)'.format(k),
+                          '\n{}: {}'.format(k,v),data,flags=re.MULTILINE)
 
     # Update the yaml file
     with open(fn, 'w') as f:
@@ -576,8 +576,10 @@ def update_config(d, config='user', config_fn='config.yaml'):
     # Open and edit the file
     update_yaml_keys(os.path.join(base,config_fn),d)
 
-    # # Reload config
-    # if sys.version_info.major == 3:
-    #     from importlib import reload
-    # reload(PYME.config)
-    
+    # Reload config
+    if sys.version_info.major == 3:
+        from importlib import reload
+    try:
+        reload(sys.modules['config'])
+    except(KeyError):
+        reload(sys.modules['PYME.config'])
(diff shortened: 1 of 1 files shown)


PYME/config.py
```diff
@@ -508,3 +508,78 @@ def get_init_filename(filename, legacy_scripts_directory=None):
...
+    """
+    import re
+
+    # Read the yaml file
+    with open(fn) as f:
+        data = f.read()
+
+    # Update the appropriate keys
+    for k, v in d.items():
+        x = re.search('^{}\s*:\s*([^(#|\n|\s)]*)'.format(k),data,flags=re.MULTILINE)
```
David-Baddeley: I thought regex strings needed to either use the `r'my regex'` form or escape the backslashes ... Is this tested?

It also looks as though this will stop on any whitespace (which might not be desirable - it would be good to allow string parameters with whitespace in them). 

I'd try this pattern: `r'^{}\s*:.*$'`.format(k)

zacsimile: This is tested in py3. The `r` might be necessary on py2. Your regex will also work, but it might delete a comment on the same line. I think all we'd need to do is remove the `\s` in the final block to not stop on any whitespace. I admittedly only tested with non-whitespace elements after the `:`.

PYME/config.py
```diff
@@ -508,3 +508,78 @@ def get_init_filename(filename, legacy_scripts_directory=None):
...
+        if x is None:
+            data += '\n{}: {}'.format(k, v)
+        else:
+            data = re.sub('^{}\s*:\s*([^(#|\n|\s)]*)'.format(k),
+                          '\n{}: {}'.format(k,v),data,flags=re.MULTILINE)
```
David-Baddeley: using `.format()` won't give you valid yaml in most cases (booleans are a case in point, and quite common in our config). A workaround would be to `json.dumps()` the values before the string substitution (which works because json is a subset of yaml). `yaml.dump()` doesn't work because it puts lots of new lines etc ... in the resulting string.

zacsimile: Just double-checked: this does work for booleans on py3. For safety, I will switch to `json.dumps`.

PYME/config.py
zacsimile: I stand corrected. `$` appears to match comments as well (it's the end of the line).

PYME/config.py
zacsimile: Tried switching to `json.dumps` and it wrapped everything in quotes, which made the regex not work. Also doesn't seem consistent with YAML. On my machine it works with the `.format()`, including with booleans.

David-Baddeley:
```python
In [25]: json.dumps(True)                                                                                                            
Out[25]: 'true'

In [26]: '{}'.format(True)                                                                                                           
Out[26]: 'True'

In [30]: yaml.dump(yaml.load('foo: true'))                                                                                           
Out[30]: '{foo: true}\n'

In [31]: yaml.dump(yaml.load('foo: True'))                                                                                           
Out[31]: '{foo: true}\n'
```

So booleans are supposed to be lower case, but pyyaml tolerates upper case when reading.

Also, I don't get any quotes
```python
In [32]: 'foo: {}'.format(True)                                                                                                      
Out[32]: 'foo: True'

In [33]: 'foo: {}'.format(json.dumps(True))                                                                                          
Out[33]: 'foo: true'
```

zacsimile: Ah, I misunderstood your earlier message and thought you'd suggested `json.dumps` on the key as well as the value. The key is where I was getting the extra quotes. I've added `json.dumps` back in for the value. While it seems we can get away without `json.dumps` for booleans, it's likely safer for other data types.

PYME/config.py
zacsimile: Hmm, further testing seems to indicate this search string actually does delete comments on the same line as the variable. Do we care about this?

PR #591, opened 2020-11-06
allow dh5view SUPERTILE:PYME-CLUSTER:///tileseries.pcs loading

barentine:
Addresses issue #wanting to grab a high level of supertile from the cluster quickly without writing two new recipe modules

enhancement
**Proposed changes:**
- if we don't have pyramid metadata when we hit supertile from filename make the pyramid. Note we already have a warning in place for the non-lightweightness of this (though for small series, e.g. 300 frames, it's really not bad)

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

### DIFF
--- a/PYME/IO/DataSources/SupertileDatasource.py
+++ b/PYME/IO/DataSources/SupertileDatasource.py
@@ -86,6 +86,9 @@ class SupertileDataSource(BaseDataSource):
         overlap = int(qp.get('overlap', [1])[0])
         
         mdh = MetaDataHandler.load_json(os.path.join(tile_base, 'metadata.json'))  # TODO - does this need to be posixpath?
+        if 'Pyramid.TileSize' not in mdh.keys():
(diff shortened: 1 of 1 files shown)


barentine:
@David-Baddeley any thoughts on this? I'm working on #648 and reminded that this would be convenient

David-Baddeley:
We talked about this directly, but not really a fan for a number of reasons:

- It's not portable across schemas (i.e. will only work with a PYMECluster .pcs series, not e.g. .tiff)
- It relies on exploiting the fact that .pcs folders have a metadata file in the same place as tile pyramids. This feels super fragile.
- I don't like implicitly creating pyramids, and think it is bad for both performance and generating readable/understandable code.

David-Baddeley:
We now have recipe tools for creating pyramids, so this should be superfluous

Issue #581, opened 2020-11-04
[ENHANCEMENT] concatenate tiled series


barentine:
**Is your feature request related to a problem? Please describe.**
I have a lot of tiled overview images that I want to view as one supertile source (just for viewing at the moment)

** *Potential soutions* **

**pyme-cluster:// - aggregate ImageOutput module:**
Add to HTTPDataServer aggregate_h5 endpoint such that we could enable `pyme-cluster:// - aggregate` scheme for `output.ImageOutput`. This would not entail much in the dataserver, but rather in `IO.h5File.H5File`, where `put_file` would need to have a bit more logic on pzf frames (since frame numbers will overlap).

However, because `PointScanner` events, `ScannerXPos, ScannerYPos` only include time stamps and not frame numbers, and we use `mdh['StartTime']` to map to frames, we would need to spoof the event times when appending them and associated frames (would also need a lock taken out in part of `H5File.put_file` to update a max_frame / max_time variable in case we try and push frames from multiple sources at once. This might be easiest in a separate method to `put_file` that can take events, frames, and metadata all at the same time.

**SuperClusterPZFDataSource:**
A datasource for the superlative extraordinaire who wants to make a SupertileDataSource from only the finest input datasource. Would be grosser from a ui perspective in my opinion, as the other I can make happen behind the scenes easily in a recipe I'm already running for detection/queueing on individual tile overviews, while this would I would somehow need to generate an arbitrarily long list of inputs for a recipe module.

**tile_super_pyramid:**
A method in tile_pyramid which directly takes as input a list of tile series and generates a mondo pyramid, never writing out a singular full concatenated tile series.

**modify the tile viewer to accept multiple inputs**
might be the most pragmatic if not least powerful

**Are you prepared to work on it?**
always

barentine:
modifying the tileviewer might be super simple from the openseadragon perspective as we just need to add more tile sources. 

adding a `multitileviewer.py` and associated `multitileviewer.html` really might be straightforward

David-Baddeley:
I think any aggregation of separate tile pyramids is going to be a bit tricky due to the fact that it's highly unlikely that the tile origins will align on the same grid.  The 3 options I'd consider would be:

a) save everything to the one pyramid (this could be done from separate, per-well protocols, but ultimately all writing into the same pyramid structure). This would ensure tiles are aligned and make viewing a cinch. Might not be practically realisable until we have a distributed version of the tile pyramid though.

[...]

PR #593, opened 2020-11-06
small tile_pyramid improvements (ADOffset subtraction, xy passing)

barentine:
Addresses issue collating xy positions is easier than collating mappings, if one wanted to be bold with concatenating tile datasources into the same pyramid. See #581  (it ended up being really simple to make a concatenated datasource and tile pyramid it - will put that datasource in a notebook PR to example_notebooks rather than something legitimate to use frequently)

enhacnement
**Proposed changes:**
- add docstring to `tile_pyramid.tile_pyramid`
- allow x y positions to be passed as either (current) `piecewiseMap`s or `numpy.ndarray`
- at least subtract the `ADOffset` when building pyramid if a full darkmap hasn't been provided

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4


barentine:
handled dark/offset as discussed @David-Baddeley

Issue #584, opened 2020-11-05
[ENHANCEMENT] Make it easier to visualize and export measurements from PYMEVis


zacsimile:
**Is your feature request related to a problem? Please describe.**
Measurements, particularly cluster measurements, can be difficult to use/plot from within PYME. It is often easier for users to export this data to CSV. The menu item `File > Save measurements` offers hope where none is to be found. To export, one needs to type `pipeline.selectedDataSource.toDataFrame().to_csv('my_csv_name.csv')` into the shell. 

**Describe the solution you'd like**
- Make `File > Save measurements` not dependent on `objectMeasures`
- Make it easy to export data to CSV (`File > export`)
- Make it easy to plot two variables from a data source against one another, either as a popup or in a new tab on PYMEVis (`View > Plot variables`)

**Are you prepared to work on it?**
Yes

David-Baddeley:
Probably best handled by reinstating the array viewer.

David-Baddeley:
PYME.ui.recArrayView as base

David-Baddeley:
Going to merge #624 in the interim, and put integrating with #594 on the todo list. Will fix some of the making sure it displays on top of #633.

Wishlist for future work (copied in part from comments on #624, some of this already in #594):

- sorting
- vertical recipe integration
- make sure this displays for measurements
- graphing
- summary statistics?
- to excel / to jupyter buttons??? (single click to open a notebook pre-populated with code to load this data)???
- fitting?? Integrate with e.g. line profile stuff?

PR #594, opened 2020-11-07
New module for visualizing pipeline data in a table

kkhchung:
Addresses issue #584.

enhancement

**Proposed changes:**
Module for viewing pipeline data in a table.
Start by calling `pipelineTable.Plug(MainWindow)` in VisGUI shell.
Support cell highlighting and sorting.
Support basic hist/scatter plots.

The pipeline is copied to a recarray for fast data access. Needs to be refreshed when the pipeline changes.

**Checklist:**

- [ ] Tested with numpy=1.14
1.16.6 and 1.19.1
2.7 and 3.7
4.04

### DIFF
--- a/PYME/DSView/modules/pipelineTable.py
+++ b/PYME/DSView/modules/pipelineTable.py
@@ -1,0 +1,496 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Mon Aug 25 16:14:38 2014
+
+@author: Kenny
+
+#PipeViewer
+
+Addon to VisGUI
+View the pipeline in a tabular format. Support sorting and highlighting by similarity.
+
+Global variables:
+*****************
+VisFr: Reference to the VisGUI MainWindow.
+pipeline: Reference to VisGUI pipeline. Reroute to use any arbitrary data. Require Dict-like object.
+    A new cached verions is built everytime the table is refreshed.
+viewingGrid: Reference to the wx.grid.Grid derived Grid GUI. Uses a derived PyGridTableBase for its base data.
+setupPanel: Reference to the setup panel.
+accEntryList: List for keyboard shortcuts.
+
+"""
+
+import wx
+import wx.grid
+import numpy as np
+import scipy as sp
+from scipy import misc, stats
+import collections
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Thanks! In the medium term I'd be interested in either adapting this as a generic viewer for tabular data (both the VisGUI pipeline and, e.g., the output of measurements on pixel based images), or creating a new tabular viewer that is more or less a hybrid of this and the existing `recArrayView` - copying some features but potentially omitting others, e.g. sort*.  A few quick thoughts/questions to help me better understand the code below:

- the use of global variables to store references to the pipeline (along with other stuff) has the potential to be problematic as it means that there can only ever be one table viewer (i.e. you can't have e.g. one viewer for the main pipeline and one for the results of some analysis on a generated image).
- file position in hierarchy seems a bit off - this would seem to be a VisGUI plugin, but is in the dh5view plugins folder (this is something that I can easily change on merging, just wanted to clarify intent)
- how necessary for performance is the caching? In general access to the pipeline output should be pretty fast and it would be nice to avoid a copy if we can. I can see an argument that the caching might be important when displaying sorted output (I can think of a more memory-efficient way of doing this but want to know I'm on the right track before suggesting it). 

(*) the reason why I might want to omit sort in a generic tabular viewer is that it would be nice if what you saw in the viewer was the same as what you got when you saved a .csv or similar when running a recipe automatically. I'm not yet decided on how important that is.

kkhchung:
- I was just using this to view the pipeline so I wasn't worried about using global variables. It probably does need quite a bit of rewrite for use as a generic tabular viewer.
- I have no opinion on the file hierarchy since I'm not familiar with it at all. Sorry if I put it in the wrong place.
- Without caching there is a delay with updating the table when scrolling. Not unbearable depending on how you use it. I left the code in if you want to comment it in for testing. [Link](https://github.com/python-microscopy/python-microscopy/pull/594/files#diff-88588d40e947133cda0947dc1eb80bfbc2e68c5b46356641be6968e7a717fb9aR384-R386)
- The sort function is performed by the wx grid class in the GUI. I don't think the underlying object is actually sorted. Right now if you right click on the column heading it cycles between ascend, descend and unsorted.

Issue #162, opened 2020-02-27
Cannot scroll through docstrings within Bakeshop on Windows 10


zacsimile:
When the docstring shows up in the add module window in Bakeshop, if it's longer than the window, a scroll bar pops up, but the scroll bar doesn't work.

David-Baddeley:
What wx version, python version? (I suspect this is a wx4 bug, but hard to be sure)

zacsimile:
wxpython 4.0.4
python 3.6.7

David-Baddeley:
Platform? Any console output that might be useful?

zacsimile:
Platform is Windows 10. I didn't see any useful console output, but maybe I can force it.

zacsimile:
No console output at all.

David-Baddeley:
Working fine with wx4, py3 on OSX.

PR #595, opened 2020-11-07
Bug fix so that scrolling in the recipe module dialog is possible

kkhchung:
Addresses issue #162  .

bugfix

**Proposed changes:**
Change the parent in the `HtmlWindow` constructor from the wx.Dialog to the wx.Panel
Tested on Windows 10

**Checklist:**

- [ ] Tested with numpy=1.14
2.7 and 3.7
4.0.4
- [x] Does the PR avoid variable renaming in existing code, whitespace changes, and other forms of tidying? [There is a place for code tidying, but it makes reviewing much simpler if this is kept separate from functional changes]

### DIFF
--- a/PYME/recipes/recipeGui.py
+++ b/PYME/recipes/recipeGui.py
@@ -234,7 +234,7 @@ class ModuleSelectionDialog(wx.Dialog):
 
         hsizer.Add(self.tree_list, 1, wx.EXPAND|wx.ALL, 2)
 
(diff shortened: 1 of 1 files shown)


zacsimile:
Heck yeah! Thanks, @kkhchung! This has bothered me for ages, but never enough to figure out how to fix it. Confirmed this works on Windows and doesn't change the behavior on Mac (which has always worked).

David-Baddeley:
Thanks! Well done spotting a pretty non-obvious bug.

Issue #596, opened 2020-11-09
[BUG] dbscan eps default is silly


barentine:
**Describe the bug**
```
 Error whilst running Analysis>Clustering>DBSCAN Clump
==============================================
python-microscopy=20.11.07
python=3.6.9, platform=linux
numpy=1.16.5, wx=4.0.4 gtk2 (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/home/smeagol/code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/home/smeagol/code/python-microscopy/PYME/LMVis/Extras/clusterAnalysis.py", line 93, in OnClumpDBSCAN
    recipe.add_modules_and_execute([clumper,])
  File "/home/smeagol/code/python-microscopy/PYME/recipes/base.py", line 878, in add_modules_and_execute
    self.execute()
  File "/home/smeagol/code/python-microscopy/PYME/recipes/base.py", line 641, in execute
    m.execute(self.namespace)
  File "/home/smeagol/code/python-microscopy/PYME/recipes/localisations.py", line 443, in execute
    self.searchRadius, self.minClumpSize)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/sklearn/cluster/_dbscan.py", line 144, in dbscan
    est.fit(X, sample_weight=sample_weight)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/sklearn/cluster/_dbscan.py", line 313, in fit
    raise ValueError("eps must be positive.")
ValueError: eps must be positive.


```

**To Reproduce**
1. open PYMEVis
2. Clustering > DBSCAN
3. OK

PR #597, opened 2020-11-09
make dbscan defaults not zero / doc

barentine:
Addresses issue #596 .

**Checklist:**

- [x] Tested


David-Baddeley:
This is case where I'm not sure we can define a sensible default. In an ideal world we wouldn't set a default but rather force the user to actually think about what they want to put here, at least for search radius. This would require some UI changes (potentially just the trait validation routine). 

In the interim this seems reasonable.

Issue #599, opened 2020-11-09
[BUG] recipe rules using inputsByTask get all tasks assigned to a single node


barentine:
**Describe the bug**
launching a recipe from the cluster over multiple series, using e.g. `run` view, results in all (in this case 22 tasks) getting assigned to a single node

**To Reproduce**
1. run either clusterUI backend
2. navigate to the run view
3. glob multiple series, select a recipe, launch
4. watch them all get assigned to the same node

**Expected behavior**
bidding on tasks to actually bid on tasks rather than rules

**Additional context**
failed recipe, but
<details> 
```
===========================================================================
Error in rule: 000014-4b2f026e1cc846aaab3988a3d26e03ff running task: 21 on drno:19399

taskDescr:
----------
{'id': '000014-4b2f026e1cc846aaab3988a3d26e03ff~21', 'type': 'recipe', 'inputs': {'input': 'pyme-cluster:///Bergamot/2020_10_27/EMC07_02_S1/A1/000/analysis/27_10_series_T.h5r'}, 'taskdefRef': 'pyme-cluster://villains/RECIPES/emc0702s1-map-and-omero.yaml'}

Traceback:
----------
Traceback (most recent call last):
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/taskWorkerHTTP.py", line 340, in computeLoop
    recipe.execute()
  File "/home/ubuntu/code/python-microscopy/PYME/recipes/base.py", line 640, in execute
    m.check_inputs(self.namespace)
  File "/home/ubuntu/code/python-microscopy/PYME/recipes/base.py", line 137, in check_inputs
    raise MissingInputError('Input "%s" is missing from namespace; keys: %s' % (input, keys))
PYME.recipes.base.MissingInputError: Input "Events" is missing from namespace; keys: ['FitResults']



===========================================================================
Error in rule: 000014-4b2f026e1cc846aaab3988a3d26e03ff running task: 20 on drno:19399

taskDescr:
----------
{'id': '000014-4b2f026e1cc846aaab3988a3d26e03ff~20', 'type': 'recipe', 'inputs': {'input': 'pyme-cluster:///Bergamot/2020_10_27/EMC07_02_S1/A1/000/analysis/27_10_series_M.h5r'}, 'taskdefRef': 'pyme-cluster://villains/RECIPES/emc0702s1-map-and-omero.yaml'}

Traceback:
----------
Traceback (most recent call last):
[...]

PR #601, opened 2020-11-10
ruleserver bidding updates

David-Baddeley:
Addresses issue #599.

Mostly enhancement - make the ruleserver enforce rather than rely on good bidding behaviour.

**Proposed changes:**

Keep current (fast and simple) behaviour if all bids are local, implement actual bidding if they are not. Actual bidding is probably not particularly fast and not something we'd want to be doing on every localisation task, hence the short-circuit behaviour if all costs under a given threshold.

**NOTE:** this is completely untested at this point, and both the bid lifetime and threshold might need tweaking. Throwing up in current state to assist discussion on #599, but needs testing prior to merge.

### DIFF
--- a/PYME/cluster/rulenodeserver.py
+++ b/PYME/cluster/rulenodeserver.py
@@ -224,7 +224,8 @@ class NodeServer(object):
                     if n_tasks >= n_tasks_to_request:
                         break
                         
-                if n_tasks < n_tasks_to_request:
+                if n_tasks < 1:
+                    # only bid on non-local if we haven't found any local tasks
                     #logger.debug('Found %d local tasks' % n_tasks)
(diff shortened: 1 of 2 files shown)


PYME/cluster/rulenodeserver.py
```diff
@@ -224,7 +224,8 @@ def _update_tasks(self):
                     if n_tasks >= n_tasks_to_request:
                         break
                         
-                if n_tasks < n_tasks_to_request:
+                if n_tasks < 1:
+                    # only bid on non-local if we haven't found any local tasks
```
barentine, line 228: I like this change, easy enough to pull other tasks on the next loop iteration

PR #606, opened 2020-11-13
consolidate verify results filename in nameUtils

barentine:
Addresses issue same function in a couple files. No functional changes

**Proposed changes:**
- move verify_cluster_results_filename to nameUtils instead of each of our task-post approaches

**Checklist:**

- [x] This PR is just tidying, but did not go wild re: camelCase, etc.


David-Baddeley:
I think we're down to having it in one place (rules.py) now which makes the refactoring redundant. Likely easier to close than resolve conflicts.

Issue #617, opened 2020-11-18
[BUG]PYMEClusterOfOne not working


HelenSun09:
**Describe the bug**
Setting icon
INFO:PYME.cluster.cluster_of_one:Launching data server: root=C:\Users\Helen Sun\PYMEData
INFO:PYME.cluster.cluster_of_one:Launching rule server
'C:\Users\Helen' is not recognized as an internal or external command,
operable program or batch file.
'C:\Users\Helen' is not recognized as an internal or external command,
operable program or batch file.
INFO:PYME.cluster.cluster_of_one:Launching node server
'C:\Users\Helen' is not recognized as an internal or external command,
operable program or batch file.
INFO:PYME.cluster.cluster_of_one:Shutting down cluster
Shutting down cluster
Traceback (most recent call last):
  File "C:\Users\Helen Sun\miniconda3\envs\pyme\Scripts\PYMEClusterOfOne-script.py", line 33, in <module>
    sys.exit(load_entry_point('python-microscopy', 'console_scripts', 'PYMEClusterOfOne')())
  File "c:\users\helen sun\python-microscopy\PYME\cluster\cluster_of_one.py", line 179, in main
    cluster.launch(gui=options.ui, clusterUI=options.clusterui)
  File "c:\users\helen sun\python-microscopy\PYME\cluster\cluster_of_one.py", line 100, in launch
    self._launch_ruleserver_ui()
  File "c:\users\helen sun\python-microscopy\PYME\cluster\cluster_of_one.py", line 77, in _launch_ruleserver_ui
    webbrowser.open_new_tab(list(ruleservers.values())[0])
IndexError: list index out of range

**To Reproduce**
1. Use an windows 10 system, python = 3.6
2. Go to miniconda3
3. conda activate pyme
4. PYMEClusterOfOne

barentine:
Thanks for posting @HelenSun09. This can happen if the computer name is too long, see #548 and #546. In another terminal, what happens if you run the command `PYMERuleServer`? If anything is printed, would you mind posting it here in full?

barentine:
Oh, sorry, ignore that. There's a space in your home directory file name, and we'll need to handle that in the PYMEClusterOfOne launch script at the very least.

HelenSun09:
Thank you so much for the fix @barentine.

David-Baddeley:
Not tested with spaces in path, but likely to work. Can be re-opened if needed

HelenSun09:
[...]

PR #619, opened 2020-11-19
double quote sys.executable to handle spaces in paths

barentine:
Addresses issue #617 .

bit of both - allow people with spaces in their python path to run PYMEClusterOfOne
**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6


David-Baddeley:
Is this checked on both windows and non-windows platforms?  Just smells a little like something which might have platform dependant behaviour ...

barentine:
I’ve only tested on osx 10.15

David-Baddeley:
(after testing on win10)

Issue #618, opened 2020-11-19
[BUG] PYMEClusterOfOne Dataserver opening on port 0 does not get found by clusterUI / clusterIO.get_status


barentine:
**Describe the bug**
running PYMEClusterOfOne results in no dataserver showing up in clusterUI (and no files showing up in clusterUI) and no cluster freespace update in PYMEAcquire. However, running `PYMEClusterOfOne` followed by opening another terminal and running `PYMEDataServer` opens a datserver which is discovered

difference is the PYMEClusterOfOne call forces to port 0

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] osx 10.15
 - Python version: 3.6
 - numpy version: 1.16
 - wxpython version [if a UI bug]: 4.0.4

barentine:
port isn't actually the issue - problem is that local flag forces registration with sql nameserver HTTPDataServer, while clusterIO checks config settings for get_ns (in case you have hybrid ns disabled)

PR #620, opened 2020-11-19
Check hybridns config before choosing nameserver for local HTTPDataServer

barentine:
Addresses issue #618 .

bugfix
**Proposed changes:**
- if we are not using hybrid nameserver in clusterIO, make sure PYMEDataServer registers with zeroconf even if we are just running local

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

### DIFF
--- a/PYME/cluster/HTTPDataServer.py
+++ b/PYME/cluster/HTTPDataServer.py
@@ -898,11 +898,12 @@ def main(protocol="HTTP/1.0"):
 
     if options.advertisements == 'local':
         # preference is to avoid zeroconf on clusterofone due to poor
-        # performance on windows on crowded networks (for oldish versions at least)
+        # performance on crowded networks
         if config.get('clusterIO-hybridns', True):
             ns = sqlite_ns.getNS('_pyme-http')
         else:
             # if we aren't using the hybridns, we are using zeroconf in clusterIO
+            # TODO - warn that we might run into performance issues???
             ns = pzc.getNS('_pyme-http')
         server_address = ('127.0.0.1', int(options.port))
         ip_addr = '127.0.0.1'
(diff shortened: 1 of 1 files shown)


PYME/cluster/HTTPDataServer.py
```diff
@@ -897,7 +897,13 @@ def main(protocol="HTTP/1.0"):
     os.chdir(options.root)
 
     if options.advertisements == 'local':
-        ns = sqlite_ns.getNS('_pyme-http')
+        # preference is to avoid zeroconf on clusterofone due to poor
+        # performance on windows on crowded networks (for oldish versions at least)
```
David-Baddeley:
```suggestion
        # performance on crowded networks
```
performance might be improved in more recent versions, but it will still be an issue - it's fundamentally not the zeroconf modules fault but rather a number of poorly behaved clients that generate excessive traffic. The only real solution is to run zeroconf on another port.

David-Baddeley: also not strictly windows specific (although potentially worse on windows)

PYME/cluster/HTTPDataServer.py
```diff
@@ -897,7 +897,13 @@ def main(protocol="HTTP/1.0"):
...
+        if config.get('clusterIO-hybridns', True):
+            ns = sqlite_ns.getNS('_pyme-http')
+        else:
+            # if we aren't using the hybridns, we are using zeroconf in clusterIO
```
David-Baddeley, line 905:
```suggestion
            # if we aren't using the hybridns, we are using zeroconf in clusterIO
            # TODO - warn that we might run into performance issues???
```

PYME/cluster/HTTPDataServer.py
David-Baddeley: Worst case I've encountered so far is a network with a) lots of windows machines, b) lots of itunes installs on windows, and c) lots of labview installs, as each of these generates a bunch of excess mDNS traffic.

PR #624, opened 2020-11-21
make the recarray viewer also handle tabular data (and add save)

David-Baddeley:
Addresses issue #584.

**Proposed changes:**

Makes the existing recarray view handle tabular data as well and adds a save button.
- currently implemented as a drop in replacement for the couple of recarray views
- launchable by clicking on a tabular item in the pipeline

TODOs:

- launch / add tab automatically for certain analyses
- integrate plotting and potentially sorting from #594
- a bit laggy for full localisation data.

### DIFF
--- a/PYME/DSView/modules/blobFinding.py
+++ b/PYME/DSView/modules/blobFinding.py
@@ -175,7 +175,7 @@ class BlobFinder(Plugin):
         self.objPosRA = numpy.rec.fromrecords(self.dsviewer.view.points, names='x,y,z')
 
         if self.vObjPos is None:
-            self.vObjPos = recArrayView.recArrayPanel(self.dsviewer, self.objPosRA)
+            self.vObjPos = recArrayView.ArrayPanel(self.dsviewer, self.objPosRA)
             self.dsviewer.AddPage(self.vObjPos, caption='Object Positions')
         else:
             self.vObjPos.grid.SetData(self.objPosRA)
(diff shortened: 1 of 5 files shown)


zacsimile:
I gave it a whirl. Pretty cool! A few thoughts.

- The title should read `Data table - <Data Source Name>`. This is especially important if you launch more than one of these views at a time.
- The window that pops up is too large. Need to drag around a bit to resize, kind of inconvenient. 
- Save should default to `.csv`, as I think this is the more common use case. You can already save to `.hdf` from the `File` menu.
- I wish I could click on the line in between tiles (and the area between the line and the data source name) to launch the tabular view. It's slightly annoying to have to double-click just on the data source name only, and I think this will make this feature less discoverable. A bigger target would be helpful.
- It would be nice if this worked on the `Data pipeline` view down the left side of the viewing window.
- A bit weird that the values all appear editable, but clearly are not
- Clicking on a column to sort would be nice

David-Baddeley:
**Save should default to .csv, as I think this is the more common use case. You can already save to .hdf from the File menu.**
Might arguably be the more common at present, but I'd like, e.g. a save results, open in jupyter workflow to become more common. HDF saving is going to give you much better performance (both saving and opening again) and is probably the best option if you are using either jupyter or matlab for subsequent processing, .csv is the excel fallback.  Note that most use cases do not currently have any saving available from the menu.

**I wish I could click on the line in between tiles (and the area between the line and the data source name) to launch the tabular view. It's slightly annoying to have to double-click just on the data source name only, and I think this will make this feature less discoverable. A bigger target would be helpful.**
Should only need a single-click on the label. Could look at making lines clickable too, but might be a decent bit of work (and hard to maintain across different recipe viewers). Ultimately clicking on a label in the recipe is a somewhat secondary use case - we'd ideally pop this up or add as a tab whenever we generate results that it makes sense for the user to see (e.g. `Analysis-->Measure objects` should pop up one of these with the results. I.E discoverability should be less of an issue.

**It would be nice if this worked on the Data pipeline view down the left side of the viewing window.**
**A bit weird that the values all appear editable, but clearly are not**
Will call these two lowish priority for now.

**Clicking on a column to sort would be nice**
Needs some form of caching for acceptable performance (and a way of sorting the row labels). Punting until we integrate more of #594

PR #625, opened 2020-11-21
update point generation for PAINT and add online help

csoeller:

This in an enhancement of the synthetic data generation plugin.

**Proposed changes:**

This enables the PAINT mode if simulation (eventify2).

In addition, I have attempted to lay a simple infrastructure for online help. This could probably be further streamlined and will be subject to comment by David et al.

This is as yet incomplete as I need info for a few more things to provide proper help texts:

- the meaning of the scale factor parameter and if it should be exposed
- details for the wormlike source parameters and possibly for the file source to be able to construct suitable online help

I appreciate that full online help for all modules is a tough ask but hopefully shows what we may aspire to. Also, I anticipate this module being used by a biology student and it would be good to have an explanation with the functionality.

**Checklist:**

- [ ] Tested with numpy=1.14

Not yet
Should not...
NA
Should not
not to my knowledge

### DIFF
--- a/PYME/LMVis/Extras/pointSetGeneration.py
+++ b/PYME/LMVis/Extras/pointSetGeneration.py
@@ -108,7 +108,7 @@ class ImageSource(PointSource):
     helpInfo = {
         'points_per_pixel' : '''
 Select average number of points (dye molecules or docking sites) per pixel in the image mask.
-The number can be a floating point fraction, e.g. 0.1 or 3.7, a suitable choice will depend
+The number is a floating point fraction, e.g. 0.1, and shouldn't exceed 1. It is used for Monte-Carlo rejection of positions and larger values (>~0.2) will result in images which have visible pixel-grid structure because the Monte-Carlo sampling is no longer a good approximation to random sampling over the grid. If this is a problem for your application / you can't get high enough density without a high acceptance fraction, use an up-sampled source image with a smaller pixel size. 
 on the pixel size of your mask image.
 ''',
         'image' : '''
@@ -209,7 +209,6 @@ The background intensity in units of photons, typically in the range from a few 
         'mode' : '''
 With the simulation mode you can choose between STORM or PAINT mode. 
 This parameter effects how event rate changes with time (it stays constant in PAINT mode).
-It has little effect otherwise.
 '''
     }
 
@@ -329,7 +328,6 @@ It has little effect otherwise.
 def Plug(visFr):
     """Plugs this module into the gui"""
     visFr.pt_generator = Generator(visFr)
-
 
 
 
--- a/PYME/LMVis/Extras/pointSetGeneration.py
+++ b/PYME/LMVis/Extras/pointSetGeneration.py
@@ -105,6 +105,37 @@ class ImageSource(PointSource):
     #name = Str('Density Image')
     #foo = Enum([1,2,3,4])
 
+    helpInfo = {
+        'points_per_pixel' : '''
+Select average number of points (dye molecules or docking sites) per pixel in the image mask.
+The number can be a floating point fraction, e.g. 0.1 or 3.7, a suitable choice will depend
+on the pixel size of your mask image.
+''',
+        'image' : '''
+Select an image from the list of open images.
+Note that you need to open or generate the mask image you want to use so that this
+list is not empty. The mask image will be normalised for the purpose of the simulation,
+with its maximum set to 1.
+''',
+    }
+    
+    def helpStr(self, name):
+        def cleanupHelpStr(str):
+            return str.strip().replace('\n', ' ').replace('\r', '')
+        
+        return cleanupHelpStr(self.helpInfo[name])
+
+    def default_traits_view( self ):
+        from traitsui.api import View, Item, EnumEditor, InstanceEditor
+
+        traits_view = View(Item('points_per_pixel',help=self.helpStr('points_per_pixel'),
+                                tooltip='mean number of marker points per pixel'),
+                           Item('image',help=self.helpStr('image'),
+                                tooltip='select the mask image from the list of open images'),
+                           buttons = ['OK', 'Help'])
+        
+        return traits_view
+                           
     def getPoints(self):
         from PYME.simulation import locify
         print((self.image))
@@ -139,8 +170,6 @@ class ImageSource(PointSource):
 
 
 
-
-
 class Generator(HasTraits):
     meanIntensity = Float(500)
     meanDuration = Float(3)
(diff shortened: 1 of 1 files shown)


csoeller:
This shows the help in action, mouse tool tips also should work.

Screenshot 2020-11-21 at 13 47 45

David-Baddeley:
I'm not quite sure how the help gets shown ... is `helpInfo` a magic traits keyword?

csoeller:
No, the way I have coded this is purely to keep things in a sort of cleanish way. The only way that traits appears to understand this is via the `Item` `help` and `tooltips` keywords. But giving longer help text in situ looks awfully messy.

If you can think of a cleverer way of structuring this happy to see that instead.

Both help and tooltips are shown in mouse overs (help on variable names and tooltips on edit item).


The help is shown when you use the help button.

David-Baddeley:
Cool, had just managed to work it out. Tempted to merge in the meantime then sort out later, but an ideal solution would have the following properties:

- Have the docs in the module docstring so the documentation is also available for programatic usage and gets into our generated html documentation, rather than in a traitsui specific way. This is how recipe modules are currently documented
- Have a wraparound traits editor which shows this documentation (the recipe editor currently does this when selecting modules to add in `recipes.recipeGUI.OnSelect()`, but it would be nice to show this when editing exisiting modules and also for traits objects outside the recipe context)
- What I imagine is something where clicking a help button generates a new column in the editing dialog which show the generated help text from the docstring in a similar way to how they are shown in the recipe editor.

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -105,6 +105,37 @@ class ImageSource(PointSource):
     #name = Str('Density Image')
     #foo = Enum([1,2,3,4])
 
+    helpInfo = {
+        'points_per_pixel' : '''
+Select average number of points (dye molecules or docking sites) per pixel in the image mask.
+The number can be a floating point fraction, e.g. 0.1 or 3.7, a suitable choice will depend
```
David-Baddeley:
```suggestion
The number is a floating point fraction, e.g. 0.1, and shouldn't exceed 1. It is used for Monte-Carlo rejection of positions and larger values (>~0.2) will result in images which have visible pixel-grid structure because the Monte-Carlo sampling is no longer a good approximation to random sampling over the grid. If this is a problem for your application / you can't get high enough density without a high acceptance fraction, use an up-sampled source image with a smaller pixel size. 
```

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -105,6 +105,37 @@ class ImageSource(PointSource):
     #name = Str('Density Image')
     #foo = Enum([1,2,3,4])
 
+    helpInfo = {
+        'points_per_pixel' : '''
+Select average number of points (dye molecules or docking sites) per pixel in the image mask.
```
David-Baddeley: mask is technically inaccurate here as the image is a density image, not necessarily a binary one. Acceptance probability = image*points_per_pixel.

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -105,6 +105,37 @@ class ImageSource(PointSource):
...
+The number can be a floating point fraction, e.g. 0.1 or 3.7, a suitable choice will depend
+on the pixel size of your mask image.
+''',
+        'image' : '''
+Select an image from the list of open images.
+Note that you need to open or generate the mask image you want to use so that this
```
David-Baddeley: Not technically a mask

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -139,37 +170,83 @@ def refresh_choices(self):
...
+        'meanIntensity' : '''
+This parameter specifies the mean number of photons in an event.
+Typically values are in the range from 100 to several 10000.
+''',
+        'meanDuration' : '''
+The mean duration of events which is specified in units of frames.
+''',
+        'meanTime' : '''
+This parameter, the mean time of the series in frame units, is related to the series duration.
+For PAINT mode this is half the duration of the series, for STORM simulation mode the relationship is a little more complex.
```
David-Baddeley: in both cases, `meanTime` is the average time at which you expect to get events (i.e. target value of `np.mean(pipeline['t'])`) . STORM mode draws event time from an exponential distribution, PAINT from a uniform one.

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -139,37 +170,83 @@ def refresh_choices(self):
...
+        'meanEventNumber' : '''
+This parameter specifies the mean number of times an event occurs at a single marker location.
+''',
+        'backgroundIntensity' : '''
+The background intensity in units of photons, typically in the range from a few tens to hundreds of photons.
+''',
+        'mode' : '''
+With the simulation mode you can choose between STORM or PAINT mode. 
+This parameter effects how event rate changes with time (it stays constant in PAINT mode).
+It has little effect otherwise.
```
David-Baddeley:
```suggestion

```

It's only effect is to change the distribution from which event times are chosen - STORM=exponential, PAINT=uniform.

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -139,37 +170,83 @@ def refresh_choices(self):
...
-                        Item('meanTime'),
+                        Item('meanIntensity',tooltip='mean photon number of events',
+                             help=self.helpStr('meanIntensity')),
+                        Item('meanDuration',tooltip='mean duration of events in units of frames',
+                             help=self.helpStr('meanDuration')),
+                        Item('meanEventNumber',tooltip='mean number of times events occurs at a single marker location',
+                             help=self.helpStr('meanEventNumber')),
+                        Item('meanTime',tooltip='mean time of the series, roughly related to series duration, in frame units',
+                             help=self.helpStr('meanTime')),
+                        Item('scaleFactor',tooltip='unsure about this parameter'),
```
David-Baddeley: Probably leave out of view ... this is the standard deviation of the PSF in pixels. It is used (in combination with the background intensity, which is per pixel) to derive a detection threshold.

PYME/LMVis/Extras/pointSetGeneration.py
csoeller: I will add a sentence to this end. Note that the typical user will be an experimenter to whom the concept of a mean time may mean very little. So it will be helpful to give something related to what they know about as well, which IMHO is series duration.

csoeller:
Thanks for comments, @David-Baddeley. I have tried to take into account in latest commit.

Also fully onboard with the docstring idea. I suppose we would need a way to associate the docstrings from function parameters with traits varaibles to match this up properly. Would be good to come up with a suitable strategy and support infrastructure to eventually encourage code and plugin writers to gradually adopt this throughout the code base.

Issue #622, opened 2020-11-20
[BUG] Running pointwise colocalization on a 3-color image throws non-descriptive error


zacsimile:
**Describe the bug**
If we run `Analysis>Pointwise Colocalisation` on a 3-color image, we get the following traceback.

```
Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/Extras/pointwiseColoc.py", line 25, in OnPointwiseColoc
    distColoc.calcDistCorr(self.visFr.pipeline.colourFilter, *(self.visFr.pipeline.colourFilter.getColourChans()[::1]))
  File "/Users/zachcm/Code/python-microscopy/PYME/Analysis/Colocalisation/distColoc.py", line 47, in calcDistCorr
    hAA = DistHist.distanceHistogram(xA, yA, xA, yA, nbins, binsize)
TypeError: an integer is required (got type str)
```

This is because we pass all 3 color channel names as `*args`, and so the variable `nbins` gets replaced by a string.

**To Reproduce**
1. Open any 3-channel (or greater) file
2. `Analysis>Pointwise Colocalisation`

**Expected behavior**
This is open to discussion. I can see us addressing this in a number of ways.

- Tell the user to use a 2-color image and don't run (seems unnecessarily restrictive).
- Clip the length of `*args` passed to 2. This will work, but the user will not be able to analyze the color channels of their choosing.
- Have a `PYME.DSView.modules.coloc.ColocSettingsDialog` pop up to allow the user to select the two color channels of their choice.
- Extend the capabilities of `Pointwise Colocalisation` to run through every combination of channels and pop up multiple graphs.

Curious to hear your thoughts on this, @David-Baddeley.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15.7
 - Python version: 3.6.10
 - numpy version: 1.16.6
 - wxpython version [if a UI bug]: 4.0.4

David-Baddeley:
Popping up a chooser is probably the best, if also the most work. At a pinch, truncating along with printing/displaying a warning might be acceptable I don't really like the do all possible combinations option.

zacsimile:
Chooser implemented in #628.

PR #628, opened 2020-11-21
Add ColocSettingsDialog to pointwiseColoc

zacsimile:
Accounts for images with … 3+ color channels

Addresses issue #622.

Yes

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/LMVis/Extras/pointwiseColoc.py
+++ b/PYME/LMVis/Extras/pointwiseColoc.py
@@ -21,8 +21,18 @@ class PointwiseColocaliser:
         #from PYME import mProfile
         #mProfile.profileOn(['distColoc.py'])
         from PYME.Analysis.Colocalisation import distColoc
+        from PYME.DSView.modules.coloc import ColocSettingsDialog
+
+        chans = self.visFr.pipeline.colourFilter.getColourChans()[::1]
+
+        dlg = ColocSettingsDialog(self.visFr, names=chans, show_bins=False)
+        dlg.ShowModal()
+
+        selected_chans = [chans[int(x)] for x in dlg.GetChans()]  # re-index into chans b/c calcDistCorr expects full channel names
+        dlg.Destroy()
+
         #A vs B
-        distColoc.calcDistCorr(self.visFr.pipeline.colourFilter, *(self.visFr.pipeline.colourFilter.getColourChans()[::1]))
+        distColoc.calcDistCorr(self.visFr.pipeline.colourFilter, *selected_chans)
         #B vs A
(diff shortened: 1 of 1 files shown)


zacsimile:
One problem with this is it will not work if a user tries this with a single color channel. Ideally we'd show a dialog in `OnPointwiseColoc`

```
if len(chans) == 0:
            import wx
            md = wx.MessageDialog(self.visFr, 'Not enough colour channels', 'Pointwise colocalization requires 2 colour channels', wx.OK)
            md.ShowModal()
            return
```

But this will still cause the recipe to execute, so there won't be a rollback. I could add a `raise` in there, but that will generate two popups. Is there a good way to alert a user to a problem with a recipe and then rollback the recipe (not add it to the pipeline) without throwing an error in the context manager?

PYME/LMVis/Extras/pointwiseColoc.py
```diff
@@ -21,8 +21,18 @@ def OnPointwiseColoc(self, event):
         #from PYME import mProfile
         #mProfile.profileOn(['distColoc.py'])
         from PYME.Analysis.Colocalisation import distColoc
+        from PYME.DSView.modules.coloc import ColocSettingsDialog
+
+        chans = self.visFr.pipeline.colourFilter.getColourChans()[::1]
```
David-Baddeley: What is the `[::1]` doing?

PYME/LMVis/Extras/pointwiseColoc.py
```diff
@@ -21,8 +21,18 @@ def OnPointwiseColoc(self, event):
...
+
+        dlg = ColocSettingsDialog(self.visFr, names=chans, show_bins=False)
+        dlg.ShowModal()
```
David-Baddeley: Check return and abort if user cancelled

David-Baddeley:
not sure what the problem with the single colour check above is ... As it stands, this doesn't add anything to the pipeline.

PYME/LMVis/Extras/pointwiseColoc.py
zacsimile: Not sure. It's a historic thing that I just copied over.

zacsimile:
You're right. I must have run something else before this.

PYME/LMVis/Extras/pointwiseColoc.py
zacsimile: Weirdly returns `ID_OK` if you click `OK` or close the dialog. Have you run into this before, @David-Baddeley? I can't see a single good reason for this.

Issue #613, opened 2020-11-18
[BUG] Measure objects has prominent placement but it's unclear how to use it


zacsimile:
**Describe the bug**
`Analysis>Measure objects` is one of the first things you see when you open the analysis menu, and one of the most likely things you'd want to use. The same tool(s) in ImageJ (`Analyze>Measure` or `Analyze>Analyze Particles`) measure area/perimeter of binary clusters. Presumably our Measure objects should also work on clusters. However, after clustering, this is the traceback

```
 Error whilst running Analysis>Measure objects
==============================================
python-microscopy=20.11.07
python=3.6.10, platform=darwin
numpy=1.16.6, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/Extras/objectMeasurements.py", line 88, in OnMeasure
    ids = set(pipeline.mapping['objectID'].astype('i'))
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 923, in __getitem__
    return self.resultsSource[keys]
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 923, in __getitem__
    return self.resultsSource[keys]
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 723, in __getitem__
    return self.resultsSource[key][self.Index][sl]
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 879, in __getitem__
    res = np.array(self.resultsSource[key])[self.Index]
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 923, in __getitem__
    return self.resultsSource[keys]
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 923, in __getitem__
    return self.resultsSource[keys]
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 468, in __getitem__
    raise KeyError('Key (%s) not found' % key)
KeyError: 'Key (objectID) not found'
```

[...]

PR #633, opened 2020-11-22
Stop-gap fixes for object measurements

zacsimile:
Addresses issue #613.

Bugfix

**Proposed changes:**
- Iterate over multiple `objectID` keys to still run measurements if we're clustering in non-standard ways
- Fix the `Delaunay` triangulation. I'm not 100% sure what was going on before, so I'm not sure this is right, but it appears to work and it makes sense.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/LMVis/Extras/objectMeasurements.py
+++ b/PYME/LMVis/Extras/objectMeasurements.py
@@ -86,6 +86,10 @@ class ObjectMeasurer:
         chans = pipeline.colourFilter.getColourChans()
 
         # If we're not using objectIDs from an image, look for other clustering labels
+        # TODO - rather than trying a few pre-set objectID alternatives, make this a dialog instead
+        # - look for objectID and carry on if present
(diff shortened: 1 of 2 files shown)


PYME/LMVis/Extras/objectMeasurements.py
```diff
@@ -85,11 +85,23 @@ def OnMeasure(self, event):
 
         chans = pipeline.colourFilter.getColourChans()
 
-        ids = set(pipeline.mapping['objectID'].astype('i'))
+        # If we're not using objectIDs from an image, look for other clustering labels
```
David-Baddeley, line 88:
```suggestion
        # If we're not using objectIDs from an image, look for other clustering labels
        # TODO - rather than trying a few pre-set objectID alternatives, make this a dialog instead
        # - look for objectID and carry on if present
        # - if not present, display a dialog "No objectID found - did you segment objects? Either cancel, 
        # segment, and come back, or choose an alternative column to use as an ID.
```

Issue #631, opened 2020-11-22
[BUG] Datasources disappear if you mess up parameters in your recipe


zacsimile:
**Describe the bug**
If you screw up parameters in your recipe so that one of your outputs returns 0 points, the datasources after this will disappear. You won't know what the problem is until you switch to the `Pipeline Recipe` tab.

![image](https://user-images.githubusercontent.com/1263313/99891402-b5cf0980-2c37-11eb-942a-b112731bbe9c.png)

**Expected behavior**
- Datasource stay selectable in Data pipeline output, but return no points
- There's a red box around the module in the Data pipeline view as well as the Pipeline recipe view

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macOS 10.15.7
 - Python version: 3.6.10
 - numpy version: 1.16.6
 - wxpython version [if a UI bug]: 4.0.4

David-Baddeley:
I'd argue that the datasources disappearing is the correct behaviour - they are not present if the pipeline doesn't execute. 

Adding a red box in the vertical view and /or greying out the data that is not present in a pipeline in both recipe views would be desirable.

David-Baddeley:
This falls into the general 'making recipes easier to debug' section of the roadmap.

PR #634, opened 2020-11-23
more recipe debug info

David-Baddeley:
Partially addresses issue #631.

debugging enhancement

**Proposed changes:**
- colour missing namespace keys differently on recipe failure
- error flagging in vertical recipe (not complete, as needs a `_layout()` call on the recipe view which is currently not triggered, but want to merge the event number update first, as it will likely conflict).

image


zacsimile:
I like the look. It's fairly intuitive. Is there a way to also have a descriptive alt text pop up when you mouse over the redded-out portions of the recipe?

David-Baddeley:
Some sort of popup text on failing recipe modules would be good. 
Could show the same error message we show in the full recipe viewer. Implementation would likely involve writing a custom caption bar for the recipe modules (as we have for the layers) and is a decent amount of work. Would give an opportunity to make the caption colouring solution much less hacky than it currently is, and potentially avoid completely rebuilding the display as much.

PR #635, opened 2020-11-23
PYMEAcquire protocol-rules GUI rewire from _rules to rules

barentine:
Addresses issue rewiring protocol-rules PYMEAcquire gui for #530.

enhancement with some bugfixes
**Proposed changes:**
- wire `PYME.Acquire.ui.rules` to `PYME.cluster.rules` instead of `PYME.cluster._rules`
- don't push the whole rule from the main thread (pushing thread init was calling the method rather than just assigning)
- default startAt to 0 rather than magic10
- when chaining rule factories, assign 'results' from outputfiles of first rulefactory to 'input' in second. This makes us consistent with default namespace for all other clusterUI / bakeshop recipe batching
- remove _rules

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4

moreOrLess

- [x] Are there any other side effects of the change?


PYME/cluster/rules.py
```diff
@@ -639,13 +650,24 @@ def chain(self, on_completion):
         """
         assert(isinstance(on_completion, RuleFactory))
         self._on_completion = on_completion
-        
+
+    @property
+    def rule_type(self):
+        return self._type
```
David-Baddeley, line 658:
```suggestion
        # TODO - rename to something like `display_name` of `display_type` to indicate that this is a UI helper function.
        # TODO - do we need a function for this, or can we just have a property?
        return self._type
```

PR #636, opened 2020-11-23
Document zoom factor

zacsimile:
Addresses issue # .

**Proposed changes:**

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/config.py
+++ b/PYME/config.py
@@ -212,7 +212,7 @@ httpspooler-chunksize : default=50, how many frames we spool in each chunk
     to. Increasing the chunksize can increase data-locality for faster analysis,
     but has spooling/writing bandwidth implications.
 
-zoom_factor : default = 1.1, adjusts zoom sensitivity by adjusting magnification factor per scroll event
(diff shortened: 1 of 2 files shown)


PYME/config.py
```diff
@@ -212,6 +212,8 @@
     to. Increasing the chunksize can increase data-locality for faster analysis,
     but has spooling/writing bandwidth implications.
 
+zoom_factor : default = 1.1, adjusts zoom sensitivity by adjusting magnification factor per scroll event
```
David-Baddeley: We need to re-name this to scope it to pymevis. May as well make it match convention as well.
```suggestion
pymevis-zoom-factor : default = 1.1, adjusts zoom sensitivity by adjusting magnification factor per scroll event
```
Will require an edit in visCore as well (I should have picked this up earlier)

zacsimile: That's fine. It will require an edit in `gl_render3D_shaders` as well. What's the `visCore` edit?

Issue #650, opened 2020-11-30
[ENHANCEMENT] supress matplotlib etc DEBUG info in dh5view (like VisGUI already does)


csoeller:
**Is your feature request related to a problem? Please describe.**

On both macOS and win10 py3 installs of PYME dh5view produces huge numbers of PYME unrelated debug statements, see screenshot:

![pyme-py3-dh5view-debug-msgs](https://user-images.githubusercontent.com/3750860/100646824-7a6ac400-3336-11eb-9f60-1586a2882c05.png)

**Describe the solution you'd like**
VisGUI has the following statements to suppress this:

```
logging.getLogger('matplotlib.font_manager').setLevel(logging.ERROR) #clobber unhelpful matplotlib debug messages
logging.getLogger('matplotlib.backends.backend_wx').setLevel(logging.ERROR)
logging.getLogger('PIL.PngImagePlugin').setLevel(logging.ERROR)
```

**Describe alternatives you've considered**
NA

**Are you prepared to work on it?**
Sure, some guidance where to insert the relevant statements would be great.

David-Baddeley:
Should be as simple as copying lines 57-61 of VisGUI.py to dsviewer.py (around line 48). The only caveat is to make sure that matplotlib and PIL have been imported before you set the log level on their loggers.

David-Baddeley:
Note: copying line 57 is not necessary for this to work, but nice anyway - means that log messages from dsviewer get tagged with the module name rather than, I think `MAIN`.

csoeller:
Cool, will raise a PR soonish...

David-Baddeley:
should be fixed with #651

PR #651, opened 2020-11-30
clobber extra debug messages

csoeller:
Addresses issue #650.

Enhancement

**Proposed changes:**

clobber extra debug messages in dsviewer.py - mimic VisGUI.py

**Checklist:**

- [ x] Tested with numpy=1.14
3.7

- [x ] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)?
No
NA
No
No

### DIFF
--- a/PYME/DSView/dsviewer.py
+++ b/PYME/DSView/dsviewer.py
@@ -45,6 +45,10 @@ except ImportError:
 
 import logging
 logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+logging.getLogger('matplotlib.font_manager').setLevel(logging.ERROR) #clobber unhelpful matplotlib debug messages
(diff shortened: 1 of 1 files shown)


PYME/DSView/dsviewer.py
```diff
@@ -45,6 +45,10 @@
 
 import logging
 logging.basicConfig(level=logging.DEBUG)
+logger = logging.getLogger(__name__)
+
+logging.getLogger('matplotlib.font_manager').setLevel(logging.ERROR) #clobber unhelpful matplotlib debug messages
+logging.getLogger('matplotlib.backends.backend_wx').setLevel(logging.ERROR)
```
csoeller, line 52: I did not add the PIL clobber statement - not sure if really applicable to dsviewer? Can add if suitable.

David-Baddeley: I'm not sure, but if you're not seeing any PIL related errors then it's probably fine.

Issue #648, opened 2020-11-25
[BUG] spooled / recipe detected tile overviews queuing shifted ROIs


barentine:
**Describe the bug**
Running a tile using the tiler panel, followed by a detection recipe in the tile viewer and `Queue from tile viewer` works, however running a tile protocol spooled with a chained detection recipe (same recipe with the addition of supertile module) and `recipes.acquisition.QueueAcquisitions`) gets bogus ROIs which are presumably shifted. Note that said recipe also saves the ROIs to disk, and `Queue ROIs from file` also does not pull legitimate ROIs (which is interesting given the difference between centre_roi_on vs. positioning.update between QueueAcquisitions and the actionUI GUI methods.

barentine:
Current status is unresolved:
two scans, circular, each 200 um radius (setting, anyway)
Both were started at (x, y) = 3500, 43200 um, with the multiview ROI already as it was for the scans.

**metadata from tiler ImagePyramid**
```
{
  "Pyramid.Depth": 3,
  "Pyramid.NTilesX": 13,
  "Pyramid.NTilesY": 13,
  "Pyramid.PixelSize": 0.115,
  "Pyramid.PixelsX": 3328,
  "Pyramid.PixelsY": 3328,
  "Pyramid.TileSize": 256,
  "Pyramid.x0": 3333.4320185778092,
  "Pyramid.y0": 43115.88078877725
}
```

**opening spooled pcs in dh5view, adding supertile module to recipe with level set to 2**
```
recipes.activeRecipe.namespace['supertile'].mdh
<NestedClassMDHandler>:

Pyramid.TileSize: 256
Pyramid.x0: 20703312.7
Pyramid.y0: 103083012.8
Pyramid.PixelSize: 0.115
Pyramid.Depth: 4
Pyramid.NTilesX: 0
Pyramid.NTilesY: 0
Pyramid.PixelsX: 0
Pyramid.PixelsY: 0
voxelsize.x: 0.46
voxelsize.y: 0.46
```

**abridged tiled pcs metadata**
```
voxelsize.x: 0.115
voxelsize.y: 0.115
voxelsize.z: 0.2
voxelsize.units: um
Camera.ROI: [180, 896, 436, 1152]
Camera.ROIHeight: 256
Camera.ROIOriginX: 180
Camera.ROIOriginY: 896
Camera.ROIWidth: 256
CameraOrientation.FlipX: False
CameraOrientation.FlipY: False
[...]

PR #654, opened 2020-11-30
use camera chip corner as origin in tile_pyramid.tile_pyramid

barentine:
Addresses issue #648.

bugfix
**Proposed changes:**
- create ImagePyramid using chip-corner referenced positions rather than raw stage positions such that `microscope.centre_rois_on` calls bring us back to the same ROI center
- use get_camera_physical_roi_origin instead of get_camera_roi_origin when preparing splitter ImagePyramid, even though it isn't the cause of #648

**Checklist:**

- [ ] Tested

### DIFF
--- a/PYME/Analysis/tile_pyramid.py
+++ b/PYME/Analysis/tile_pyramid.py
@@ -1,5 +1,5 @@
 import numpy as np
-from PYME.IO.MetaDataHandler import get_camera_roi_origin, NestedClassMDHandler
+from PYME.IO.MetaDataHandler import get_camera_physical_roi_origin, NestedClassMDHandler
 import os
 import glob
 import collections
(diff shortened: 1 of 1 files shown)


PYME/Analysis/tile_pyramid.py
```diff
@@ -576,7 +581,7 @@ def tile_pyramid(out_folder, ds, xm, ym, mdh, split=False, skipMoveFrames=False,
     weights[:, :edgeRamp, :] *= np.linspace(0, 1, edgeRamp)[None, :, None]
     weights[:, -edgeRamp:, :] *= np.linspace(1, 0, edgeRamp)[None, :, None]
     
-    roi_x0, roi_y0 = get_camera_roi_origin(mdh)
+    roi_x0, roi_y0 = get_camera_physical_roi_origin(mdh)
```
David-Baddeley:
```suggestion
    roi_x0, roi_y0 = get_camera_roi_origin(mdh)
```

PYME/Analysis/tile_pyramid.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
-from PYME.IO.MetaDataHandler import get_camera_roi_origin, NestedClassMDHandler
+from PYME.IO.MetaDataHandler import get_camera_physical_roi_origin, NestedClassMDHandler
```
David-Baddeley:
```suggestion
from PYME.IO.MetaDataHandler import get_camera_roi_origin, get_camera_physical_roi_origin, NestedClassMDHandler
```

Issue #652, opened 2020-11-30
[BUG] clusterIO.get_file OSErrors more prevalent after get_advertise_services performance regression fix


barentine:
**Describe the bug**
Performance is generally rescued by #645s, however we are dropping frames in localization analysis (and recipe work) 
![image](https://user-images.githubusercontent.com/31105780/100663760-c3198180-3324-11eb-8bed-17659b9bb863.png). Can be particularly annoying for recipe PZF sources as one frame will fail out the whole thing.

```
Error in rule: 000003-ea8d53ae02ec418faf2da28b9d1d53c2 running task: 204 on elektra:12366
taskDescr:
----------
{'id': '000003-ea8d53ae02ec418faf2da28b9d1d53c2~204', 'type': 'localization', 'taskdef': {'frameIndex': '204', 'metadata': 'PYME-CLUSTER://villains/Bergamot/2020_11_25/EMC06.05.S1/000/analysis/25_11_series_G.h5r.json'}, 'inputs': {'frames': 'PYME-CLUSTER://villains/Bergamot/2020_11_25/EMC06.05.S1/000/25_11_series_G.pcs'}, 'outputs': {'fitResults': 'http://10.150.7.3:15348/__aggregate_h5r/Bergamot/2020_11_25/EMC06.05.S1/000/analysis/25_11_series_G.h5r/FitResults', 'driftResults': 'http://10.150.7.3:15348/__aggregate_h5r/Bergamot/2020_11_25/EMC06.05.S1/000/analysis/25_11_series_G.h5r/DriftResults'}}
Traceback:
----------
Traceback (most recent call last):
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/taskWorkerHTTP.py", line 308, in computeLoop
    res = task()
  File "/home/ubuntu/code/python-microscopy/PYME/localization/remFitBuf.py", line 417, in __call__
    bufferManager.updateBuffers(md, self.dataSourceModule, self.bufferLen)
  File "/home/ubuntu/code/python-microscopy/PYME/localization/remFitBuf.py", line 113, in updateBuffers
    md['Camera.ElectronsPerCount'])
  File "/home/ubuntu/code/pyme-warp-drive/warpdrive/buffers.py", line 41, in __init__
    self.slice_shape = [int(d) for d in self.data_buffer.dataSource.getSliceShape()]
[...]

PR #656, opened 2020-11-30
Make service removal less zealous

David-Baddeley:
Addresses issue #652.

bugfix, increase timeout on `is_port_open()` to avoid erroneous removal due to server congestion.

may not be a full fix, but if the server takes >10s to respond it may as well be dead. TODO - how can we handle this more robustly??


David-Baddeley:
Added a couple more robustness improvements

David-Baddeley:
As a bonus, our tests are passing again ...

PR #661, opened 2020-12-02
recipe module and datasource to lazily swap color and time dimensions

barentine:
Addresses issue #people loading e.g. lifs converted to tif in Fiji often end up with color loaded as dim 2 rather than 3.

enhacnement
**Proposed changes:**
add a datasource to swap the z/t dim with c dim

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

### DIFF
--- a/PYME/IO/DataSources/SwapColorAndSliceDataSource.py
+++ b/PYME/IO/DataSources/SwapColorAndSliceDataSource.py
@@ -1,0 +1,52 @@
+from .BaseDataSource import BaseDataSource, DefaultList
+import numpy as np
+
+
+class DataSource(BaseDataSource):
+    moduleName = 'SwapColorAndSliceDataSource'
+    def __init__(self, datasource):
+        if len(datasource.shape) < 4:
+            raise ValueError('no')
+        self._datasource = datasource
+        self.sizeC = self._datasource.shape[3]
+        
+    def _getNumFrames(self):
+        self._datasource.shape[3]
+    
+    @property
+    def shape(self):
+        return DefaultList(self.getSliceShape() + (self.sizeC, int(self.getNumSlices()/self.sizeC)))
+    
+    def getSlice(self, ind):
+        return self[:,:,ind % self.shape[2], ind // self.shape[2]].squeeze()
+    
+    def __getitem__(self, keys):
+        keys = list(keys)
+        if len(keys) == 3:
+            keys.insert(2, slice(None))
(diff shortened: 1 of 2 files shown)


David-Baddeley:
My gut feeling it that this would be better and more intuitively handled during the import. As it stands, we don't make any real attempt to parse ImageJ metadata, and the solution is probably to do that. What happens if we load the lif directly through bioformats?

David-Baddeley:
I wonder if using a more up to date version of `tifffile` instead of our locally hacked version might not solve this without having to re-invent the wheel.

### Potential issues with using updated version:

- no support for lazy IO / keeping data out of memory when saving (we can currently turn, e.g. an 4GB HDF into an 8 GB TIFF whilst only using a few MB of memory. With stock `tifffile` we would need to pull the whole data set into memory.
- would need to push cluster awareness outside the tiff loading code (adds a bit of overhead)
- have the internals changed too much for our lazy loader to work?

If we go down this route, it would likely make sense to initially use the stock version for loading and our customised version for saving. Suspect updating is the right thing to do, but it might be less work just to cobble together an imagej metadata parser.

barentine:
the lif can apparently be a bit of a nightmare with bioformats if you acquired it using their scan/click/auto run (can't remember the name, this is all second hand, though the converted tifs I'm running with are now first-hand ;) )

David-Baddeley:
If going the datasource route, to avoid logic duplication, it would probably just be better to write a proxy datasource which passed through `getSlice()`, `getNumSlices()` and `getSliceShape()` to the source data and simply redefined `sizeC` and `additionalDims`.

The way datasources currently work is that they use `additionalDims=='TC'` or `additionalDims=='CT'` to decide how to interpret the data.

barentine:
just a note that our tif loading gets even more silly if we get a typical fiji stack which is multichannel and z. I'm getting all of it stuff into dim 2, cycling between all 4 colors on a step, then the next.

David-Baddeley:
Although changing to, e.g., an updated tifffile might help, I wonder if it makes sense to also offer a `File->Import` menu option as well as `File->Open` where we show a dialog allowing people to set the dimension order (i.e. XYTC or XYCT) and the number of colour channels. That would let people correctly load .tifs with missing / messed up metadata and would mean that we didn't need to constantly chase ImageJ changes. Might not be much use for batch processing though.

barentine:
would be useful for batch processing if we stuffed the backend into a recipe module

PR #662, opened 2020-12-02
make dh5view recipe somehow accessible from shell for debugging

barentine:
Addresses issue #no handle to module collection if one wants to play with a namespace item in the shell, etc.

enhancement
**Proposed changes:**
- add a method to get the recipe plugin modulecollection from the shell

**Checklist:**

- [ ] Tested with numpy=1.14

1.16

3.6

4.0.4


David-Baddeley:
Already accessible as `recipes.activeRecipe` in the dsviewer shell

barentine:
whoops. Thanks @David-Baddeley

Issue #674, opened 2020-12-05
[BUG] Segfault on Analysis->Measure objects in py3


zacsimile:
**Describe the bug**

```
Traceback (most recent call last):
  File "/Users/zachcm/opt/miniconda3/envs/pyme/lib/python3.7/site-packages/traits/trait_notifiers.py", line 522, in _dispatch_change_event
    self.dispatch(handler, *args)
  File "/Users/zachcm/opt/miniconda3/envs/pyme/lib/python3.7/site-packages/traits/trait_notifiers.py", line 484, in dispatch
    handler(*args)
  File "/Users/zachcm/Code/python-microscopy/PYME/ui/layerFoldPanel.py", line 98, in _refresh
    self.Refresh()
RuntimeError: wrapped C/C++ object of type LayerCaptionBar has been deleted
hl update
lw update
hl update
/Users/zachcm/Code/python-microscopy/PYME/Analysis/points/objectMeasure.py:80: RuntimeWarning: Mean of empty slice.
  measurements['xPos'] = object[:,0].mean()
/Users/zachcm/opt/miniconda3/envs/pyme/lib/python3.7/site-packages/numpy/core/_methods.py:170: RuntimeWarning: invalid value encountered in true_divide
  ret = ret.dtype.type(ret / rcount)
/Users/zachcm/Code/python-microscopy/PYME/Analysis/points/objectMeasure.py:81: RuntimeWarning: Mean of empty slice.
  measurements['yPos'] = object[:,1].mean()
/Users/zachcm/Code/python-microscopy/PYME/Analysis/points/moments.py:26: RuntimeWarning: Mean of empty slice.
  xm = x.mean()
/Users/zachcm/Code/python-microscopy/PYME/Analysis/points/moments.py:27: RuntimeWarning: Mean of empty slice.
  ym = y.mean()
/Users/zachcm/Code/python-microscopy/PYME/Analysis/points/moments.py:40: RuntimeWarning: invalid value encountered in true_divide
  m[i,j] = ((xc**i)*(yc**j)).sum()/x.size
/Users/zachcm/Code/python-microscopy/PYME/ui/recArrayView.py:55: wxPyDeprecationWarning: Using deprecated class. Use GridTableBase instead.
  gridlib.PyGridTableBase.__init__(self)
[...]

PR #675, opened 2020-12-05
Make tabular.keys() return a list

zacsimile:
Addresses issue #674.

Bugfix

**Proposed changes:**
- `dict` types were causing us to index into the wrong parts of memory in py3 (`GetValue`), switched to `list`

**Additional notes**
We'll want to re-trigger package building after accepting this fix.

### DIFF
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -669,7 +669,7 @@ class RecArraySource(TabularBase):
     _name = 'RecArray Source'
     def __init__(self, recordArray):
         self.recArray = recordArray
-        self._keys = self.recArray.dtype.names
+        self._keys = list(self.recArray.dtype.names)
 
     def keys(self):
         return self._keys
@@ -712,7 +712,7 @@ class DictSource(TabularBase):
                 raise ValueError('Columns are different lengths')
         
     def keys(self):
-        return self._source.keys()
+        return list(self._source.keys())
     
     def __getitem__(self, keys):
         key, sl = self._getKeySlice(keys)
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Would it be better to make `tabular.keys()` return a list so we don't run into this elsewhere?

I struggle to see how this could be causing a segfault though.

zacsimile:
It would be useful to make `tabular.keys()` return a list. I think it's segfaulting in `GetValue` on ` str(self._tabular[self._tabular.keys()[col]][row])`. We know `self._tabular.keys()[col]` fails, but then it still tries to index into `self._tabular`. Also supporting this theory: changing this line got rid of the segfault.

David-Baddeley:
There's still no way it should segfault (unless there is a particularly nasty python or numpy bug). The iterator / list issue should cause a python exception which can be handled as other exceptions are. It may not matter, but I'm pretty sure we're not fixing the segfault here.

zacsimile:
Won't it return a number associated with the error and then try to index into tabular with that? I guess it might not, it might just get handled like a normal error.

Either way, there was a segfault, and there isn't after changing this line (the other line change fixed the `TypeError: 'odict_keys' object is not subscriptable` error). I'm about to pass a different fix where I change the tabular keys that were not lists to lists.

David-Baddeley:
Nope, it will raise an exception - the array will never be indexed. Even if the array was indexed, numpy should catch any out of bounds errors.

My suspicion is that this is actually a wx error (wx has a habit of segfaulting when things don't go it's way) caused by an exception being thrown in out handler).

zacsimile:
Ah, fair enough. I forgot about the `IndexError`. I agree, it shouldn't segfault from indexing. Must be in `wx`.

Issue #669, opened 2020-12-04
[BUG] Pos: slider not responsive using arrow keys or intermittently responsive using mouse


mdlessard75:
After loading a multi-dimensional image (3D, multi-color, etc.) the viewing window intermittently updates the z position/channel while clicking and moving the Pos: slider. Selecting the slider and using the right and left arrow keys results in no change/update of the viewing window.

**Expected behavior**
I expect that moving the Pos: slider with the mouse or with the arrow keys will result in an updated image representing the selected z plane or color channel.

**Platform**
 - OS: Windows 10
 - Python version: 3.7.9
 - numpy version: 1.19.2
 - wxpython version [if a UI bug]: 4.0.4

zacsimile:
Gave this a quick go binding the mouse keys. For some reason `self.Bind(wx.EVT_KEY_DOWN, self.OnKeyPress)` doesn't seem to bind key pressing in `PYME.DSView.modules.playback.PlayPanel`. Is this superseded somewhere @David-Baddeley?

David-Baddeley:
Behaviour of arrow keys will be platform dependant, and shouldn't rely on explicit key bindings. On platforms where sliders do respond to key strokes (Win, linux) it should just do the right thing without any specific binding. IE - you shouldn't need to bind the keys (this probably also explains why explicitly binding them doesn't do anything - they are already consumed by the underlying control and don't propagate up to stack to where you are binding them).

Having a quick look this looks like a refresh/update issue. At some point I reduced the amount of refreshes in the viewer to improve performance (display was being refreshed multiple times and it was causing flicker / etc ...), suspect I might have taken it too far. Reproducible on OSX and Linux, py2.7 and py3.7.

For reference, pgdown/pgup and mouse wheel will also both scroll through z when the image has focus.

David-Baddeley:
[...]

PR #676, opened 2020-12-06
bind arrow-key based scrolling

David-Baddeley:
Addresses issue #669.

Probably best considered a bugfix. Make z/t position update in dsviewer when cursor keys are used with slider.

### DIFF
--- a/PYME/DSView/modules/playback.py
+++ b/PYME/DSView/modules/playback.py
@@ -58,6 +58,8 @@ class PlayPanel(wx.Panel):
 
         self.slPlayPos = wx.Slider(self, -1, 0, 0, 100, style=wx.SL_HORIZONTAL)
         self.slPlayPos.Bind(wx.EVT_SCROLL_THUMBRELEASE, self.OnPlayPosChanged)
+        self.slPlayPos.Bind(wx.EVT_SCROLL_LINEUP, self.OnPlayPosChanged)
+        self.slPlayPos.Bind(wx.EVT_SCROLL_LINEDOWN, self.OnPlayPosChanged)
 
         self.bSeekStart = wx.BitmapButton(self, -1, self.bmStartSeek)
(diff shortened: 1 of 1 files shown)


zacsimile:
I have to hit `fn` and the up or down arrow key for this to work. left/right keys don't work at all. It's weird that the scroll wheel works when I'm moused over the image but not moused over the slider

David-Baddeley:
On OSX? OSX doesn't support moving sliders with the keyboard / have a concept of slider focus (also doesn't have pgup/pgdown except as bound as fn+arrows). Basically this is win & linux only.

David-Baddeley:
We could try and hack those keys in for OSX, but I'm not sure were we'd have to bind them to get them captured.

zacsimile:
Yeah, on OSX. That's too bad, I'd like this feature on OSX.

zacsimile:
Yeah binding them in the slider itself didn't work when I gave it a go.

David-Baddeley:
I think you'd have to bind them in the playback panel (the parent window), which is kinda gross. Whatever you bind them to has to be able to capture focus. Would need to test for OSX/darwin and special case as windows and linux are likely to get confused as they handle these things properly.

zacsimile:
Oh sorry, I did bind `EVT_KEY_DOWN` in the `PlayPanel`. I put a print statement in the bound function, but it was never called.

David-Baddeley:
Just checked on my OSX, and pgdown/pgup  (fn+arrow_down, fn+arrow_up) work even after clicking in the playback panel, which means that nothing in the playback panel captures focus so they are still handled by the main viewer. This is obviously desireable, and we might not want to do anything. Not sure I'd want to bind the straight arrows, however, as they might be used for something else (at one point they changed x and y positions, although I might have changed that to the "gamer" combos of either asdw or jkli)

Issue #678, opened 2020-12-06
[BUG] PYMEAcquire fails in Linux with wxPython>=4.1.0 (wxWidgets>=3.1)


carandraug:
PYMEAcquire is incompatible with wxPython 4.1.0 release, at least on Linux and GTK.

**To Reproduce**

Steps to reproduce the behaviour:
1. Start `PYMEAcquire` from command line (so it picks the defaul `init.py` with demo devices)
2. Note that it won't start and the following erros:

```
ERROR:PYME.Acquire.acquiremainframe:Error whilst initializing Simulation UI GUI
Traceback (most recent call last):
  File "/home/carandraug/src/PyME/PYME/Acquire/acquiremainframe.py", line 238, in doPostInit
    cm.run(self, self.scope)
  File "/home/carandraug/src/PyME/PYME/Acquire/ExecTools.py", line 205, in run
    self.codeObj(parent, scope)
  File "/home/carandraug/src/PyME/PYME/Acquire/Scripts/init.py", line 92, in sim_controls
    dsc = simui_wx.dSimControl(MainFrame, scope.simcontrol)
  File "/home/carandraug/src/PyME/PYME/Acquire/Hardware/Simulator/simui_wx.py", line 425, in __init__
    self._init_ctrls(parent)
  File "/home/carandraug/src/PyME/PYME/Acquire/Hardware/Simulator/simui_wx.py", line 216, in _init_ctrls
    2)
wx._core.wxAssertionError: C++ assertion "!(flags & wxALIGN_CENTRE_HORIZONTAL)" failed at /tmp/pip-install-2004mlpb/wxpython/ext/wxWidgets/src/common/sizer.cpp(2106) in DoInsert(): Horizontal alignment flags are ignored in horizontal sizers
[...]
ERROR:PYME.Acquire.acquiremainframe:Error whilst initializing Camera controls GUI
Traceback (most recent call last):
  File "/home/carandraug/src/PyME/PYME/Acquire/acquiremainframe.py", line 238, in doPostInit
    cm.run(self, self.scope)
  File "/home/carandraug/src/PyME/PYME/Acquire/ExecTools.py", line 205, in run
    self.codeObj(parent, scope)
  File "/home/carandraug/src/PyME/PYME/Acquire/Scripts/init.py", line 100, in cam_controls
    scope.camControls['Fake Camera'] = AndorControlFrame.AndorPanel(MainFrame, scope.cam, scope)
[...]

PR #680, opened 2020-12-07
Fix 678 wxpython 4.1

carandraug:
Addresses issue #678 .

**Proposed changes:**

This fix was discussed on issue #678. I believe this will no break on wxPython 3 but I did not actually test that. This chang removes some alignment flags that are ignored in wxWidgets but that since wxPython 4.1 raise assertion errors. I did test this on wxPython 4.0.7, 4.1.0, and 4.1.1.

I found two types of alignment flags being ignored. 1) alignment flags that are meaningless in the specific sizer, e.g., vertical alignment flags makes no sense on vertical sizers; 2) alignment flags used together with the expand flag (the expand and align flag act on the window, so if the window is expanded there is no alignment to be made.

I did not remove all instances of those ignored flags, only the instances that were raising errors to me because I'm afraid to break the other parts of the code that I'm not using. However, grepping the code for this, suggests this happens in many places:

    $ grep -r wx.EXPAND PYME/ | grep wx.ALIGN | wc -l
    68

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/AndorIXon/AndorControlFrame.py
+++ b/PYME/Acquire/Hardware/AndorIXon/AndorControlFrame.py
@@ -73,15 +73,15 @@ class AndorPanel(afp.foldingPane):
         self.rbSingleShot.SetValue(False)
         #self.rbSingleShot.SetToolTipString('Allows multiple channels with different integration times and good shutter synchronisation')
         self.rbSingleShot.Bind(wx.EVT_RADIOBUTTON,self.OnRbSingleShotRadiobutton)
-        sbAqMode.Add(self.rbSingleShot, 1, wx.EXPAND|wx.ALIGN_CENTER_VERTICAL, 0)
+        sbAqMode.Add(self.rbSingleShot, 1, wx.EXPAND, 0)
 
         self.rbContin = wx.RadioButton(cp, -1, 'Continuous')
         self.rbContin.SetValue(True)
         #self.rbContin.SetToolTipString('Allows fastest speeds, albeit without good syncronisation (fixable) or integration time flexibility')
         self.rbContin.Bind(wx.EVT_RADIOBUTTON, self.OnRbContinRadiobutton)
-        sbAqMode.Add(self.rbContin, 1, wx.EXPAND|wx.ALIGN_CENTER_VERTICAL, 0)
+        sbAqMode.Add(self.rbContin, 1, wx.EXPAND, 0)
 
-        vsizer.Add(sbAqMode, 0, wx.EXPAND|wx.ALIGN_CENTER_HORIZONTAL|wx.TOP, 5)
+        vsizer.Add(sbAqMode, 0, wx.EXPAND|wx.TOP, 5)
 
         #self.bUpdateInt = wx.Button(id=wxID_ANDORFRAMEBUPDATEINT,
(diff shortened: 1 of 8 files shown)


PYME/Acquire/Hardware/AndorIXon/AndorControlFrame.py
```diff
@@ -73,15 +73,15 @@ def _createCollapsingPane(self):
         self.rbSingleShot.SetValue(False)
         #self.rbSingleShot.SetToolTipString('Allows multiple channels with different integration times and good shutter synchronisation')
         self.rbSingleShot.Bind(wx.EVT_RADIOBUTTON,self.OnRbSingleShotRadiobutton)
-        sbAqMode.Add(self.rbSingleShot, 1, wx.EXPAND|wx.ALIGN_CENTER_VERTICAL, 0)
```
David-Baddeley, line 76: I should have mentioned the tricky bit on the initial issue - it's only an error if being added to a `BoxSizer` (or `StaticBoxSizer` that was created as `wx.BoxSizer(wx.VERTICAL)`. As `sbAqMode` is a horizontal sizer, the `wx.ALIGN_CENTER_VERTICAL` flag is OK (and highly desirable - the GUI looks very unpolished if it's absent).

Likewise the `ALIGN_CENTER_HORIZONTAL` flag on a vertical sizer. Most of the usages of these flags are not an error, it's just the few where we've either changed from a horizontal sizer to a vertical (or vice-versa), or have copy-pasted sizer additions without thinking about what type of sizer we are adding to.

David-Baddeley: Oops, posted this without reading the `wx.EXPAND` bit. Need to think a bit more about it.

David-Baddeley:
Does the combination with `wx.EXPAND` cause an error? I have a feeling in some of these cases we should potentially be dropping `wx.EXPAND` rather than `wx.ALIGN_CENTER_VERTICAL` (I think I'd been using expand with the intention of permitting growth along the sizer direction, rather than orthogonal to it, but might have been doing so in error), but not 100% sure.

carandraug:
> Does the combination with wx.EXPAND cause an error?

Yes. You get errors like `wx._core.wxAssertionError: C++ assertion "!(flags & (wxALIGN_BOTTOM | wxALIGN_CENTRE_VERTICAL))" failed at /tmp/pip-install-2004mlpb/wxpython/ext/wxWidgets/src/common/sizer.cpp(2114) in DoInsert(): Vertical alignment flags are ignored with wxEXPAND`

Every single flag that this PR remove was a flag that was raising an error mentioning that it was ignored for some reason.

> I have a feeling in some of these cases we should potentially be dropping wx.EXPAND rather than wx.ALIGN_CENTER_VERTICAL (I think I'd been using expand with the intention of permitting growth along the sizer direction, rather than orthogonal to it, but might have been doing so in error), but not 100% sure.

The error message says that the alignment flags were being ignored so that's what I removed. I don't know about the look feel you want, not sure how you want to fix this then.

David-Baddeley:
If they were all raising errors, that's good enough for me (and means that they won't change the existing look and feel). Which one we *should* have been using might be something for me to file away and address if/when we're tweaking the appearance again.

Issue #64, opened 2018-09-18
saving multicolor ImageStacks as h5 drops a color channel


David-Baddeley:
**[Original report](https://bitbucket.org/david_baddeley/python-microscopy/issue/64) by Andrew Barentine (Bitbucket: [AndrewBarentine](https://bitbucket.org/AndrewBarentine), ).**

----------------------------------------

Saving multicolor ImageStacks as tif's is functional, however the 0th color channel is the only one that gets saved when saving h5 (either from the menu -> save as, or from the shell, ImageStack.Save('[something].h5')

David-Baddeley:
**Original comment by David Baddeley (Bitbucket: [david_baddeley](https://bitbucket.org/david_baddeley), ).**

----------------------------------------

I'm going to call this a missing feature, rather than a bug. For most use cases, OME TIFF is a better choice for multicolour stacks. To support multicolour stacks in .h5 would require quite a fundamental re-think of how we store images in HDF (not a bad idea in the long run, but not a pressing concern).

We should probably disable the export as HDF option for multi-colour images (or display a warning), however.

barentine:
I don't actually have permission to close this one @David-Baddeley , but it's all set. Thanks!

PR #698, opened 2020-12-09
initial work on an xyztc data source

David-Baddeley:
Initial steps in a longstanding item on the roadmap, namely the transition to an xyztc data model from an xy(z/t)c model. Also relevant to #64.

Plan is to create an xyztc data source for every input data source and initially have this live beside `image.data` so that recipe modules, display, etc ... can use it if they prefer, before finally moving everything over.

### DIFF
--- a/PYME/IO/DataSources/BGSDataSource.py
+++ b/PYME/IO/DataSources/BGSDataSource.py
@@ -229,6 +229,14 @@ class DataSource(BaseDataSource):
         self.bgRange = bgRange
         self.dataStart = 0
         
+    @property
+    def additionalDims(self):
+        return self.datasource.additionalDims
+    
+    @property
+    def sizeC(self):
+        return self.datasource.sizeC
+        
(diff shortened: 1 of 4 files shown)


David-Baddeley:
bit more detail on migration plan:

- add an `image.voxels` attribute which is an XYZTC data source (sitting on top of our existing xyt datasources, and initially having `shape[2]=1` - i.e. XY(Z)TC unless we have information from the metadata to suggest otherwise)
- change the viewer to use `image.voxels` rather than `image.data` and add a z slider rather as well as a t slider
- change recipe modules etc to use `image.voxels`
- make `image.data` throw a deprecation warning.
- potentially remove `image.data` (there are arguments for keeping it around as it better represents disk storage)

For dask etc ... compatibility, add a `.chunks` property to provide chunk hinting.

David-Baddeley:
merging in current state so as to get multicolour hdf in, still more work to go on final stuff

Issue #697, opened 2020-12-09
[BUG] DCAM multiview view-change subarray


barentine:
posting quick during meeting - apologies for incompleteness
**Describe the bug**
Invalid subarrays for multiview orca flash which are valid if you go from full chip to those views, but not from other active views
```
DEBUG:PYME.IO.clusterIO:Request for b'http://10.150.7.4:15348/Bergamot/2020_8_13/EMC0103S2/000/' failed with error: 404
DEBUG:PYME.Acquire.Hardware.Camera:setting chip ROI
-128361 128361 -1
(1597343378, 44004) 62824 0
DEBUG:PYME.Acquire.Hardware.HamamatsuDCAM.HamamatsuORCA:Setting ROI: x0 744.0, y0 896.0, w 256.0, h 256.0
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\frameWrangler.py", line 227, in onExpReady
    d = self.getFrame()
DEBUG:PYME.Acquire.Spooler:Disconnecting from frame source
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\frameWrangler.py", line 198, in getFrame
    self.cam.ExtractColor(cs,0)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\Camera.py", line 1038, in ExtractColor
    self.camera_class.ExtractColor(self, self.chip_data, mode)
DEBUG:PYME.Acquire.Spooler:Frame source should be disconnected
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Hardware\HamamatsuDCAM\HamamatsuORCA.py", line 300, in ExtractColor
    raise DCAMZeroBufferedException()
PYME.Acquire.Hardware.HamamatsuDCAM.HamamatsuORCA.DCAMZeroBufferedException
DEBUG:PYME.Acquire.ui.rules:htsms-tile
PatchedServiceInfo(type='_pyme-taskdist._tcp.local.', name='PYMERuleServer: smeagol._pyme-taskdist._tcp.local.', address=b'\n\x96\x07\x82', port=15346, weight=0, priority=0, server='PYMERuleServer: smeagol._pyme-taskdist._tcp.local.', properties={}) b'\n\x96\x07\x82'
DEBUG:root:Successfully created recipe rule
Traceback (most recent call last):
[...]

PR #699, opened 2020-12-10
DCAM orca: potential fix for shifting position out of subarray bounds

barentine:
Addresses issue #697 .

bugfix
**Proposed changes:**
- turn off subarray mode before applying new subarray settings
- leave subarray mode on even if someone sets full chip

**Checklist:**

- [x] Tested


barentine:
Works great - thanks for the catch @David-Baddeley

PR #703, opened 2020-12-13
Remove cutoff at object size of 100 nm^2

zacsimile:
We had an arbitrary cutoff of 100 nm^2 in Analysis > Measure Objects. Switched to a cutoff of infinity.

### DIFF
--- a/PYME/LMVis/Extras/objectMeasurements.py
+++ b/PYME/LMVis/Extras/objectMeasurements.py
@@ -105,7 +105,7 @@ class ObjectMeasurer:
         pipeline.objectMeasures = {}
 
         if len(chans) == 0:
-            pipeline.objectMeasures['Everything'] = objectMeasure.measureObjectsByID(pipeline.colourFilter, 10,ids,key)
+            pipeline.objectMeasures['Everything'] = objectMeasure.measureObjectsByID(pipeline.colourFilter, np.inf,ids,key)
 
             from PYME.ui import recArrayView
             f = recArrayView.ArrayFrame(pipeline.objectMeasures['Everything'], parent=self.visFr, title='Object Measurements')
@@ -126,7 +126,7 @@ class ObjectMeasurer:
             for ch, i in zip(chans, range(len(chans))):
                 pipeline.colourFilter.setColour(ch)
                 #fitDecayChan(colourFilter, metadata, chanNames[i], i)
-                pipeline.objectMeasures[chanNames[i]] = objectMeasure.measureObjectsByID(pipeline.colourFilter, 10,ids,key)
+                pipeline.objectMeasures[chanNames[i]] = objectMeasure.measureObjectsByID(pipeline.colourFilter, np.inf,ids,key)
             
             pipeline.colourFilter.setColour(curChan)
 


David-Baddeley:
The `sizeCutoff` parameter is poorly named, and is not, as you have assumed, a threshold on object size, but rather on edge length in the tessellation used to calculate object areas and perimeters. `sizeCutoff=np.inf` will force the area to be evaluated as the convex hull of the points comprising the object, and, as a result will consistently overestimate area and underestimate the perimeter. 

It's a little hard to know how we should set this parameter - for most super-res data `sizeCutoff=10 nm` will give more accurate results than `sizeCutoff=np.inf`, but the former is an arbitrary empirical value and will run into issues for small clusters which don't really have enough localisations to reliably define an area. `sizeCutoff=np.inf` is potentially a more conservative choice, with the caveat that it will generate the kinds of errors discussed above. Ultimately the right solution might be to expose the parameter in the GUI along with a good explanation of what it does. 

Another slightly better approach might be to, e.g. set it automatically based on the standard deviation of the point cloud and the number of points it contains (we could probably find a statistical way of finding a good value based on those parameters, but it would need a bit more research).

David-Baddeley:
Other options: 
- base the area and perimeter metrics on the segmentation mask, not directly on the points (although this would not work for, e.g. dbscan clustering).
- render a density image using just the segmented points, and threshold this to get area and perimeter.

zacsimile:
Ahh, I see, yes. I got a bit confused because

```
    A = maximum(s01, s12, s02)

    cutInd = A < sizeCutoff**2

    #area
    A = 0.5*sqrt((s_01*s_01).sum(1)*(s_12*s_12).sum(1) - ((s_01*s_12).sum(1)**2))
```

just made me think A = area in my head. It's quite clear from the lines before this is not the case.

A cut of 10 nm throws away an awful lot of objects, especially on sparser data sets. I think we should consider making this user-adjustable or automatically set, as you suggested.

Why not use the convex hull, though? `sizeCutoff` is an additional hyperparameter that kind of goes against our argument that you should use a thresholded image for segmentation because the limitations of automatic thresholds are well-understood. If we add this on top, we're not just using well-understood thresholding modes. I'd argue that any point caught by the density threshold should be included in the cluster, regardless of how far away it is.

David-Baddeley:
I agree that the the arbitrary parameter is bad, and that it should at least be adjustable. That said, its important to use something other than the convex hull if you want to get decent results (especially if you have, e.g. used dbscan to get your clusters in the first place). If you have objects with any form of complex structure (the simplest would be a banana shape), the convex hull is going to over-estimate the area by a lot. Such structures are the rule rather than the exception in biology (e.g. RyR clusters, synapses, caveolin clusters etc ...).  

It's important to note that we are not dropping points from the structure, but rather pruning triangles to move inwards from the convex hull to the true structure. It is possible that you might also drop a few points if the impingent triangles are all too large, but dropping a significant number of points is generally a sign that your labelling is too sparse to get a remotely reliable estimate of area and perimeter. For small or sparse clusters, the standard deviation of the point positions and/or the radius of gyration are better measures. If the cluster is either small enough or sparsely labeled enough that the length threshold is giving problems, then there is no accurate way to get a shape-based measure of area or perimeter, and you have to fall back to statistical descriptors such as the std. deviation of the points. In such cases, it's arguably better to give a obviously wrong value (e.g. 0) than a made up value from the convex hull. 

In practice if I'm after areas I often just use image based quantification, but the benefit of the object measures as it stands is that it doesn't care how you segmented the objects. In an ideal world, we'd probably have the area and perimeter estimation off by default, so you'd have to turn them on in the GUI (this should force you to think a little about how and why the parameters are being calculated).

David-Baddeley:
On re-reading the code, `sizeCutoff=10` corresponds to a 100nm edge length. To get there, you have to have a super, super sparse cluster - i.e. it's either an artefact or your data is crappy enough that your resolution is somewhere around 500nm-1um.

David-Baddeley:
Also that code is super-gross and needs documenting (for context, it did start out as an area threshold, then became an edge-length threshold).

zacsimile:
OK, I see your point. 100 nm would be quite a bit of space in a super-resolution cluster. In practice, that's probably a reasonable cutoff. 

Better documentation would help a lot. We might also want to provide clearer feedback to users as to what is going on. A particularly confusing thing is to have a bunch of clusters pop up with area 0 in the resulting measures with no explanation. If you compare this result to the labeled image and/or points, it's hard to make sense of what happened.

David-Baddeley:
Just looked at the code again, it is actually only 10 (we compare with the squared edge length and I'm trying to be clever and avoid a `sqrt()`) - we should probably increase the default to 30-100. 30 might be a good choice base on typical localisation precision arguments.

That said, you probably want the density to be high enough that 10 is OK if you want to trust the measurements.

David-Baddeley:
Working on some doc changes.

David-Baddeley:
With better documentation in #705 

At this point do we?
- change the default to 100nm and call it done?
- add a dialog (which we should have anyway for selecting object ID column)?
- do some adaptive stuff (something like `3*np.median(edge_lengths)` would probably be a good start).

barentine:
Throwing this out there after only partially following: is it not better to jt-density map, threshold, calculate area on the segmented areas and map those segmented areas back to the localizations rather than fiddling with edge dropping on the vertices? Alternatively, can one histogram the vertice edge lengths and use e.g. otsu's method rather than a fixed value?

zacsimile:
I think we do a hybrid of all 3, @David-Baddeley. Add a dialog with 100 nm as the default, but have a checkbox that says "guess for me" or something like that and overrides the 100 nm with the adaptive calculation.

Issue #537, opened 2020-10-25
[PROPOSAL] Multiview channel splitting module with ImageStack output(s), for multicolor overviews


barentine:
**Is your feature request related to a problem? Please describe.**
Need to tile overview in two colors (at the moment, just two colors, not two planes), and eventually do some image analysis on resulting supertile datasources. 

**Describe the solution you'd like to implement**
- run a tiled acquisition protocol with two multiview channels on (different colors)
- run that series through this new recipe module which will split that ImageStack into two ImageStacks, one for each multiview channel
- run each of those ImageStacks through the tilepyramid output module
- open each of those pyramids as supertile datasources, and do whatever analysis I want on them (fingers crossed chromatic shifts aren't awful)

**Describe alternatives you've considered**
- adding support for two-color tilepyramid creation. That might be cool as it could easily account for integer-pixel constant lateral shifts between the two color channels. That would be a bit more work, and in any case I think we would still need a multiview-specific module to split out the color channels. So implementing this still makes sense and I might as well try the full pipeline once we're there.

barentine:
maybe a MultiviewDatasource on the backend makes this more acceptable memory wise

barentine:
Note that datasource is different than the current MultiviewDataSource which serves to crop multiview stuff for the old dcimgspooler. I've forgotten what we use in unsplitting for e.g. PSF calibrations and whether we already have that in a recipe module. Wasn't thinking. Note it looks like there's a float-indexing error or two lurking in UnsplitDataSource that I'll PR shortly

barentine:
conversation with David:
- refactor get unsplit data source from DSView/modules/splitter into a recipes/processing recipe module (which is maybe just a thin wrapper around further refactoring that logic somewhere else)

PR #707, opened 2020-12-15
extract multiview channel recipe module

barentine:
Addresses issue #537 

enhancement
**Proposed changes:**
- make a recipe module to easily split multiview views into separate imagestacks. goal here is to allow multiview tile series to be split and generate two separate image pyramids, two separate supertile sources, and either do subsequent analysis or recombine as a 'normal' stack.

**Checklist:**

- [ ] Tested with numpy=1.14

1.19

3.7

4.0.4

### DIFF
--- a/PYME/recipes/multiview.py
+++ b/PYME/recipes/multiview.py
@@ -381,6 +381,26 @@ class CalibrateShifts(ModuleBase):
 
 
 class ExtractMultiviewChannel(ModuleBase):
+    """Extract a single multiview channel
+
+    Parameters
+    ----------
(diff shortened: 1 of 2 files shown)


barentine:
I suppose it would also be possible to crop and slot in with each of the views as a color / channel. That might be a bit cleaner, but kind of odd as I have multiple views of the same color. Not necessarily wrong for it to go there, but simply offering to extract one, with the disclaimer that it then can't be used for other multiview-specific things seems OK too (one could also join channels if they wanted it in as multicolor).

PR #710, opened 2020-12-15
add output_dir setting/passing for remaining recipe clusterUI views

barentine:
Addresses issue only the templated recipe view being able to save outputs at the moment.

bugfix
**Proposed changes:**
- add output path text entry to the forms and pass it in views / run

**Checklist:**

- [ ] Tested with numpy=1.14

1.19

3.6

4.0.4


barentine:
note that uncommenting / wiring up the select button would be neat, but is not included in this PR currently

PR #717, opened 2020-12-16
thread-polled get_status to hide latency on randomly slow windows requests

barentine:
Addresses issue #545. forgot to comment out the freespace update last time I merged main and it was pretty painful - finally crossing this one off the to-do list.

enhancement
**Proposed changes:**
- use a thread to poll clusterIO.get_status to hide any latency in retrieving it.

**Checklist:**

- [ ] Tested with numpy=1.14
1.16
- [ ] 
3.6

### DIFF
--- a/PYME/Acquire/SpoolController.py
+++ b/PYME/Acquire/SpoolController.py
@@ -281,8 +281,8 @@ class SpoolController(object):
 
         """
         if self.spoolType == 'Cluster':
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Merging for sake of expediency, but current structure doesn't feel super intuitive. TODO - rewrite using e.g. a `StatusPoller` class to improve maintainability??

PR #726, opened 2020-12-19
Make it harder to crash pco. cameras

zacsimile:
General fixes to the pco. camera class. I'm having a reasonably difficult time crashing the camera at this point.  

It is still possible to crash the camera after a while if you run z-stacking with a low frame average (30 frames). None of the standard errors that would be caught by us or pco. cause this: I've verified it's it's not a problem with `memcpy` and not a problem with the image number we pass to the camera in the `cam.image()` call. I do think it's something with pco.recorder that isn't caught by the log, which I think means it's something sinister with the timer or the DLL.

These camera fixes resolve all other previously-identified sources of crashing and make the pco. generally usable.

### DIFF
--- a/PYME/Acquire/Hardware/pco/pco_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_cam.py
@@ -199,22 +199,29 @@ class PcoCam(Camera):
         if y1 < y0:
             y0, y1 = y1, y0
 
+        # Don't let the ROI go out of bounds
+        # See pco.sdk manual chapter 3: IMAGE AREA SELECTION (ROI)
+        x0 = np.clip(x0, 1, lx_max-dx+1)
+        y0 = np.clip(y0, 1, ly_max-dy+1)
+        x1 = np.clip(x1, 1+dx, lx_max)
+        y1 = np.clip(y1, 1+dy, ly_max)
+
         # Don't let us choose too small an ROI
-        if (x1-x0+1) < lx_min:
+        if (x1-x0) < lx_min:
             logger.debug('Selected ROI width is too small, automatically adjusting to {}.'.format(lx_min))
-            guess_pos = x0+lx_min-1
+            guess_pos = x0+lx_min
             # Deal with boundaries
             if guess_pos <= lx_max:
                 x1 = guess_pos
             else:
-                x0 = x1-lx_min+1
-        if (y1-y0+1) < ly_min:
+                x0 = x1-lx_min-1
+        if (y1-y0) < ly_min:
             logger.debug('Selected ROI height is too small, automatically adjusting to {}.'.format(ly_min))
-            guess_pos = y0+ly_min-1
+            guess_pos = y0+ly_min
             if guess_pos <= ly_max:
                 y1 = guess_pos
             else:
-                y0 = y1-ly_min+1
+                y0 = y1-ly_min-1
 
         # Round to a multiple of dx, dy
         # TODO: Why do I need the 1 correction only on x0, y0???
@@ -222,13 +229,6 @@ class PcoCam(Camera):
         y0 = 1+int(np.floor((y0-1)/dy)*dy)
         x1 = int(np.floor(x1/dx)*dx)
         y1 = int(np.floor(y1/dy)*dy)
-
-        # Don't let the ROI go out of bounds
-        # See pco.sdk manual chapter 3: IMAGE AREA SELECTION (ROI)
-        x0 = np.clip(x0, 1, lx_max-dx+1)
-        y0 = np.clip(y0, 1, ly_max-dy+1)
-        x1 = np.clip(x1, 1+dx, lx_max)
-        y1 = np.clip(y1, 1+dy, ly_max)
 
         self.cam.sdk.set_roi(x0, y0, x1, y1)
         self._roi = [x0, y0, x1, y1]
--- a/PYME/Acquire/Hardware/pco/pco_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_cam.py
@@ -76,7 +76,7 @@ class PcoCam(Camera):
     def ExtractColor(self, chSlice, mode):
         # Somehow this check matters... shouldn't this be taken care of by ExpReady???
         if (not self.recording) or (self.GetNumImsBuffered() < 1):
-            return True
+            raise RuntimeError('Trying to grab a frame when there is no frame to grab.')
 
         if self.n_read < self.buffer_size:
             curr_frame = self.n_read
--- a/PYME/Acquire/Hardware/pco/pco_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_cam.py
@@ -207,7 +207,7 @@ class PcoCam(Camera):
         y1 = np.clip(y1, 1+dy, ly_max)
 
         # Don't let us choose too small an ROI
-        if (x1-x0) < lx_min:
+        if (x1-x0+1) < lx_min:
             logger.debug('Selected ROI width is too small, automatically adjusting to {}.'.format(lx_min))
             guess_pos = x0+lx_min
             # Deal with boundaries
@@ -215,7 +215,7 @@ class PcoCam(Camera):
                 x1 = guess_pos
             else:
                 x0 = x1-lx_min-1
-        if (y1-y0) < ly_min:
+        if (y1-y0+1) < ly_min:
             logger.debug('Selected ROI height is too small, automatically adjusting to {}.'.format(ly_min))
             guess_pos = y0+ly_min
             if guess_pos <= ly_max:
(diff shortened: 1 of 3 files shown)


barentine:
What’s the integration time on the 30 frame per step z stack crashing? You’re doing this with the z stack menu (above Time/Blinking series)?

zacsimile:
This is with any integration time. Done with z-stepping in the Time/Blinking series box. There are other ways to crash the camera, too, I'm sure. This is just the easiest one. It's pretty clearly due to one of the memory errors we don't catch (we catch `memcpy` and accessing an improper frame on the camera). I think also we're not crashing the camera, but the software/software buffer we read from. The camera appears to keep recording after the program crashes.

zacsimile:
I've resorted to putting a print statement in every single line of the SDK and it never crashes in the same place twice. I don't even get a segfault error, PYMEAcquire just quits and the camera seems to keep recording (it's hot, and I can't reconnect to it without rebooting it because it says it's busy).

PYME/Acquire/Hardware/pco/pco_cam.py
```diff
@@ -68,25 +74,21 @@ def CamReady(self):
         return self.initalized
 
     def ExtractColor(self, chSlice, mode):
-
         # Somehow this check matters... shouldn't this be taken care of by ExpReady???
-        if self.GetNumImsBuffered() < 1:
+        if (not self.recording) or (self.GetNumImsBuffered() < 1):
             return True
```
David-Baddeley: The `return True` here is super unsafe (or will at the very least result in data corruption) - because we return our result by writing into the `chslice` input argument, returning from this method with no exception set is implicitly stating that you have put a new frame into `chslice`. If you are in this method and there is no data available, the only safe thing to do is to raise an exception.

David-Baddeley: I know this is not new in this PR - but thought it worth flagging for debug efforts in general.

zacsimile: Makes sense. I'll push a commit fixing this in a second.

PYME/Acquire/Hardware/pco/pco_cam.py
```diff
@@ -68,25 +74,21 @@ def CamReady(self):
...
-
         # Grab the image (unused metadata _)
         image, _ = self.cam.image(curr_frame)
 
-        # print('curr_frame: {}, recorder_image_number: {}, num_ims_buffered: {}, num_ims_read: {}'.format(curr_frame, meta['recorder image number'], self.cam.rec.get_status()['dwProcImgCount'], self.n_read))
-
         ctypes.cdll.msvcrt.memcpy(chSlice.ctypes.data_as(ctypes.POINTER(ctypes.c_uint16)),
-                   image.ctypes.data_as(ctypes.POINTER(ctypes.c_uint16)), chSlice.nbytes)
+                   image.ctypes.data_as(ctypes.POINTER(ctypes.c_uint16)), 
+                   self.__image_frame_bytes)
```
David-Baddeley: This is super unsafe and likely to generate segfaults and/or corruption as it introduces the possibility of copying / writing over the end of the chSlice array. Because you are copying **into** `chSlice`, it **must** be the arbiter of the amount to be copied.   

You could theoretically get read segfaults on the camera image buffer with the old code if the sizes didn't match, but these are safer than writing to memory you don't own. For safety, consider the following:

- revert to old code
- add a line above such as:

```python
assert(self.__image_frame_bytes == chSlice.nbytes)
```

which will throw an exception if things are out of sync, but should prevent segfaults / corruption.

zacsimile: Interesting. I originally had `self.__image_frame_bytes` as `chSlice.nbytes`, but I changed it because the `self.__image_frame_bytes` approach is what we do in `HamamatsuORCA`, so I assumed this was the preferred system.

Either way, a memory error here (in `memcpy`) is caught. I know, I've made lots of them.

zacsimile: I did push a commit making this change, though.

David-Baddeley: I  guess we should change the Orca code to0 - in general the Zyla is a (much) safer reference than the Orca.

If `memcpy` is catching things, it can only catch out of process memory issues - you still risk over-writing large chunks of in-process memory and causing data (and potentially even bytecode) corruption (it's completely unaware of python-internal memory allocation).

zacsimile: Probably true, but every time I've made a memory error here only ExtractColor gets messed up. There's no segfaults, just a bunch of tracebacks telling me I screwed up and the program keeps rolling until I quit. It's a very noticeable, repeatable error, and it tends to happen right away rather than 10 minutes into imaging (which is the error I experience with the z-stacking with a low number of frames to average with this camera).

PYME/Acquire/Hardware/pco/pco_cam.py
```diff
@@ -68,25 +73,21 @@ def CamReady(self):
...
         # Somehow this check matters... shouldn't this be taken care of by ExpReady???
-        if self.GetNumImsBuffered() < 1:
-            return True
+        if (not self.recording) or (self.GetNumImsBuffered() < 1):
+            raise RuntimeError('Trying to grab a frame when there is no frame to grab.')
 
         if self.n_read < self.buffer_size:
             curr_frame = self.n_read
         else:
             curr_frame = self.buffer_size-self.GetNumImsBuffered()
```
David-Baddeley, line 83: This logic looks wrong for a circular buffer ... shouldn't it be something like `n_read % self.buffer_size`

zacsimile: It's a ring buffer, rather than a circular buffer, so the position of the latest frame moves as data populates (appends to the end of the buffer and rotates through the rest of the positions before disappearing).

David-Baddeley:
I'm wondering if there might be a race condition somewhere that is behind some of your issues ...

zacsimile:
Certainly could be a race condition, I just don't know where it is. I never get a `segfault`, the program just quits. I can't get it to reliably fail in the same spot.

zacsimile:
I think there were a few things going. This latest code fixes the easy bugs--the `self.recording` flag solved a lot of the obvious stuff (the `pco` backend build a `recorder` object separate from the camera). Regardless of the remaining error(s), this is at least working with correct ROIs now. It's definitely a step up from my first pass at the pco. camera.

zacsimile:
How do I debug a race condition in this instance?

PYME/Acquire/Hardware/pco/pco_cam.py
zacsimile: The `Zyla` uses `chSlice` size, but not `nbytes`. Should be fairly similar to `__image_frame_bytes` actually, since `chSlice` size gets set by the chip/roi size. Identical, I think, since both use `cam.GetPicWidth()` and `cam.GetPicHeight()` to set size.

zacsimile:
One other thing to note is this only crashes when I have the cluster on. I'm not sure if that's due to the z-stacking just getting stuck when I spool to file or because the cluster+camera is what creates the race condition.

It is strange on spooling to file that I purge the buffer overflow a couple of times but eventually stop purging. Is this a problem with the pco. specifically or with any camera?

image

I tried testing with the simulator but I can't do z-stacking with it. I get the following traceback.

```
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\ui\HDFSpoolFrame.py", line 470, in OnBStartSpoolButton
    self.spoolController.StartSpooling(fn, #stack=stack, #compLevel = compLevel,
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\SpoolController.py", line 463, in StartSpooling
    self.spooler.StartSpool()
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\Spooler.py", line 134, in StartSpool
    self.protocol.Init(self)
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\protocol.py", line 220, in Init
    scope.state.setItem(self.piezoName, self.zPoss[self.pos], stopCamera=self.require_camera_restart)
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\microscope.py", line 169, in setItem
    return self.setItems({key:value}, stopCamera, force)
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\microscope.py", line 232, in setItems
    self.scope().frameWrangler.start()
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\frameWrangler.py", line 449, in start
    self.cam.StartExposure()
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\Hardware\Simulator\fakeCam.py", line 499, in StartExposure
    eventLog.logEvent('StartAq', '')
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\eventLog.py", line 73, in logEvent
    evl.logEvent(eventName, eventDescr, timestamp = timestamp)
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\HDFSpooler.py", line 92, in logEvent
    ev['Time'] = sp.timeFcn()
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\Spooler.py", line 265, in fakeTime
    return self.tStart + self.imNum*self.fakeCamCycleTime
AttributeError: 'Spooler' object has no attribute 'tStart'
```

Issue #708, opened 2020-12-15
[BUG] fresh install, `PYMEClusterOfOne --clusterUI` silently fails clusterUI if django isn't manually installed


barentine:
**Describe the bug**
django is an optional dependency (I dissent, but not a big deal), `PYMEClusterOfOne --clusterUI` fails silently if one doesn't have django installed

**To Reproduce**
1. install PYME
2. `PYMEClusterOfOne --clusterUI
3. browser redirects to page not found
4. try running clusterUI standalone and see modulenotfound Django error

**Expected behavior**
warn people they forgot to install django

barentine:
might also become relevant for packaging, particularly related to #657, but already an issue on 3.7:
```
$ conda install django=1.11
Collecting package metadata (current_repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Collecting package metadata (repodata.json): done
Solving environment: failed with initial frozen solve. Retrying with flexible solve.
Solving environment: \ 
Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
failed                                                                                                                                                 

UnsatisfiableError: The following specifications were found
to be incompatible with the existing python installation in your environment:

Specifications:

  - django=1.11 -> python[version='>=2.7,<2.8.0a0|>=3.5,<3.6.0a0|>=3.6,<3.7.0a0']

Your python: python=3.7

If python is on the left-most side of the chain, that's the version you've asked for.
When python appears to the right, that indicates that the thing on the left is somehow
not available for the python version you are constrained to. Note that conda will not
change your python version to a different minor version unless you explicitly specify
that.

```

barentine:
note that pip installing any old django 1.11 won't work, but after some convincing (supposed to use django 2.x for py37) they backported a fix https://github.com/django/django/pull/10654

David-Baddeley:
The above mentioned packaging / conda issue is one of the reasons *why* django is an optional dependency.

- django is generally not backwards compatible between versions (i.e. we need to either pin or spend a lot of time constantly re-writing the bits of the code which use django so they work with the most recent version).
- such pinning is highly likely to result in un-resolvable environments.

For the main issue, I agree that it would be nice to warn people if they don't have django installed.

David-Baddeley:
We should probably try and see if the code runs on a more recent django and update pins if needed

barentine:
[...]

PR #730, opened 2020-12-21
make clusterUI work on django 2.0, 2.1, warn django not present

David-Baddeley:
Addresses issue #708.

- minor django version compat changes for clusterUI code
- make cluster_of_one warn when run with `--clusterUI` flag and you don't have django 

Needs a bit more work to support 2.2.x


barentine:
Thanks for this David!

PR #735, opened 2020-12-22
Add option to display mesh normals

zacsimile:
Mostly for debugging purposes, but kind of nifty. Allows a user to optionally display mesh normals by vertex or by face.

image


David-Baddeley:
Thinking aloud here, but should the option to display normals be shown in the layer controls (which are already pretty busy)? I'm wondering if (as it's more of a debug option) it wouldn't make sense to just make it toggle-able in the shell (as we do for outlines in shaded and flat views).

zacsimile:
That seems reasonable.

zacsimile:
Changed. Can be called from the shell as `glCanvas.layers[1].display_normals=True` or similar.

David-Baddeley:
I'm wondering if we should think (medium/long term) about having 3 levels of UI complexity -e.g.  standard, pro, debug and choose what to display based on these.

zacsimile:
I love this idea. I think we could go even further and display different levels of complexity based on a chosen workflow, the same way Adobe switches up available tools depending on what you say you're trying to make.

Issue #743, opened 2020-12-29
[BUG] rule based localization not saving events to results h5r


barentine:
**Describe the bug**

**To Reproduce**
1. run automatic localization posting rules on spool stop
2. open results in PYMEVis
3. no acquisition events
4. See acquisition events file is on the cluster with the original series
5. wonder what happened but file this issue before really looking into it since I'm imaging today

barentine:
At the moment we push events in `prepare`, which more closely matches the old pusher which was only set up to launch when the series was already complete. This fails at the moment because of a race condition which we seem to lose each time, in that I connected the rule pushing to the onSpoolStop signal, which gets fired in `HTTPSpooler` before the events files are pushed to the cluster by the spooler. The intermediate refactor pushed events when the series datasource complete property returned true (see [here](https://github.com/python-microscopy/python-microscopy/pull/635/files#diff-b10a3d380c10f04818a3cc6044e08f586ca53f706a3d46f8f387658a93536707L296)). Given that onSpoolStop disconnects, calls protocol Finish which may still log events, etc. this seems like the safest approach, rather than swapping order in StopSpool

barentine:
solved with #781

PR #744, opened 2020-12-29
make sure we copy events to results file for localization rules

barentine:
Addresses issue #743 .

bugfix
**Proposed changes:**
move events copying to ensure it happens after the datasource is completed (i.e. so the rule can be posted before the datasource is finished acquiring). Would be nice if this didn't require overriding both `push` and `_poll_loop`, but I think at the moment I think it's 'fine' / would require a major reworking elsewhere to change this which probably isn't worthwhile.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/cluster/rules.py
+++ b/PYME/cluster/rules.py
@@ -540,7 +540,7 @@ class LocalisationRule(Rule):
         #set up results file:
         logging.debug('resultsURI: ' + self.worker_resultsURI)
         clusterResults.fileResults(self.worker_resultsURI + '/MetaData', self.mdh)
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I really, really dislike the duplication of the base-class logic for `push()` and `_poll_loop()` the removal of which was one of the main reasons for refactoring the rule pushing stuff in the first place. Think there might be a better/cleaner way of doing this which avoids the duplication - e.g. an over-ridable function which gets called from the base class when it detects that all the data has been spooled / the series is complete. Can possibly tie this in with a solution for #736.

David-Baddeley:
superseded by #781

Issue #725, opened 2020-12-18
Should sample metadata be available directly in rendered images rather than through the `Source.*` metadata subkeys?


barentine:
**Describe the bug**
I try to access 'Sample.SlideRef' metadata of a density mapping (histogram) and KeyError because we don't copy entries and overwrite voxelsize, we just give it a new one.

**To Reproduce**
1. do a density mapping
2. hope to find metadata from the localizations
3. not find it

David-Baddeley:
We do record the `SourceImageID`, which should in theory be enough to find the the metadata about the localisations, if in a somewhat roundabout way. 

This probably should fall into a more general "What should be propagated discussion" - it probably does make sense to propagate all the `'Sample.*'` keys, although I I'm not sure about any of the other metadata from the localisation file.

David-Baddeley:
I should note that the copy all entries and over-write where needed strategy that is adopted by a lot of modules is a quick and dirty solution, not the way things **should** work. Modules should only copy things which are objectively the same, but not anything that isn't  (potentially preserving the input metadata under the `'Source'` key).

David-Baddeley:
Actually I correct myself, for rendered images it should all there under the `'Source.*'` keys - i.e. you'd need to use `Source.Sample.XXX` instead of `Sample.XXX`

David-Baddeley:
Changed the title - this isn't a bug, as the metadata is available. There could be an argument, however, for also propagating sample related metadata up to the top level of the metadata hierarchy.

barentine:
I would argue we are still looking at the same sample, and therefore it would be nice not to need to know a-priori whether you need to dig or not. are you OK with me explicitly passing through sample related keys then?

barentine:
should be set

PR #752, opened 2021-01-06
pass sample metadata through to rendered images

barentine:
Addresses issue #725 .

enhancement 
**Proposed changes:**
pass sample metadata through to the top tree of rendered image metadata

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/LMVis/renderers.py
+++ b/PYME/LMVis/renderers.py
@@ -32,7 +32,7 @@ import numpy as np
 
 renderMetadataProviders = []
 
-PASS_THROUGH_KEYS = [
+SAMPLE_MD_KEYS = [
     'Sample.SlideRef',
     'Sample.Creator',
     'Sample.Notes',
@@ -40,8 +40,8 @@ PASS_THROUGH_KEYS = [
     'AcquiringUser'
 ]
 
-def add_pass_through_metadata(old_mdh, new_mdh):
-    pass_through = [k for k in PASS_THROUGH_KEYS if k in old_mdh.keys()]
+def copy_sample_metadata(old_mdh, new_mdh):
+    pass_through = [k for k in SAMPLE_MD_KEYS if k in old_mdh.keys()]
     for k in pass_through:
         new_mdh[k] = old_mdh[k]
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Ultimately it might make sense to handle this in the base recipe module rather than here (as rendering is only one place where we are likely to run into this). Might be OK in interim. Will sleep on it.

PYME/LMVis/renderers.py
```diff
@@ -32,6 +32,19 @@
 
 renderMetadataProviders = []
 
+PASS_THROUGH_KEYS = [
```
David-Baddeley:
```suggestion
SAMPLE_MD_KEYS = [
```

PYME/LMVis/renderers.py
```diff
@@ -32,6 +32,19 @@
...
+    'Sample.SlideRef',
+    'Sample.Creator',
+    'Sample.Notes',
+    'Sample.Labelling',
+    'AcquiringUser'
+]
+
+def add_pass_through_metadata(old_mdh, new_mdh):
```
David-Baddeley:
```suggestion
def copy_sample_metadata(old_mdh, new_mdh):
```

PYME/LMVis/renderers.py
```diff
@@ -171,6 +184,7 @@ def _genJitVals(self, jitParamName, jitScale):
 
     def Generate(self, settings):
         mdh = MetaDataHandler.NestedClassMDHandler()
+        add_pass_through_metadata(self.pipeline.mdh, mdh)
```
David-Baddeley:
```suggestion
        copy_sample_metadata(self.pipeline.mdh, mdh)
```

PYME/LMVis/renderers.py
```diff
@@ -32,6 +32,19 @@
...
+PASS_THROUGH_KEYS = [
+    'Sample.SlideRef',
+    'Sample.Creator',
+    'Sample.Notes',
+    'Sample.Labelling',
+    'AcquiringUser'
+]
+
+def add_pass_through_metadata(old_mdh, new_mdh):
+    pass_through = [k for k in PASS_THROUGH_KEYS if k in old_mdh.keys()]
```
David-Baddeley:
```suggestion
    pass_through = [k for k in SAMPLE_MD_KEYS if k in old_mdh.keys()]
```

PYME/LMVis/renderers.py
```diff
@@ -221,6 +235,7 @@ class ColourRenderer(CurrentRenderer):
     
     def Generate(self, settings):
         mdh = MetaDataHandler.NestedClassMDHandler()
+        add_pass_through_metadata(self.pipeline.mdh, mdh)
```
David-Baddeley:
```suggestion
        copy_sample_metadata(self.pipeline.mdh, mdh)
```

Issue #757, opened 2021-01-07
[BUG] multiview calibration generation breaking on delaunay dimensionality issue


barentine:
**Describe the bug**
```
 Error whilst running Corrections>Multiview>Calibrate Shifts
==============================================
python-microscopy=20.12.10
python=3.6.9, platform=linux
numpy=1.16.5, wx=4.0.4 gtk2 (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/home/smeagol/code/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/home/smeagol/code/python-microscopy/PYME/LMVis/Extras/multiviewMapping.py", line 205, in OnCalibrateShifts
    sm = calibration_module.apply_simple(self.pipeline.selectedDataSource)
  File "/home/smeagol/code/python-microscopy/PYME/recipes/base.py", line 206, in apply_simple
    return self.apply(*args, **kwargs)[next(iter(self.outputs))]
  File "/home/smeagol/code/python-microscopy/PYME/recipes/base.py", line 185, in apply
    self.execute(namespace)
  File "/home/smeagol/code/python-microscopy/PYME/recipes/multiview.py", line 365, in execute
    dy_err[ii - 1, :])
  File "/home/smeagol/code/python-microscopy/PYME/Analysis/points/twoColour.py", line 265, in genShiftVectorFieldQ
    wonky = findWonkyVectors(nx, ny, nsx, nsy, tol=2*err_sx.mean())
  File "/home/smeagol/code/python-microscopy/PYME/Analysis/points/twoColour.py", line 411, in findWonkyVectors
    T = tri.Triangulation(x,y)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/matplotlib/tri/triangulation.py", line 54, in __init__
    self.triangles, self._neighbors = _qhull.delaunay(x, y)
ValueError: x and y arrays must have a length of at least 3
```

**To Reproduce**
1. take multiview localizations
2. fold it
3. calibrate shifts
4. see error

barentine:
hmm, a simple test recipe works fine
```
import os
from PYME.recipes.base import ModuleCollection
from PYME.recipes import multiview, localisations

recipe = ModuleCollection()
recipe.loadInput('pyme-cluster:///Bergamot/2021_1_7/000/analysis/reg.h5r')

[...]

PR #758, opened 2021-01-07
fail early if we don't have enough clusters to calibrate shifts

barentine:
Addresses issue makes it harder to feel like a fool, e.g. #757 .

enhancement
**Proposed changes:**
fail before we get into delaunay triangulation calls if we don't have enough to make a triangle

### DIFF
--- a/PYME/recipes/multiview.py
+++ b/PYME/recipes/multiview.py
@@ -318,6 +318,8 @@ class CalibrateShifts(ModuleBase):
         # Generate raw shift vectors (map of displacements between channels) for each channel
         mol_list = np.unique(clump_id)
         n_mols = len(mol_list)
+        if n_mols < 3:
+            raise ValueError('Need at 3 clusters containing points from each channel - try increasing search radius')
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I wonder if it's also worth considering some more general QC criteria here - 3 points is not a lot to generate a shift-field from, and as you have a stated goal of automating a lot of this for end-users I'd be tempted to either:

- fail on a larger number (e.g. 20)
- generate some form of warning
- and/or output some form of quality metric for the shift calibration.

There's something to be said for manually doing the shift analysis as you can see pretty quickly if it's crap. 

In the medium-long term, I wonder if it's worth automatically acquiring data from two ROIs on the bead slide, using one to correct the other, and then quantifying the residuals.

Probably shouldn't stop me merging, but thought it was worth flagging up.

barentine:
Good ideas @David-Baddeley. I was thinking one thing that would be _really_ nice would be reports for shiftmaps and astigmatism calibrations (particularly since I still have a that silent 'message' about mpld3 failing and don't get astigmatism calibration plots on my desktop)

Issue #759, opened 2021-01-07
[BUG] pyme-cluster:// scheme for output modules fails if the directory doesn't exist on the local node


barentine:
**Describe the bug**
![image](https://user-images.githubusercontent.com/31105780/103946806-1db4b300-5105-11eb-9017-31a89b8ff026.png)
![image](https://user-images.githubusercontent.com/31105780/103946856-31f8b000-5105-11eb-832a-2e5d767b4419.png)
![image](https://user-images.githubusercontent.com/31105780/103946891-4046cc00-5105-11eb-9af1-845c90125e0f.png)

for the following recipe
```
- localisations.AddPipelineDerivedVars:
    inputEvents: ''
    inputFitResults: FitResults
    outputLocalizations: input
- tablefilters.FilterTable:
    filters:
      LLH:
      - -250.0
      - 0.0
      error_x:
      - 0.0
      - 20.0
      error_y:
      - 0.0
      - 20.0
    inputName: input
    outputName: filtered
- multiview.Fold:
    input_name: filtered
    output_name: folded
- multiview.CalibrateShifts:
    input_name: folded
    output_name: shiftmap
    search_radius_nm: 500.0
- output.HDFOutput:
    filePattern: '{output_dir}/{file_stub}.sf'
    inputVariables:
      shiftmap: shift_map
    scheme: pyme-cluster://

```

barentine:
@David-Baddeley wouldn't be tough to check clusterIO dir exists for pyme-cluster:// scheme, if we wanted to preserve the default crash on trying to write a table if the directory tree doesn't already exist. Otherwise we could just make it. Any preference? The first seems more safe

David-Baddeley:
I think we should just create it if it doesn't exist - that's the default cluster logic.

PR #768, opened 2021-01-11
consistently make sure we create an output directory for output modules

barentine:
Addresses issue #759 .

enhancement
**Proposed changes:**
- make sure we make the output file directory (for pyme-cluster:// and File schemes) 
- use the exists_ok flag (smack my wrist if we're not py3 yet)
- untab a bonus docstring indent

**Checklist:**

- [ ] Tested 

**side note**
does `ReportForEachOutput` work? Where does env get defined?


David-Baddeley:
- unfortunately can't quite break py2 yet - particularly on something like this which is used quite a lot.
-  `os.path.dirname()`, as was used in the bits which do this already is a bit more readable than `.split()[0]`
- `ReportForEachOutput` definitely works / worked from the bakeshop. Possibly not on the cluster.

barentine:
thanks @David-Baddeley. I PR'd what I meant re: env in #770

David-Baddeley:
as mentioned in #770, suspect `ReportForEach` regressed in 68453d8.

Issue #775, opened 2021-01-12
[BUG] events handling issue with multiview extract channel?


barentine:
![image (5)](https://user-images.githubusercontent.com/31105780/104349614-26283780-54d1-11eb-8022-0f1fa1944d60.png)
**Describe the bug**
Not immediately sure how this happens since the crop datasource should pass through getEvents

```
- rcnn_segment.RCNNDetect:
    device: /gpu:0
    input_name: supertile
    intensity_factor: 0.1
    model_dir: /home/ubuntu/keras_models
    output_mask: masks
    output_rois: detected
- supertile.TilePhysicalCoords:
    input_measurements: detected
    input_supertile: supertile
    measurement_units: um
    output_name: physical_coords
- output.HDFOutput:
    filePattern: '{output_dir}/{file_stub}_detections.hdf'
    inputVariables:
      filtered: filtered
      overlap_identified: og_detected
      subset: roi_locations
    scheme: pyme-cluster://
- measurement.IdentifyOverlappingROIs:
    input: physical_coords
    output: overlap_identified
- tablefilters.FilterTable:
    filters:
      rejected:
      - -0.5
      - 0.5
      x_width:
      - 0.0
      - 30.0
      y_width:
      - 0.0
      - 30.0
    inputName: overlap_identified
    outputName: filtered
- acquisition.QueueAcquisitions:
    action_server_url: http://10.150.7.131:9393
    input_positions: subset
    max_duration: 200.0
    spool_settings:
      maxFrames: 30000
      protocol: htsms-flow.py
      stack: true
      subdirectory: '{file_stub}'
- supertile.Supertile:
    input_name: extracted
    level: 2
    output_name: supertile
- output.ImageOutput:
    filePattern: '{output_dir}/{file_stub}_detections.h5'
    inputName: masks
    scheme: pyme-cluster://
- tablefilters.RandomSubset:
    input: filtered
    num_to_select: 30
    output: subset
- multiview.ExtractMultiviewChannel:
    input_name: input
    output_name: extracted
    view_number: 1

```

**To Reproduce**
1. run the above recipe on a tile pcs

**Expected behavior**
same recipe sans the extract multiview channel module, on a single view tile pcs, works fine

David-Baddeley:
Not much, but error message implies that `events` is a list, rather than a `recarray`. This could be due to loading from json.

David-Baddeley:
Doesn't explain why the same recipe works without the multiview module though.

[...]

PR #776, opened 2021-01-12
slow events loading timeout

barentine:
Addresses issue #775 .

draft pr, testing hack at the moment. once again, something like #161 w/ array format for io would be grand


barentine:
yep, this works and fixes #775

I know you won't like it @David-Baddeley . I wouldn't say that I like it either, but with a sample on the scope half-imaged I'm a bit easier to convince ;)

Issue #693, opened 2020-12-08
[BUG] recipe rule tasks not being marked as complete?


barentine:
**Describe the bug**
Looks like recipe rule tasks are not getting marked as complete. This can be particularly weird when you are queuing acquisition tasks from a recipe rule, and it works, and then it works again, and then it works a third time ;)

**To Reproduce**
1. run a recipe rule
2. see the task get assigned
3. see the CPU usage on a node ramp up
4. See the output file save, see the output acquisition queues queue
5. look at the rulenodeserver temrinals and see 'time elapsed' debug statement from the module you ran, nothing bad
6. unpause your acquisition queue and let the scope run wild
7. wait 10 minutes and see the same tasks get queued again, now with a 'timeout' marked on the ruleserver
![image](https://user-images.githubusercontent.com/31105780/101527502-15c5ef80-395c-11eb-8382-d400aac2ddcc.png)

**Expected behavior**
mark recipe rule tasks complete when they've executed error free and saved outputs error free

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: win10 intrument, ubuntu cluster
 - Python version: 3.6
 - numpy version: 1.16 and 1.19
 - wxpython version [if a UI bug]: 4.0.4

**Additional context**
would make sense in my eyes to change the default number of retries for recipe rules to 1. This should also probably be paired with `optimal-chunck-size=1` in the recipe rule default template, as mentioned in #599

barentine:
Note the comment about changing default number of retries for a recipe rule would generally mean allowing the number of retries to be specified in a rule template, rather than current code which allows it to be spec'd with a config option on the ruleserver and is global for all rules

David-Baddeley:
I'm not so sure there is a case for having a different number of retries for recipes, although there could be one for having a longer timeout. I definitely don't think that we should be specifying the number of retries in the rule template, as this adds a whole new level of complexity where it shouldn't be present (i.e. if something makes this necessary, it's a bug we should fix elsewhere). When submitting a rule, you shouldn't have to think about how many retries you should attempt, but rather just expect it too run.

barentine:
That would certainly be nice ;) Having a look at that now

David-Baddeley:
[...]

PR #780, opened 2021-01-13
timeout taskWorker io loop so we loop every ~5 s even if there are no tasks in the queue

barentine:
Addresses issue #693.

bugfix
**Proposed changes:**
- don't hang on `while len(tasks)==0`
- don't hang on get tasks request which currently doesn't have a timeout

**Checklist:**

- [x] Tested with osx 10.15, py3.7

### DIFF
--- a/PYME/cluster/taskWorkerHTTP.py
+++ b/PYME/cluster/taskWorkerHTTP.py
@@ -206,15 +206,11 @@ class taskWorker(object):
                     if not r.status_code == 200:
                         logger.error('Returning task failed with error: %s' % r.status_code)
 
-    def _get_tasks(self, local_queue_name):
+    def _get_tasks(self, timeout=5):
         """
 
-        Query nodeserver for tasks and place them in the queue for this worker, if available
-
-        Parameters
-        ----------
-        local_queue_name : str
-            computer name prepended by 'PYMENodeServer: '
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Well spotted! Not sure timing out the get() on the client side is going to be a robust solution though (this will potentially result in tasks getting lost if the timeout occurs while tasks are in "transit" from the nodeserver). Better solutions might be to either timeout on the nodeserver and return an empty list, or have separate threads for getting tasks and returning results.

David-Baddeley:
Just double checked - there's already a timeout on the nodeserver, so we should be good. Currently set to 60s (which might be a little on the long side, but probably not by much). Need to double check whether this timeout can be cumulative across all workers (i.e. 12 mins for 12 cores ...).

David-Baddeley:
nope, no locking so timeout should not be cumulative ...

David-Baddeley:
correction:

- we should do away with the `while len() == 0:`
- we should **not** add a timeout to the `get()`

David-Baddeley:
Will merge now (timeout removed) to fix bug

**TODO:** - split i and o loops for more timely feedback.

David-Baddeley:
going to tag i/o separation onto #781 because it's quick and simple

PR #782, opened 2021-01-13
silence print statement everytime we check for tasks

barentine:
Addresses issue my bad. Guess how I debugged this earlier ;)


David-Baddeley:
not 100% sure it's your debug print, seems reasonable to remove anyway

Issue #741, opened 2020-12-23
[BUG] clusterIO failing decent number of frames in localization tasks at high bandwidths/large folder size


barentine:
**Describe the bug**
performance is high and we're keeping up, but dropping quite some frames on io errors
![image](https://user-images.githubusercontent.com/31105780/103029203-46468580-4527-11eb-8f2e-52d9a67bde52.png)

```
===========================================================================
Error in rule: 000083-95aa046a0c394811a3df841956cc6c53 running task: 3500 on elektra:7858

taskDescr:
----------
{'id': '000083-95aa046a0c394811a3df841956cc6c53~3500', 'type': 'localization', 'taskdef': {'frameIndex': '3500', 'metadata': 'PYME-CLUSTER://villains/Bergamot/2020_12_23/EMC0607_try2/B1/000/analysis/23_12_series_L.h5r.json'}, 'inputs': {'frames': 'PYME-CLUSTER://villains/Bergamot/2020_12_23/EMC0607_try2/B1/000/23_12_series_L.pcs'}, 'outputs': {'fitResults': 'http://10.150.7.10:15348/__aggregate_h5r/Bergamot/2020_12_23/EMC0607_try2/B1/000/analysis/23_12_series_L.h5r/FitResults', 'driftResults': 'http://10.150.7.10:15348/__aggregate_h5r/Bergamot/2020_12_23/EMC0607_try2/B1/000/analysis/23_12_series_L.h5r/DriftResults'}}

Traceback:
----------
Traceback (most recent call last):
  File "/home/ubuntu/code/python-microscopy/PYME/cluster/taskWorkerHTTP.py", line 308, in computeLoop
    res = task()
  File "/home/ubuntu/code/python-microscopy/PYME/localization/remFitBuf.py", line 420, in __call__
    bufferManager.updateBuffers(md, self.dataSourceModule, self.bufferLen)
  File "/home/ubuntu/code/python-microscopy/PYME/localization/remFitBuf.py", line 94, in updateBuffers
    self.dBuffer = buffers.dataBuffer(DataSource(md.dataSourceID, md.taskQueue), bufferLen)
  File "/home/ubuntu/code/python-microscopy/PYME/IO/DataSources/ClusterPZFDataSource.py", line 58, in __init__
    self.mdh.update(json.loads(clusterIO.get_file(mdfn, self.clusterfilter)))
[...]

PR #788, opened 2021-01-17
extend default directory listing and get_file timeouts

barentine:
Addresses issue #741 

makeshift bugfix?
**Proposed changes:**
- extend timeouts on clusterIO.get_file and clusterIO._listSingleDir

**Checklist:**

- [x] Tested

### DIFF
--- a/PYME/IO/clusterIO.py
+++ b/PYME/IO/clusterIO.py
@@ -206,10 +206,10 @@ class _LimitedSizeDict(OrderedDict):
                 self.popitem(last=False)
 
 
-_locateCache = _LimitedSizeDict(size_limit=500)
-_dirCache = _LimitedSizeDict(size_limit=100)
+_locateCache = _LimitedSizeDict(size_limit=5000)
+_dirCache = _LimitedSizeDict(size_limit=500)
 DIR_CACHE_TIME = 1
-_fileCache = _LimitedSizeDict(size_limit=100)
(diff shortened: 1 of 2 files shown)


barentine:
works like a charm. ~400x 20k series in with no failures

barentine:
note that for a throttled push script to avoid difficulties encountered with #789, increasing the put_file timeout from 1 s to 5 s is also helpful

David-Baddeley:
I'm a bit nervous about increasing the timeouts as we rely on the operations completing quickly for consistency. On some level, increasing timeouts simply masks the fact that something is not working properly. I wonder if a reasonable compromise might be to increase the timeouts, but also log an error/warning if the operation takes longer than the original timeouts (or potentially a reduced value in the directory listing case, which has already been increased). For `put_file` at least, we already calculate `dt`.

barentine:
Long term I think we might need to make things somewhat adaptive. Everything worked great for 10k frames per series, got strained at 20k, and at 40k is just a disaster. People will surely want to crank longer series than that. Unless we implement an internal/hidden chunking in clusterIO, we'll probably need to rework something

David-Baddeley:
Slightly reduced version merged with #796, as the history on this was a bit messed up and it was easier to start from scratch.

I'm not at all convinced that increasing cache sizes will make for a meaningful improvement in performance and would want that to be well argued. I think they are already pretty generously sized compared to their lifetimes.

PR #790, opened 2021-01-18
More bugfixes for acquisition on py3

David-Baddeley:
@barentine - this makes quantization **off** by default in the spool frame (as this is a safer option - particularly on systems with EMCCD cameras or other cases where the noise model may not have been setup correctly). Current behavior / high throughput behavior can be set with the `Acquire-quantize_by_default` config option. Will need a suitable modification to HT init file to set this config option prior to merging on the HT system.

### DIFF
--- a/PYME/Acquire/ui/HDFSpoolFrame.py
+++ b/PYME/Acquire/ui/HDFSpoolFrame.py
@@ -165,7 +165,7 @@ class PanSpool(afp.foldingPane):
         hsizer.Add(self.cbCompress, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
     
         self.cbQuantize = wx.CheckBox(pan, -1, 'Quantization')
-        self.cbQuantize.SetValue(config.get('Acquire-quantize_by_default', False))
+        self.cbQuantize.SetValue(config.get('spooler-quantize_by_default', False))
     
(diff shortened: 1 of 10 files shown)


barentine:
With all the 'do this right' things we enforce I would not have expected us to brush off not getting electrons per count and offsets right. I thought we'd written in a pop-up dialogue for when you try to quantize without proper metadata?

That said, I don't mind it being disabled by default - probably makes sense even. I think `spooler-quantize_by_default` might be a bit more inline with our other spooler config keys though. Regardless - thanks for the heads up!

PYME/Acquire/ui/HDFSpoolFrame.py
```diff
@@ -164,7 +165,7 @@ def _comp_pan(self, clp):
         hsizer.Add(self.cbCompress, 0, wx.ALL | wx.ALIGN_CENTER_VERTICAL, 5)
     
         self.cbQuantize = wx.CheckBox(pan, -1, 'Quantization')
-        self.cbQuantize.SetValue(True)
+        self.cbQuantize.SetValue(config.get('Acquire-quantize_by_default', False))
```
David-Baddeley:
```suggestion
        self.cbQuantize.SetValue(config.get('spooler-quantize_by_default', False))
```

David-Baddeley:
have changed to `spooler-quantize_by_default`. I also thought we had logic to pick up on it but still ended up getting bung data saved (EMCCD) when I thought I had the right metadata and figured it was safest to turn it off by default.

Issue #789, opened 2021-01-17
[ENHANCEMENT] allow ridiculous numbers of localization tasks to be queued


barentine:
**Is your feature request related to a problem? Please describe.**
Should you manually queue analysis for thousands of series rather than handling them as they come in, you can kill the ruleserver ungracefully when it runs out of RAM for task status arrays

<details>  <summary> Traceback </summary>

```
DEBUG:ruleserver:release_rule_tasks(ruleID = 003269-bb1d660964a94e4ba242ddf291369e50, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003271-c651e1e81a674eb5969361284090d9ee, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003272-6fa95b8bd1e847a68d365a39c1b92c31, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003273-dfbd923b40584294a3cfecca9246c913, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003274-097ca3435fb646e183eec724569e7a9e, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003275-fd9fe4422cb74957813238e2186a97e5, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003276-3800b32f396148d6bc1054d80fc5cd4e, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003277-3910c7cfbe4849c1a2ddf303001a8acf, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003278-ea228e19ae5c4e529f32edcaa671e240, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003279-f89dd659f65f4faf8369ec2b50107919, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003280-d3bd358ec15d465fb1d640a1618dfe1a, release_start=0, release_end=20000)
DEBUG:ruleserver:release_rule_tasks(ruleID = 003281-8e71878c986d416194f5ea78b5256d58, release_start=0, release_end=20000)
[...]

PR #791, opened 2021-01-18
if we know the final number of frames in rules.LocalisationRule, let the ruleserver know

barentine:
Addresses issue #789 .

enhancement
**Proposed changes:**
- don't allocate more resources on the ruleserver than necessary if we know the number of frames already

**Checklist:**

- [x] Tested

### DIFF
--- a/PYME/cluster/rules.py
+++ b/PYME/cluster/rules.py
@@ -568,7 +568,8 @@ class LocalisationRule(Rule):
         self._next_release_start = self.start_at
         numTotalFrames = self.ds.getNumSlices()
         self.frames_outstanding=numTotalFrames - self._next_release_start
-        
+        if self.data_complete:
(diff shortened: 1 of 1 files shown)


PYME/cluster/rules.py
```diff
@@ -568,7 +568,8 @@ def prepare(self):
         self._next_release_start = self.start_at
         numTotalFrames = self.ds.getNumSlices()
         self.frames_outstanding=numTotalFrames - self._next_release_start
-        
+        if self.data_complete:
+            return dict(max_tasks=self.ds.getNumSlices())
```
David-Baddeley, line 572:
```suggestion
            # TODO - add to a post_args dict rather than just returning here??
            return dict(max_tasks=numTotalFrames)
```

avoid one `listdir()` call

Issue #786, opened 2021-01-17
[PR Proposal] switch clusterUI recipe/localization pushing to use Rules


barentine:
**Is your feature request related to a problem? Please describe.**
I'm frustrated when I run localization on the cluster manually and the old pusher starts at frame 10 no matter what, when I want frame 0. 
```
DEBUG:ruleserver:release_rule_tasks(ruleID = 000578-ac0661eacdb540b5a983f3e584c858a2, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000579-8e403050cc594a3889947e73db8b9779, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000580-9ebe24b9f9d84548be754141b774134e, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000581-07056e3c3a234136860a8dd08f06a2d7, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000582-a7243d6fe32644eaa2f112edf7c18215, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000583-9307bcd25b314532afe7677f516ca0ba, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000584-6ebc1b6f082c46b5bd05f90e20e58396, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000585-917813f0d54c405eb912cb69aa902143, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000586-d0002973bdad4348b2cf57de92ea4881, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000587-ded1b52c5443439fb82df7122413e445, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000588-84502c7ec74e4b7880e25826eddf9173, release_start=10, release_end=19999)
DEBUG:ruleserver:release_rule_tasks(ruleID = 000589-66ad5311c87d4011b4c67e98b46b91c7, release_start=10, release_end=19999)

```

additionally, moving everything to rules would reduce duplication, we could consider removing (or loft-ing) the old pushers

**Describe the solution you'd like to implement**
[...]

PR #792, opened 2021-01-18
use localization rule factory in clusterUI localization view

barentine:
Addresses issue #786 .

enhancement
**Proposed changes:**
- use the localization rule factory to instantiate rules to push themselves

**Checklist:**

- Tested with numpy=1.19
- Tested on python 3.7

### DIFF
--- a/PYME/cluster/clusterUI/localization/views.py
+++ b/PYME/cluster/clusterUI/localization/views.py
@@ -122,7 +122,7 @@ def localize(request, analysisModule='LatGaussFitFR'):
 
     #print json.dumps(f.cleaned_data)
     # NB - any metadata entries given here will override the series metadata later: pass analysis settings only
-    analysisMDH = MetaDataHandler.DictMDHandler()
+    analysisMDH = MetaDataHandler.NestedClassMDHandler()
     analysisMDH.update(f.cleaned_data)
 
     #print request.GET
@@ -168,4 +168,3 @@ def localize(request, analysisModule='LatGaussFitFR'):
         raise RuntimeError('Failed to push %d of %d series' % (nFailed, nSeries))
 
     return HttpResponseRedirect('/status/queues/')
-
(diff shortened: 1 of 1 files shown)


PYME/cluster/clusterUI/localization/views.py
```diff
@@ -145,7 +150,8 @@ def localize(request, analysisModule='LatGaussFitFR'):
         for seriesName in seriesToLaunch:
             try:
                 if USE_RULES:
-                    HTTPRulePusher.launch_localize(analysisMDH, seriesName)
+                    context = {'seriesName': seriesName}
+                    rule_factory.get_rule(context=context).push()
```
David-Baddeley, line 154: this is being called incorrectly - should be `rule_factory.get_rule(seriesName=seriesName).push()`

barentine: Pretty sure that'd throw a TypeError unexpected kwarg since `get_rule` only takes `self` and `context`

David-Baddeley: oops, you're right. Does raise the question of what the behaviour should be ....

David-Baddeley: ie: would it make more sense for the signature of `get_rule(self, context)` to be `get_rule(self, **context)`

barentine: can we merge @David-Baddeley ? I don't personally care what the get_rule signature is, but if you change it we should update acquire/ui/rules as well

PYME/cluster/clusterUI/localization/views.py
```diff
@@ -122,7 +122,7 @@ def localize(request, analysisModule='LatGaussFitFR'):
 
     #print json.dumps(f.cleaned_data)
     # NB - any metadata entries given here will override the series metadata later: pass analysis settings only
-    analysisMDH = MetaDataHandler.NestedClassMDHandler()
+    analysisMDH = MetaDataHandler.DictMDHandler()
```
David-Baddeley:
```suggestion
    analysisMDH = MetaDataHandler.NestedClassMDHandler()
```

PR #796, opened 2021-01-20
increase clusterIO timeouts, make long times generate warnings

David-Baddeley:
potentially helpful for issue #741.

Alternative to #788 which has undesirable cache changes and also seems to have obtained bits of other PRs / have a messed up history.


David-Baddeley:
note - commit message is screwy

Issue #794, opened 2021-01-19
[PROPOSAL] add endpoint to actionserver to allow multiple acquisition events to queue at once


barentine:
**Is your feature request related to a problem? Please describe.**
We currently have a race condition, and if you have multiple recipes running which will queue acquisition events, you might get them interweaved, which can result in e.g. spooling a series to the wrong directory, etc. etc.
![image](https://user-images.githubusercontent.com/31105780/105077814-0bb70680-5a5b-11eb-90fe-397bb8e06b66.png)

**Describe the solution you'd like to implement**
add an endpoint for queueing multiple actions, and a lock to the action server so you can queue actions of the same nice in the desired order

David-Baddeley:
I'd been thinking about this problem in a slightly different context, specifically what happens when you are actually using `nice` values to, e.g., insert a high priority task into the queue. Such a task could easily execute between a `state.update` and a `startSpooling`. Adding a new queueing endpoint and lock wouldn't solve this variant of the problem.

I think the right approach is probably going to be to reformulate the tasks so that the combination of state update and action appear as one task ... i.e. something like:

```
with scope.state=...:
    spool_series()
```

barentine:
That would be interesting. I've found it useful/necessary to put the starting state in the front of all my protocols. Would this replace that, or would we want both? I could see this state definition combined with your suggestion in #738  replacing preflights.

In any event, I don't think it would be the end of the world to queue multiple tasks at once rather than throttling a for loop so we don't lock the GUI (as I currently do). Thoughts on that in the mean time David?

barentine:
I could also see being even more flexible with making the action queue task format be a list of chained actions. The typical one would be to update position followed by an acquisition, but could be nice if you wanted to e.g.
`[state.update({...}), spoolController.StartSpooling(protocol='dic-stack'), spoolController.StartSpooling(protocol='smlm')]`,

PR #798, opened 2021-01-20
Some ActionManager improvements

David-Baddeley:
See issue #794 and discussion at https://forum.image.sc/t/documentation-of-queued-actions-in-pymeacquire/46538/4.

Still a bit of a work in progress / draft concept, but replaces actions which get defined as an `eval`able string with callable objects. Adds support for associating state updates with tasks so they don't get separated via a simple chaining setup. Still needs work on UI and web interface.

### Firm bits
- making actions callable objects
- function to queue multiple actions

### Subject to change
- state and action coupling. Philosophically I think I'd prefer a declarative approach - "do this action **with** a specified state" than the imperative "set this state **then** do this action" approach that's coded in the PR. The reason for going for the latter in the interim is that I couldn't immediately think of a compelling way of expressing "centre the ROI on **then**  ..." in a declarative way.

### Example of new interface

```python
my_actions = [UpdateState({'Camera.ROI' : [50, 50, 200, 200]}),
      SpoolSeries(maxFrames=500, stack=False),
      UpdateState({'Camera.ROI' : [100, 100, 250, 250]}).then(SpoolSeries(maxFrames=500, stack=False)),
      ]

ActionManager.queue_actions(my_actions)
```


David-Baddeley:
Note - the `.then()` stuff is going to make, e.g. json serialisation for a web interface a bit nasty.

barentine:
the queueItems.sort() call in actionUI currently hits
```
'<' not supported between instances of 'FunctionAction' and 'FunctionAction'
```

PR #799, opened 2021-01-24
py3 (and other) fixes for pymeacquire web ui

David-Baddeley:
Bring PYMEAcquire webui up to date with Py3 and more recent jupyter, various other fixes/improvements to webui.


David-Baddeley:
Note these changes are not strictly backwards compatible, but as the webui is strictly alpha, that probably doesn't matter.

David-Baddeley:
@barentine - this is ready to merge, and should address #902 as well. The one caveat is that it will likely require a change to recipes which queue acquisitions (due to the new parameterisation of `SpoolControler.start_spooling()` and, as a consequence `actions.SpoolSeries`). Holding off until you've given it a glance. 

Main change is that `start_spooling()` and `SpoolSeries` now take all the series settings in a `settings` dict rather than as individual parameters. Rationale behind this is to permit them to be passed in the body of a REST request, and as a consequence support non-string typed parameters. Changing recipes should be pretty simple.

barentine:
Cool, thanks for pushing on this @David-Baddeley and for flagging me to change my recipes. I'd say go ahead and merge - I'll keep you posted if I run into anything.

I notice you dropped the subdirectory and extra_metadata docstrings. Not sure if this was intentional - I know you aren't a big fan of either of them, but I think it's generally more helpful if you write that you don't like it but leave the doc, otherwise we just end up with more undocumented functionality to break or duplicate later

David-Baddeley:
The doc changes were not intentional - likely a result of having initially not included them in the unified`settings` parameter.

PR #801, opened 2021-01-25
relax zeroconf pin

barentine:
Addresses issue works up until the service info constructor drops the address bit in 0.27.0
enhancement
**Proposed changes:**
relax the pin on zeroconf, following the to-do / most of us using later versions by now anyway

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ ujson
 jinja2
 toposort
 pymecompress >=0.2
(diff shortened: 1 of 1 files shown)


David-Baddeley:
It's largely anecdotal, but I recall some intermediate versions **not** working. As far as I remember this was platform specific (either win or OSX) - I think linux was fine. For that reason, I'd like to do some pretty robust testing before relaxing the pin.

PR #802, opened 2021-01-25
Interlock / triggering off of focus lock IR camera saturation

barentine:
Addresses issue #burning things 

enhancement
**Proposed changes:**
- add an interlock server / client
- trigger a microscope hault on my focus lock camera mostly saturating

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/focus_locks/reflection_focus_lock.py
+++ b/PYME/Acquire/Hardware/focus_locks/reflection_focus_lock.py
@@ -110,7 +110,8 @@ class ReflectedLinePIDFocusLock(PID):
     """
     def __init__(self, scope, piezo, p=1., i=0.1, d=0.05, sample_time=0.01, 
                  mode='frame', fit_roi_size=75, min_amp=0, 
-                 max_sigma=np.finfo(float).max):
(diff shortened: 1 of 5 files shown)


David-Baddeley:
Will probably do some renaming/refactoring/commenting before merging. 

Concerns:

- this feels a bit htsms specific in a few places
- interlock is not a particularly accurate description of the function - it's more of a failsafe
- adding additional HTTP server threads is not ideal (this should go away once we have the acquire server stuff better integrated, but code should be commented to indicate this)

Longer term:
- as currently coded, this is susceptible to the focus lock process crashing.

barentine:
All good points David. I'm fine with a rename etc. as you see fit. I hoped I wrapped enough of the scope attributed based kills in try's such that if someone else wanted to use it they wouldn't have to have all my stuff. I thought one option for flexibility would be to use a task list protocol for it so people could write their own shutdown procedure, but given that I had a sample in the wing and an objective not to cook I figured we could cross that bridge later.

How is the acquire server stuff working these days, and re: susceptible to the focus lock process crashing, what is the risk of tying it into the same server that is running the microscope we are trying to failsafe?

Issue #803, opened 2021-01-25
[BUG] django version ratcheting is incomplete


barentine:
**Describe the bug**
django 2.1.7:
```
Traceback (most recent call last):
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/home/smeagol/code/python-microscopy/PYME/cluster/clusterUI/localization/views.py", line 104, in settings
    'categorized_fields' : fields_by_category})
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/shortcuts.py", line 36, in render
    content = loader.render_to_string(template_name, context, request, using=using)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/backends/django.py", line 61, in render
    return self.template.render(context)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/base.py", line 171, in render
    return self._render(context)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/base.py", line 163, in _render
    return self.nodelist.render(context)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/base.py", line 937, in render
    bit = node.render_annotated(context)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/base.py", line 904, in render_annotated
    return self.render(context)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/defaulttags.py", line 209, in render
    nodelist.append(node.render_annotated(context))
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/base.py", line 904, in render_annotated
    return self.render(context)
  File "/home/smeagol/miniconda3/lib/python3.6/site-packages/django/template/defaulttags.py", line 209, in render
    nodelist.append(node.render_annotated(context))
[...]

PR #804, opened 2021-01-25
django widgets fix

barentine:
Addresses issue #803 

bugfix
**Proposed changes:**
`TextInput._format_value` doesn't exist, and renderer must be passed even if null


David-Baddeley:
cool, looks like the new name is present from 1.10 so this is backwards compatible.

PR #806, opened 2021-01-26
PYMEAcquire rule ui refactoring (Pr779+)

David-Baddeley:
A few mods on top of #779 

@barentine - nothing too dramatic, but thought you might want to have a quick glance before I merge

Refactoring probably best seen in context of future vision for analysis chaining (outlined below). In that context I'd like to keep this PR as silo-ed as possible so that we can get it in but build v2 without having to worry about breaking the HTSMS workflow. 

Vision:

- Single page for editing analysis chains
- List of steps in chain on left, editor for selected step (recipe or localisation) on right
- Editors modify the `analysisMDH` or `recipe` of the relevant `RuleFactory` (lets you edit a rule chain rather than having to create from scratch)
- Do not use the global `analysisSettings` object for localisation settings in the rule chain (problematic for flexible rule chains, as it forces all localisation settings to be the same across all acquisitions)
- Do not choose rule chains based on protocol (this seems very fragile, not super intuitive, and doesn't scale to non-protocol acquisitions - e.g. z stacks, some tiling forms, future extensions to deterministic timing) - would probably require a new argument to 'StartSpooling()' instead.
- Loading and saving of rule chains
- 2 analysis start timing options - 'on_trigger' and 'on_completion' (with the 'triggger' being via either a protocol task or manually). No 'on_start' when this is not in the protocol, as I don't think this is possible to do robustly. Potentially a 3rd option for no automated analysis (although this could also be indicated by not having a rule chain selected).
- 2 start modes - 'full auto' and 'via dh5view'


PYME/ui/AUIFrame.py
```diff
@@ -129,8 +129,18 @@ def CreateFoldPanel(self):
         self.Update()
         self._mgr.Update()
     
-    def select_page_by_name(self, name):
-        """set a page to be active, using just it's caption
+    def _select_page_by_name(self, name):
+        """
+        WARNING - ADDED FOR TEMPORARY FUNCTIONALITY - WILL PROBABLY BE REMOVED
+        
+        FIXME - this is likely broken for page captions with spaces (where name != caption), this includes the use case
```
barentine, line 136: ask me how I know it wasn't broken for the use case I added it for ;)

barentine:
Just had a quick look - looks good. I'll PR later if I run into anything. Thanks!

PYME/ui/AUIFrame.py
David-Baddeley: I thought it must have worked, but couldn't work out why it did ... looking at the AUI code it's as follows:

`_mgr.GetPaneInfo(bad_name)` returns `NonePaneInfo`

`NonePaneInfo.notebook_id == -1`

So the code as it stands will select the **last** notebook. Meaning that it will work if and only if there is only one notebook in the UI. This is usually, but not always the case.

barentine: It works for `Simulation Settings`, `Queued Actions`, and `Chained Analysis`. On wx 4.0.4 if that matters

David-Baddeley: If you, e.g. drag the acquisition tasks on top of the hardware panel to maximise screen estate

image

or e.g. drag the shell and queued actions panels to a new notebook at the bottom  so you can interact with them whilst still looking at the preview image you'll run into problems. It's a bit of an edge case, but my point is that the current code relies on a quirk.

barentine: Didn't actually know you could do that. It still worked for me in the use case you screenshot, and the `bad_name` case documented in the code, but you're right that you can derail it if you make a new notebook by dragging the shell and queued actions to the bottom, after which it doesn't work regardless of caption

Issue #816, opened 2021-02-01
[BUG]Cannot use dSView.module.camera_maps


HelenSun09:
**Describe the bug**

In the miniconda3 prompt where I typed "dh5view":
c:\users\helen sun\python-microscopy\PYME\DSView\splashScreen.py:117: wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  MemBitmap = wx.EmptyBitmap(s.GetWidth(), s.GetHeight())
c:\users\helen sun\python-microscopy\PYME\DSView\dsviewer.py:144: wxPyDeprecationWarning: Call to deprecated item. Use Append instead.
  tmp_menu.AppendMenu(-1, 'Save &Results', self.save_menu)
WARNING:PYME.DSView.modules:Plugin [arrayView] injects into dsviewer namespace, could result in circular references
WARNING:PYME.DSView.modules:Plugin [shell] injects into dsviewer namespace, could result in circular references
0
WARNING:PYME.DSView.modules:Plugin [metadataView] injects into dsviewer namespace, could result in circular references
ERROR:PYME.recipes.modules:Could not import skimage
Traceback (most recent call last):
  File "c:\users\helen sun\python-microscopy\PYME\recipes\modules.py", line 25, in <module>
    from . import skfilters
  File "c:\users\helen sun\python-microscopy\PYME\recipes\skfilters.py", line 14, in <module>
    import skimage.filters as skf
ModuleNotFoundError: No module named 'skimage'
0
Creating fold panel
c:\users\helen sun\python-microscopy\PYME\DSView\scrolledImagePanel.py:52: wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  MemBitmap = wx.EmptyBitmap(s.GetWidth(), s.GetHeight())
c:\users\helen sun\python-microscopy\PYME\DSView\arrayViewPanel.py:1154: wxPyDeprecationWarning: Call to deprecated item ImageFromData. Use :class:`Image` instead.
  img = wx.ImageFromData(ima.shape[1], ima.shape[0], ima.ravel())
c:\users\helen sun\python-microscopy\PYME\DSView\arrayViewPanel.py:513: wxPyDeprecationWarning: Call to deprecated item BitmapFromImage. Use :class:`wx.Bitmap` instead
  im2 = wx.BitmapFromImage(im)
Creating fold panel
Traceback (most recent call last):
[...]

PR #817, opened 2021-02-01
make rulenodeserver safer for directory trees with spaces

barentine:
Addresses issue #816 .

depends on your point of view
**Proposed changes:**
- enclose profile directory in double quotes
- use sys.executable instead of `python`

**Checklist:**

- [x] Tested on my machine which doesn't actually have spaces in root/python directories


David-Baddeley:
this modifies the old, unused, version of the nodeserver, so is unlikely to fix any current issue.

barentine:
added the real one. Thanks @David-Baddeley

PR #820, opened 2021-02-02
add sum projection

barentine:
Addresses issue #someone just asked me if we can and I was embarrassed to tell them to 'Mean' and then scale up or scale everything else down
enhacenment
**Proposed changes:**

**Checklist:**

- [ ] Tested

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -1971,7 +1971,7 @@ class Projection(Filter):
     TODO - make this more efficient - we currently force the whole stack into memory
     """
     
-    kind = Enum(['Mean', 'Max', 'Median', 'Std', 'Min'])
+    kind = Enum(['Mean', 'Sum', 'Max', 'Median', 'Std', 'Min'])
     axis = Int(2)
     
     processFramesIndividually = False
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I guess adding a sum projection might not hurt - I do feel that it's a bit redundant though and struggle to think of a case where it is preferable to choose sum over mean. If we do add a sum option, it would be nice if we had a way to push users towards mean instead. 

Mean was chosen over sum as it's a lot more robust in it's results when you have, e.g. variable stack sizes) - if you were e.g., thresholding the projected image a mean projection will work with the same threshold over different input sizes whereas a sum projection won't. The same applies to, e.g. offset parameters in deconvolution and/or fitting. Parameters for a mean projection will also be roughly the same as for a single image (and close to what you'd use for median, min, and max projections). My gut feeling is that, if you have the choice of mean projection and sum projection you want mean projection 95% of the time.

barentine:
The answer to the question ‘how much X is in this cell?’ is sum not mean. I agree with you, 95% of the time mean is better, but I’m not trying to do everyone’s analysis for them at the expense of withholding the tool they need (or would like, anyway) for the job

PR #830, opened 2021-02-09
Convert keep_largest_connected_component to optionally keep n largest

zacsimile:
Maybe should convert to keep by size? Or should that be a new function? Should size be defined by the number of elements or by the "radius" of the object (1/2 largest distance between any two vertices in the mesh)?

### DIFF
--- a/PYME/experimental/_triangle_mesh.pyx
+++ b/PYME/experimental/_triangle_mesh.pyx
@@ -2358,20 +2358,25 @@ cdef class TriangleMesh(TrianglesBase):
             # iteration's search will be part of another component.
             component += 1
 
-    def keep_largest_connected_component(self):
+    def keep_largest_connected_component(self, n=1):
         # Find the connected components
         self.find_connected_components()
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Might be worth eventually refactoring roughly along the following lines:

```python

@property
def connected_components(self):
    """ identifies connected components and returns some form of index which lets you identify which vertices/halfedges/etc
    belong to a given component.

    Could be a list of lazy `Component` proxy objects which had indices etc, quantification functions (size, number of vertices
    etc ... available as methods/properties - would probably argue for computing these as needed.
   """ 

def keep_connected_components(self, indices : iterable):
    """ keep the connected compontents having the given indices/labels"""
```

That would give the end user a way of selecting the ones they want, and also (eventually) facilitating quantifications - e.g. volume, surface area, etc ... that people might want to do on meshes.

keep largest 10 would then work out as, e.g.:
```python
mesh.keep_connected_components(np.argsort([c.n_vertices for c in mesh.connected_components])[-10:])
```

PR #831, opened 2021-02-09
add cropping recipe module

barentine:
Addresses issue #no recipe module for cropping.

enhancement
**Proposed changes:**
- add a recipe module wrapping the crop datasource
- add a unit test for said module
- template applyFilter on filtermodule (irrelevant, but there. oops)

**Checklist:**

- [x] Tested with numpy=1.19 python 3.7 wx=4

### DIFF
--- a/PYME/IO/DataSources/CropDataSource.py
+++ b/PYME/IO/DataSources/CropDataSource.py
@@ -38,7 +38,7 @@ class DataSource(BaseDataSource):
             x0 = xrange[0] if xrange[0] > -1 else self.dataSource.shape[0] + xrange[0]
             x1 = xrange[1] if xrange[1] > -1 else self.dataSource.shape[0] + xrange[1]
             assert x1 - x0 > 0 and x1 - x0 <= self.dataSource.shape[0]
-            assert x0 < self.dataSource.shape[0] and x0 > 0
+            assert x0 < self.dataSource.shape[0] and x0 >= 0
             assert x1 < self.dataSource.shape[0] and x1 > 0
             self.xrange = (x0, x1)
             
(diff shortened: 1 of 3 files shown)


PYME/recipes/base.py
```diff
@@ -1411,6 +1428,29 @@ def execute(self, namespace):
...
+    t_range = List(Int)([0, -1])
+    output = Output('cropped')
+
+    def execute(self, namespace):
+        from PYME.IO.DataSources.CropDataSource import DataSource
+        from PYME.IO.MetaDataHandler import DictMDHandler
+        
+        im = namespace[self.input_stack]
+        mdh = DictMDHandler(im.mdh)
+        mdh['Crop.XRange'] = self.x_range
```
David-Baddeley, line 1445: `cropping.xslice`, `cropping.yslice`,  and `cropping.zslice`  are currently used to indicate cropping in the metadata (not to say that this is necessarily right, but I'm reluctant to introduce new conflicting conventions.)

PYME/recipes/base.py
```diff
@@ -1411,6 +1428,29 @@ def execute(self, namespace):
...
+        mdh['Crop.YRange'] = self.y_range
+        mdh['Crop.TRange'] = self.t_range
```
David-Baddeley, line 1447: Definitely need to update the origin metadata as well.

PYME/IO/DataSources/CropDataSource.py
```diff
@@ -29,20 +29,41 @@ def __init__(self,dataSource, xrange=None, yrange=None, trange=None):
         #self.unmixer = unmixer
         self.dataSource = dataSource
         
-        if xrange is None:
+        sizeC = self.dataSource.sizeC
+
+        if xrange is None or tuple(xrange) == (0, -1):
```
David-Baddeley, line 34: if you want to use -1 indexing etc ... it would probably be better to use a `slice()` object. `slice.indices` would likely also let you eliminate much of the boilerplate in each of the cases below.

David-Baddeley: but I'm not sure I'm particularly keen on negative indexing in this case.

Issue #838, opened 2021-02-10
[BUG] Don't run recipes if I hit cancel


zacsimile:
If I click on a recipe tile in the `Pipeline Recipe` tab in `PYMEVis` and the dialog pops up and then I hit cancel, the recipe re-runs. This shouldn't happen.

David-Baddeley:
Which traits / traitsui version? This smells a little like a traits error.

David-Baddeley:
Actually, there is no way of robustly handling a cancel - traits are updated as you edit them. This makes cancel functionally equivalent to OK. Have removed the cancel button as it was confusing.

David-Baddeley:
PS - I assume it only runs if you have changed a parameter

zacsimile:
It runs even if I don’t change a parameter. If I just hit cancel, it runs.

David-Baddeley:
That's odd, as it should check if new_params==old_params and only run if they are different.

David-Baddeley:
see `ModuleBase.edit_no_invalidate()`

zacsimile:
You sufficiently convinced me to try again. I even add a print statement "Executing..." to a few of the `execute()` commands in some recipes (e.g. DBSCAN). If I click and then hit cancel, I get the print statement, even if I change no parameters.

David-Baddeley:
I've found the bug here which is making things execute when they shouldn't. Fix to follow (probably the *real* fix to this issue - I might have closed prematurely)

zacsimile:
Thanks for pushing on this, @David-Baddeley!

Issue #832, opened 2021-02-09
[BUG] can click-through recipe configure traits dialogs, opening new ones on accident


barentine:
**Describe the bug**
in the recipeGUI, at least in PYMEImage, I'm hitting a funny bug where I can accidentally open a fair number of recipe module configuration dialogs while trying to edit the first one I pulled up

**To Reproduce**
1. PYMEImage something
2. Click on 'Recipe'
3. add some stuff
4. click a module to configure it
5. click in that dialog into a field which happens to be over the tile for another recipe module in the gui behind it
6. notice that you get another recipe module dialog open (behind?) the one you were working on in the first place (maybe it's only for the same one you have up? Didn't try others)

**Expected behavior**
one dialog at a time

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] osx 10.15
 - Python version: 3.7.7
 - numpy version: 1.19.2
 - wxpython version [if a UI bug]: 4.0.4

zacsimile:
Yep, thanks for flagging that @barentine. Same issue. I ran into this in PYMEVis.

David-Baddeley:
Hmm, this looks like an error with traitsui and/or matplotlib. Might be hard to work around without significant re-writes to recipe display and/or editor.

Simplest potential fix would be to make the pop-up editors modal, but not sure that will be enough.

csoeller:
I can also attest to this behaviour. A bit annoying, can be worked around (move window before doing anything with it), but would be good to fix for sure.

David-Baddeley:
A bit more on potential solutions:

- try making the dialogs modal (this might also improve cancel handling - see #838). Potentially combine with a new editor dialog which shows the module help (on TODO list in any case).
- make editor pop up as a panel on the side/bottom of the recipe display rather than in a separate window
- change to using the html based recipe editor (advantageous in that we'd only need to maintain one editor, but potentially problematic on linux and/or windows, also not there yet)

David-Baddeley:
should be fixed with #869 (at least until we refactor stuff properly)

PR #843, opened 2021-02-11
remove cancel button as it doesn't cancel.

David-Baddeley:
Addresses issue #838.

Removes the non-functional cancel button - there is no easy way to implement cancel, as parameters get changed as they are edited, so it's better not to display it.

### DIFF
--- a/PYME/recipes/base.py
+++ b/PYME/recipes/base.py
@@ -423,7 +423,7 @@ class ModuleBase(HasTraits):
                     [Item('_'),] +
                     self._view_items(params) +
                     [Item('_'),] +
-                    [Item(tn) for tn in outputs], buttons=['OK', 'Cancel']) #TODO - should we have cancel? Traits update whilst being edited and cancel doesn't roll back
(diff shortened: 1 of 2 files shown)


zacsimile:
I feel like not having a `Cancel` button is going to drive users crazy, esp. with #832...

David-Baddeley:
Well it's better to not have one than have one which means `OK` - a proper cancel would be better, but unfortunately not doable with the way traits works ...

zacsimile:
I agree it's confusing to have a `Cancel` button that doesn't cancel. Why can't we use the same code we use to rollback the pipeline when a user hits cancel initially adding a recipe?

David-Baddeley:
because the rollback on addition works by never adding it to the pipeline in the first place - we create the module, but don't actually add it unless we click OK. I do worry that we might have broken that cancel here though ...

zacsimile:
Got it. I thought it like froze the previous state of the pipeline and rolled back to that if the module was canceled. Thought that might work the same. Sounds like I was wrong, but you found the fix. Thanks!

PR #849, opened 2021-02-12
allow shortcircuit for pyme-cluster:///

barentine:
Addresses issue `unifiedIO.local_or_temp_filename('pyme-cluster:///something') will not work if you have a localserverfilter config'd, which is I think more cruel than we need to be there - if you specify you want it from a given cluster fine, but if you don't include it that's just annoying

enhancement
**Proposed changes:**

**Checklist:**

- [x] Tested with numpy=1.19 and 3.7


David-Baddeley:
We could also consider using the `serverfilter in name` logic that is used elsewhere in clusterIO (rather than testing for equality or '')

Issue #822, opened 2021-02-04
[ENHANCEMENT] retries on failed software trigger frames


barentine:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when the ocassional tile overview locks on
```
DEBUG:PYME.Acquire.HTTPSpooler:Starting spooling: Bergamot/2021_2_2/EMC0104S3_4/D4.pcs
'http://10.150.7.4:15348/Bergamot/2021_2_2/EMC0104S3_4/D4.pcs/metadata.json'
INFO:PYME.Acquire.htsms.rule_ui:protocol name : htsms-tile
DEBUG:PYME.Acquire.Utils.pointScanner:Current positions: {'x': 36000.0, 'x_target': 36000.0, 'y': 27000.0, 'y_target': 27000.0, 'z': 49.9697, 'z_target': 50.0}
Warning: not keeping up with camera, giving up with 0 frames still in buffer
nse
ERROR:PYME.Acquire.Hardware.HamamatsuDCAM.HamamatsuORCA:DCAM error for dcamwait_start with return value -2147482879: Lost Frame detected!
```

**Describe the solution you'd like**
when this happens it can be rescued through the shell with
```
scope.frameWrangler.stop()
scope.frameWrangler.Prepare()
scope.frameWrangler.start()
scope.cam.FireSoftwareTrigger()
```
I do not know how necessary the framewrangler restart is, but in the past when we've tried simply pausing and following with another software trigger we haven't had much luck. If I hit this again I'll try just with the trigger first and see what happens.

**Describe alternatives you've considered**
failure

David-Baddeley:
a couple of quick thoughts:

- it would probably be good to handle this on the DCAM side
- might be worth replacing the `logger.error(...)` call on line 349 of HamamatsuORCA with `logger.exception()` so we get a traceback and a slightly more informative message.
- potential options for handling on the camera end would be:
    - call `fire_trigger()`again if in triggered mode??
    - return a null frame
    - propagate the exception upwards

for error checking, in the camera driver it might be worth refactoring `DCAMException` as, e.g. `DCAMException(fn, ret_val, message='')`, and keeping `ret_val` as, e.g. an `.error_code` attribute so that the exception handler can check the error code before deciding whether to ignore or propagate the exception.

[...]

PR #850, opened 2021-02-12
move metadata / event pushing in HTTPSpooler into doStopLog

barentine:
Addresses issue race condition mentioned in #822 .

**Proposed changes:**
- move metadata and events put into doStopLog. This moves them into `Spooler.StopSpool` rather than after, but importantly puts them after disconnecting from frame source, after protocol.finish() which can still generate events, but _before_ `self.spoolOn = False`.

**Checklist:**

- [x] Tested

### DIFF
--- a/PYME/Acquire/HTTPSpooler.py
+++ b/PYME/Acquire/HTTPSpooler.py
@@ -236,6 +236,26 @@ class Spooler(sp.Spooler):
         else:
             clusterIO.put_file(self.seriesName + '/metadata.json', self.md.to_JSON().encode(), serverfilter=self.clusterFilter)
     
+    def doStopLog(self):
+        sp.Spooler.doStopLog(self)
+        # save events and final metadata
+        # TODO - use a binary format for saving events - they can be quite
+        # numerous, and can trip the standard 1 s clusterIO.put_file timeout.
+        # Use long timeouts as a temporary hack because failing these can ruin
+        # a dataset
+        if self._aggregate_h5:
+            clusterIO.put_file('__aggregate_h5/' + self.seriesName + '/final_metadata.json',
+                               self.md.to_JSON().encode(), self.clusterFilter)
+            clusterIO.put_file('__aggregate_h5/' + self.seriesName + '/events.json',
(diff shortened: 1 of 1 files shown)


barentine:
note that this also avoids the events/rule pushing race condition I ran into without any clusterpzfdatasource special casing

David-Baddeley:
Unfortunately this might not work as intended as the existence of `final_metadata.json` / `events.json` is used in pushers to confirm whether spooling is complete (and, by inference, is assumed to occur **after** all the frames have been pushed). 

Hard to be 100% certain if this will cause any issues, but there is a small chance that it will give rise to yet another race condition down the track. It might be safer to leave writing these files where it is and add a new `spool_complete` flag to check in the return from `StartSpooling()` instead. Will see if I can quickly pull together an alternative.

barentine:
ah, I see your point. I do kind of think this is still the right place to push the events/metadata though, and wonder if it should just be combined with a spool_complete signal for the pushers to connect?

David-Baddeley:
making the pushers aware of the spooler completion (rather than relying on file existence checks) is on the TODO list, and would be nice, but will only ever work for pushers which are in-process - and not for pushers which are running elsewhere - e.g. triggered from ClusterUI, or from the ruleserver, e.g.,  with a rule that monitors directories for changes (If you notice the rule structure, `IntegerIDRule` is currently the only rule type implemented on the ruleserver, but it derives from a `Rule` base class for a reason- the long term plan has always been to add GLOB based rules - i.e. generate tasks for all files matching this pattern - as well).

barentine:
I guess I don't see how we're answering anything other than in-process pushers right now? A non in-process pusher e.g. on clusterUI will still check for the events file

--edit-- sorry, this was completely missing your last point. I'm there now!

David-Baddeley:
And a non in-process pusher will be fine as long as the events and/or
metadata files land after the frames. Changing that order will mess up non
in process pushers.

David-Baddeley:
Closing as issue addressed in #857 in a slightly safer / more paranoid way

Issue #840, opened 2021-02-10
[BUG] actions.FunctionAction comparison


barentine:
**Describe the bug**
```
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\ui\mytimer.py", line 42, in Notify
    a()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\ActionManager.py", line 184, in Tick
    self.currentTask = self.actionQueue.get_nowait()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\queue.py", line 192, in get_nowait
    return self.get(block=False)
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\queue.py", line 174, in get
    item = self._get()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\queue.py", line 230, in _get
    return heappop(self.queue)
TypeError: '<' not supported between instances of 'FunctionAction' and 'FunctionAction'
Traceback (most recent call last):
  File "c:\users\bergamot\code\python-microscopy\PYME\ui\mytimer.py", line 42, in Notify
    a()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\ActionManager.py", line 184, in Tick
    self.currentTask = self.actionQueue.get_nowait()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\queue.py", line 192, in get_nowait
    return self.get(block=False)
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\queue.py", line 174, in get
    item = self._get()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\queue.py", line 230, in _get
    return heappop(self.queue)
TypeError: '<' not supported between instances of 'FunctionAction' and 'FunctionAction'
```

David-Baddeley:
This looks like it has to be a `nice` collision - is there a race condition when adding actions from multiple workers? Potentially resolvable with some locking on the endpoint.

barentine:
closing, will reopen if we hit this again

barentine:
this is still very much an issue

barentine:
I thought maybe it was queueing these turnAllLasersOff  with `np.iinfo(int).max` for their nice, but 
```
In [5]: np.iinfo(int).max + (time.time() * 1e-10) == np.iinfo(int).max
Out[5]: False

[...]

PR #854, opened 2021-02-14
enforce bounds on nice

David-Baddeley:
Addresses issue #840.

as in the title - keep nice to a reasonable range to prevent floating point precision issues

### DIFF
--- a/PYME/Acquire/ActionManager.py
+++ b/PYME/Acquire/ActionManager.py
@@ -93,7 +93,9 @@ class ActionManager(object):
             a dictionary of arguments to pass the function    
         nice : int (or float)
             The priority with which to execute the function. Functions with a
-            lower nice value execute first.
+            lower nice value execute first. Nice should follow the convention of the unix function with the same name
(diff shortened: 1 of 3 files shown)


barentine:
would you mind also changing the line in `recipes.acquisition` if you're going to hard-throw like that? I changed it local to 100 when I commented #840. Didn't know you thought of nice as being negative, nor that there's a unix function by the same name. Is that convention intuitive enough to replicate here?

David-Baddeley:
I'm not sure that -ve nices are that intuitive, and/or if it really makes sense to follow the unix nice range. We could potentially also do, e.g. 0-20, or potentially even 0-100 (100 would put the precision break point at ~100us, which is still OK with our 1ms forced increment). I do think that you're probably doing something wrong if you need more than ~20 priorities.

I figured you'd know better than I would where it might have used with out of bounds values, which is why it's not merged yet ...

barentine:
Thanks for doing that David. Just `acquisition` and `tile_panel` I think

PR #858, opened 2021-02-16
Test fixes

David-Baddeley:
Addresses issue fix a bunch of broken tests


David-Baddeley:
merged early as it also fixes #861

PR #860, opened 2021-02-16
clusterUI upload files and glob select files url fix

barentine:
Addresses issue #803.

### DIFF
--- a/PYME/cluster/clusterUI/clusterbrowser/views.py
+++ b/PYME/cluster/clusterUI/clusterbrowser/views.py
@@ -110,7 +110,7 @@ def upload(request, directory):
     request.upload_handlers.insert(0, TemporaryFileUploadHandler(request))
     return upload_files(request, directory)
 
-@csrf_protect
+@csrf_exempt
 def upload_files(request, directory):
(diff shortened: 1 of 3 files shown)


PYME/cluster/clusterUI/clusterbrowser/views.py
```diff
@@ -110,7 +110,7 @@ def upload(request, directory):
     request.upload_handlers.insert(0, TemporaryFileUploadHandler(request))
     return upload_files(request, directory)
 
-@csrf_protect
+@csrf_exempt
```
David-Baddeley: This is definitely not something we should be exempting (the fact that we had specifically asked to protect it might be a clue). In practice we should probably protect this endpoint better - i.e. enforce authentication once that is fully rolled out.

barentine: re:clues `upload` just above it is exempt

David-Baddeley:
```suggestion
@csrf_protect
```
```suggestion
@csrf_exempt
```

Issue #861, opened 2021-02-16
[BUG] cannot open h5r files


csoeller:
**Describe the bug**
Hi @David-Baddeley, @barentine, @zacsimile, I am not sure if I have a hosed installation or if something changed in latest git. Opening any `.h5r` I now get:

```
Traceback (most recent call last):

  File "/Users/csoe002/opt/anaconda2/envs/pyme-default-plain/lib/python2.7/site-packages/wx-3.0-osx_cocoa/wx/_misc.py", line 1367, in Notify

    self.notify()

  File "/Users/csoe002/opt/anaconda2/envs/pyme-default-plain/lib/python2.7/site-packages/wx-3.0-osx_cocoa/wx/_core.py", line 16869, in Notify

    self.result = self.callable(*self.args, **self.kwargs)

  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/LMVis/visCore.py", line 754, in OpenFile

    self.pipeline.OpenFile(filename, **args)

  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/LMVis/pipeline.py", line 825, in OpenFile

    self.recipe.execute()

  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/recipes/base.py", line 718, in execute

    m.execute(self.namespace)

  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/recipes/localisations.py", line 291, in execute

    cached_output.mdh = output.mdh

AttributeError: can't set attribute
```

Nothing to look at in the viewer at this stage, obviously.

**To Reproduce**

Open any `h5r` file.

**Expected behavior**
Not like this.

csoeller:
Looks we need to add a setter for the `mdh` property to the cached class. With that it works again. Patch below.

```patch
diff --git a/PYME/IO/tabular.py b/PYME/IO/tabular.py
index f0f08cbbb..a78b02176 100644
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -917,6 +917,10 @@ class CachingResultsFilter(TabularBase):
     def mdh(self):
         return self.resultsSource.mdh
 
+    @mdh.setter
+    def mdh(self, mdh):
+        self.resultsSource.mdh = mdh
+
 @deprecated_name('mappingFilter')
 class MappingFilter(TabularBase):
     _name = "Mapping Filter"
```

barentine:
[...]

PR #862, opened 2021-02-16
caching results filter metadata setter

barentine:
Addresses issue #861.

bugfix
**Proposed changes:**

not sure if you envision anything else with the mdh property @David-Baddeley but I needed this on the cluster via git so am submitting @csoeller 's patch while here. Thanks @csoeller !


barentine:
see #858

PR #863, opened 2021-02-17
add file handler to PYMEAcquire logger

barentine:
Addresses issue #853

enhancement
**Proposed changes:**
- get the root logger that we just set up rather than a __file__ logger
- add a rotatingfilehandler to said logger
- try and set file logging to INFO, though seems to still write debug to file...

### DIFF
--- a/PYME/Acquire/PYMEAcquire.py
+++ b/PYME/Acquire/PYMEAcquire.py
@@ -88,7 +88,12 @@ def main():
     
     from PYME import config
     
-    logger = logging.getLogger(__name__)
+    logger = logging.getLogger()
+    fh = logging.handlers.RotatingFileHandler(filename=os.path.join(config.get('dataserver-root'), 'LOGS', 'PYMEAcquire.log'), 
+                                              mode='w', maxBytes=1e6, backupCount=1)
+    fh.setLevel(logging.INFO)
+    logger.addHandler(fh)
+
     parser = OptionParser()
     parser.add_option("-i", "--init-file", dest="initFile",
                       help="Read initialisation from file [defaults to init.py]",


David-Baddeley:
As discussed earlier, PYMEAcquire logging is configured using `logging.json` in the PYMEAcquire directory (or a per-site alternative as specified by an environment variable).

Multiple potential issues with the alternative approach here:
- `logging.getLogger()` should be called with the module name (`__name__`)
- although `dataserver-root` is somewhat special cased with a default, it's probably not a good to rely on any config option being defined. When spooling to a cluster (in particular) it's probably not great to assume that the value of `dataserver-root` has any meaning on the acquisition computer (which is not running a dataserver). It's also a specific parameter for a specific spooling backend, and PYMEAcquire (on a general level) should not rely on anything to do with the specific spoolers beyond actually using them to spool.
- what we'd discussed was making the console log use `logging.INFO` as it's level, and the file logger using `logging.DEBUG`. This is the other way round.

barentine:
Hey David, thanks again for pointing out the file yesterday. It was a long day for me (09 - 01) and in the end I just needed this to work so I could image. With the dictconfig I ran into some confusion over how to set up a filehandler, given it takes a filename and I didn't want to hardcode that in a json. Are you suggesting I should instead just write my own json and use an env variable? I thought we were generally interested in file logging for everyone.

I would think the way we cab get what you want is to dictconfig a debug-level logger as we have now, but add a console handler set to info and a file handler set to debug. Point well taken about dataserver-root, I assume using config['datadir'] or so would be preferable. as to 'should use __name__' in PYMEAcquire I'm not quite sure on that. Would that mean just adding our handlers for the default logger, still using other loggers everywhere, and magically logging'll take care of it for us?

barentine:
While we're not using this PR, I'll note that it had some issues. Note sure if we'll hit this again in the future or not, but suspect clearing out wx/clusterIO 404s on directory listings will go some way
```
--- Logging error ---
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\logging\handlers.py", line 72, in emit
    self.doRollover()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\logging\handlers.py", line 173, in doRollover
    self.rotate(self.baseFilename, dfn)
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\logging\handlers.py", line 113, in rotate
    os.rename(source, dest)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\Users\\Bergamot\\PYMEData\\LOGS\\PYMEAcquire.log' -> 'C:\\Users\\Bergamot\\PYMEData\\LOGS\\PYMEAcquire.log.1'
Call stack:
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\threading.py", line 884, in _bootstrap
    self._bootstrap_inner()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "c:\Users\Bergamot\Anaconda2\envs\chaos\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\frameWrangler.py", line 288, in _poll_loop
    self.onExpReady()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\frameWrangler.py", line 236, in onExpReady
    self.onFrame.send(sender=self, frameData=d)
  File "c:\users\bergamot\code\python-microscopy\PYME\contrib\dispatch\dispatcher.py", line 174, in send
    for receiver in self._live_receivers(sender)
  File "c:\users\bergamot\code\python-microscopy\PYME\contrib\dispatch\dispatcher.py", line 174, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\HTTPSpooler.py", line 293, in OnFrame
    sp.Spooler.OnFrame(self)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Spooler.py", line 237, in OnFrame
    self.StopSpool()
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\Spooler.py", line 175, in StopSpool
    self.onSpoolStop.send(self)
  File "c:\users\bergamot\code\python-microscopy\PYME\contrib\dispatch\dispatcher.py", line 174, in send
    for receiver in self._live_receivers(sender)
  File "c:\users\bergamot\code\python-microscopy\PYME\contrib\dispatch\dispatcher.py", line 174, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\SpoolController.py", line 506, in SpoolStopped
    self.onSpoolStop.send(self)
  File "c:\users\bergamot\code\python-microscopy\PYME\contrib\dispatch\dispatcher.py", line 174, in send
    for receiver in self._live_receivers(sender)
  File "c:\users\bergamot\code\python-microscopy\PYME\contrib\dispatch\dispatcher.py", line 174, in <listcomp>
    for receiver in self._live_receivers(sender)
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\htsms\rule_ui.py", line 103, in on_spool_stop
    self.on_spool_event('spool stop')
  File "c:\users\bergamot\code\python-microscopy\PYME\Acquire\htsms\rule_ui.py", line 120, in on_spool_event
    logger.info('protocol name : %s' % protocol_name)
Message: 'protocol name : htsms-staggered'
Arguments: ()
```

David-Baddeley:
Hmm, that will be because both the main PYMEAcquire process and the drift correction process end up using the same log file. Will likely effect #896 as well - should turn probably turn logging to file off by default (or find a way of subscripting the log name with the PID).

Issue #846, opened 2021-02-11
[BUG] writing table from Measure2D generates broken csv and hdf


csoeller:
**Describe the bug**
Generated table of measurements with Measure2D module. Saving using the table viewer seems to result in csv and hdf that are unreadable by all readers that I tried.

If these are indeed bugs (and not errors in myusage) then PYME is doing a poor job in preventing and detecting regressions. I know it's hard but this dents confidence. If it is all my fault I'd be happy to learn what the issues area and hang my head in shame.

**To Reproduce**
1. Run yaml recipe (see below)
2. Open generated table and save either in cvs or hdf format
4. Despair trying to read generated files

Question: is there now any other way to save these measurements?

**Expected behavior**
Before I could just open these with excel, pandas csv and hdf readers.

**Screenshots**

Table viewer:

![Screenshot 2021-02-11 at 11 59 30](https://user-images.githubusercontent.com/3750860/107636068-02891600-6c64-11eb-9a83-239740dc0dd5.png)

Unreadable csv in Emacs:

![Screenshot 2021-02-11 at 11 59 48](https://user-images.githubusercontent.com/3750860/107635969-e2f1ed80-6c63-11eb-9529-2a62dd903c26.png)

Unreadable hdf in HDFViewer:

![Screenshot 2021-02-11 at 11 56 24](https://user-images.githubusercontent.com/3750860/107636005-eb4a2880-6c63-11eb-8feb-32f33cfa53b7.png)

**Platform (please complete the following information - or copy-paste from error dialog if shown):**

   -  python-microscopy=20.12.10
   -  python=2.7.17, platform=darwin
   - numpy=1.16.6, wx=3.0.2.0 osx-cocoa (classic)

**Additional context**

yaml recipe
```yaml
- processing.FlexiThreshold:
    inputName: input
    method: fractional
    outputName: thresh
    parameter: 0.8
- processing.BinaryFillHoles:
    inputName: thresh
    outputName: tfilled
- processing.Label:
    inputName: tfilled
    minRegionPixels: 35
    outputName: labels
- measurement.Measure2D:
    inputIntensity: input
    inputLabels: labels
    measureContour: false
    outputName: measurements
```

David-Baddeley:
[...]

Issue #845, opened 2021-02-11
[BUG] Measure2D breaks when using 'output' as target name


csoeller:
**Describe the bug**

I believe one could previously select 'output' as the name for the result of the Measure2D recipe module and it would overlay the contours on the image. This seems now broken. Not sure if this is a known bug.

**To Reproduce**
1. run yaml recipe
4. See error as below

 
```
Traceback
=========
Traceback (most recent call last):
  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/ui/progress.py", line 107, in func
    return fcn(*args, **kwargs)
  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/DSView/modules/recipes.py", line 158, in RunCurrentRecipe
    self.dsviewer.pipeline.OpenFile(ds = cache)
  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/LMVis/pipeline.py", line 825, in OpenFile
    self.recipe.execute()
  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/recipes/base.py", line 718, in execute
    m.execute(self.namespace)
  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/recipes/localisations.py", line 169, in execute
    mdh = fitResults.mdh
  File "/Users/csoe002/Documents/src/PYME-src/python-microscopy/PYME/IO/tabular.py", line 195, in __getattr__
    raise AttributeError("'%s' has not attribute '%s'" % (self.__class__, item))
AttributeError: '<class 'PYME.IO.tabular.CachingResultsFilter'>' has not attribute 'mdh'
```

**Expected behavior**
Should just overlay contours.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
- python-microscopy=20.12.10
- python=2.7.17, platform=darwin
- numpy=1.16.6, wx=3.0.2.0 osx-cocoa (classic)

**Additional context**

yaml recipe

```yaml
- processing.FlexiThreshold:
    inputName: input
    method: fractional
    outputName: thresh
    parameter: 0.8
- processing.BinaryFillHoles:
    inputName: thresh
    outputName: tfilled
- processing.Label:
    inputName: tfilled
    minRegionPixels: 35
    outputName: labels
- measurement.Measure2D:
    inputIntensity: input
    inputLabels: labels
    measureContour: true
    outputName: measurements
```

[...]

PR #866, opened 2021-02-17
fix pipeline wrapping of measurements + replace newlines in CSV output

csoeller:
Addresses issues #845 and #846.

Remaining bugfix tweaks for above issues.

**Proposed changes:**

- fix pipeline wrappings for measurements
- for CSV output replace newlines (and delimiters) in stringified measures as ML strings

**Checklist:**

- [x ] Tested with numpy=1.14
- [ x] Tested on python 2.7 and 3.6
- [ N/A] Tested with wx=3.x and wx=4.x [if UI code]

### DIFF
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -824,7 +824,10 @@ class Pipeline:
         
         self.recipe.execute()
         self.filterKeys = {}
-        self.selectDataSource('filtered_localizations') #NB - this rebuilds the pipeline
+        if 'filtered_localizations' in self.dataSources.keys():
+            self.selectDataSource('filtered_localizations') #NB - this rebuilds the pipeline
+        else:
+            self.selectDataSource(self.dataSources.keys()[0])
 
         # FIXME - we do this already in pipelinify, maybe we can avoid doubling up?
         self.ev_mappings, self.eventCharts = _processEvents(ds, self.events,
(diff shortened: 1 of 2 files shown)


csoeller:
## Testing

- These fixes have been tested with `Measure2D` output and they now write easily readable CSV.
- In addition, the pipeline wrapping of the measures has been fixed which previously barfed on the lack of a `filtered_localizations` data source.
- testing with `h5r` files suggests that the fix in `pipeline.py` has no obvious unintended side effects

PYME/LMVis/pipeline.py
```diff
@@ -824,7 +824,10 @@ def OpenFile(self, filename= '', ds = None, clobber_recipe=True, **kwargs):
         
         self.recipe.execute()
         self.filterKeys = {}
-        self.selectDataSource('filtered_localizations') #NB - this rebuilds the pipeline
+        if 'filtered_localizations' in self.dataSources.keys():
+            self.selectDataSource('filtered_localizations') #NB - this rebuilds the pipeline
+        else:
+            self.selectDataSource(self.dataSources.keys()[0])
```
David-Baddeley: `self.dataSources.keys()[0]` won't work on py3 (dict_keys is not subscriptable). It's also a little fragile, as `dict.keys()` is not guaranteed to have any specific order. We know we add the data source as `FitResults` (see line 793), so this is a safe interim (re. py3/ordering) alternative, although this all needs to be revisited at some point as it's a mess of hacks.

```suggestion
            # TODO - replace / remove this fallback with something better. This is currently required
            # when we use/abuse the pipeline in dh5view, but that should ideally be replaced with
            # something cleaner. This (and case above) should probably also be conditional on `clobber_recipe`
            # as if opening with an existing recipe we would likely want to keep selectedDataSource constant as well.
            self.selectDataSource('FitResults')
```

PR #885, opened 2021-02-23
add gmm / bgmm modules

barentine:
Addresses issue # .

enhancement
**Proposed changes:**
- add BIC-optimized GMM module
- add Bayesian Gaussian Mixture module

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/recipes/pointcloud.py
+++ b/PYME/recipes/pointcloud.py
@@ -411,6 +411,11 @@ class GaussianMixtureModel(ModuleBase):
     output_labeled: PYME.IO.tabular
         input source with additional column indicating predicted component
         membership of each point
+    
+    Notes
+    -----
+    Directly implements or closely wraps scikit-learn mixture.GaussianMixture
+    and mixture.BayesianGaussianMixture
     """
     input_points = Input('input')
     n = Int(1)
--- a/PYME/recipes/pointcloud.py
+++ b/PYME/recipes/pointcloud.py
@@ -408,6 +408,7 @@ class BICGaussianMixtureModel(ModuleBase):
     n_max = Int(1)
     covariance = Enum(('full', 'tied', 'diag', 'spherical'))
     output_labeled = Output('labeled_points')
+    
     def execute(self, namespace):
         from sklearn.mixture import GaussianMixture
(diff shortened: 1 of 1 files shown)


PYME/recipes/pointcloud.py
```diff
@@ -381,3 +384,98 @@ def execute(self, namespace):
...
+    """Fit a Gaussian Mixture to a pointcloud, varying the number of components
+    and using the Bayesian Information Criterion to determine the optimum
+    number to use in the final model.
+
+    Parameters
+    ----------
+    input_points: PYME.IO.tabular
+        points to fit. Currently hardcoded to use x, y, and z keys.
+    n_max: Int
+        maximum number of Gaussians in the model
```
David-Baddeley: specify that the number which yield the best fit will be used.

PYME/recipes/pointcloud.py
```diff
@@ -381,3 +384,98 @@ def execute(self, namespace):
...
+        n_components = range(1, self.n_max + 1)
+        bic = np.zeros(len(n_components))
+        for ind in range(len(n_components)):
+            gmm = GaussianMixture(n_components=n_components[ind], 
+                                  covariance_type=self.covariance)
+            gmm.fit(X)
+            bic[ind] = gmm.bic(X)
+            logger.debug('%d BIC: %f' % (n_components[ind], bic[ind]))
+
+        best = n_components[np.argmin(bic)]
```
David-Baddeley: TODO - warn if the best fit is the highest tried n_components (or if excessively noisy)

PYME/recipes/pointcloud.py
```diff
@@ -381,3 +384,101 @@ def execute(self, namespace):
             pass
         
         namespace[self.outputName] = res
+
+
+@register_module('BICGaussianMixtureModel')
+class BICGaussianMixtureModel(ModuleBase):
```
David-Baddeley: worth considering a refactor to `GaussianMixtureModel` with 'BIC selection of `n_components` as an option (could be an enum, with 2 options to start with, BIC and manual, which would let us expand to other selection methods later).

Issue #695, opened 2020-12-09
[BUG] pymeimage recipe tab won't let me connect recipes in sequence 


zacsimile:
**Describe the bug**
If I add a more than one module to the `pymeimage` recipe I get a circular reference issue and can't link recipe elements.

![image](https://user-images.githubusercontent.com/1263313/101665738-5e90ad80-3a1b-11eb-8b38-4bb99a8de0a9.png)
 

Hitting `Apply Text Changes` gives me the error.

```
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipeGui.py", line 502, in OnApplyText
    self.recipes.UpdateRecipeText(self.tRecipeText.GetValue())
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipeGui.py", line 619, in UpdateRecipeText
    self.activeRecipe.update_from_yaml(s)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 870, in update_from_yaml
    return self._update_from_module_list(l)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 823, in _update_from_module_list
    mod.set(**md)
  File "/Users/zachcm/opt/miniconda3/envs/pyme/lib/python3.7/site-packages/traits/util/deprecated.py", line 31, in wrapper
    return fn(*args, **kw)
  File "/Users/zachcm/opt/miniconda3/envs/pyme/lib/python3.7/site-packages/traits/has_traits.py", line 1376, in set
    trait_change_notify=trait_change_notify, **traits
  File "/Users/zachcm/opt/miniconda3/envs/pyme/lib/python3.7/site-packages/traits/has_traits.py", line 1368, in trait_set
    setattr(self, name, value)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/traits.py", line 24, in validate
    raise TraitError('Assigning "%s" to input "%s" would result in a circular reference (value is in module outputs).' % (value, name))
traits.trait_errors.TraitError: Assigning "filtered_image" to input "inputName" would result in a circular reference (value is in module outputs).
```

I also get this error on adding the `Threshold` recipe tile initially. It seems to think I'm creating a circular reference to `filtered_image` even when I properly change input/output names.

**Expected behavior**
I should be able to add recipes and connect them in sequence.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
[...]

Issue #692, opened 2020-12-08
[ENHANCEMENT] Make modules searchable in bakeshop `Add Module`


zacsimile:
**Is your feature request related to a problem? Please describe.**
It's often difficult to find the module I'm looking for when I open a pipeline and press `Add Module`. There's an increasingly long list of functions (the SKF functions take up a bunch of space), and they're not sorted particularly well (e.g. what is in `localisations` vs. `measurement` is a tossup). I tend to remember the name of the module I want rather than where it's located in the list, and it takes a minute to find the module I want.

**Describe the solution you'd like**
I'd like a search bar where it says `Module name` in the image below that live-filters the list based on what I type.

![image](https://user-images.githubusercontent.com/1263313/101519819-a0552180-3951-11eb-9d2a-785ebfcc37c4.png)

**Are you prepared to work on it?**
Yes

David-Baddeley:
Great idea (you can already do a bit of filtering by starting to type in the tree view, but it's pretty flaky and might not work on all platforms).

More generally we should probably do a bit of sorting and better document what goes where. The general rule is image data -> measurement, point data > localisations.

zacsimile:
Let's add to this "reorganize the recipes". Why is `FilterTable` not under `filters`? Why do we need `tablefilters`?

David-Baddeley:
to `filters` and `tablefilters` it's image data vs tabular data - we could definitely change the names to, e.g. `image_filters` and `table_filters` or similar.

I hadn't realised how much of a problem the lack of filtering was on wx4 until recently - it works quite nicely on wx3 without changes, but should be good on wx4 with #886

PR #886, opened 2021-02-24
Avoid spurious circular reference detection, output name collisions

David-Baddeley:
Addresses issue #695.

Both - also includes UI controls to filter module list as finding a specific module is painful on wx4


David-Baddeley:
image

David-Baddeley:
so should also address #692

PR #891, opened 2021-02-24
Update entry points for PYMEVis

zacsimile:
Alias `PYMEVisualize` and `PYMEVisualise` to `PYMEVis`.


David-Baddeley:
We really need to find a way of avoiding the duplication of this file between here and the conda recipes.

PR #893, opened 2021-02-25
make sure we have a csrf cookie

David-Baddeley:
Addresses issue #803.


barentine:
thanks for figuring this out @David-Baddeley !

PR #900, opened 2021-02-27
timeout poll loop in rule pushing

barentine:
Addresses issue #if something goes wrong our rule threads keep checking if a datasource is complete/rule is complete forever.

sure
**Proposed changes:**
- if e.g. an events file never makes it, eventually raise an error in the rule pushing thread and give up rather than continuing and burdening clusterIO/PYMEAcquire indefinitely
- remove the 1.5 s wait as mentioned in the TODO

**Checklist:**
- [x] tested on scope
- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/cluster/rules.py
+++ b/PYME/cluster/rules.py
@@ -309,18 +309,18 @@ class Rule(object):
         else:
             logging.error('Failed to mark rule complete with status code: %d' % r.status_code)
 
-    def _poll_loop(self):
+    def _poll_loop(self, update_timeout=15):
         logging.debug('task pusher poll loop started')
-        # wait until clusterIO caches clear to avoid replicating the results file.
-        # TODO - we shouldn't need this any more as results are being pushed to a specific server rather than using a PYME-CLUSTER:// URI
-        time.sleep(1.5)
-    
+        
+        last_update = time.time()
         while (self.doPoll == True):
             try:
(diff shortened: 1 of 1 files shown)


David-Baddeley:
If we do have a default timeout, I'd want it to be rather more conservative - at least several minutes. That said, I'm not convinced a timeout is desirable, especially in the base Rule class. 

I'd want to be convinced that a) this solves a pressing need and b) this is the most sensible way to solve that. 

Examples where you'd run into issues with a short timeout are a little constructed at present, but could include, e.g. pushing 3D frames (rather than 2D), and having a single rule process e.g. the results of multiple analyses. It does feel, however, that we could be introducing an additional constraint that has the potential to bite us later. If we were to e.g. put tile reconstruction into rules, I could potentially see us hitting this constraint.

David-Baddeley:
For the specific PYMEAcquire problem, I suspect it would be better to subclass `LocalizationRule` with a version which looked to the spooler rather than the cluster to determine if a series was complete.

barentine:
on the one hand having a timeout as argument means you _could_ change it, but I think you're right that subclassing LocalizationRule would be smartest and additionally be more performant to `lambda : self.spooler.spool_complete` than clusterIO grep for the events file.

barentine:
going to alternative fix with the spoollocal rule

PR #902, opened 2021-02-27
allow stack settings to be per-series

barentine:
Addresses issue #osmalarity changes cell height

enhancement
**Proposed changes:**
- grab stack settings for the protocol fresh each time so we can spec it for a single series without changing the global stacksettings

**Checklist:**

- [x] Tested on sim PYMEAcquire

### DIFF
--- a/PYME/Acquire/SpoolController.py
+++ b/PYME/Acquire/SpoolController.py
@@ -414,7 +414,8 @@ class SpoolController(object):
         if stack:
             protocol = protocol_z
             protocol.dwellTime = z_dwell
-            #print(protocol)
+            stack_settings = stack if hasattr(stack, 'keys') else None
+            protocol.set_stack_positions(stack_settings)
         else:
             protocol = protocol
 
--- a/PYME/Acquire/protocol.py
+++ b/PYME/Acquire/protocol.py
@@ -195,10 +195,11 @@ class ZStackTaskListProtocol(TaskListProtocol):
     
     def set_stack_positions(self, stack_mdh=None):
         stack_mdh = dict() if stack_mdh is None else stack_mdh
-        step_size = stack_mdh.get('StackSettings.StepSize', scope.stackSettings.GetStepSize())
-        self.zPoss = np.arange(stack_mdh.get('StackSettings.StartPos', scope.stackSettings.GetStartPos()),
-                               stack_mdh.get('StackSettings.EndPos', scope.stackSettings.GetEndPos()) + .95 * step_size,
-                               step_size * scope.stackSettings.GetDirection())
+        start = stack_mdh.get('StackSettings.StartPos', scope.stackSettings.GetStartPos())
+        stop = stack_mdh.get('StackSettings.EndPos', scope.stackSettings.GetEndPos())
+        direction = 1 if stop > start else -1
+        step = direction * stack_mdh.get('StackSettings.StepSize', scope.stackSettings.GetStepSize())
+        self.zPoss = np.arange(start, stop + .95 * step, step)
 
         if self.slice_order != 'saw':
             if self.slice_order == 'random':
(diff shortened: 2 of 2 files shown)


barentine:
Note there is a small bug in current protocol.py in setting z position since the step size add on doesn't include the direction, so if you are doing a top-down stack you won't get the last slice 
```
In [4]: np.arange(0, 5 + 0.95, 1)
Out[4]: array([0., 1., 2., 3., 4., 5.])

In [5]: np.arange(5, 0 - 0.95, -1)
Out[5]: array([5., 4., 3., 2., 1., 0.])

In [6]: np.arange(5, 0 + 0.95, -1)
Out[6]: array([5., 4., 3., 2., 1.])
```

now also addressed in this pr

PYME/Acquire/SpoolController.py
```diff
@@ -414,7 +414,8 @@ def StartSpooling(self, fn=None, stack=None, compLevel=None,
         if stack:
             protocol = protocol_z
             protocol.dwellTime = z_dwell
-            #print(protocol)
+            stack_settings = stack if hasattr(stack, 'keys') else None
```
David-Baddeley, line 417: Let me rest on it for a bit, but not 100% comfortable with the re-purposing of the boolean  `stack` parameter. Can see arguments both ways ...

PYME/Acquire/protocol.py
```diff
@@ -192,11 +192,14 @@ def __init__(self, taskList, startFrame, dwellTime, metadataEntries=[], prefligh
             self.slice_order = slice_order
 
         self.require_camera_restart = require_camera_restart
-
-    def Init(self, spooler):
-        self.zPoss = np.arange(scope.stackSettings.GetStartPos(),
-                               scope.stackSettings.GetEndPos() + .95 * scope.stackSettings.GetStepSize(),
-                               scope.stackSettings.GetStepSize() * scope.stackSettings.GetDirection())
+    
+    def set_stack_positions(self, stack_mdh=None):
```
David-Baddeley, line 196: a) see PR #799 for re-factored `StackSettings` object. Any function to set one of these should either take a `StackSettings` object, or a dict of arguments (as provided by the http endpoints), **not** a metadata handler.  

A dedicated object/handler is needed as some of the properties are computed rather than directly read, depending on stack mode (hard ends, or middle & number).  It makes sense that the logic for computing these is in one place and not duplicated. Notably, this code will probably not do what you want as the default mode is middle & number. Protocol then uses it's own `StackSettings` object/the one that it gets passed rather than the scope one, leaving the rest of the code the same.

a) having this as a separate function (rather than in `Init()`) feels a bit off, but could grow on me.

barentine:
@David-Baddeley any status updates here?

David-Baddeley:
See latest iteration of #799. Needs a bit of testing / refinement, and maybe another iteration to remove some of the gross aspects. Should however give nice easily called endpoints for both action queueing and webui.

Major un/semi-resolved design choice as follows:

*Should the z dwell time be part of the stack settings, or the spooler settings?* Current version of the #799 opts for the worst of both worlds and lets it be either with significant grossness as to who takes priority for default values.

Arguments **for the dwell time to be part of the stack settings**:
- seems like a logical place for it
- tidy, easy to grasp as an endpoint for an end user

Arguments **against** (effectively the status-quo where the z-dwell settings are a property of localisation series rather than z-stepping):
- you are unlikely to want to share it between different modalities (i.e. the chance of wanting the same z-dwell for a localisation series and a widefield stack of the same structure is virtually nil, whereas sharing other stack parameters is quite likely/common/desirable if you want to do any form of correlative or multi-channel imaging.)
- putting dwell time in the stack settings means that you can't easily share stack settings between modalities (or if you do you end up with garbage spaghetti logic like I have in #799.

I think there are are some pragmatic short-term things which would let us get #799 in though (even if this is a bit of a brain fart)

David-Baddeley:
Have a pragmatic interim solution to the above.

David-Baddeley:
should be redundant to changes in #799

PR #908, opened 2021-03-01
Localization rule using spooler rather than clusterIO for series information

barentine:
Addresses issue #900.

enhancement
**Proposed changes:**
- subclass `LocalisationRule` with a spooler connected version for cleaner live pushing from PYMEAcquire

**Checklist:**

- [ ] Tested

### DIFF
--- a/PYME/cluster/rules.py
+++ b/PYME/cluster/rules.py
@@ -566,11 +566,14 @@ class LocalisationRule(Rule):
         #time.sleep(1.5) #moved inside polling thread so launches will run quicker
 
         self._next_release_start = self.start_at
-        numTotalFrames = self.ds.getNumSlices()
-        self.frames_outstanding=numTotalFrames - self._next_release_start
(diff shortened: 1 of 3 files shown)


PYME/cluster/rules.py
```diff
@@ -616,8 +622,66 @@ def get_new_tasks(self):
             
             return release_start, release_end
         
+
+class SpoolLocalLocalizationRule(LocalisationRule):
+    def __init__(self, spooler, seriesName, analysisMetadata, resultsFilename=None, startAt=0, serverfilter=clusterIO.local_serverfilter, **kwargs):
+        from PYME.IO import MetaDataHandler
```
David-Baddeley, line 629:
```suggestion
        # TODO - reduce duplication of `LocalisationRule.__init__()` and `LocalisationRule._setup()`
        from PYME.IO import MetaDataHandler
```

PR #910, opened 2021-03-02
Add removal of previous versions to PYME installation instructions

zacsimile:
We now recommend removing previous installations of PYME before installing a new version.

### DIFF
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -163,6 +163,8 @@ msvc for python.
 Removing a PYME install
 =======================
 
+To upgrade PYME instead of starting fresh, enter ``conda update python-microscopy`` in the Anaconda prompt. Otherwise, read on.
+
 To remove an executable installer on Windows, go to **Start Menu > Settings > Apps**, find `python-microscopy` under
 **Apps & Features**, select it and press *Uninstall*. 
 
--- a/docs/Installation/Installation.rst
+++ b/docs/Installation/Installation.rst
@@ -3,7 +3,9 @@
 Installation
 ************
 
-The best to install PYME will depend on your background and whether you are already using python on your computer.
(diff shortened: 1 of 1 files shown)


docs/Installation/Installation.rst
```diff
@@ -158,4 +160,16 @@ msvc for python.
...
 
+Removing a PYME install
+=======================
+
+To upgrade PYME instead of starting fresh, enter ``conda update python-microscopy`` in the Anaconda prompt. Otherwise, read on.
+
+To remove an executable installer on Windows, go to **Start Menu > Settings > Apps**, find `python-microscopy` under
+**Apps & Features**, select it and press *Uninstall*. 
 
+For executable installers on Mac, and conda installations on Windows, Mac and Linux, removing the conda envrionment
```
David-Baddeley: - executable installers on mac don't install into an evironment
- I think our instructions recommend using `PYME` as the env name

zacsimile: * They install into an environment on mine, just defined by path name
* Current instructions use `pyme` http://python-microscopy.org/doc/Installation/Installation.html

zacsimile: * I get rid of my packaged install via `conda remove -p /Users/zachcm/python-microscopy -all`

zacsimile: Do you recommend switching to `rm -rf /Users/zachcm/python-microscopy`? That's the other way I can think of doing it.

David-Baddeley: `conda remove -p /Users/zachcm/python-microscopy -all` will work if you are running conda from another miniconda/anaconda installation, but not if it's the only conda you have installed.

zacsimile: Yeah I guess that's the only way I've tried it. So switch to `rm -rf`?

PR #925, opened 2021-03-04
(temporarily) increase directory cache time in clusterIO when listSingleDir takes > 1 s

barentine:
Addresses issue #clusterIO performance.

**Proposed changes:**
things are fine until they get slammed, when things get slammed it takes longer than 1 s to listsingle, and that's when it'd be really nice to ease off a bit. We don't even necessarily take a performance hit if we're backed up on the dataservers since chances are a higher fraction of directories are stagnant/we're done writing to them for the rest of forever, so an older cache should get us far.

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
closing in favor of simply defaulting to longer timeouts on the dir cache

PR #926, opened 2021-03-05
More dataserver / clusterIO performance tweaks

David-Baddeley:
- use expired directory cache entries is it's unlikely we'd get fresh ones in a reasonable time
- tweaks to `PZFDataSource` to make it behave better for fully spooled series.


barentine:
is this set then David?

Issue #867, opened 2021-02-17
[PROPOSAL] Distributed pyramid support


David-Baddeley:
**Is your feature request related to a problem? Please describe.**
Pyramid creation (and subsequent operations on pyramids) are time consuming, especially for large tiled areas. It should be possible to store and process these in parallel ... There are currently a couple of barriers, the two most critical being: 
 
- pyramid creation requires modifying existing tiles which is not really compatible with the write-once nature of the cluster file system.
- data from the camera does not align 1:1 with tiles (due to both chosen tile sizes, and the need for overlap). This means that a) a single frame will touch multiple tiles and b) tiles will be touched by multiple different frames (due to overlap). This makes some operations intrinsically not thread / process safe (i.e. we need locking and/or some additional smarts).

Once created, and you are just reading tiles, everything should work fine with the current clusterIO.

A few subsidiary issues (easier):

- ensuring data locality for creation of higher levels in tile pyramid
- if division of tile into frames happens on the server, will need to touch data on multiple different nodes.

## pyramid creation I/O [server side]

There are a couple of solutions here:
### Add a custom endpoint to HTTPDataserver
as is done for the aggregation endpoints

**pros:**

- reasonably simple
- precedent in, e.g. __aggregate_h5
- keeps everything in one server
- easy to, e.g., update directory caches

**cons:**

- tiles are quite a lot larger (more overhead) than what is currently handled by aggregate endpoints
- significant processing is required when adding a frame (load impacted tiles, decompress, add new frame data and weights, compress, save). I'm more than a little concerned about putting significant processing in the dataservers for performance reasons.
- thread-safety - data server is multi-threaded (each request handled in a separate thread), tiling (by it's nature) is not really thread safe, as things might need to deal with multiple overlapping tiles. 

### Have a separate server specifically for tile pyramids

**pros:**

- keeps processing out of the data server
- can potentially be single-threaded to avoid concurrency issues

**cons:**

- another process to run and another protocol
- very hard to keep directory caches etc ... up to date
- processing still happens on the same CPU (you avoid the GIL, but still competing for memory and CPU resource)

### Hybrid
After writing the above I'm leaning towards a hybrid solution, outlined as follows:

- custom end point in dataserver
- endpoint drops incoming data onto a `queue.Queue`
- single thread pulls data from queue and does processing
    - could start out with a python implementation
[...]

PR #929, opened 2021-03-05
Distributed pyramid support

adrianhollow:
Addresses issue # .

enhancement

**Proposed changes:**

Add support for distributed image pyramids on a PYME cluster, specifically:
- `DistributedImagePyramid` and `PartialPyramid` subclasses of `ImagePyramid` which overwrite specific methods for updating, building and accessing tiles and layers. A `DistributedImagePyramid` represents the microscope side where frames are received as input and a cluster server has a corresponding `PartialPyramid` which only contains the chunks it is responsible for. During creation, `DistributedImagePyramid` objects send tile data to the cluster servers via HTTP and the `PartialPyramid`s on each server uses that data to update its tiles and layers as far as the chunks allow. Then, the top layer in the `PartialPyramid`s is aggregated in `DistributedImagePyramid` and used to build the final levels.
- Endpoints in `HTTPDataServer` which correspond to operations in `PartialPyramid` and are used by `DistributedImagePyramid`.
- Bool trait in Supertile recipe which indicates whether to use a regular pyramid or a distributed one.
- A `load_pyramid` function for loading pyramids from directories which determines from `metadata.json` if a pyramid is distributed or not.

**Checklist:**

- [ ] Tested with numpy=1.14

- [x] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)? ([s. issue #867](https://github.com/python-microscopy/python-microscopy/issues/867))


David-Baddeley:
Haven't had time to look at this in detail yet, (hopefully later tomorrow). A quick skim has turned up the following - I was somewhat off with my refactoring suggestions.  What I really meant was something like this (where you chop the chunk out of the frame in the initial function before passing it to `update_base_tile()`.

```python
def update_base_tiles_from_frame(self, x, y, frame, weights):

    ...

    for tile_x in tile_xs:
        for tile_y in tile_ys:
            # find co-ordinates of tile (or partial tile chunk) within image frame
            xs = max(tile_x * self.tile_size - x, 0)
            xe = min((tile_x + 1) * self.tile_size - x, frameSizeX)

            ys = max((tile_y * self.tile_size) - y, 0)
            ye = min(((tile_y + 1) * self.tile_size) - y, frameSizeY)
            
            # find co-ordinates of chunk within tile
            xst = max(x - tile_x * self.tile_size, 0)
            xet = min(xst + (xe - xs), self.tile_size)
            
            yst = max(y - tile_y * self.tile_size, 0)
            yet = min(yst + (ye - ys), self.tile_size)
            
    self.update_base_tile(0, x, y, tile_x, tile_y, frame[xs:xe, ys:ye], weights[xs:xe, ys:ye], xst, xet, yst, yet)

    self.pyramid_valid = False

def update_base_tile(self, layer, x, y, tile_x, tile_y, chunk, weights, xst, xet, yst, yet):
    # TODO - can probably infer xet, yet as, e.g. xst + chunk.shape[0]
    # TODO - take xs, ys, frame_shape and calculate/lookup weights here rather than taking as argument (to reduce data transfer)
    acc_ = self._acc.get_tile(layer, tile_x, tile_y)
    occ_ = self._occ.get_tile(layer, tile_x, tile_y)

    if (acc_ is None) or (occ_ is None):
        acc_ = np.zeros([self.tile_size, self.tile_size])
        occ_ = np.zeros([self.tile_size, self.tile_size])

    acc_[xst:xet, yst:yet] += chunk
    occ_[xst:xet, yst:yet] += weights
    self._acc.save_tile(layer, tile_x, tile_y, acc_)
    self._occ.save_tile(layer, tile_x, tile_y, occ_)

    self._clean_tiles(tile_x, tile_y)
``` 

As it stands, you'll be sending much more data than is needed.

David-Baddeley:
This could explain some, but not all of your performance differences (depending on your raw frame size, you'll end up sending each frame multiple times - at least 4, but potentially much more).

David-Baddeley:
json serialising the data is also going to be super slow - could easily add up to 2 orders of magnitude. I had thought of transmitting the chunks as PZF, in the message body (as for the standard put file), with the other parameters (x, y, etc ...) encoded in the URL.

`json.dumps(data.to_list())` will be particularly bad for image data (converts a 2 byte uint16 into anywhere from 3 to 7 bytes, with 5-6 being typical on our data). It will be even worse on the weights with the standard float printing precision of around 8sf. What's really killing you however, is likely to be iterating over python lists of pixels.

PZF will typically turn your 2 bytes per pixel into ~0.4 bytes per pixel on average (using quantisation and Huffman coding).

adrianhollow:
Thanks @David-Baddeley , using PZF instead of JSON and building the weights on the server side crunches down the overhead by an order of magnitude.
However, on an example with ca. 10k frames, the `distributed_pyramid` still takes 20x as much time as `tile_pyramid`. `mProfile` shows me that `DistributedPyramid` spends most of this time with PUT requests [here](https://github.com/adrianhollow/python-microscopy/blob/4090489ec5a8bbe9c4cb480356e0b31c0291ed5d/PYME/Analysis/distributed_pyramid.py#L646) and [here](https://github.com/adrianhollow/python-microscopy/blob/4090489ec5a8bbe9c4cb480356e0b31c0291ed5d/PYME/Analysis/distributed_pyramid.py#L493). Since I'm testing with `PYMEClusterOfOne`, I don't think it's much of a network issue and more of a data transfer bottleneck.
From a pure runtime perspective, I suspect it might be better to parallelize pyramid building locally via multi-threading/-processing to avoid transfering data over a network.

David-Baddeley:
Thoughts:
- for a small pyramid building locally is definitely preferred (even in a single-threaded implementation). The arguments for distributing are more what to do with it afterwards, and scalability (building locally runs into issues once you run out of local SSD storage).
- round trip times / latency are probably (at least part of) the issue with the HTTP PUTs (`requests.put` waits until it has received a reply from the server before program flow continues). The standard series spooler addresses this in two ways: Spooling from multiple threads, and batching and streaming the puts (i.e. putting the next frame before it's received the reply from the previous one). There is a bunch of low-level socket code in `clusterIO.put_files()` which handles this. These latency problems are likely to be even more of an issue for tiling than for normal spooling as you are subdividing frames and making more PUTs per frame.
- beware profiler overhead - `mProfile` is not a light-weight profiler, so it might be worth checking that the results also hold up when you use `cProfile (the default python profiler).

David-Baddeley:
With stuff like this, you might find that the thread-aware `fProfile` (now with some docs in #949) is more useful than `mProfile`. 

It's not in there yet, but, in theory you should be able to combine profile files from multiple processes and see how the call chain progresses between processes.

adrianhollow:
Thanks @David-Baddeley , I have repurposed the code of `clusterIO.put_files` in `Distributed_Pyramid.multi_put` which shaves off about 45% of the previous runtime. There is minimal new overhead (< 0.1%) because `Distributed_Pyramid.update_base_tiles_from_frame` now identifies all chunks at once and assigns them to the servers so that `multi_put` can be executed on all the chunks a server is responsible for.
The runtime estimate is qualitatively the same with `cProfile` and when tracking time manually and indicates a bottleneck [here](https://github.com/adrianhollow/python-microscopy/blob/5e433b60414069938dffe948109e0378f591cba3/PYME/Analysis/distributed_pyramid.py#L517), meaning that `Distributed_Pyramid` is really spending most the time waiting for a response from `HTTPDataServer`.
There might be a slowdown due to 1. network transfer, 2. threading on the server side (multiple requests per chunk, one request per tile) or 3. synchronization of `PartialPyramid.update_queue` where received tile data is stored until the tile is updated. However, when I use `cProfile` in `HTTPDataserver`, it won't track the exact method calls I expect like `_update_base_tiles`, but rather some library methods. I assume that it's the framework which delegates incoming requests to the corresponding endpoints.
```
   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000  269.062  269.062 /home/tphan/miniconda3/envs/mypyme/lib/python3.7/http/server.py:385(handle_one_request)
        8    0.000    0.000  269.062   33.633 {method 'readline' of '_io.BufferedReader' objects}
        1    0.000    0.000  269.062  269.062 /home/tphan/miniconda3/envs/mypyme/lib/python3.7/socket.py:575(readinto)
        1  269.062  269.062  269.062  269.062 {method 'recv_into' of '_socket.socket' objects}
```

David-Baddeley:
Sorry it's taken me so long to get to this. A few thoughts:

- there are way too many additions to the data server (this should be able to be handled by ideally one or potentially 2 new endpoints. All the get endpoints are definitely not needed as these should comfortably be handled as clusterIO gets of existing files, the only real place that should need changing on the server is to allow atomic additions to the base tile contents.
- This is partially because I don't want the added complexity, but also because by re-implementing the gets you eliminate lots of caching optimisations and end up multiplying the IO load considerably.

Will try and do some refactoring to clean the dataserver side of things up a bit ...

David-Baddeley:
@adrianhollow - I think it's late enough in the piece that it's probably on me to to the final bit to get this working, but will try and keep some feedback here / in the commit history in case you are interested. I've found another couple of potential performance bottlenecks which could well explain why the performance is not as expected.

- you do an awful lot of processing in your overridden `update_base_tiles_from_frame()` (including things like mask generation and labelling etc ... which are unlikely to be fast
- IO still happens in the one thread. What makes the standard spooling code fast is that there is a separate thread talking to each server - this lets us write to all the servers simultaneously maxing out the bandwidth to each. It also goes a long way towards masking the round-trip latency.
- you spool all the tiles from a single frame (that are going to a single server) in one group, but that will still be a fairly small chunk compared to the 50 frames at once that get sent by HTTPSpooler.
- as it stands, you perform the weighting on the client side, but avoid transmitting the weights by reconstructing them server side. If the weighting was performed server side (see commits I'm about to push), we would shift some of the computation there. We would also be able to avoid using a floating point data type for the tile data, and could use our fancy compression.

adrianhollow:
Thank you @David-Baddeley for taking the time for this! I think those bottlenecks make sense and can definitely be bypassed. I'm sorry that I haven't been as active lately, but I have to push on other things at the moment.

David-Baddeley:
#### Current status:

- Runs locally spooling tiles to cluster of one. These are viewable in the tileviewer.
- Not tested for performance compared to status quo
- Not tested for opening the generated series on a real cluster (without local short-cutting)
- Contains an important BigSur bugfix (required for running any of the cluster stuff on BigSur)
- Contains a bugfix for a previously unknown bug in PYMEDataServer (all subsequent puts on a session fail after a 405 response) that might be relevant for general stability.
- Still some scope for further performance improvements (sending data as uint16 rather than float and compressing)
- Still scope for some refactoring. The new `TileSpooler` class or a slight variation thereof could be used to good effect in the HTTPSpooler as well (should offer slightly better performance by having read and write operations in different threads. For that reason, it probably belongs in clusterIO or somewhere similar, rather than in the tiling code.
- The tiling stuff is really IO not analysis, we might want to look at shifting it

The above points notwithstanding, I'd be in favour of merging soon and clobbering the rest as we get around to it, particularly to get the BigSur and DataServer fixes in. It's fairly low risk (I'd say much lower risk than the xyztc stuff), but does touch the dataservers, the current tiling, and (to a minimal extent) the spooling. @barentine - tagging you as you're most likely to be hit by any oddities.

David-Baddeley:
On a simulated 15x15 tile series with one server node, times are within spitting distance of the local tiler (2.44s vs 2.14s respectively).

PR #936, opened 2021-03-07
Supertile optional directory

adrianhollow:
Addresses issue # .

enhancement
**Proposed changes:**
At the moment, the `Supertile` module builds an image pyramid in a temporary directory and passes it to a `SupertileDataSource` which usually uses only one pyramid level. In some cases, it might make sense to save the whole pyramid in a permanent directory for later use.

- Add a `FileOrURI` to the module which specifies a permanent directory location. By default, it is ignored and a temporary directory is used.

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
This gets tricky as recipe modules are not supposed to have side effects (i.e. the only way of saving stuff is via the specific `output` modules). The reason for this is that as soon as you have side effects, it's hard to ensure that running the same recipe twice will give the same results. Side effects can also make execution order important, and allow you to introduce implicit dependencies between steps which will not be captured in the recipes dependency graph which we use to set execution order (and potentially in the future parallelisation). If you are familiar with functional programming, it's useful to think of recipes as being declarative, functional, programs.  

The `Supertile` module is going to disappear and I should probably make the **DO NOT USE** warning on it a bit more explicit. In general the correct way to use Supertiles within recipes is as follows:

- generate a tile pyramid on disk
- run the recipe against a  SUPERTILE datasource pointing to that pyramid (i.e. use a filename of the form `SUPERTILE:/path/to/pyramid?level=x&stride=y&overlap=z)

The current `Supertile` module is a quick and dirty hack to get something up and running for analysing pre-spooled (but not created) pyramids on the cluster, and adding additional functionality to it makes it harder to properly replace. At the time, I think the major obstacle in the way of a proper solution was the lack of recipe chaining, which is now fixed. Hopefully spooled but not created pyramids go away in the near future, but I guess it's always going to be useful to be able to create a pyramid from a spooled dataset. With that in mind, a suitable replacement would look like this.

- A Pyramid recipe **output** module (takes a series, saves a pyramid to disk, does not make that pyramid available within that recipe). Effectively file format conversion via recipe.
- Chain a recipe which uses a SUPERTILE datasource against this pyramid.

David-Baddeley:
i.e. something like 

```python
@register_module('TilePyramid')
class TilePyramid(OutputModule):
    inputName = Input('input')
    base_tile_size = Int(256)
    filePattern = '{output_dir}/{file_stub}_pyramid/'

    def save(self, namespace, context={}):
        """
        Save series as a tile pyramid

        Parameters
        ----------
        namespace : dict
            The recipe namespace
        context : dict
            Information about the source file to allow pattern substitution to generate the output name. At least
            'basedir' (which is the fully resolved directory name in which the input file resides) and
            'file_stub' (which is the filename without any extension) should be resolved.

        Returns
        -------

        """
        from PYME.Analysis import tile_pyramid
        from .output import _ensure_output_directory
        out_dirname = self._schemafy_filename(self.filePattern.format(**context))
        _ensure_output_directory(out_dirname)

        stack = namespace[self.inputName]
        x, y = tile_pyramid.get_position_from_events(stack.events, stack.mdh)

        tile_pyramid.tile_pyramid(out_dirname, stack.data, x, y,stack.mdh,pyramid_tile_size=self.base_tile_size)
```

David-Baddeley:
The other thing we'd probably need to do is make the pyramid use unified IO for loading tiles etc ...

David-Baddeley:
will PR

Issue #931, opened 2021-03-06
[BUG] event copying to h5r fails with spoollocallocalizationrule


barentine:
**Describe the bug**
The events do not seem to be automatically copied into the h5r files at the moment. 
```
DEBUG:root:pytables file open: /data/PYMEData/Bergamot/2021_3_4/C1/analysis/4_3_series_BO.h5r
Traceback (most recent call last):
  File "/home/ubuntu/code/python-microscopy/PYME/IO/h5rFile.py", line 148, in _appendToTable
    table = getattr(self._h5file.root, tablename)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/tables/group.py", line 818, in __getattr__
    return self._f_get_child(name)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/tables/group.py", line 698, in _f_get_child
    self._g_check_has_child(childname)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/tables/group.py", line 395, in _g_check_has_child
    % (self._v_pathname, name))
tables.exceptions.NoSuchNodeError: group ``/`` does not have a child named ``Events``

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/ubuntu/code/python-microscopy/PYME/IO/h5rFile.py", line 246, in _pollQueues
    self._appendToTable(tablename, np.hstack(recs))
  File "/home/ubuntu/code/python-microscopy/PYME/IO/h5rFile.py", line 158, in _appendToTable
    expectedrows=500000)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/tables/file.py", line 1055, in create_table
    chunkshape=chunkshape, byteorder=byteorder)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/tables/table.py", line 808, in __init__
    descr_from_dtype(nparray.dtype)
  File "/home/ubuntu/miniconda3/lib/python3.6/site-packages/tables/description.py", line 786, in descr_from_dtype
    for name in dtype_.names:
TypeError: 'NoneType' object is not iterable
```

I originally thought this was due to potentially having a low pytables version, and we need at least 3.2.x, but I actually have 3.4.2 and 3.6.1 on my cluster

**To Reproduce**
1. chain a localization rule push in pymeacquire
2. everything seems fine
3. wonder where your events are in the h5rs, see that they're in fact there in the pcs
[...]

PR #940, opened 2021-03-08
spoollocallocalizationrule: file events in a format we know what to do with

barentine:
Addresses issue #931

bugfix
**Proposed changes:**
convert events json to structured array so it gets handled correctly in `clusterResults.fileResults`

**Checklist:**

- [x] Tested on scope


David-Baddeley:
serialising to json and using pandas to load that will not be great performance wise - adding a `to_recarray` or similar to the event handler using logic like that in `h5file.put_file()` - i.e. 

```python
events_array = np.empty(len(events), dtype=EVENTS_DTYPE)
            
for j, ev in enumerate(events):
    events_array['EventName'][j], events_array['EventDescr'][j], events_array['Time'][j] = ev
```

is likely to be much faster and cleaner.

PR #922, opened 2021-03-04
Some dataserver performance tweaks

David-Baddeley:
- make h5rfile vstack multiple appends before passing them to pytables
- avoid running makedirs if file already exists
- profiling fixes

suspect it'll be subtle, but might help when dataservers are being hammered with results.

PR #948, opened 2021-03-11
Start time fix

David-Baddeley:
Addresses issue with event timing when there are lots of initial (-1) protocol tasks (should also generally improve robustness of timing).

Also includes a few event handling and display fixes.

### DIFF
--- a/PYME/Acquire/Spooler.py
+++ b/PYME/Acquire/Spooler.py
@@ -134,11 +134,10 @@ class Spooler:
         self.imNum = 0
 
         self.protocol.Init(self)
-
-        self._collect_start_metadata()
         
         # record start time when we start receiving frames.
         self.tStart = time.time()
+        self._collect_start_metadata()
         self.frameSource.connect(self.OnFrame, dispatch_uid=self._spooler_uuid)
         
         self.spoolOn = True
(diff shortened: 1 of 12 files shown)


David-Baddeley:
@barentine - you might be interested in this

PYME/Acquire/Spooler.py
```diff
@@ -133,14 +133,14 @@ def StartSpool(self):
 
         self.imNum = 0
 
-        # record start time here in case protocol init tasks generate events (prob only effects simulator).
-        self.tStart = time.time()
-
         self.protocol.Init(self)
 
         self._collect_start_metadata()
-
```
barentine, line 142: This probably wont work quite right. I think we need to set `self.tStart` before `_collect_Start_metadata` where we log it

David-Baddeley: good catch ...

David-Baddeley:
@barentine - have you actually run acquisition with this already? Or should I check on the simulator prior to merging ...

barentine:
I was using it, yes. Am now concerned that you found other things to fix in a87af8e

barentine:
though if I understand at quick glance that appears to be a) simulator specific, b) display related, and c) only an issue for spooling to h5 files. Correct?

David-Baddeley:
Yep, 3 separate small regressions, one simulator specific. An event display regression in dh5view, and a PYMEClusterOfOne regression which I think is actually separate to the changes in this PR (think it was #922).

PR #950, opened 2021-03-12
check for subdirectories instead of detection h5s in requeue_missed

barentine:
Addresses issue if a node is MIA the spooled subdirectories are better to look for than the h5's. Maybe runaway memory concerns
enhancement
**Proposed changes:**
- look for spooled subdirectories for wells rather than _detections.h5 since the subdirectories get spread across all nodes while the h5's disappear if a node goes missing
- ulimit workers to 90% of ram. I wrote/tested this briefly in a notebook, committed it, then forgot it was there. I'll include it for discussion, but I don't think it actually addresses the problem I encounter where two detection tasks getting assigned to the same node seems to knock it over. a better approach seems to me to be for RAM-heavy recipe module to check whether they can execute. Something like 
```
class MemoryAwareMRCNNDetect(MRCNNDetect):
    n_retries = Int(1)
    fractional_memory_needed = Float(0.5)
    sleep_time = Float(60)

    def execute(self, namespace):
        import psutil
        n_tries = 0
        while n_tries < n_tries:
            if (1 - psutil.virtual_memory().percent) > self.fractional_memory_needed:
                MRCNNDetect.execute(self, namespace)
                return
            else:
                time.sleep(self.sleep_time)

        raise MemoryError('We do not have adequate memory to execute')
```
except it'd be nice if the module calculated for itself how much it'll likely draw

**Checklist:**

- [ ] Tested

### DIFF
--- a/PYME/Acquire/ui/tile_panel.py
+++ b/PYME/Acquire/ui/tile_panel.py
@@ -363,20 +363,12 @@ class MultiwellProtocolQueuePanel(wx.Panel):
 
         logger.debug('requeuing missed wells')
         time.sleep(sleep)
-
-        spooldir = self.scope.spoolController.dirname
-        detections_pattern = posixpath.join(spooldir, '[A-Z][0-9]*_detections.h5')
-        imaged = clusterIO.cglob(detections_pattern)
-        imaged_wells = [im.split('/')[-1].split('_detections.h5')[0] for im in imaged]
-        logger.debug('imaged %d wells' % len(imaged_wells))
-
+        
         x_wells, y_wells, names = self._get_positions(n_x, n_y, x_spacing, y_spacing, start_pos)
         x_wells, y_wells, names = self._pop_wells(x_wells, y_wells, names, self._drop_wells)
-        to_pop = [fn.split('_')[0] for fn in imaged_wells]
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Can't you just use ulimit on PYMERuleNodeServer when you launch it?  Should apply to the server itself and all children and would avoid the somewhat messy special case logic within the nodeserver script. This feels like something which should be done on a per -cluster basis, rather than baked into PYME/

David-Baddeley:
Re memory-awareness etc ... (this is mostly thinking aloud)  holy grail would look something like the following:

- recipe modules record peak memory usage (and potentially execution time), along with input shape
- use this data to train a regression model to predict memory usage and computational time for a complete recipe.
- use this in nodeserver bidding to make sure that a node didn't bid on two memory intensive tasks (also a neat way of potentially making stuff smarter than the current bid on N tasks heuristics - i.e. changing to bid on 10s worth of tasks or similar).

For a more pragmatic/short-term solution ... I'm not sure the `time.sleep()`ing within a recipe module is going to be particularly robust. Maybe pushing all recipe tasks to one worker on each node (I think we discussed this earlier).

barentine:
I hadn't realized ulimit would pass on to child processes with separate PIDs etc., but that does sound like a good solution.

Issue #664, opened 2020-12-03
[BUG] packages behind, cannot install on macOS 11 BigSur


barentine:
**Describe the bug**
numpy, pyopengl, etc. seem to be having some issues getting bigsur compatible packages up. When someone tried to install PYME, they got the following traceback. Opening this ticket as a place to track when:
- [ ] we support macOS 11
- [ ] we have documented the incompatibility in the install docs in the meantime

```
/Applications/python-microscopy/lib/python3.6/site-packages/PYME/misc/extraCMaps.py:73: MatplotlibDeprecationWarning: 
The revcmap function was deprecated in Matplotlib 3.2 and will be removed two minor releases later. Use Colormap.reversed() instead.
cmapdat_r = cm.revcmap(ndat[cmapname])
Traceback (most recent call last):
File "/Applications/python-microscopy/lib/python3.6/site-packages/OpenGL/platform/darwin.py", line 38, in GL
mode=ctypes.RTLD_GLOBAL 
File "/Applications/python-microscopy/lib/python3.6/site-packages/OpenGL/platform/ctypesloader.py", line 45, in loadLibrary
return dllType( name, mode )
File "/Applications/python-microscopy/lib/python3.6/ctypes/__init__.py", line 348, in __init__
self._handle = _dlopen(self._name, mode)
OSError: ('dlopen(OpenGL, 10): image not found', 'OpenGL', None)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
File "/Applications/python-microscopy/bin/VisGUI", line 7, in 
from PYME.LMVis.VisGUI import main
File "/Applications/python-microscopy/lib/python3.6/site-packages/PYME/LMVis/VisGUI.py", line 72, in 
from PYME.LMVis import visCore
File "/Applications/python-microscopy/lib/python3.6/site-packages/PYME/LMVis/visCore.py", line 17, in 
from PYME.LMVis import gl_render3D as gl_render
File "/Applications/python-microscopy/lib/python3.6/site-packages/PYME/LMVis/gl_render3D.py", line 30, in 
from OpenGL import GLUT
File "/Applications/python-microscopy/lib/python3.6/site-packages/OpenGL/GLUT/__init__.py", line 2, in 
from OpenGL.raw.GLUT import *
File "/Applications/python-microscopy/lib/python3.6/site-packages/OpenGL/raw/GLUT/__init__.py", line 6, in 
from OpenGL.raw.GLUT.constants import *
[...]

PR #954, opened 2021-03-14
Add Big Sur patch to entry points

zacsimile:
Addresses issue #664.

Not 100% Sur we need this in `cluster_of_one` but I figured better safe than Surry.


David-Baddeley:
We might additionally need to add to the dataserver, ruleserver, and task worker entry points if numpy gives us issues

Issue #956, opened 2021-03-15
[BUG] download pcs series fails because temporary file is deleted in file request to cluster


adrianhollow:
**Describe the bug**
Downloading pcs series from ClusterOfOne as h5 fails as in #746, but with this error message:
```
File "/home/tphan/miniconda3/envs/mypyme/lib/python3.7/site-packages/django/core/handlers/exception.py" in inner
  34.             response = get_response(request)

File "/home/tphan/miniconda3/envs/mypyme/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  126.                 response = self.process_exception_by_middleware(e, request)

File "/home/tphan/miniconda3/envs/mypyme/lib/python3.7/site-packages/django/core/handlers/base.py" in _get_response
  124.                 response = wrapped_callback(request, *callback_args, **callback_kwargs)

File "/home/tphan/thesis/python-microscopy/PYME/cluster/clusterUI/clusterbrowser/views.py" in file
  43.             response['Content-Length'] = os.path.getsize(outf.name)

File "/home/tphan/miniconda3/envs/mypyme/lib/python3.7/genericpath.py" in getsize
  50.     return os.stat(filename).st_size

Exception Type: FileNotFoundError at /files/download/tphan/2021_3_15/15_3_series_A.pcs/
Exception Value: [Errno 2] No such file or directory: '/tmp/tmphwec6l04.h5'
```
**To Reproduce**
1. Run `PYMEClusterOfOne --clusterUI`
2. Navigate to pcs series in clusterUI
3. try to download as h5
4. see error

**Expected behavior**
 download the pcs as an h5

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: ubuntu 20.04
 - Python version: 3.7.9
 - numpy version: 1.19.2
 - wxpython version [if a UI bug]: 4.0.4

David-Baddeley:
This is super odd - the temporary file should be open as long as the `with ...` context manager is open. This looks like a bug/regression in `NamedTemporaryFile` (or `os.stat`)

David-Baddeley:
Do we also get this error on py3.6?

David-Baddeley:
What about other platforms?

David-Baddeley:
I suspect the solution to this might have to involve writing our own replacement for `NamedTemporaryFile`, using our own filename within a temporary directory as `NamedTemporaryFile` is even more broken on Windows (I though there was an issue for this but can't seem to find it).

PR #957, opened 2021-03-15
Dont delete tmpfile in filerequest

adrianhollow:
Addresses issue #956 .

bugfix
**Proposed changes:**
Avoid deleting temporary file upon close. Resolves the error in #956, but may lead to piling temporary files as requests come in.

**Checklist:**

- [ ] Tested with numpy=1.14

- [x] Are there any other side effects of the change? May lead to wasted storage space (s. above).


David-Baddeley:
We definitely want the temporary file to be deleted when we are done, otherwise we fill our disk up with junk. What we don't want is for it to be deleted before we are done. I think the issue is that we are returning the open file, so the context-manager completes (and removes the file) before we finish writing it over the http session. 

The simplest approach might not be to use a context manager (although I'd need to think about implications). 

A brute force approach would be to read the file into memory and return a `BytesIO` instance from the request handler, although that would only work for files which fit in memory.

PR #959, opened 2021-03-15
Make dumbbell test more permissive

zacsimile:
Our tests must pass


David-Baddeley:
The other way of addressing this which we had talked about was setting the random seed for the fit tests (we also get occasional failures on the other fits, this one is just the most common).

zacsimile:
I remembered, I just wasn't sure exactly what to set when I looked at the code. This is a quick fix. I'm not opposed to the long fix.

David-Baddeley:
was going to try and fix globally, but can't seem to get the tests to fail on my computer. I wonder if it depends on the precise version of pytest (which I think we have pinned in the test setup) and/or if numpy/pytest now does something sneaky with random seeds).

David-Baddeley:
I wonder if it's due to numpy rng changes ...

Issue #915, opened 2021-03-02
[ENHANCEMENT] Add Help>Documentation to all menus


zacsimile:
Add a `Help>Documentation` option to our program menus in all PYME software that links to our web docs.

PR #960, opened 2021-03-16
Add Help>Documentation to PYMEVis

zacsimile:
Addresses issue #915.

Enhancement

**Proposed changes:**
- Add link to online documentation to PYMEVis `Help` menu

There's no help menu in the other programs at the moment. Ideally we'd create a default help menu that was in all of our programs and included `Documentation` and `About`. Can I add this change to this PR, @David-Baddeley? Or would you prefer to leave help menus out of `PYMEImage` and `PYMEAcquire`?

### DIFF
--- a/PYME/LMVis/VisGUI.py
+++ b/PYME/LMVis/VisGUI.py
@@ -276,6 +276,11 @@ class VisGUIFrame(AUIFrame, visCore.VisGUICore):
 
         wx.adv.AboutBox(dlg)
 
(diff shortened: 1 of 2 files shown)


David-Baddeley:
I think having a global `Help->Documentation` everywhere would be good (potentially refactored along with Help->About into PYME.ui so that we don't just have duplicate code in a bunch of places). 

Open question:

- should Help->Documentation point to the global docs, or to the program/component specific docs?

PR #965, opened 2021-03-18
Sim enhancements

David-Baddeley:
A bit of refactoring in advance of making the simulation code more flexible.

Also:

- add an signed distance function based point generation function to locify. @zacsimile - can discuss where this should end up in the medium term, but seems to fit here for now  and didn't want to introduce a `ch_shrinkwrap` dependency just yet (suspect optimum might be to have the SDF stuff in it's own module).
- fixes a bug in sherical shell code


PYME/Analysis/points/spherical_harmonics.py
```diff
@@ -374,7 +374,7 @@ def set_fitting_points(self, x, y, z):
     def _scale_fitting_points(self):
         self.standard_deviations, self.principal_axes = coordinate_tools.find_principle_axes(self.x_c, self.y_c, self.z_c,
                                                                                              sample_fraction=self.sampling_fraction)
-        self.scaling_factors = np.max(self.standard_deviations) / (self.standard_deviations)
+        self.scaling_factors = (self.standard_deviations) / np.max(self.standard_deviations)
```
David-Baddeley: @barentine - this is the change I mentioned to the scaling. It inverts the scaling factors, but I think this is correct as we multiply the co-ordinates by them (i.e. we should blow up our z when fitting so that it has a similar range to x and y). 

Before this change, it would seem that we double-down and make the smallest axis even smaller.

David-Baddeley: argh, no I was wrong in my reading. Initial code was correct. I knew it would be good to get another view on this.

zacsimile:
Can we merge this? I would like to use it

David-Baddeley:
just need to resolve some conflicts ...

zacsimile:
Thanks!

PR #966, opened 2021-03-20
use pyueye in htsms_focus_lock

barentine:
Addresses issue #windows update seemed to have messed up my focus lock scope/camera, though rolling it back didn't fix it. I was using the from 'OMEINC', version 4.80.5.

The frameWrangler.GetFPS was returning ~60 while the camera was set to 285 and then slowy decreasing until it was < 1 Hz. Downloading the latest driver from IDS (4.94.1527.0, which now says from IDS not OMEINC) did not initialize with the ucam480 camera class in our code, so I implemented the pyueye bindings, updating a function or two which are now 'obsolete' in their API. 

That was fine, but did not actually fix things. I noticed I could get higher than 60 Hz, but if I got too fast the waiting for frame started failing and if we got behind on that things compounded. I thought maybe we weren't unlocking on-camera memory when we failed, or maybe we needed a little time.sleep after that failure in the poll loop, but neither fixed things. What _did_ was removing the time.sleep(0.005) in the poll loop, which I guess makes some sense as get down to an integration time which is on the order of a ms. I did not check whether the same modification fixes the ucam480 code for my camera (UI3240-LE_NIR).

enhancement in general, bugfix for me
**Proposed changes:**
- use pyueye for a camera class closely mirroring the ucam480 code

todo:
simplify the buffer stuff - frame accumulations in the camera code seems really weird/useless. For such a commonly used camera class (or, potentially used anyway) would maybe benefit from being slightly simplified from where the ucam480 code stands.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/ueye.py
+++ b/PYME/Acquire/Hardware/ueye.py
@@ -1,4 +1,13 @@
+'''
+Alternative bindings for IDS ueye cameras using the pyueye module.
 
+See also the uc480 module which talks directly to the ueye DLLs without an intermediate python shim. The uc480
+implementation has been more widely used and tested and might be more complete. At present, however, there might be issues using
+uc480 with the most recent versions of the ueye SDK, hence this module. As we don't want to maintain 2 drivers
+for the same camera, this module might disappear once api compatibility issues in uc480 are resolved (i.e. use
+uc480 instead of this if at all possible).
+
+'''
 from PYME.Acquire.Hardware.Camera import Camera
 from pyueye import ueye
 import ctypes
--- a/PYME/Acquire/Hardware/ueye.py
+++ b/PYME/Acquire/Hardware/ueye.py
@@ -1,0 +1,442 @@
+
+from PYME.Acquire.Hardware.Camera import Camera
+from pyueye import ueye
+import ctypes
+import threading
+import logging
+import queue
+import numpy as np
+import time
+from PYME.Acquire import eventLog as event_log
+
+
+logger = logging.getLogger(__name__)
+
+def GetError(camera_handle):
+    error = ctypes.c_int()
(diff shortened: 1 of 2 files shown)


David-Baddeley:
I'm a little wary about having 2 competing implementations of the same camera driver in the code (although can see advantages). As I see it, this a) introduces a new dependency, and b) might not work with some of the older cameras, whereas the old code might not work with some of the more recent ueye drivers. Ideally we'd have one set of code with minimal dependencies which worked with everything, but failing that we need to clearly signal (with comments etc) which code is preferred for people setting stuff up. Let me have a bit of a think.

Accumulation in the original code is to aid use in a focus lock because a lot of the cameras run by the driver are 8 bit, with significant dark/read noise, and as a result some form of averaging is often required to get good SNR (potentially more important for a transmitted lock than for a reflection based one). There could be a strong argument for refactoring it out into a shim/virtual camera that sits between the software and an actual camera adapter, but also an argument for keeping it until such a shim is available.

barentine:
Ideally it would be just one class, but given the number of cameras supported on these drivers and that we typically make a new file for each camera on other drivers, maybe it's not so crazy to just comment clearly what it is and which driver versions it supports.

You'll notice I didn't fully implement e.g. binning - there's still more that could/should be done. I do think it'd be worth simplifying the buffer stuff by dropping the accumulation code in the camera - if that's necessary for a focus lock with low signal it might be cleaner to implement it there.

csoeller:
Hi @barentine, came across the newer ueye driver via `pyueye` as I am having a few issues with the older ueye driver myself, now that we are using latest IDS software and a python 3/ win 10 install. What is the best way to install `pyueye`, is it just `pip install`. Sorry to use a merged PR for discussion, perhaps open as another discussion.

Issue #971, opened 2021-03-25
[ENHANCEMENT] figure making streamlining / docs


barentine:
**Is your feature request related to a problem? Please describe.**
SMLM dataset -> 3D multicolor image suitable for a figure should be something we do easily/make accesible.

At the moment, with projection from the GUI only working for xy and otherwise just showing a slice (see #811) I feel a little ridiculous with the process:

first running this recipe on my localization files

```
- localisations.DensityMapping:
    colours:
    - chan0
    - chan1
    inputLocalizations: fitResults
    jitterScale: 30.0
    manualXYBounds:
    - -50.0
    - 1100.0
    - 27800.0
    - 27800.0
    outputImage: g2d
    pixelSize: 15.0
    renderingModule: Gaussian
    xyBoundsMode: manual
- output.RGBImageOutput:
    colorblindFriendly: true
    filePattern: '{output_dir}/{file_stub}.png'
    inputName: g2d
    scaling: percentile
- output.ImageOutput:
    filePattern: '{output_dir}/{file_stub}_g2d.h5'
    inputName: g2d
- localisations.DensityMapping:
    colours:
    - chan0
    - chan1
    inputLocalizations: fitResults
    jitterScale: 30.0
    jitterScaleZ: 30.0
    manualXYBounds:
    - -50.0
    - 1100.0
    - 27800.0
    - 27800.0
    outputImage: g3d
    pixelSize: 15.0
    xyBoundsMode: manual
    zBoundsMode: min-max
    zSliceThickness: 15.0
- output.ImageOutput:
    filePattern: '{output_dir}/{file_stub}_g3d.h5'
    inputName: g3d
```

and then
open _g2d.h5 in PYMEImage. In the shell toggle crosshairs and scale bar off
view.crosshairs=False
view.showScaleBar=False

set display to 1:2

chan0 LUT to ‘c’, histogram to [0, 0.13]
chan1 LUT to ‘m’, histogram to [0, 0.13]

save as _2d.png

select a box for a cross section and click File > Export Cropped
open the _g3d.h5 of that same ROI 
switch to the shell tab, and enter

```
from PYME.IO.image import ImageStack
im = ImageStack(data=np.swapaxes(image.data[:, 874:1049, :, :], 1, 2), mdh=image.mdh)
```

where 874: 1049 are the ‘y’ bounds I selected, and the 1, 2 correspond to swapping y and z to get an xz image with y slices.

Now again in the shell 
```
[...]

PR #972, opened 2021-03-26
stacking

David-Baddeley:
Some z stacking changes, xyztc  acquisition base class

### DIFF
--- a/PYME/DSView/modules/projections.py
+++ b/PYME/DSView/modules/projections.py
@@ -1,0 +1,45 @@
+
+
+def project(dsviewer, axis, type, crop=False):
+    from PYME.recipes import processing
+    from PYME.DSView import ViewIm3D
+    
+    if crop:
+        from .cropping import crop_2D
+        x0, x1, y0, y1, z0, z1 = dsviewer.do.sorted_selection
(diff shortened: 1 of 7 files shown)


David-Baddeley:
also has some projection stuff in here related to #971 as was on the wrong branch when committing.

PYME/DSView/modules/projections.py
```diff
@@ -0,0 +1,45 @@
...
+    else:
+        im = dsviewer.image
+    
+    ViewIm3D(processing.Projection(kind=type, axis=axis).apply_simple(im), parent=dsviewer)
+
+
+def Plug(dsviewer):
+    # type: (PYME.DSView.dsviewer.DSViewFrame) -> object
+    
+    dsviewer.AddMenuItem('Processing>Project>Mean', 'x', lambda e : project(dsviewer, axis=0, type='Mean'))
```
zacsimile, line 21: I read through this and my first thought was "which project is this for?" Can we change the menu name to `Processing>Projections`? It's slightly more intuitive. Very minor point, but I think a helpful one for users.

David-Baddeley: Seems reasonable (I guess I try to make menu options verbs, but think this would be clearer the other way)

Issue #970, opened 2021-03-25
[BUG] Fit Info panel does not work properly?


csoeller:
**Describe the bug**
In dh5view I run a `Test this frame` command and immediately get the following error in the console, Not sure if relevant to the problem at hand though.

```
Traceback (most recent call last):

  File "/Users/csoe002/Documents/src/PYME-src/pyme-py3/PYME/contrib/wxPlotPanel.py", line 69, in _onIdle

    self._SetSize()

  File "/Users/csoe002/Documents/src/PYME-src/pyme-py3/PYME/contrib/wxPlotPanel.py", line 86, in _SetSize

    float( pixels[1] )/self.figure.get_dpi() )

  File "/Users/csoe002/opt/anaconda2/envs/pyme-py3/lib/python3.7/site-packages/matplotlib/figure.py", line 902, in set_size_inches

    raise ValueError(f'figure size must be positive finite not {size}')

ValueError: figure size must be positive finite not [0.   6.86]

```

**To Reproduce**
1. `Test this frame`
2. Click on event that was detected
3. go to `Fit Info` panel
4. ROI of event shows up for a second but then all plot panels are blank.

**Expected behavior**
Would have expected to see event and some plots in `Fit Info` panel.

**Screenshots**

<img width="1217" alt="Screenshot 2021-03-25 at 12 16 31" src="https://user-images.githubusercontent.com/3750860/112472460-0e88ed00-8d65-11eb-92d3-d140dfe49fe7.png">

<img width="1217" alt="Screenshot 2021-03-25 at 12 16 26" src="https://user-images.githubusercontent.com/3750860/112472496-1a74af00-8d65-11eb-9ef6-401f308d2fc7.png">

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: macos 10.15
 - Python version: py37
 - numpy version: ??
 - wxpython version [if a UI bug]: 4.X

**Additional context**

David-Baddeley:
Hmm, which matplotlib version? 

Best guess is that matplotlib has got fussier with what it accepts in `set_size_inches()`, causing it to bail whilst we are creating the plot panel (before the plot panel has been resized). Best guess at a fix / workaround.

- comment out `self._SetSize()` on line 41
- optionally change`self._resizeflag=True` a couple of lines below 

[...]

PR #973, opened 2021-03-27
Fitinfo fix

csoeller:
Addresses issue #970.

bugfix

**Proposed changes:**

The logic of `fitDispPanel.draw` was not ok and also some issue with index arithmetic for recent py3/numpy in plotting command. `fitDispPanel.draw` is called by the `wxPlotPanel` parent method `_onIdle` without an index argument which led with the original implementation of `draw` to a clearing of all plot panels which always persisted.

Admittedly, I am not sure how important the `_onIdle` `draw` call is.

Currently has quite a few debug logging statements which could be removed if felt necessarily.

`wxPlotPanel` not yet modified since the `set_size_inches` value errors are inconsequential.

Also added a `genFitImage` function for `LatGaussfitFR` so that we can use full Fit Info functionality now.

**Example Screenshot of working implementation**

Screenshot 2021-03-27 at 15 31 39

**Checklist:**

- [ x] Tested with numpy=1.19.2
- [x ] Tested on python 3.7
- [ x] Tested with wx=4.x [if UI code]

### DIFF
--- a/PYME/DSView/fitInfo.py
+++ b/PYME/DSView/fitInfo.py
@@ -408,6 +408,7 @@ class fitDispPanel(wxPlotPanel.PlotPanel):
 #            self.subplot1.set_xticks([0, ed.max()])
 #            self.subplot1.set_yticks([0, numpy.floor(a.max()/float(numpy.diff(ed[:2])))])
             if not i is None:
+                # only clear the panels if we have been called with a valid fitResults index
                 self.subplot1.cla()
                 self.subplot2.cla()
                 self.subplot3.cla()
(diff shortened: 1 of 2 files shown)


David-Baddeley:
We need a way of re-setting / clearing when we click away, should possibly still be clearing on `None` and/or working up an alternative way of handling things. 

Will merge for now, as it's definitely an improvement over the status quo. Should look at updated wx embedding docs to see if we can't make things generally cleaner though (in theory we shouldn't need to explicitly redraw any of the contents on sizing etc, - I'd expect matplotlib to do that for us.

Issue #988, opened 2021-04-06
[SUPPORT] is action manager and associated code in a usable state?


csoeller:
Note: Support is also available by emailing support@python-microscopy.org, choose this option if you would like the question and answer to be publicly searchable.

**What are you trying to do?**
I am trying to use the action manager / spoolcontroller code. I found, to my slight dismay, that the interface code that I used and extensively tested in December (implemented in https://github.com/csoeller/PYME-extra/blob/master/PYMEcs/Acquire/Actions/custom.py) has stopped working.

Similarly, the `Queue Actions` action manager UI currently has a button to issue a `spoolController.StartSpooling` command which gives an error since all syntax seems to have changed to the `start_spooling` etc syntax. I also have trouble to figure out how to use the `max_frames` setting to `start_spooling` which gives me

```
TypeError: start_spooling() got an unexpected keyword argument 'max_frames'
```

My questions are:

- is the action manager interface in a sane state?
- did I miss that these syntax changes etc were instigated?
- do we have docs that explain things properly?

**Where are you stuck?**
Explanations see above.

David-Baddeley:
The action manager is in a state of ongoing, active, development - it is usable and should be fairly robust, but falls under the category of "internals" rather than "API". In other words whilst there is an ongoing commitment that to maintain the functionality encapsulated in the `ActionManager`, there is no guarantee that the **programatic** interface to that functionality will be static.  This will likely change as the implementation details shake out a bit more (I think the current iteration has a fair chance of sticking around a bit, but there are still a couple of non-finalised design choices which might result in small changes). 

The "internals" status notwithstanding, the latest changes are pretty small, and it should be really easy to adapt to the new interface (basically you pass a dictionary of settings rather than a constantly growing/changing set of keyword arguments - the `max_frames` argument is now one of the keys in the settings dictionary). This is documented in the `start_spooling()` docstring. The relevant changes came in https://github.com/python-microscopy/python-microscopy/pull/799, which was potentially merged a bit earlier than might otherwise have been the case as, in addition to refactoring etc ... it happened to fix a few bugs/feature requests.

David-Baddeley:
[...]

PR #991, opened 2021-04-07
decorators for adding api status to class/function docs

David-Baddeley:
In response to #988

**Proposed changes:**

Currently a bit of a straw-man concept for documenting the API status of various bits of PYME to

a) help writers of 3rd party code determine what is safe to use/call
b) act as mental reminders when maintaining/modifying code when backwards compatibility should be maintained
or there should be a deprecation path.

The decorators will add an `API Status` section to the class/function docstring

Would be great to get some feedback on whether this is likely to be useful / make sense.

@csoeller, @barentine, @zacsimile

**TODOS:**

Probably need to change apidoc/napoleon settings to make it recognize the custom docstring section

### DIFF
--- a/PYME/apistatus.py
+++ b/PYME/apistatus.py
@@ -1,0 +1,51 @@
+"""
+
+This module defines a series of decorators used for documenting functions API status / maturity. The hope is that this
+will provide a simple way of indicating which parts of PYME are safe to use from third party code (and which parts are
+volatile).
+
+"""
+
+
+def _api_annotation(status_blurb):
+    def dec(obj):
+        """Decorator to mark the API status of a function/class
+        
+        Parameters
+        ----------
+        obj : object to be decorated
+        
+        """
+        
+        obj.__doc__ = obj.__doc__ + '\nAPI Status\n-----------\n' + status_blurb
+        
+        return obj
+    
+    return dec
+
+api = _api_annotation('''
+**API:** This is considered part of the external API and expected to be used 3rd party code. Any backwards incompatible
+future changes will by clearly signalled with deprecation warnings for several releases prior to making breaking changes.
+''')
+
+internal = _api_annotation('''
+**Internal:** This is not expected to be called from 3rd party code and may change without notice, use at you own risk.
+''')
+
+experimental = _api_annotation('''
+**Experimental:** This is experimental functionality, or functionality under development. It might become part of the API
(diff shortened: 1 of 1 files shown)


David-Baddeley:
proposed usage:

```python

from PYME import apistatus

@apistatus.api
def some_function_we_are_going_to_maintain(...)
    # a function that we expect to be used in 3rd party code / plugins

@apistatus.internal
def _my_magic_function(...)
   # something that we do not expect to (ever) be called from external code
  # NOTE: a leading underscore will always imply @apistatus.internal, whether annotated or not
  # but is used for things that are private to a specific module/class, even within PYME. The purpose of
  # the decorator is actually for classes/functions which might represent an internal API of sorts - 
  # i.e. public within PYME, but not recommended for external use.

```

David-Baddeley:
On the super-long time scale TODO - add a flag to the class/method which alters how APIDoc parses/formats it.

csoeller:
This seems like a good approach. Its presence in the code will already signal to anybody who seriously works on code what the current/intended status is - as this generally needs looking into the source code anyway. If it eventually shows up in docs good, too, but probably less essential then in the code base. Would also remind any core developer touching certain interfaces may require extra care.

zacsimile:
I think this is a reasonable middle-ground between writing clear documentation and saying, "it's in the code, good luck". Documentation would eliminate the need for these decorators, but we probably won't document everything anytime soon.

We should definitely make sure the meaning of these flags is in the online docs.

David-Baddeley:
@zacsimile - Documentation would not strictly remove the need for the decorators. 

- The alternative of manually putting an API status section in each docstring would result in a lot of extra work, and would be error prone and/or not done.
- High level documentation (as might be needed) is not directly tied to the functions in question and might not track. 
- People are not necessarily going to remember to read the API summary when modifying functions - to have a flag that will be immediately visible when editing a function should be very useful
- the decorator approach can be extended to a) automatically collect and summarize all functions that form part of the API and b) potentially introduce C/I checks on modifications to API functions.

zacsimile:
I think this is worth merging. Can we also add a `decorators.rst` file describing this and put it under "API Documentation" in `hacking.rst`?

David-Baddeley:
Major issue with this at present is how sphinx/numpydoc treats docstring sections which don't match it's expectations ...

David-Baddeley:
looks like https://github.com/sphinx-doc/sphinx/pull/8658 might have a solution ...

PR #1006, opened 2021-04-14
add SelectLargestLabel filter module

barentine:
Addresses issue #just wanting the largest label if one is trying to e.g. grab a nucleus mask when there's an edge of a neighbor in the field of view.

enhancement
**Proposed changes:**
add SelectLargestLabel module

**Checklist:**

- [x] Tested 
- [ ] with numpy=1.14

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -120,7 +120,11 @@ class SelectLabel(Filter):
 
 @register_module('SelectLargestLabel') 
 class SelectLargestLabel(Filter):
-    """Creates a mask corresponding to all pixels with the largest label"""
+    """Creates a mask corresponding to all pixels with the largest label
+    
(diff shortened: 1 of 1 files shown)


PYME/recipes/processing.py
```diff
@@ -118,6 +118,19 @@ def applyFilter(self, data, chanNum, frNum, im):
     def completeMetadata(self, im):
         im.mdh['Processing.SelectedLabel'] = self.label
 
+@register_module('SelectLargestLabel') 
+class SelectLargestLabel(Filter):
+    """Creates a mask corresponding to all pixels with the largest label"""
```
```suggestion
    """Creates a mask corresponding to all pixels with the largest label
    
    NOTE: the input image must be a labeled image (e.g. the output of `Processing.Label`) in which contiguous
    areas have unique integer IDs
    """
```

PR #1007, opened 2021-04-14
allow spherical harmonic shell tabular to be loaded in recipe, add shell coordinate module

barentine:
Addresses issue #fitting a shell using a shell channel but wanting to know the scaled radial positions of points in another channel.

enhancement
**Proposed changes:**
- fix to valueerror when m is negative, as numpy doesn't like (-1)**anything, only (-1.0)**anything.
- refactor loading slightly so we can make a shell from a tabular type which we e.g. throw into the recipe namespace at least when shell fitting in PYMEVis gui
- module to add scaled spherical coordinates and a normalized radius mappings

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
note that this needs metadata passthrough on the addshellmappedcoords. Have it local

Issue #809, opened 2021-01-28
[ENHANCEMENT] BigTIFF support??


David-Baddeley:
Should we consider adding BigTIFF support? 

Given the fact that many of our files are >4GB it might make sense to offer BigTIFF support for, e.g. export to ImageJ. 

As ImageJ & Matlab (the two most likely export targets) can read HDF5 I don't think this is necessarily a priority, but it came up in #784 and should potentially be on our radar.

David-Baddeley:
Actually, our version of tifffile should have automaticaly saved as bigtiff if the data size was big enough to need it - maybe we just need to do a bit of debugging.

David-Baddeley:
closing this as I suspect it is fixed .... can reopen if needed

Issue #607, opened 2020-11-14
[BUG] ImageJ multichannel tiff loading/ view scaling on PYMEVis load raw data


Hannahmar:
I clicked on PYMEVis on my Windows 10 laptop, I went to File>Open raw/prebleached data, I opened a two channel TIFF, I went to GLComp and it was dark, Chan 0 was also blank. Images from both channels were mixed and appeared in Slices. Also what is GLComp?
I ran the Gaussian filter, and after a few minutes a blank window appeared and I couldn't see the image.
Not sure what I did but after zooming in and out, also adjusting display, I could see the image clearly on my second try.

![20201113_182523](https://user-images.githubusercontent.com/74436462/99131380-fa77f680-25e0-11eb-850e-0255c1bd1577.jpg)

David-Baddeley:
I suspect there are a number of things going on here

- File>Open raw/prebleached is designed for opening raw data that was used to generate a set of localisations that you have already opened. The "Chan 0" and "GLComp / composite" panes will the same zoom and position from the points viewer. This means that you might not get something sensible if: a) there are no points loaded or b) the metadata for either points or image is wrong (i.e. the software gets the relative scaling and/or position wrong). The voxelsize of the loaded data would be the most likely culprit here. The "Slices" tab is not effected as it ignores the scaling in the points view.
 - Most software saves multi-channel tiff stacks as a single XYZ stack with the colour channels interleaved (i.e. for a 3 colour tiff images 1,2,3 are the different channels of the 1st z slice, images 4,5,6 are the second z slice etc .... Unfortunately there is no standard way of indicating that this is what has been done, so you need to rely on something in the image metadata to tell you. How this is specified differs depending on what program saved the tifs. We currently tend to get the colour channels right if you save as OME TIFF (e.g. by using the bioformats exporter, rather than the standard save as tiff function in FIJI) or save as tiff from PYME, but don't do so well at guessing what is going on if the data doesn't have OME metadata. **Where did your multi-channel tiff come from?** 

For reference, the tabs in the image viewer are as follows:

**Chan 0, Chan 1 etc ... :** The individual colour channels, with zoom and position synced with the points viewer 

**Composite:** A composite of all colour channels, synced to the points viewer (only shown if the software detects multiple channels)

**Slices:** An unsynchronised view of the data (also supports x-z slices and a few extra annotation functions, which is why it's shown here)

[...]

PR #1016, opened 2021-04-20
Further xyztc data model improvements

David-Baddeley:
Follows on from the initial PR on this branch, to add an xyztc data source to images by default


David-Baddeley:
I'm sure there are a few issues related to this, but can't seem to find them. I think #607 and #809 are somewhat related, but feel there should be others.

David-Baddeley:
@zacsimile @csoeller - I want to merge this before it diverges too far, as well as some important new functionality, it has loads of neat bugfixes etc in it. There are some pretty big changes, however, which so would be very grateful if one of you could give it a try before I merge with master ...

To get the full functionality you'll want to install `tifffile`, `zar`, `dask` and `neuroglancer` on top of your existing `pyme-depends` environment, although you probably don't need all of them (`tifffile` makes the most difference, fixing our longstanding issues with loading .tiff from Fiji/ImageJ).  I think that all except for dask are pip only.

csoeller:
Thanks @David-Baddeley for the heads up, I'll try to get to it this week, probably will install it in its own conda env so that switching between implementations is real easy.

zacsimile:
Very exciting, @David-Baddeley! I gave this a go. Nothing seems broken, but I tried Fiji's `File>Open Samples>Mitosis (5D Stack)`, saved the tif, and then tried opening it in `dh5view` and it opened flat (all z, no t, no c). `dh5view` also prompted me for voxelsize, which is definitely stored in this tif.

csoeller:
Hey @David-Baddeley and @zacsimile, I have now built the xyztc branch of David's fork and was going to ask how should I test.

Anyway, started with something simple (I thought) and experiences are not quite so smooth yet. I decided to give it a spin using some of our example data, 2 colour, start from h5r, render that, save and try to open with dh5view. It seems like I am running into a bunch of small bugs. I will double check if I made some silly mistakes but just in case, how to best report bugs/issues? Can we use an issue tracker in your fork?

David-Baddeley:
@csoeller I'd probably just keep feedback /bugs etc as comments on this
thread, if that's OK. I thought I had already tried most of the steps in
your example, but potentially not.

David-Baddeley:
@zacsimile Have you installed tifffile? You need that to open the Fiji
tiffs un-flattened. Reading the Fiji pixel sizes is still on the TODO.

zacsimile:
Hi @David-Baddeley. I do have `tifffile` installed. I tried opening this file as a `.tiff` (`bioformats`) instead of `.tif` and it worked better. I now have `z` and color, but I still don't get `t`. `z` and `t` are flattened together.

zacsimile:
Ah sorry, @David-Baddeley, I didn't have `tifffile` installed. Even though I told it to install earlier, I guess it didn't. It does work with `tifffile` installed.

I'd recommend not making this a silent error. Also, we should make it work with `bioformats`, too.

David-Baddeley:
Huh - what version of Fiji - is it up to date? My success was with a fully
up to date Fiji, using scifio (which might need to be enabled in the
options). Obviously we'd like to open all fiji/imagej tiffs though.

zacsimile:
@David-Baddeley what was the reasoning behind adding `tifffile` as a new dependency given that we already had `bioformats` working? Just that it's all Python?

David-Baddeley:
Ultimately we should put tifffile in pyme-depends and fail if not present -
I added the fallback to the old/current behavior to make things smoother in
the interim.

csoeller:
Ok, here is the first little issue. I rendered the file as described and saved as tif. Then tried to open with dh5view and got:

```
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/scrolledImagePanel.py", line 58, in OnPaint
    self.renderer(MemDC)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/arrayViewPanel.py", line 494, in DoPaint
    im = self.Render(fullImage)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/arrayViewPanel.py", line 1071, in Render
    seg = ds[slice_key +  (chan,)].squeeze().T
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/IO/DataSources/BaseDataSource.py", line 314, in __getitem__
    out[:, :, zi, ci, ti] = self.getSlice(slice_idx)[keys[0], keys[1]]
ValueError: could not broadcast input array from shape (604,515) into shape (603,514)
/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/scrolledImagePanel.py:52: wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  MemBitmap = wx.EmptyBitmap(s.GetWidth(), s.GetHeight())
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/scrolledImagePanel.py", line 58, in OnPaint
    self.renderer(MemDC)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/arrayViewPanel.py", line 494, in DoPaint
    im = self.Render(fullImage)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/DSView/arrayViewPanel.py", line 1071, in Render
    seg = ds[slice_key +  (chan,)].squeeze().T
  File "/Users/csoe002/Documents/src/PYME-src/pyme-xyztc/PYME/IO/DataSources/BaseDataSource.py", line 314, in __getitem__
    out[:, :, zi, ci, ti] = self.getSlice(slice_idx)[keys[0], keys[1]]
ValueError: could not broadcast input array from shape (604,515) into shape (603,514)
```

Observations:

1. the file opened ok using the current default branch dh5view
2. I realised that I did not have tifffile installed when I first wrote and then read the tif.
3. I installed tifffile (and the other recommended modules, where tifffile installed ok with conda)
4. Still same problem reading the file when tifffile is installed.
5. Then tried to render a new image (different ROI) and read with dh5view, this time it worked.

Two possibilities:
a. not having tifffile etc when rendering caused the problem
b. nothing to do with tifffile presence/absence, but a problem that is jiust triggered sometimes depending on some pixel/ROI arithmetic issue.

David-Baddeley:
Tifffile has a lot of advantages when compared with bioformats:
- it's safer (we can be much more certain that it will install)
- it's **much** faster
- it will usually have a lower memory footprint
- it is essentially already a dependency, albeit in an older version

David-Baddeley:
@csoeller - I thought I'd clobbered that display issue. It's not tiff specific, and occurs when the zoom is <1 and the image is smaller than the window, and the image size is odd. A workaround is to resize the window such that the window is smaller than the image (potentially useful in debugging)

csoeller:
@David-Baddeley, thanks, I can confirm that this resising approach works. Other, potentially unrelated issues:

1. With recent PYME I struggle with scaling the display suitably, all the things that I try (stretch or M in the histogram panel) leave deeply saturated patches, screenshot below.
2. in the new xyztc branch the profiling does not work at the moment, maybe known? See also screenshot below.

Screenshot 2021-05-05 at 20 28 39

David-Baddeley:
- profile plotting is definitely a new bug (and one I might not have thought to test)
- the scaling may or may not be specific to this branch. There have been a couple of changes which might effect this, both related to improving performance for large images.
    - **change 1 (in master):** We now downsample the data to 10k pixels when calculating the histograms (effectively `data[::int(data.size/1e4)]`). This effectively downsamples on a regular grid and might miss the bright regions in sparse images. This change has been in for a while (was added with sCMOS camera support).
    - **change 2 (in this branch): ** Instead of downsampling uniformly, we take a 100x100 pixel area around the current position (as indicated by the cyan crosshairs). The rationale behind this is to support large (terravoxel), tiled datasets where you don't want to load an entire x-y slice. As it stands, I'm not entirely sure that the histograms update when you change xy position unless you also change z,  clicking on a structure and hitting 'm' in the image view (not the histogram) does work, however.

none of these feels quite right, and I wonder if we should do something like using the whole image up to a certain size and then switching to the ROI base approach, or doing some aggregate - ROI + downsampled whole image for small images, ROI only for large.

csoeller:
Ok, great. The scaling feels a little weird, previously with `stretch` I could force scaling with a maximum value at the top end (I think). While the 1e4 downscale may be a good general solution, it would be good to also have some command that forces the old behaviour. Could you come up with something like that? Maybe F (=force) could do it?

I will continue exercising the new xyztc code over the next few days/weekend and see if anything else comes up.

How can I actually try the new features that this data model enables? I am a little ignorant about that aspect.

David-Baddeley:
There are not a huge number of new features at present - the major ones visible to the user at present are:
- export to .tif and have ImageJ read the channels  correctly etc ...
- import multichannel etc ... tiffs from ImageJ and have them import correctly
- support for large, terra-voxel image pyramids in .zarr format
- export to h5 and preserve colour channels, time steps.

Under the hood, it makes things more consistent, more flexible, and adds support for data which has both z-stepping and a temporal aspect. If you use `image.data_xyztc` or `image.voxels` programatically (rather than `image.data`), the dimensions/indexing should always have a consistent meaning, rather than the 3rd dimension being either z or t with no way of knowing with certainty which it was.

At this stage I'm mostly interested in making sure we haven't broken anything which is part of the normal workflow. So far the broadcasting (which is hopefully now fixed) and line profile bugs fall into this category, and the display scaling one possibly needs more attention (although on some level we are trading broken for broken, so it might not be blocking).

zacsimile:
I haven't hit any more errors but I also don't use `dh5view` that frequently so I may just not be stress-testing it enough.

David-Baddeley:
Remaining known issues:

[ ] diagonal composite
[ ] make composite

Untested

[ ] deconvolution
[ ] unsplit
[ ] tiling

PR #1017, opened 2021-04-22
density estimate for spherical harmonic shell normalized radius

barentine:
Addresses issue # .

enhancement
**Proposed changes:**
estimate the expected distribution of normalized radius for a uniform distribution inside a spherical harmonic shell

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/recipes/surface_fitting.py
+++ b/PYME/recipes/surface_fitting.py
@@ -364,6 +364,8 @@ class SHShellRadiusDensityEstimate(ModuleBase):
     Also estimates the volume of the shell, the anisotropy of the shell, and
     the standard deviation along its principle axes (if filled with a uniform
     distribution)
+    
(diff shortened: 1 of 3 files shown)


PYME/recipes/surface_fitting.py
```diff
@@ -354,6 +354,106 @@ def execute(self, namespace):
...
+class SHShellRadiusDensityEstimate(ModuleBase):
+    """
+    Estimate the normalized radius histogram for a uniform distribution within
+    the shell.
+    TODO: make more generic re: SDF / other surfaces
+
+    Also estimates the volume of the shell, the anisotropy of the shell, and
+    the standard deviation along its principle axes (if filled with a uniform
+    distribution)
+    """
```
David-Baddeley, line 369:
```suggestion
    
    TODO - rename/refactor. This has too much logic in the recipe module itself, it also feels very special case and the name might make it hard to discover.
    """
```

PR #1024, opened 2021-05-06
allow config for local tasks only

barentine:
Addresses issue #really slamming nodes on big RAM-slaying recipes makes them not so responsive to network I/O / susceptible to timeouts.

enhancement
**Proposed changes:**
- allow whether to look for nonlocal tasks to be config'd in rulenodeserver

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/cluster/rulenodeserver.py
+++ b/PYME/cluster/rulenodeserver.py
@@ -225,7 +225,7 @@ class NodeServer(object):
                     if n_tasks >= n_tasks_to_request:
                         break
                         
-                if n_tasks < 1:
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Should really be fixed by: 

a) making sure that recipe tasks are not stupidly large 
b) improving recipe costings

But this should be acceptable as a short/medium term workaround. a) needs a lot of work to be practical for some tasks, but as a general rule of thumb, if a single recipe task takes > ~1 min or needs >1GB ram, it should be refactored into multiple tasks on chunks which each of which can be completed quickly. Will create a roadmap issue on where I think recipes should be going to improve large dataset support.

Issue #657, opened 2020-12-01
Py 3.8 and 3.9 compatibility issues 


David-Baddeley:
Starting as somewhere to put known issues/status on py=3.8 and 3.9. Add as encountered.

## Python==3.8

- PYMEVis appears to work 
- Some issues in dialogs (e.g. generate image) on most recent wx due to wx getting fussier. Think this also effects py3.6, 3.7
- Deprecated `time.clock()` calls need replacement (in `PYMEAcquire.mytimer`). This needs to either be fudged or wait on py2 retirement as suitable replacement calls were introduced in 3.3. PYMEAcquire does not work.
- Need to build stuff manually.

## Python==3.9

- basically no conda packages for anything (including basics such as numpy)

zacsimile:
Re: wx. We can start fixing most of those wx complaints by moving to modern wx syntax. We can do this once we drop wx3.

David-Baddeley:
The specific wx issues I was running into can and should be fixed whilst maintaining backwards compatibility (specific issue is using `wx.ALIGN_CENTER_VERTICAL` in a vertical sizer which used to be silently ignored). Think I have an uncommitted / un pushed fix on my linux computer.

barentine:
Are we not one month away from dropping wx3 support? #337

David-Baddeley:
My point was more that we should fix the specific vertical alignment problem now (as it's also likely to bite us on 3.6 and 3.7). 

As to the deadline for dropping wx3 support, it's now looking like that will slip a month or two. We might need to hold off a little on throwing the final switch, and saying we will merge breaking code, as I still haven't got the microscope machine here fully ported. This down is a combination of things, and partly the result of trying to combine the transition to py3 with a switch to win10, but the bottom line is that that has to be running happily on py3 for a month or two before I'm happy to pull the plug.

zacsimile:
I understand not wanting to switch off py2 for a couple of months, but wx3 is only supported on Mac (https://anaconda.org/david_baddeley/wxpython/files). Is this safe to drop in a month? It would be cool to start moving to the new wx syntax.

David-Baddeley:
Unfortunately not, conda is so broken on py27 at the moment that the only safe way to update is to do so in a way which doesn't require updating **any** of the dependencies.

zacsimile:
[...]

PR #1033, opened 2021-05-12
Adjust display option flags to be non-conflicting for wx 4.1

zacsimile:
Addresses comment in issue #657.

Bugfix

**Proposed changes:**
- Remove extraneous alignment flags.

### DIFF
--- a/PYME/DSView/DisplayOptionsPanel.py
+++ b/PYME/DSView/DisplayOptionsPanel.py
@@ -125,7 +125,7 @@ class OptionsPanel(wx.Panel):
             hsizer.Add(self.cbScale, 0, wx.ALL|wx.ALIGN_CENTER, 5)
             vsizer.Add(hsizer, 0, wx.ALL|wx.ALIGN_CENTER, 0)
         else:
-            vsizer.Add(self.bOptimise, 0, wx.LEFT|wx.RIGHT|wx.TOP|wx.ALIGN_CENTER|wx.EXPAND, 5)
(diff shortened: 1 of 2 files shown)


PYME/DSView/DisplayOptionsPanel.py
```diff
@@ -125,7 +125,7 @@ def __init__(self, parent, displayOpts, horizOrientation=False, thresholdControl
             hsizer.Add(self.cbScale, 0, wx.ALL|wx.ALIGN_CENTER, 5)
             vsizer.Add(hsizer, 0, wx.ALL|wx.ALIGN_CENTER, 0)
         else:
-            vsizer.Add(self.bOptimise, 0, wx.LEFT|wx.RIGHT|wx.TOP|wx.ALIGN_CENTER|wx.EXPAND, 5)
```
David-Baddeley, line 128: In this case, the `wx.LEFT|wx.RIGHT|wx.TOP` refer to padding (go with the following 5) and should not be removed.

PYME/DSView/DisplayOptionsPanel.py
```diff
@@ -125,7 +125,7 @@ def __init__(self, parent, displayOpts, horizOrientation=False, thresholdControl
...
+            vsizer.Add(self.bOptimise, 0, wx.EXPAND, 5)
```
David-Baddeley:
```suggestion
            vsizer.Add(self.bOptimise, 0, wx.LEFT|wx.RIGHT|wx.TOP|wx.EXPAND, 5)
```

Issue #1035, opened 2021-05-14
[BUG] Downloading Distance transform colocalization raw histogram data into .txt file. 


Carmen-Rodilla:
I am trying to save the raw histogram data from the Distance transform colocalization of the coloc module as indicated in the page (https://www.python-microscopy.org/doc/howtos/distancecolocalization.html) for further quantification. After download, the document appears to be empty. Is there any other way -apart from the described in the page- to get the data? 
Thank you!

zacsimile:
Thanks for the bug report @Carmen-Rodilla! I've looked into this a bit. Following steps in https://www.python-microscopy.org/doc/howtos/distancecolocalization.html. When I try to use the signal threshold, I get

```
Traceback (most recent call last):
  File "c:\users\zrc4\code\python-microscopy\PYME\DSView\DisplayOptionsPanel.py", line 296, in OnSignalFracThresh
    t = thresholding.signalFraction(self.do.ds[:,:,:,i], frac)
  File "c:\users\zrc4\code\python-microscopy\PYME\IO\DataSources\BaseDataSource.py", line 291, in __getitem__
    raise RuntimeError('Must provide a 5-dimensional slice')
RuntimeError: Must provide a 5-dimensional slice
```

When I hit `File > Save` and try to save the histogram txt, I get the following traceback.

```
Traceback (most recent call last):
  File "c:\users\zrc4\code\python-microscopy\PYME\DSView\dsviewer.py", line 324, in OnSave
    self.image.Save()
  File "c:\users\zrc4\code\python-microscopy\PYME\IO\image.py", line 1252, in Save
    self.filename = dataExporter.ExportData(self.data_xyztc, self.mdh, self.events, defaultExt=self.defaultExt, filename=filename, progressCallback=progressCallback)
  File "c:\users\zrc4\code\python-microscopy\PYME\IO\dataExporter.py", line 647, in ExportData
    exp.Export(ds, filename, slice(0, ds.shape[0], 1), slice(0, ds.shape[1], 1), slice(0, ds.shape[2], 1),mdh, events, origName, progressCallback=progressCallback)
  File "c:\users\zrc4\code\python-microscopy\PYME\IO\dataExporter.py", line 429, in Export
    dat = [data[xslice, yslice, zslice, chan].squeeze() for chan in range(data.shape[3])]
  File "c:\users\zrc4\code\python-microscopy\PYME\IO\dataExporter.py", line 429, in <listcomp>
[...]

PR #1036, opened 2021-05-14
Fixes for Issue1035, plus other stuff I ran into on the way

zacsimile:
Addresses issue #1035.

Bugfix

**Proposed changes:**
- XYZTC stuff in the display options panel and data exporter
- Byte/String stuff in the data exporter 
- Some align center vertical flag removal, since I'm now betaing everything in py38/wx4.1

### DIFF
--- a/PYME/DSView/DisplayOptionsPanel.py
+++ b/PYME/DSView/DisplayOptionsPanel.py
@@ -285,7 +285,7 @@ class OptionsPanel(wx.Panel):
         
         
         for i, hClim in enumerate(self.hcs):
-            t = thresholding.isodata_f(self.do.ds[:,:,:,i], bin_spacing=bin_spacing)
+            t = thresholding.isodata_f(self.do.ds[:,:,:,:,i], bin_spacing=bin_spacing)
             hClim.SetValueAndFire((t,t))
 
     def OnSignalFracThresh(self, event):
@@ -293,7 +293,7 @@ class OptionsPanel(wx.Panel):
         frac = max(0., min(1., float(self.tPercThresh.GetValue())))
         self.do.ThreshMode = ['fractional',frac]
         for i, hClim in enumerate(self.hcs):
-            t = thresholding.signalFraction(self.do.ds[:,:,:,i], frac)
+            t = thresholding.signalFraction(self.do.ds[:,:,:,:,i], frac)
             hClim.SetValueAndFire((t,t))
 
 
--- a/PYME/DSView/dsviewer.py
+++ b/PYME/DSView/dsviewer.py
@@ -142,7 +142,7 @@ class DSViewFrame(AUIFrame):
         #a submenu for modules to hook and install saving functions into
         self.save_menu = wx.Menu()
         self._menus['Save'] = self.save_menu
-        tmp_menu.AppendMenu(-1, 'Save &Results', self.save_menu)
+        tmp_menu.Append(-1, 'Save &Results', self.save_menu)
         
         #tmp_menu.AppendSeparator()
         #tmp_menu.Append(wx.ID_CLOSE, "Close", "", wx.ITEM_NORMAL)
--- a/PYME/DSView/dsviewer.py
+++ b/PYME/DSView/dsviewer.py
@@ -118,7 +118,7 @@ class DSViewFrame(AUIFrame):
 
         if self.image.mdh and 'ChannelNames' in self.image.mdh.getEntryNames():
             chan_names = self.image.mdh.getEntry('ChannelNames')
-            if len(chan_names) == self.image.data.shape[3]:
+            if len(chan_names) == self.image.data_xyztc.shape[4]:
                 self.do.names = chan_names
 
         #self.vp = ArraySettingsAndViewPanel(self, self.image.data, wantUpdates=[self.update], mdh=self.image.mdh)
--- a/PYME/DSView/modules/coloc.py
+++ b/PYME/DSView/modules/coloc.py
@@ -172,12 +172,12 @@ class Colocaliser(Plugin):
         chans = dlg.GetChans()
         use_mask = dlg.GetUseMask()
 
-        zs, ze = (0, self.image.data.shape[2])
-        if (self.image.data.shape[2] > 1) and restrict_z:
+        zs, ze = (0, self.image.data_xyztc.shape[3])
+        if (self.image.data_xyztc.shape[3] > 1) and restrict_z:
             zs, ze = dlg.GetZrange()
             
         zs,ze = (0,1)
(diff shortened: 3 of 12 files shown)


PYME/DSView/dsviewer.py
```diff
@@ -142,7 +142,7 @@ def __init__(self, image,  parent=None, title='', mode='LM',
         #a submenu for modules to hook and install saving functions into
         self.save_menu = wx.Menu()
         self._menus['Save'] = self.save_menu
-        tmp_menu.AppendMenu(-1, 'Save &Results', self.save_menu)
+        tmp_menu.Append(-1, 'Save &Results', self.save_menu)
```
David-Baddeley, line 145: We should probably change these to use the AUIFrame methods ...

PYME/DSView/modules/coloc.py
```diff
@@ -189,8 +189,8 @@ def OnColoc(self, event=None, restrict_z=False, coloc_vs_z = False):
 
         #assume we have exactly 2 channels #FIXME - add a selector
         #grab image data
-        imA = self.image.data[:,:,zs:ze,chans[0]].squeeze()
-        imB = self.image.data[:,:,zs:ze,chans[1]].squeeze()
+        imA = self.image.data_xyztc[:,:,zs:ze,:,chans[0]].squeeze()
```
David-Baddeley: Is all time points what we want??

PYME/DSView/modules/coloc.py
```diff
@@ -316,8 +316,8 @@ def OnColoc(self, event=None, restrict_z=False, coloc_vs_z = False):
         
         im.mdh['Colocalisation.Channels'] = names
         im.mdh['Colocalisation.Thresholds'] = [tA, tB]
-        im.mdh['Colocalisation.Pearson'] = pearson
-        im.mdh['Colocalisation.Manders'] = [MA, MB]
+        im.mdh['Colocalisation.Pearson'] = np.array(pearson)
```
David-Baddeley: these should be scalars, casting them to an array will cause problems with serialisation. If casts are needed, they should be to `float`.

David-Baddeley:
A few points:

- although we probably can break wx3 compat, it would be nice to keep these breaking changes in separate PRs
- I think you've been ever so slightly too eager on the xyztc conversions. The old `image.data` should still work, and my strategy has been to leave the old methods on things (e.g. coloc) until we change them to something that is fully 5D aware and makes sense. For a lot of these, just taking all t-values does not make sense.

PYME/LMVis/layers/image_layer.py
```diff
@@ -272,7 +272,7 @@ def update_from_datasource(self, ds):
         
         if not self._im_key == im_key:
             self._im_key = im_key
-            self._im = ds.data[:,:,self.slice, self.channel].astype('f4').squeeze()# - c0)/(c1-c0)
+            self._im = ds.data_xyztc[:,:,self.slice,:,self.channel].astype('f4').squeeze()# - c0)/(c1-c0)
```
David-Baddeley: This will break badly if you ever give it 5D data (self._im must be a 2D array). It's probably best to stick with the collapsed z/t for now (potentially explicitly if you want to remove the deprecation warning). A 5D version needs another parameter analogous to slice. 

This probably also makes the changes to the visgui dsviewer plugin are also a bit premature.

PYME/DSView/DisplayOptionsPanel.py
```diff
@@ -285,15 +285,15 @@ def OnIsodataThresh(self, event=None, bin_spacing='linear'):
         
         
         for i, hClim in enumerate(self.hcs):
-            t = thresholding.isodata_f(self.do.ds[:,:,:,i], bin_spacing=bin_spacing)
+            t = thresholding.isodata_f(self.do.ds[:,:,:,:,i], bin_spacing=bin_spacing)
```
David-Baddeley, line 288: Probably OK for now, but this might need a slightly more complicated solution, as I'm not sure that `do.ds` is strictly guaranteed to be xyztc.

zacsimile:
> A few points:
> 
> * although we probably can break wx3 compat, it would be nice to keep these breaking changes in separate PRs
> * I think you've been ever so slightly too eager on the xyztc conversions. The old `image.data` should still work, and my strategy has been to leave the old methods on things (e.g. coloc) until we change them to something that is fully 5D aware and makes sense. For a lot of these, just taking all t-values does not make sense.

Fair enough. I'll restore some of the old `image.data` calls. I definitely was too eager. Although we should discuss how we want to handle these cases (all t? t=0? something else?)

Are any of these changes breaking for wx3? I have difficulty finding wx3 documentation.

PYME/LMVis/layers/image_layer.py
zacsimile: Will revert

PYME/DSView/modules/coloc.py
zacsimile: Definitely needs a cast. Will switch to `float`.

PYME/DSView/modules/coloc.py
zacsimile: No idea. Will revert for now

PYME/DSView/dsviewer.py
zacsimile: Sure

David-Baddeley:
Don't rush to revert too much, I think the gl view definitely needs it, not
sure on others.

David-Baddeley:
Also don't worry about changing the menu stuff... That was more of a side
comment.

zacsimile:
OK, sounds good.

zacsimile:
So, I made those changes, and there's no errors thrown in the process, but I think EDT colocalization is wrong right now. I'm trying it on the data we used in the PYMEVis publication and it's not producing the same results. Thoughts? The problem appears to persist, even if I revert everything. I'd actually noticed it before I fixed anything and then didn't think about it again until now.

image

image

image

zacsimile:
For reference, this is what we get if I download our packaged installer and do the same thing.

image
image
image

David-Baddeley:
For the `t=?` cases, we might want to use `t=do.tp` (i.e. the currently shown time point) in the interim, but ultimately convert to a time-aware recipe module and call that. That's the approach I started for making composites (although the viewer code doesn't call the recipe module yet, the re-factored logic used by both viewer and recipe is common).

David-Baddeley:
For reference, the breaking wx changes are the `wx.Cursor`, `wxMenu.Append()`, and `InsertItem()` and `SetItem()` changes. None of the type-overridden stuff works on wx3. Will merge with these anyway, but it's nice to know where we have made breaking changes.

PR #1039, opened 2021-05-19
pco. camera using pco.sdk only

zacsimile:
New version of the pco. camera that bypasses the pco.recorder stuff, which seems to not play nice with the Python garbage collector.

This works! I am able to image and it's responsive (potentially full frame rate) and doesn't crash.

However, I decided not to do what we do with the Zyla because pco.sdk makes a big point of saying don't use the camera with threading. As such, we're not buffering exactly as you'd expect. Also, we can only create 16 buffers max. Doesn't seem to be a problem practically, just a lot lower than what I'm used to.

But it works.

### DIFF
--- a/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
@@ -1,0 +1,285 @@
+# -*- coding: utf-8 -*-
+
+"""
+Created on Sun May 16 2021
+
+@author: zacsimile
+"""
+
+from PYME.Acquire.Hardware.Camera import Camera
+from PYME.Acquire.Hardware.pco import pco_sdk
+
+import numpy as np
+import logging
+logger = logging.getLogger(__name__)
+
+import ctypes
+
+timebase = {pco_sdk.PCO_TIMEBASE_NS : 1e-9, 
+            pco_sdk.PCO_TIMEBASE_US : 1e-6, 
+            pco_sdk.PCO_TIMEBASE_MS : 1e-3}  # Conversions 
+
+class camReg(object):
+    """
+    Keep track of the number of cameras initialised so we can initialise and
+    finalise the library.
+    """
+    numCameras = -1
+
+    @classmethod
+    def regCamera(cls):
+        if cls.numCameras == -1:
+            pco_sdk.reset_lib()
+
+        cls.numCameras += 1
+
+    @classmethod
+    def unregCamera(cls):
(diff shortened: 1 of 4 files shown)


zacsimile:
I could switch `add_buffer_ex` to `add_buffer_extern` calls if we want to make a larger buffer. Will just create buffers in the software rather than on the camera. Might be the move.

PYME/Acquire/Hardware/pco/pco_sdk_cam.py
```diff
@@ -0,0 +1,405 @@
...
+        self.SetIntegTime(0.025)
+        self.SetAcquisitionMode(self.MODE_CONTINUOUS)
+        self._cam_type = pco_sdk.get_camera_type(self._handle)
+        self.SetHotPixelCorrectionMode(pco_sdk.PCO_HOTPIXELCORRECTION_OFF)
+
+    @property
+    def noise_properties(self):
+        return self.noiseProps
+
+    def ExpReady(self):
```
David-Baddeley, line 130: As written, this blocks waiting for frames. When `ExpReady()` was polled from the GUI thread that would have been a major problem. If we can get away with it now , it's definitely cleaner.

David-Baddeley: The general assumption made in the code however, is that `ExpReady()` is lightweight and non-blocking, so I'm worried that this will come back to bite us.

zacsimile: I think you are right. It does work at the moment, but I suspect it can cause problems going faster. I could try the Zyla architecture. Maybe it will just work. The pco.sdk manual makes a big thing out of not using threading but it might be fine.

David-Baddeley: I'd be tempted to just put a comment in the code, push it until it breaks, and then fix it when it becomes a problem.

zacsimile: That would also be fine. I was able to take z-stacks with this today, so it's at least working well enough to do what I want for now.

zacsimile: I've added a comment.

zacsimile:
Zyla architecture seems to work fine. I think as long as we're not making calls to the camera in two threads at once we're good to go.

zacsimile:
This is now quite mature and ready to be reviewed again @David-Baddeley.

PR #1040, opened 2021-05-19
Prevent mesh collapse on remesh

David-Baddeley:
@zacsimile - Seems to prevent the skeletonization problems we had on remeshing.

Works by making the target edge length adapt to the local curvature at that edge.  Currently uses the dot product between the vertex normals at either end of the edge to derive the weighting.

```latex
target_edge_length_i = dot(halfedge_i.vertex.normal, halfedge_i.twin.vertex.normal)*target_edge_length 
```

This functions to reduce the edge length / make the mesh finer where the normals between the vertices at either end of an edge are not aligned (i.e. we are in an area of high curvature). Probably best considered a work in progress for now because of the performance regression.

### TODOs:

- optimise for performance (this is definitely a performance regression as currently implemented)
- finalise the scaling factor (this is the first thing I attempted, but might be a bit conservative - we might, e.g., want to raise the dot product to some power to amplify the effect)
- alternatively, introduce curvature as an independent metric for division rather than scaling the edge length - e.g. divide whenever the dot product is < a desired angle.

### DIFF
--- a/PYME/experimental/_triangle_mesh.pxd
+++ b/PYME/experimental/_triangle_mesh.pxd
@@ -99,9 +99,9 @@ cdef class TriangleMesh(TrianglesBase):
     cdef object _K
     cdef public object smooth_curvature
 
-    cdef _set_chalfedges(self, halfedge_t[:])
-    cdef _set_cfaces(self, face_d[:])
-    cdef _set_cvertices(self, vertex_d[:])
+    cdef _set_chalfedges(self, np.ndarray)#halfedge_t[:])
+    cdef _set_cfaces(self, np.ndarray)#face_d[:])
+    cdef _set_cvertices(self, np.ndarray)#vertex_d[:])
     cdef bint _check_neighbour_twins(self, int)
     cdef bint _check_collapse_fast(self, int, int)
     cdef bint _check_collapse_slow(self, int, int)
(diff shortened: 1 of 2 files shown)


zacsimile:
Thanks! This has been on my TODO list for a while. Long term, I like option 3 or a variant on it the best. My working idea is to subdivide whenever the curvature at vertex is > 2/[mean edge length to the neighbors]. This (roughly) keeps the curvature from exceeding 1/R where we assume R is the radius of a sphere fit through the point and its neighbors (i.e. we try to prevent estimating curvatures higher than achievable by our local sampling). Part of the reason I like this is that we have to calculate the curvature at every vertex during each iteration anyway.

David-Baddeley:
To keep things moderately cheap, I think we should still aim for a metric based on the vertex normals rather than a full curvature calculation, but making the curvature subdivision independent of lengths is probably a good idea.

zacsimile:
Sorry, my point was we're already making a curvature calculation most of the time anyway, so I was advocating using that for free.

David-Baddeley:
We don't recalculate curvatures within a remesh cycle / after each split/merge (which was why I wanted something lighter-weight)

zacsimile:
Ah, I see. Yes, good point.

David-Baddeley:
I'm going to merge this as I think performance improvements (59a509a) more than offset the regressions, and it's definitely an improvement on the status quo.

Issue #1041, opened 2021-05-19
[ENHANCEMENT] get some clear instructions up somewhere how to update the one-click installer based installs


csoeller:
**Is your feature request related to a problem? Please describe.**
I am increasingly recommending that people use the one-click installer for PYME. The issue is that I am not sure where to point them on
1. how to update that install when fixes are made available
2. how to find out that fixes have been made available

**Describe the solution you'd like**
Initially a doc page to link to on these questions would be a good starting point. More longer term a simpler way friendly to non-experts would be desirable, possibly along the line of @zacsimile's #483 or something different altogether.

**Describe alternatives you've considered**
If we say in the PYMEVisualise paper we would like to become the ImageJ of SMLM data processing then we will eventually have to bite this bullet.

**Are you prepared to work on it?**
Certainly, once a strategy has been agreed.

David-Baddeley:
updating is covered briefly at 

https://python-microscopy.org/doc/Installation/Installation.html#updating

but could potentially use a little expanding (mostly to explain what to do when `conda update python-microscopy` fails).

Update **notification** has been on the TODO list for ages, and the proposed implementation is sketched out in some detail in #483.

In short:

- when we make a release, update a file on python-microscopy.org which gives the current release number 
- request this file on  program startup, and check against the currently running pyme version
- show a dialog that says an update is available, and gives instructions on how to update. This will probably initially involve running `conda update python-microscopy`, or `conda update -S python-microscopy` in the console to avoid issues over-writing loaded, compiled, modules, but it would be nice to make this more seamless.

Basically we just need to write it, and decide how annoying we should be if someone skips an update (i.e. how long do we wait before telling them again that an update is available).

Duplicate of #483

David-Baddeley:
I think we can close this with update notification now in

Issue #483, opened 2020-10-06
[ENHANCEMENT] Add Help>Update to PYMEVis, PYMEImage, PYMEAcquire


zacsimile:
**Is your feature request related to a problem? Please describe.**
It would be nice to add a graphical method of updating PYME. Asking users to open `Anaconda prompt` and type `conda update python-microscopy` is a bit much for folks who installed using single-click.

**Describe the solution you'd like**
Add a `Help>Update` menu item that calls `conda update python-microscopy` for the user, and maybe prompts the user to relaunch on success.

**Describe alternatives you've considered**
Do nothing.

**Are you prepared to work on it?**
Yep

zacsimile:
I'm not sure of the best way to implement this. The quick and dirty way is 

```
import subprocess
ret = subprocess.call('conda update python-microscopy -y', shell=True)
if ret == 0:
    # success, offer a restart
```

The problem is, this will be successful even if the update wasn't necessary. Ideally we'd compare versions first. Shockingly, I cannot find an easy way to just check if an update exists for a `conda` package without doing anything else. Thoughts?

David-Baddeley:
The solution I had in mind is as follows: 
- put a file (or potentially CGI endpoint) on python-microscopy.org - e.g. current_version.txt. 
- update this whenever we do a release
- client applications fetch this on startup and compare contents to PYME.version, potentially prompting for an upgrade

The advantage of this approach is that it gives us usage metrics (we get to count the number of requests for that file on the server - potentially also sorted by unique IP) for grants etc without adding too much intrusive telemetry. There's an uncommitted (or potentially in PYMEnf) `pyme_phone_home.py` which has a very rough start to this.

David-Baddeley:
There are a few wrinkles in actually implementing the upgrade as we have to detect if they are using a conda-, pip, or source install, so as a first step I'd probably just display a message box indicating that an update was available.

zacsimile:
I'm not sure if we do need to make those detections. I think it might be safe to say the update only works for packaged installs as a first pass.

How do we check if an update is available? Do we compare `PYME.version.version` to the `conda` channel? PyPI? GitHub?

zacsimile:
I suppose `git` is the obvious choice since we trigger builds for pip and conda from here.

zacsimile:
OK, here's how I propose doing this, following https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#releases.

```
import requests
import PYME.version
[...]

PR #1042, opened 2021-05-19
update checking

David-Baddeley:
Preliminary work on #483, #1041

### TODOs

- put current_version.json on the website
- make it update when we make a release
- have an opt-out after being prompted once to update
- add a check_for_updates menu item or "an update is available" menu item.
- have a global opt-out?
- actually call this when launching programs.

### DIFF
--- a/PYME/DSView/modules/vis3D.py
+++ b/PYME/DSView/modules/vis3D.py
@@ -26,6 +26,7 @@ import matplotlib.cm
 
 from PYME.ui.AUIFrame import AUIFrame
 import wx.lib.agw.aui as aui
+from PYME.LMVis import view_clipping_pane
 
 class MeshViewFrame(AUIFrame):
     def __init__(self, *args, **kwargs):
@@ -38,12 +39,21 @@ class MeshViewFrame(AUIFrame):
 
         
         self.panesToMinimise = []
+        
         self.layerpanel = layer_panel.LayerPane(self, self.canvas, caption=None, add_button=False)
         self.layerpanel.SetSize(self.layerpanel.GetBestSize())
         pinfo = aui.AuiPaneInfo().Name("layerPanel").Right().Caption('Layers').CloseButton(
             False).MinimizeButton(True).MinimizeMode(
             aui.AUI_MINIMIZE_CAPT_SMART | aui.AUI_MINIMIZE_POS_RIGHT)#.CaptionVisible(False)
         self._mgr.AddPane(self.layerpanel, pinfo)
+
+        self.panesToMinimise.append(pinfo)
+        
+        self._clip_pane = view_clipping_pane.ViewClippingPanel(self, self.canvas)
+        pinfo = aui.AuiPaneInfo().Name("ClippingPanel").Right().Caption('Clipping').CloseButton(
+            False).MinimizeButton(True).MinimizeMode(
+            aui.AUI_MINIMIZE_CAPT_SMART | aui.AUI_MINIMIZE_POS_RIGHT)
+        self._mgr.AddPane(self._clip_pane, pinfo)
 
         self.panesToMinimise.append(pinfo)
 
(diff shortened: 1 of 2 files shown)


David-Baddeley:
image

image

(currently have `current_version.json` set to something that triggers the message for testing)

zacsimile:
Looks like a solid start! Thanks, @David-Baddeley.

David-Baddeley:
Have added automatic checking, and a config option to opt-out. @zacsimile, @csoeller - Might be good to get some feedback as to whether we are too pushy / not pushy enough. As it stands.

- we check for updates each time you launch PYMEVis or PYMEImage (implicitly also if you launch PYMEAcquire, but that doesn't prompt by default). This can be disabled by setting the `PYME.config` `check_for_updates` key to `False`.
- if (and only if) we have not already prompted for the new version, we display a dialog which says that there is a new version. This dialog will only ever get shown once across all PYME programs for each update.
- We have added an option to the `Help` menu to manually check for updates. This will always show the update dialog if an update is available.
- if an update is available, we also add a menu item to the `Help` menu, called `Update to xx.xx.xx`. For now the functionality is completely equivalent to `Check for updates`.

The open questions:
- are we being too conservative with only prompting once - should we offer 'ignore' (which delays for a day) and 'don't ask again' options instead.
- will the http requests slow down opening too much? We might open the programs a lot, and this could get annoying if, e.g. you are not on a network and you have to wait for the request to time out. Should we, e.g. limit our queries to once per day instead?
- Can we use this to do user-number estimation for grant reporting etc ... (do we need to send some form of unique ID with the request?, how can we do this without compromising user privacy?).

csoeller:
This looks very good @David-Baddeley, I gave it a try yesterday.

In terms of feedback on the general approach:

- I would be tempted to take a page out of the Fiji book, i.e. have a a button `Remind me later` that ensures another check is auto-performed >24h later or similar
- for now leave the config variable as the main way to completely disable automatic checks, in future possibly add an `Options` menu entry that allows GUI setting of config variables, including this one
- in the mid future add a button `update` to the message that invokes the functionality linked to  menu entry `Update to xx.xx.xx` once that actually can perform an update
- as you say an `don't ask again` button that disables checking until the next new version has come out and brings up a message box that says so (i.e. not `don't ask again ever` which would remain as the config variable functionality).
- I'd opt for the "once per day" checking limit in any case as this could otherwise become too annoying

I am sure some points can be debated.

The user statistics seem interesting but I would be paranoid that not accidentally some kind of privacy invasion can be construed by users (I know you mention that point). This could hurt uptake quite strongly.

zacsimile:
I agree with all of @csoeller's recommendations. 

One way of getting user statistics might be to just count the number of fetch requests for each version of PYME. Presumably unique users aren't downloading this too many times (obviously there will be some multiple downloads, but not too many version-to-version). Certainly unique installs are unlikely to upgrade more than once.

David-Baddeley:
We tried the fetch requests route for the CZI grant - with 4+ different ways people can get the package it was almost impossible to get meaningful data out of it. I really would like to have a more robust (i.e. actually usable) way of estimating user numbers, and think this is critical for ensuring that continued development is supported. 

If we do add a unique ID to the update checks, I'd want to make it a deliberate user choice (i.e. present a dialog before sending the info the first time). Tentative dialog text as follows:

----------------------

### Help us ensure that PYME development is supported

When applying for funding to support PYME development, it is very important to have a good estimate of user numbers. To facilitate this, we would like to send the following information with update requests:

**A unique installation identifier:** A random number generated the first time PYME checks for updates, this enables us to count the number of computers that are requesting updates. We do not link this identifier to any personal data, and by using a random number that is unique to PYME, rather than e.g. a MAC address, make it hard for a 3rd party to make such links. 

**The version of PYME you are currently using:** This gives us information to help decide when it is safe to e.g. deprecate old features.

----------------------
Pretty much every package that does automated update checking sends some user information - a lot of packages (Windows, MacOS, Chrome, Word, Anything from Adobe, etc ... are much, much more intrusive).

David-Baddeley:
Moved to #1050 to keep git history sane

PR #1047, opened 2021-05-23
yaml string quoting

### LINKED ISSUE #1045: [BUG] Problem saving file paths on Windows with `update_yaml_keys()`
**Describe the bug**
`update_yaml_keys()` writes "\\\\" as "\\", and then yaml.safe_load loads this as "\\". As such, file paths on Windows become a problem when we store them in a configuration YAML. If I save a path to my file as written, it will load improperly the next time (i.e. ``C:\\\\path\\\\to\\\\my\\\\file" saves as "C:\\path\\to\\my\\file", loads as "C:\\path\\to\\my\\file"). Workaround is to swap "\\" for "\\\\" when saving file path to YAML. Generally swapping "\\\\" for "\\\\\\\\" in `update_yaml_keys()` feels a bit unsafe, though. Is there a better way to work around this problem?

Edit: I had to redo the backslashes in this GitHub editor so you could actually see what I mean. Same problem I'm describing in our code is happening here. Hah.

David-Baddeley:
fixes #1045 - quote strings appropriately so that slashes are not treated as escape characters. Note we could also not quote at all (slightly unsafe if we begin with a special char)

### DIFF
--- a/PYME/config.py
+++ b/PYME/config.py
@@ -555,11 +555,16 @@ def update_yaml_keys(fn, d, create_backup=False):
     # Update the appropriate keys
     for k, v in d.items():
         x = re.search(r'^{}\s*:.*$'.format(k),data,flags=re.MULTILINE)
+        if isinstance(v, str):
+            v_s = '\'' + v + '\''
+        else:
+            v_s = json.dumps(v)
+        
         if x is None:
-            data += '\n{}: {}'.format(k, json.dumps(v))
+            data += '\n{}: {}'.format(k, v_s)
         else:
             data = re.sub(r'^{}\s*:.*$'.format(k),
-                          '{}: {}'.format(k,json.dumps(v)),data,flags=re.MULTILINE)
+                          '{}: {}'.format(k,v_s),data,flags=re.MULTILINE)
 
     # Update the yaml file
     with open(fn, 'w') as f:


zacsimile:
Ran into a problem with this.

```
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\sre_parse.py", line 1039, in parse_template
    this = chr(ESCAPES[this][1])
KeyError: '\\U'

  File "c:\users\scope_baby\code\pyme-oidic\oidic\pymeacquire_modules\oidic_calibration_frame.py", line 202, in on_save
    self.lc_ch_set.write_oidic_config()
  File "c:\users\scope_baby\code\pyme-oidic\oidic\pymeacquire_modules\liquid_crystal_calibration.py", line 255, in write_oidic_config
    update_yaml_keys(config_file, self._oidic_config)
  File "c:\users\scope_baby\code\python-microscopy\PYME\config.py", line 566, in update_yaml_keys
    data = re.sub(r'^{}\s*:.*$'.format(k),
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\re.py", line 210, in sub
    return _compile(pattern, flags).sub(repl, string, count)
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\re.py", line 327, in _subx
    template = _compile_repl(template, pattern)
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\re.py", line 318, in _compile_repl
    return sre_parse.parse_template(repl, pattern)
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\sre_parse.py", line 1042, in parse_template
    raise s.error('bad escape %s' % this, len(this))
re.error: bad escape \U at position 16
```

with this dictionary

```
{'lc_cal_file': 'C:\\Users\\scope_baby\\Downloads\\lc_config_data.xls', 'lc_cal_file_source': 'mbl', 'lc_voltage_dir0': 1.0, 'lc_voltage_dir1': 4.0, 'lc_voltage_dir0_zero': 2.7, 'lc_voltage_dir1_zero': 2.65, 'lc_bias': 0.15, 'wavelength': 530.0, 'shear_distance': 172.0, 'settling_time': 0.3}
```

David-Baddeley:
Seeing as single quoted strings are valid yaml which we might expect to encounter, it would be better to fix the underlying problem (lack of regex escaping in the substitution) rather than hacking around it in a way which is likely to break in the future. I suspect using `re.escape()` - i.e.  `re.sub(r'^{}\s*:.*$'.format(re.escape(k)), ...` should fix things.

zacsimile:
Tried it. The problem is with `v_s`, rather than `k`. Unfortunately, `re.escape()` also escapes `.` and other regex expressions. This makes things like filenames load in super weird (wrong). So far the `repr()` seems to be what works (#1051). It's worth noting that 
```
>>> repr('a')
"'a'"
```
which I think means we'll keep single-quoted YAML strings no problem if we go with this method.

David-Baddeley:
Ah, the traceback missed the value part of the sub call so I thought it was the key ... 

I'm slightly reluctant to rely on `repr()` as the behaviour feels a bit undocumented (it's realistically unlikely to change, but it's definitely not advertised as a string escaping function, and if it did change we'd have no legitimate complaint based on the docs). I think the underlying problem is that `re.sub` processes some backslashes but not the full escaping that you'd do with the search expression. It would seem that with `repr` you've hit on a solution that happens to put the extra escapes that `re.sub` needs (at least for the cases you've tested).

a still slightly gross option, but potentially less likely to trigger problems when it hits other special characters (e.g. &) is to prevent `re.sub` doing anything to the string by passing it a function that returns a string, rather than a string itself:
```python
In [35]: re.sub(r'^foo\s*:.*$',lambda m: r"foo: '\path\to\bar.txt'", s)
Out[35]: "foo: '\\path\\to\\bar.txt'"
```

zacsimile:
Another option is to just call `v_s = vs.replace('\\','\\\\')`.

PR #1048, opened 2021-05-24
xyztc dimension reordering, improved tiff import

### LINKED ISSUE #1043: [BUG] Import stack order message confusing
**Describe the bug**
Super minor. Not 100% sure this is a bug. I have a .tif that is XYIZTC. I wanted to move the I to C and drop the I entirely. I opened the TIF in Fiji, ran "Image>Stacks>Stack To Images". Then I went to "Image > Color > Merge Channels" and I merged the 6 images into 6 color channels. Opening it in a Jupyter notebook with `im = ImageStack(filename=fn)`, it correctly identifies 6 color channels and then loads the stack.

```
(6, 2048, 2048)
Detected OME TIFF
sh = {'Z': 1, 'T': 1, 'C': 6, 'Y': 2048, 'X': 2048}
axisOrder =  XYCZT
[2048, 2048, 1, 1, 6]
```

I can still access everything as `im.data_xyztc` as expected. So, clearly `[2048, 2048, 1, 1, 6]` is the `im.data_xyztc` order, but the way the info is presented makes the user (or at least me) assume this order corresponds to `XYCZT` (i.e. it looks like `T` has 6 channels rather than `C`).

P.S. I need a GUI enabled to call  `im = ImageStack(filename=fn)` in a Jupyter notebook, because that dialog asking for voxel size pops up. Is there a way to just pass the voxel size and bypass the dialog?

David-Baddeley:
closes #1043

### DIFF
--- a/PYME/Acquire/acquiremainframe.py
+++ b/PYME/Acquire/acquiremainframe.py
@@ -452,7 +452,7 @@ class PYMEMainFrame(AUIFrame):
     def OnFileOpenStack(self, event):
         #self.dv = dsviewer.DSViewFrame(self)
         #self.dv.Show()
-        im = dsviewer.ImageStack()
+        im = dsviewer.ImageStack(haveGUI=True)
         dvf = dsviewer.DSViewFrame(im, parent=self, size=(500, 500))
         dvf.SetSize((500,500))
         dvf.Show()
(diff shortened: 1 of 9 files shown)


David-Baddeley:
TODOs - check dimension product is constant

PR #1050, opened 2021-05-25
Update check

David-Baddeley:
Replaces #1042 which had a messed up commit history.

TODOs:

- decrease check frequency
- more automated updates (for conda installs)


zacsimile:
I think we should merge this and release another build of PYME. Our last build was in March.

David-Baddeley:
We're definitely overdue a release, not sure I'd consider this blocking though (I'd like to decrease the checking frequency before pushing this on everyone). 

I think the xyztc stuff has mostly shaken out, although suspect there are still one or two bugs in there. 
Can you think of anything else that we need to check / fix before releasing? Can we call #1035 fixed?

PR #1051, opened 2021-05-25
Use printable representation for strings in `update_yaml_keys()`

zacsimile:
This works for me, whereas the previous string fix threw an error (see #1047).


David-Baddeley:
see comment on #1047 - I think we can fix this properly.

zacsimile:
Replied to your comment in #1047. I think this might be the proper fix?

zacsimile:
Changed following @David-Baddeley's latest suggestion in #1047.

PR #1055, opened 2021-06-03
Import fixes

zacsimile:
- Get rid of extraneous alignment flags
- Open single-color matlab files properly

### DIFF
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -705,7 +705,7 @@ class Pipeline:
                 #old style matlab import
                 ds = tabular.MatfileSource(filename, kwargs['FieldNames'], kwargs['VarName'])
             else:
-                if 'Multichannel' in kwargs.keys() and kwargs.get('Multichannel'):
+                if kwargs.get('Multichannel', False):
                     ds = tabular.MatfileMultiColumnSource(filename)
                 else:
                     ds = tabular.MatfileColumnSource(filename)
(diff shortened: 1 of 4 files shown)


PYME/LMVis/pipeline.py
```diff
@@ -705,7 +705,7 @@ def _ds_from_file(self, filename, **kwargs):
                 #old style matlab import
                 ds = tabular.MatfileSource(filename, kwargs['FieldNames'], kwargs['VarName'])
             else:
-                if 'Multichannel' in kwargs.keys():
+                if 'Multichannel' in kwargs.keys() and kwargs.get('Multichannel'):
```
David-Baddeley: Could simplify a little as

```suggestion
                if kwargs.get('Multichannel', False):
```

Not really important here, but generally a better pattern as it should also be faster than `in keys()`, as it's faster to access with a missing key and handle the `KeyError` or to call `get()` with a default than to iterate keys - especially as dictionary sizes increase.

PYME/LMVis/pipeline.py
```diff
@@ -714,7 +714,8 @@ def _ds_from_file(self, filename, **kwargs):
                 field_names = kwargs.get('FieldNames', None)
                 if field_names:
                     if 'Multichannel' in kwargs.keys():
-                        field_names.append('probe')  # don't forget to copy this field over
+                        if kwargs.get('Multichannel'):
```
David-Baddeley: see above comment

PYME/LMVis/pipeline.py
zacsimile: I like it. Let's switch to this.

PR #1063, opened 2021-06-10
Add save mesh to voxel-based isosurface display window

zacsimile:


David-Baddeley:
seems like a reasonable start ...

Might want to revisit as we flesh out the mesh viewer menus etc ... to address the following:

- consistent menu positioning wrt PYMEVis (where I believe the export code is under `Mesh` menu) and potentially source image / mayavi meshing (which has STL export under the `3D` menu)
- reduce code duplication (this might be the 3rd place where mesh export functionality is exposed, with PYMEVis and the 3D menu in dh5view being the other two locations - PYMEVis and this should definitely be able to use the same code)
- consistent behaviour for multi-channel meshing (I think the `3D` menu STL exports all channels,  but suffixes the name with the channel if there is more than one channel - to me this feels like the right thing to do for meshes generated using the isosurface functionality in dh5view, but probably not in PYMVis).

PR #1066, opened 2021-06-15
onTarget tolerance adjustment in piezo init

zacsimile:
Make it possible to change the `onTarget` tolerance on piezo initialization. Needed for my less precise piezo.

Might be worth thinking about refactoring the piezo hardware code. I think this, along with many other parameters, belongs in the `BasePizeo` class, but perhaps I am wrong.

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_e816.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_e816.py
@@ -180,7 +180,7 @@ class piezo_e816(PiezoBase):
 
 import numpy as np
 class piezo_e816T(PiezoBase):
-    def __init__(self, portname='COM1', maxtravel=12.00, Osen=None, hasTrigger=False):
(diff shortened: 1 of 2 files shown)


David-Baddeley:
`target_tolerance` does not belong in the base class, as we would hope that piezo drivers provide there own, non-blocking, low-latency, hardware ,`on_target()` functionality (i.e. this is a hack to work around the non-ideal behaviour of a particular piezo controller). I could however see an argument for having a specific threaded piezo base class as I suspect the problems the threaded class works around will not be that unusual.

In terms of refactoring, we

zacsimile:
Hi @David-Baddeley. It looks like you were cut off mid-thought

David-Baddeley:
I think that belonged at the beginning of the previous sentence ...

zacsimile:
Ah, got it. Your comments make sense. Do you have an issue with how I've implemented this `targetTolerance` here? Or is this OK?

PR #1067, opened 2021-06-16
fix deconvolution for xyztc

David-Baddeley:
fixes deconvolution dimensionality issue, also makes drag and drop work in both fields in bakeshop


David-Baddeley:
image

Can now specify metadata defaults in bakeshop.

Issue #1068, opened 2021-06-17
[BUG] point set generation - image.openImages not in name space, missing import


csoeller:
**Describe the bug**

**To Reproduce**
1. Configure point generation with image source
2. Generate fluorophore positions and events

**Expected behavior**
Just do as told.

**Screenshots**
N/A

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
- python-microscopy=21.03.10
- python=3.7.6, platform=darwin
- numpy=1.19.2, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

**Error message**

```
 Error whilst running Extras>Synthetic Data>Generate fluorophore positions and events
==============================================
python-microscopy=21.03.10
python=3.7.6, platform=darwin
numpy=1.19.2, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py3/PYME/ui/progress.py", line 121, in func
    return fcn(*args, **kwargs)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py3/PYME/LMVis/Extras/pointSetGeneration.py", line 138, in OnGenPoints
    self.OnGenEvents(None)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py3/PYME/LMVis/Extras/pointSetGeneration.py", line 170, in OnGenEvents
    pipeline.imageBounds = image.openImages[self.source.image].imgBounds
NameError: name 'image' is not defined

```

David-Baddeley:
Can we call this fixed?

csoeller:
When you put back the `ed._values_changed()` check back in you check for a `ValueError` while the actual error raised is a `TypeError`. Therefore the current committed and merged code still does not work. Can you change the code as shown below? I believe after that we can close the issue.

In addition, I run into OpenGL errors when points are generated in an empty `visGUI`. This is not an issue if you generate the points in a `visGUI` instance that already shows events prior to generating the synthetic data. I probably better raise a separate issue for that?

<img width="1364" alt="Screenshot 2021-06-22 at 06 59 49" src="https://user-images.githubusercontent.com/3750860/122871953-a7f57a00-d327-11eb-902c-9b85b84761e6.png">

David-Baddeley:
[...]

PR #1069, opened 2021-06-17
fix image.OpenImages bug; fix image source config; add metadata

csoeller:
Addresses issue #1068.

bugfix

**Proposed changes:**
- fixes bug as reported in #1068
- also fixes issues with pointSetGeneration config
- adds metadata of pointSetGeneration parameters

**Checklist:**

- [ x] Tested with numpy=1.19
- [x ] Tested on python 3.7

### DIFF
--- a/PYME/LMVis/Extras/pointSetGeneration.py
+++ b/PYME/LMVis/Extras/pointSetGeneration.py
@@ -38,7 +38,7 @@ class Generator(HasTraits):
     meanEventNumber = Float(2)
     scaleFactor = Float(2) # Note: we don't expose the scale factor in the view
     meanTime= Float(2000)
-    mode = Enum(['STORM','PAINT'])
+    mode = Enum(['STORM','DNA PAINT'])
 
     sources = List([WormlikeSource(), ImageSource(), FileSource()])
 
@@ -134,7 +134,15 @@ There should be no need to modify this from the default and it is accordingly no
         self.edit_traits()
 
     def OnGenPoints(self, event):
-        self.xp, self.yp, self.zp = self.source.getPoints()
+        import wx
+        try:
+            self.xp, self.yp, self.zp = self.source.getPoints()
+        except KeyError:
+            dlg = wx.MessageDialog(self.visFr, "Could not access image, please configure image source",
+                                   'Warning!', wx.OK | wx.ICON_WARNING)
+            dlg.ShowModal()
+            dlg.Destroy()
+            return
         self.OnGenEvents(None)
 
     def OnGenEvents(self, event):
@@ -142,6 +150,7 @@ There should be no need to modify this from the default and it is accordingly no
         #from PYME.Acquire.Hardware.Simulator import wormlike2
         from PYME.IO import tabular
         from PYME.IO.image import ImageBounds
+        from PYME.IO.image import openImages
         # import pylab
         import matplotlib.pyplot as plt
         
@@ -167,7 +176,15 @@ There should be no need to modify this from the default and it is accordingly no
         ds = tabular.MappingFilter(tabular.FitResultsSource(res))
         
         if isinstance(self.source, ImageSource):
-            pipeline.imageBounds = image.openImages[self.source.image].imgBounds
+            try:
+                pipeline.imageBounds = openImages[self.source.image].imgBounds
+            except KeyError:
+                import wx
+                dlg = wx.MessageDialog(self.visFr, "Could not access image, please configure image source",
+                                       'Warning!', wx.OK | wx.ICON_WARNING)
+                dlg.ShowModal()
+                dlg.Destroy()
+
         else:
             pipeline.imageBounds = ImageBounds.estimateFromSource(ds)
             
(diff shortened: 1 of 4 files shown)


PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -134,14 +134,23 @@ def OnConfigure(self, event):
         self.edit_traits()
 
     def OnGenPoints(self, event):
-        self.xp, self.yp, self.zp = self.source.getPoints()
+        import wx
+        try:
```
csoeller: these checks are needed as otherwise one is left in an unworkable state and needs to exit the app as configuration cannot be fixed; therefore there are similar changes in all other places

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -180,6 +197,16 @@ def OnGenEvents(self, event):
         pipeline.mdh['Camera.TrueEMGain'] = 1
         pipeline.mdh['Camera.CycleTime'] = 1
         pipeline.mdh['voxelsize.x'] = .110
+        # some info about the parameters
```
csoeller, line 187: to be really useful we need to preserve the parameters events were generated with so that upon saving as hdf the info is included as metadata

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -134,14 +134,23 @@ def OnConfigure(self, event):
         self.edit_traits()
 
     def OnGenPoints(self, event):
-        self.xp, self.yp, self.zp = self.source.getPoints()
+        import wx
+        try:
+            self.xp, self.yp, self.zp = self.source.getPoints()
+        except KeyError:
+            dlg = wx.MessageDialog(self.visFr, "Could not access image, please configure image source",
```
David-Baddeley: This error messages is incorrect (and potentially confusing) for 2/3 cases - i.e. when the source is not an image.

PYME/LMVis/Extras/pointSetGeneration.py
```diff
@@ -38,7 +38,7 @@ class Generator(HasTraits):
     meanEventNumber = Float(2)
     scaleFactor = Float(2) # Note: we don't expose the scale factor in the view
     meanTime= Float(2000)
-    mode = Enum(['STORM','PAINT'])
+    mode = Enum(['STORM','DNA PAINT'])
```
David-Baddeley:
```suggestion
    mode = Enum(['STORM','PAINT'])
```

The DNA is redundant and  overly specific as there are multiple forms of non-DNA PAINT (e.g. with WGA)

PYME/LMVis/Extras/pointSetGeneration.py
csoeller: While possibly obscure the KeyError is only ever raised with the image source. Trapping the error earlier in the chain does not work. What is the tidy solution, e.g. create a new 'ImageSourceKeyError' and raise and check for that?

David-Baddeley: There's a chance that it doesn't practically matter if the `ImageSource` is the only one which could throw a `KeyError`

PYME/LMVis/Extras/pointSetGeneration.py
csoeller: I was under the impression that not all PAINT forms are described by reversible binding/unbinding acording to first order kinetics. Do they all follow the same statistics?

PYME/LMVis/Extras/pointSetGeneration.py
csoeller: That is the case in my reading of the code and thus a `KeyError` directly implies the `ImageSource`.

David-Baddeley:
Many thanks! Have made a couple of specific comments.

On a more general level, I'm a little nervous about how the exceptions are being handled - it feels like it's a bit too special cased and requires too much knowledge of the `ImageSource` in classes which aren't the `ImageSource`. It feels a little fragile. I'd prefer to handle anything we can in the `ImageSource`, but then let anything unhandled propagate so that we got full tracebacks. This might require working out *why* an exception leaves things in an unrecoverable state and fixing that, rather than just catching individual exceptions (which sounds like it'll be a game of whack-a-mole).

csoeller:
One issue with the `KeyError` related issues is the behaviour of the gui selector that I found on the mac was confusing if you had actually selected the (only available) image. I have only tried this on the mac and there I have repeatedly come across it to the level that it made using this unworkable without the current `KeyError` checks. You may be right it would be better to fix the selection gui. Can you reproduce the selection issues I describe? I can attempt to record a screencast.

PYME/LMVis/Extras/pointSetGeneration.py
David-Baddeley: In general I think they are all *supposed* to be reversible binding/unbinding with 1st order kinetics (or a reasonable approximation thereof). I suspect some have quite a slow off-rate so in practice you bleach before the label comes off, but this would still look like 1st order kinetics from a simulation perspective.

David-Baddeley:
hmm, fixing a py3 list vs enum error in `WRDictEnum` seems to make it recoverable ...

csoeller:
Fixing the py3 list vs enum helps but reverting the exception catch that I had at the bottom of `pointsets.py` has again left the implementation in an unrecoverable state once you made the config error. To reproduce, open raw image, set image source settings and DO NOT select the only open image. This raises the user error fine but triggers the error with

```python
    if ed:
            ed._values_changed()
```

on attempt to configure properly. So this needs some kind of fixing, perhaps @David-Baddeley, could you suggest what may be the best way?

**Screenshots to demonstrate error**

First, the missing image selection is picked up ok:

Screenshot 2021-06-18 at 11 06 44

But then on trying to configure properly we can't make it past this error:

Screenshot 2021-06-18 at 11 07 13

David-Baddeley:
Hmm, I was sure I'd tried that. Will attempt to reproduce, although I
seemed to have the only open image pre-selected (Ie no way of not choosing
it). Think I still clicked on the drop down however. Will try not clicking.

Putting the type error check back in wouldn't be the end of the world, I
just thought it was unnecessary after the list fix.

csoeller:
Yes, please try to reproduce. In my hands it is not an artificial situation that I created for a test but happens in real use more often than not. There is no clear visual feedback if you have selected something or not, it drove me a little mad when I first tried to use the functionality and always had to resort to restarting `visGUI` when Configure shut me out with the `NoneType` error as shown above. Maybe something to do with my wx version (4.0.4) on Catalina, not sure...

David-Baddeley:
I've put the check back in as it seems the safest option

PR #1073, opened 2021-06-23
Concatenated data source

zacsimile:
Concatenate multiple files of the same size along Z, T, C.

First pass, looking for feedback.


zacsimile:
Not sure how to handle the GUI portion of this. I'm inclined to add a `File> Import Image Sequence` option to `pymeimage`, but `pymeimage` won't open without an image selected, so I'm not quite sure what the move is. Add a `-s` command line flag that prompts `pymeimage` to glob files based on a pattern assumed from the initial file you pass on the command line? E.g. `pymeimage -s my_data_0000.h5` opens `my_data_*` in the folder?

David-Baddeley:
I suspect things are not going to work properly if datasources are actually more than 3D, at least not without some really complex `getSlice()` logic. For your use case it's definitely not necessary .... 

My inclination (at least in the first instance) would be to not care about  (i.e. ignore) input dimensionality, and not promote the input datasources to xyztc. Just assume that the input datasources have `getSlice()` and `getNumSlices()` and treat them as "flat" frame sources. You can then set the dimension stuff on the final datasource. I think the simplicity of doing it like this outweighs what you give up.

A couple of other thoughts:
- you will currently fail (potentially in a non-trivial way) if datasources do not have exactly the same number of frames 
- actually, your assert might take care of this, but for the 'opening labview data spooled to dcimg chunks' use case, I don't think it's safe to assume that all chunks will be the same size (I'd be particularly suspicious of the last chunk in the series, which could be any length from 1 frame to the full chunk size.

zacsimile:
@David-Baddeley is there a better way to deal with file opening here? I couldn't think of a faster way to do this while still keeping the dynamic size guessing.

zacsimile:
@David-Baddeley, this is ready for review!

PYME/DSView/OpenSequenceDialog.py
```diff
@@ -0,0 +1,82 @@
...
+
+        self.file_panel.on_update_file_list.connect(self.on_update)
+
+    def on_update(self, **kwargs):
+        self._datasources = []
+        self.btn_open.Enable()
+        for fn in self.file_panel.filenames:
+            image = ImageStack(filename=fn)
+            self._datasources.append(image.data_xyztc)
+        try:
```
David-Baddeley, line 69:
```suggestion
        
        try:
```

Separate from for loop for readabliltiy

PYME/IO/image.py
```diff
@@ -643,6 +643,65 @@ def _load_supertile(self, filename):
...
+            try:
+                c = int(urllib.parse.parse_qs(query)['c'][0])
+            except KeyError:
+                c = 1
+
+            if ((z==1) and (t==1) and (c==1)):
+                #we have a query string but no dimensions specified
+                z, t, c = None, None, None
+
+        if filename.find(',') == -1:
```
David-Baddeley, line 684: might need to revisit commas as a separator, but happy to deal with this when the time comes.

Issue #1054, opened 2021-06-03
[ENHANCEMENT] Make it easier to move through color channels in `pymeimage`


zacsimile:
I have some 6-channel images. In order to view each channel individually, I have to expand the window to maximum size and uncheck each channel I don't want to see. If I have more than 6 channels, I think I would be out of luck.

Two potential solutions come to mind:

- Make the Display Settings scrollable
- Allow attachment of a PlayPanel for color channels (maybe under modules? loaded by default?)

![image](https://user-images.githubusercontent.com/1263313/120668146-2e6f1800-c45c-11eb-90eb-45fb3fa65aaa.png)

David-Baddeley:
Had a quick go at making it scroll when I did the import fixes and redimension dialog as I ran into similar issues with one of your example OIDIC images. Couldn't get vertical scrolling to play nice with the horizontal sizing in the ammount of time I had. Most of the scrolling code is still in there, albeit commented out until it could be revisited.

Scrolling through colours would require quite a different view philosophy (akin to imagej multichannel vs composite) and might be a bit tricky to implement and/or require substantial viewer mods.

I'll throw another couple of options out there:
- a "concertina" as used for layers in pymevis
- making the channel controls tabbed when the number of channels exceeds the vertical space
- putting the controls into a wrap sizer so you could resize horizontally to get them displayed in two columns

David-Baddeley:
I think we can close this now ...

PR #1074, opened 2021-06-23
Convert channel display to concertina when `DisplayOptionsPanel` in a vertical layout

zacsimile:
Addresses issue #1054.

Enhancement

**Proposed changes:**
- Display channels as concertina of foldable panels when `DisplayOptionsPanel` has a vertical layout

### DIFF
--- a/PYME/DSView/DisplayOptionsPanel.py
+++ b/PYME/DSView/DisplayOptionsPanel.py
@@ -26,6 +26,7 @@ import wx.lib.agw.aui as aui
 # import pylab
 import wx.lib.scrolledpanel as scrolled
 import PYME.ui.manualFoldPanel as afp
+import PYME.ui.layerFoldPanel as lfp
 
 from matplotlib import cm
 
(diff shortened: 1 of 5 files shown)


David-Baddeley:
Will have to pull and have a look at what the GUI looks like. Due to the number of different places this is used I am a little bit cautious about modifying the existing panel rather than developing a new one alongside and substituting in on a case-by case basis. I don't think this caution is necessarily unsurmountable, but there might be a slightly higher threshold than normal to make sure we have this right.

David-Baddeley:
Some visual and sizing improvements to a) bring it closer to the layers GUI and b) make sure that the channels do sensible things when there are too many to show and when windows are resized.

David-Baddeley:
image
image

zacsimile:
Looks good! Thanks, @David-Baddeley!

David-Baddeley:
There is, however, a super odd glitch with the threshold controls showing up in the wrong place until you have expanded the settings panel.
image

Curiously, this doesn't seem to happen to a new image within dh5view

zacsimile:
That is odd. I can't replicate this on my machine. 

```
wxpython                  4.0.4            py37hebc47a7_0  
traits                    6.1.1            py37haf1e3a3_0  
traitsui                  7.1.0                      py_0 
```

Can you give me a step-by-step on how to duplicate this issue?

David-Baddeley:
Open an image on the command line (`dh5view image_name`), or use test data (`dh5view --test3d`). Glitch shows up until you expand "display settings", but goes away when you minimise it again. Does not show up when you create a new image using a command on the processing menu - e.g. processing->Gaussian Filter, or ->Redimension.

Py 3.7, wx 4.04, OSX Big Sur

zacsimile:
Can't replicate it on OSX Catalina (my python/wx are the same as yours). Is it possible this is a Big Sur problem?

David-Baddeley:
Merging, figure we can then commit a fix when we find it

David-Baddeley:
Also checked as working fine fine on MSW

PR #1078, opened 2021-06-24
Add user alert if no update is available

zacsimile:
Let the user know the update checking ran, and found nothing.

### DIFF
--- a/PYME/misc/check_for_updates.py
+++ b/PYME/misc/check_for_updates.py
@@ -99,6 +99,9 @@ def check_for_updates(gui=True, force=False):
                 
         else:
             update_available = False
+            
+            if gui:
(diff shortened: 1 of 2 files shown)


PYME/misc/check_for_updates.py
```diff
@@ -99,6 +99,9 @@ def check_for_updates(gui=True, force=False):
                 
         else:
             update_available = False
+            
+            if gui:
+                gui_prompt_no_update()
```
David-Baddeley: Calling it here will pop up a "you are running the latest version" popup for all the automatic checks as well - this is not what we want - we only want to pop up a "no updates" dialog if you specifically run help->check for updates.

PR #1079, opened 2021-07-03
Add another ORCA camera to noiseProperties

zacsimile:
Hooked up another ORCA.

Should we make `noiseProperties` a passable parameter during camera initialization? Then we could put this kind of information in the init scripts, which seems more sensible.

### DIFF
--- a/PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
+++ b/PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
@@ -86,6 +86,14 @@ noiseProperties = {
         'DefaultEMGain': 1,
         'SaturationThreshold': (2**16 - 1)
         },
+'301777' : {
+        'ReadNoise': 1.63,
+        'ElectronsPerCount': 0.47,
+        'NGainStages': 0,
+        'ADOffset': 100,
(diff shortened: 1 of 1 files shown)


David-Baddeley:
doing something saner with noise properties has been on the TODO list for ages, and there is a draft spec for what this should look like (see discussion on https://github.com/python-microscopy/python-microscopy/pull/144). I feel this has come up more recently as well.

reproducing from #144:

My proposal:

-   we modify the base camera class to specify that all cameras should implement a `.noise_properties` property that returns a dictionary of the **current** properties
- we modify the `__init__()` signature of the camera class to accept a `noise_info` parameter which can either be a path to a .yaml file (preferred), or a dictionary. The resulting noise_info dictionary **must** be keyed on the camera serial number (this should help avoid problems when swapping cameras), but what sits under that is up to the individual camera.
- we make the above parameter optional, with cameras using either a `PYME.config` option to specify the location of the .yaml file, or a fixed location within the pyme config tree by default (this allows, amongst other things, a smooth upgrade path by transferring the current hard-coded properties to a .yaml that we distribute with PYME).
- The use of a path to a .yaml allows a master file to be held on shared storage, if desired, making camera swaps etc ... easy.

Issue #1086, opened 2021-07-10
[BUG] missing tupel conversion in shape property in PYME.IO.dataWrap.ListWrapper botches shape comparison


csoeller:
**Describe the bug**
As in the bug title, error example see below.

**To Reproduce**
Replicate steps described in #1082, get the following bug:

```
Traceback
=========
Traceback (most recent call last):
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py37/PYME/ui/progress.py", line 153, in func
    return fcn(*args, **kwargs)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py37/PYME/DSView/modules/recipes.py", line 98, in RunCurrentRecipe
    self.outp = self.activeRecipe.execute(input=self.image)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py37/PYME/recipes/recipe.py", line 203, in execute
    m.execute(self.namespace)
  File "/Users/csoe002/Documents/src/PYME-src/pyme-py37/PYME/recipes/measurement.py", line 750, in execute
    assert (labels.data_xyztc.shape == intensity.data_xyztc.shape), 'Measure2D labels and intensity must be the same shape'
AssertionError: Measure2D labels and intensity must be the same shape
```

**Expected behavior**
The shapes are in fact the same but the assertion compares a tuple and a `<class 'PYME.IO.DataSources.BaseDataSource.DefaultList'>` as the `labels` shape returns such an instance due to the missing tuple conversion.

Expected no error and can be achieved by adding tuple conversion, will supply RT.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
- python-microscopy=21.07.07
- python=3.7.6, platform=darwin
- numpy=1.19.2, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

**Additional context**

David-Baddeley:
Can hopefully close this now (with merge of #1087)

PR #1087, opened 2021-07-10
ListWrapper - add missing tuple conversion

csoeller:
Addresses issue #1086.

Bugfix.

**Proposed changes:**
Just add missing tuple conversion in shape property.

**Checklist:**

- [ x] Tested with numpy=1.14


David-Baddeley:
Thanks!

Issue #1089, opened 2021-07-16
[BUG] pointsets FileSource issues: fileselector in save mode and np.load implementation issues for py3


csoeller:
**Describe the bug**
When configuring the `FileSource` the file dialog that comes up is a save dialog where it should be an open dialog. Also, loading 3 numpy arrays via a single `np.load` call seems not compatible with py3.

**To Reproduce**
1. in `LMVis` Go to `Extras>Synthetic Data>Configure`
2. Try to configure FileSource
3. Try to load from FileSource
4. points 2 and 3 have issues

   - Issue in point 2: a save dialog comes up that asks if a file should be overwritten or not.
   - Issue in point 3: get an error as np.load semantics have changed in py3

**Suggestions how to fix**

 - for point 2 add options to `File` trait, using the `exists=True` option e.g.

```patch
--- a/PYME/simulation/pointsets.py
+++ b/PYME/simulation/pointsets.py
@@ -30,13 +30,15 @@ class WormlikeSource(PointSource):
         mdh['GeneratedPoints.Source.PersistLength'] = self.persistLength
         
 class FileSource(PointSource):
-    file = File()
+    file = File(filter=['*.csv'],exists=True)
```

- for point 3 change to a more portable/versatile format, I suggest csv:

```patch
     def getPoints(self):
-        import numpy as np
-        return np.load(self.file)
+        # switched to csv based storage format
+        import pandas as pd
+        df = pd.read_csv(self.file)
+        return (df['x'].values,df['y'].values,df['z'].values)
 
     def genMetaData(self, mdh):
         mdh['GeneratedPoints.Source.Type'] = 'File'
```

I will prepare a PR with these points, the details of the coordinate format can be discussed. I will also suggest a `Extras>Synthetic Data>Save fluorophore positions` menu entry as saving generated positions can be very useful for the analysis of simulated data and allows reloading of the same positions at a later stage.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
- python-microscopy=21.07.07
- python=3.7.6, platform=darwin
- numpy=1.19.2, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

David-Baddeley:
Guessing we should be able to close with #1090

PR #1090, opened 2021-07-16
point set generation: fix dialog and change filesource format, allow saving coordinates to file

csoeller:
Addresses issue #1089.

bugfix + saving coordinates enhancement

**Proposed changes:**

- change filesource format to csv; numpy load/save formats are difficult to preview and generate using third party software while very easy with csv
- change traits `File` options to default to `File>Open` type dialog
- add `Extras>Synthetic Data>Save fluorophore positions` functionality to save generated positions in `FileSource` readable format

**Remaining issue with file dialog**

With the `exists=True` flag the file selection remains red and the ok button can't be selected in my wx version **just the first time around**, like so

Screenshot 2021-07-16 at 16 50 09

This is then worked around by pressing `Cancel` and then when you configure `Source` again everything is fine, as below and you can select what you like.

Screenshot 2021-07-16 at 16 50 26

Possibly there is another flag that makes this behave right away or this is an actual bug. But I don't feel comfortable enough with traits to come up with a better solution, e.g. a custom editor etc, was happy enough that I managed to google the flags which were hard enough to find any docs on.

**Checklist:**

- [x ] Tested with numpy=1.14
- [x ] Tested on python 3.7
- [x] Tested with wx=4.x [if UI code]

**Discussion**

Happy to discuss, bottom line is we need a fix/improvement for the `FileSource` for our use cases. Saving fluorophore positions is high on our list of needed features, so ideally we can incorporate that too. The file format(s) can clearly be dicussed, CSV seems like a good compromise that is flexible. This does not have to use the pandas interface if that were an issue.

### DIFF
--- a/PYME/LMVis/Extras/pointSetGeneration.py
+++ b/PYME/LMVis/Extras/pointSetGeneration.py
@@ -126,6 +126,7 @@ There should be no need to modify this from the default and it is accordingly no
             visFr.AddMenuItem('Extras>Synthetic Data', "Configure", self.OnConfigure)
             visFr.AddMenuItem('Extras>Synthetic Data', 'Generate fluorophore positions and events', self.OnGenPoints)
             visFr.AddMenuItem('Extras>Synthetic Data', 'Generate events', self.OnGenEvents)
+            visFr.AddMenuItem('Extras>Synthetic Data', 'Save fluorophore positions', self.OnSavePoints)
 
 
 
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Trying to decide how much I care (potentially not that much), but the change from .npy to .csv is quite a big, backwards incompatible, change. 

The principle advantage to saving as numpy is that it's significantly faster (and smaller) for large datasets when compared with CSV.  Note, the semantics on `np.load()` have not changed from py2 to py3 - we always expected a single 3xN array.

A better approach would have probably been to leave the file source (optionally renaming to NpySource) and added a CSV source, although I'm not sure how much to care in this case as I don't think it's widely used (can't be 100% sure though). Suspect this largely goes away when the simulation gets refactored as separate recipe modules to allow proper automation. This would let us piggy-back on standard recipe HDF and csv IO, both for opening and saving.  

Although not critical, and not comprehensively implemented, I try to avoid pandas where simple alternatives like `np.loadtxt` exist, both to reduce the dependency footprint, and as the numpy version (in this case) is quite a lot faster.

PR #1091, opened 2021-07-22
batched spharm density

David-Baddeley:
Addresses (or hopefully addresses) memory usage in uniform density estimates from spherical harmonic shells.

**NOTE:** as yet incomplete, as the recipe module is unmodified, posting for feedback from @barentine 

@barentine : This does not fully replace your implementation as it doesn't estimate principle axes, volume, etc ... I'd like to get a feeling for how necessary these are before adding them back in or not and altering the recipe module. Volume would be fairly simple, but I'd be reluctant to add back in the SVD stuff unless it was absolutely necessary. I suspect this was what was causing a lot of the issues in the original code - it uses a lot of memory, and has really bad computational complexity (𝑂(𝑚2𝑛+𝑚𝑛2+𝑛3)  ).

### DIFF
--- a/PYME/recipes/surface_fitting.py
+++ b/PYME/recipes/surface_fitting.py
@@ -374,8 +374,9 @@ class SHShellRadiusDensityEstimate(ModuleBase):
     input_shell = Input('harmonic_shell')
 
     r_bin_spacing = Float(0.05)
-    sampling_nm = ListFloat([75, 75, 75])
-    jitter_iterations = Int(3)
+    sampling_nm = ListFloat([75, 75, 75]) # TODO - convert this to a single float - we only use the first entry
+    #jitter_iterations = Int(3)
+    batch_size = Int(100000)
 
     output = Output('r_uniform_kde')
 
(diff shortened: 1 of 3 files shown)


PYME/recipes/surface_fitting.py
```diff
@@ -388,6 +389,33 @@ def execute(self, namespace):
...
+        })
+        
+        res.mdh = MetaDataHandler.DictMDHandler(getattr(shell, 'mdh', None))
+
+        # FIXME - this is really gross - we must not pass data / results in metadata
+        # Leaving here for now, for backwards compatibility, but needs to be fixed / removed somewhat urgently
+        # if we record sampling in the metadata, as we should, the volume calculation could easily be done in the consuming module)      
+        res.mdh['SHShellRadiusDensityEstimate.Volume'] = float(volume)
+
+        # record module parameters  - FIXME this should be under the 'Analysis' top-level metadata key.
```
barentine, line 411: Is 'Analysis' no longer reserved for localization analysis?

David-Baddeley:
current test failure is intermittent (due to test tolerance being too close to statistical variation in results)

David-Baddeley:
Note, also occurs exclusively in the inner radial bin (which is the smallest, and most error prone)

Issue #1098, opened 2021-08-10
[BUG] Localization analysis doesn't run at single-frame image


Yujin-Bao:
**Describe the bug**
I use `LatGaussFitFR` to analyse localization data from a one-frame image of 1um beads. An error shows in `pymeclusterofone` terminal.

**To Reproduce**
1. For `point finding`, I set the threshold to 5 and run `test this frame` to make sure for each bead, there is only one candidate region to be fitted, labeled by the small green square.
2. This is only a one-frame image, so set `start at` as 0, and cancel `subtract background in fit`, then click `go`
![image](https://user-images.githubusercontent.com/80548133/128936074-ab0f2a73-57f2-4f6c-8cc4-896d0cf5f1c1.png)
3. See error in `pymeclusterofone` terminal: `tables.exceptions.NoSuchNodeError: group / does not have a child named Events`.
![image](https://user-images.githubusercontent.com/80548133/128937648-4cdc7944-dae8-4f59-8726-33ddefced6b6.png)

**Expected behavior**
I expect to see 2D Gaussian fittings in each small green square as well as their acceptance and rejection.

David-Baddeley:
Hmm, there looks to be multiple things going on here, and the tracebacks are not particularly useful ...

- What format is the input data? The tracebacks would suggest that it's .h5 formatted, but lacks an events table. (This might not actually be a critical error - it might get handled fine in normal operation, but we might be seeing the error because something else fails when we try and handle it)
- The second error (NoneType is not iterable) is probably more important. The only thing that I can think of is that this might be caused by an empty array (i.e. no molecules detected and fit).

In line with this error, it's worth pointing out that to disable background subtraction, you need to both uncheck the `subtract background in fit` box **and** set the `Background:` range to `0:0`. With the settings as shown, you will be attempting to use background subtraction for object detection, but not for fitting. Revamping the GUI here to show separate "subtract background for detection" and "subtract background in fit" options (which would be much more understandable) has been on the  TODO list for a while now, but as it stands, having anything other than `0:0` in the background range implies subtract background for detection. There might also be subtle differences in how this works when testing and when analyising.

David-Baddeley:
PS - if you could share the data you're trying to fit that might help further debugging.

Yujin-Bao:
Thank you for the reply, David! 
- The format of the input data is `.tif` file, but I did create a `.md` metadata file as it suggested in the documetation. However, I only set the minimum set of metadata parameters.
[...]

PR #1104, opened 2021-08-20
Make check_name work for mac and windows

zacsimile:
Addresses issue #1098.

Yes

**Proposed changes:**
- Adjust regex to work for \\ as well as /
- Eliminate `C:\` and the like from `check_name` on Windows, similar to how we remove `PYME-CLUSTER://` from `check_uri`

Tested on Mac OS Catalina and Win 10.

### DIFF
--- a/PYME/IO/unifiedIO.py
+++ b/PYME/IO/unifiedIO.py
@@ -12,7 +12,7 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-alpha_regex = re.compile(r'^[\w/\.\-]+$')
+alpha_regex = re.compile(r'^[\w/(\\)\.\-]+$')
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Thanks!

Ultimately having backslashes in the name should be an error / cause failure on mac/nix and **and** for cluster urls on windows, so we need to switch the regex accordingly. Will see how much time I have today, might otherwise merge this in the interim.

zacsimile:
I wondered about that. One quick a dirty way to do it would be to check if the system is Windows and use this regex if it is and the original regex if it isn't, but there are cleaner ways...

David-Baddeley:
Halfway through a modified version ...

PR #1109, opened 2021-08-28
Make distance to mesh SDF accurate on the interior, smooth

zacsimile:
A few adjustments to our distance to mesh SDF.

- Convert SDF from intersection of point-plane SDFs to union of triangle SDFs
- Calculate the SDF as if the interior is outside of the mesh, to ensure the SDF inside the mesh has no discontinuities. This method does allow for SDF discontinuities outside the mesh, but these should be less of a problem.
- Add a smoothing factor to eliminate discontinuities in the derivatives of the SDF


David-Baddeley:
Will merge, we should revisit re speed so as to only consider nearest triangles

zacsimile:
Thanks! Distance to mesh does consider only the nearest 5 triangles. Is there another spot I should modify to consider only nearest triangles?

Issue #1114, opened 2021-09-14
[BUG] [21.07.07] stackSettings


barentine:
**Describe the bug**
If you install 21.07.07 on win10 from https://python-microscopy.org/downloads/automated_builds/, run PYMEAcquire (from anaconda prompt, because for some reason we dropped the PYMEAcquire (simulator) shortcut)

```
Traceback (most recent call last):
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\Acquire\acquiremainframe.py", line 238, in doPostInit
    cm.run(self, self.scope)
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\Acquire\ExecTools.py", line 205, in run
    self.codeObj(parent, scope)
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\Acquire\Scripts\init.py", line 120, in fake_dmd
    from PYMEnf.Hardware import FakeDMD, DMDGui
ModuleNotFoundError: No module named 'PYMEnf'
DEBUG:PYME.Acquire.acquiremainframe:Loading GUI component for Laser controls
DEBUG:PYME.Acquire.acquiremainframe:Loading GUI component for Focus Keys
DEBUG:PYME.Acquire.acquiremainframe:Loading GUI component for Action manager
0
DEBUG:PYME.Acquire.acquiremainframe:Loading GUI component for Tiling
DEBUG:PYME.Acquire.acquiremainframe:Run all custom GUI init tasks
0
Traceback (most recent call last):
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\ui\mytimer.py", line 42, in Notify
    a()
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\Acquire\acquiremainframe.py", line 170, in _check_init_done
    self.doPostInit()
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\Acquire\acquiremainframe.py", line 279, in doPostInit
    self.seq_d = seqdialog.seqPanel(self, self.scope)
  File "C:\Users\aesb\AppData\Local\python-microscopy\lib\site-packages\PYME\Acquire\ui\seqdialog.py", line 214, in __init__
    self.stackSettings = self.scope.stackSettings
AttributeError: 'microscope' object has no attribute 'stackSettings'
```

[...]

PR #1116, opened 2021-09-15
stack_settings_tweak

David-Baddeley:
ensure pymeacquire loads when webu is not present


David-Baddeley:
should hopefully resolve #1114

PR #1124, opened 2021-09-27
spherical harmonics RAM improvements

barentine:
Addresses issue expensive spherical harmonics + radial density estimate recipes stalling out when batched by the ~thousands. See alternative, #1091, which addresses the density estimate only. I was working through TODO's/documenting #1091, got tripped up thinking we might be missing a normalization and after awhile of looking at that decided I would have another look at the original's RAM usage and found that it wasn't the (only) culprit.

On my laptop, using atmonitor to look at PYMEVis ram usage in time (not super high time resolution, so these measurements are all ~), I found we were maybe a bit too focused on the radial density estimate. I hit about 1.4 GB RAM running the spharm followed by radial density estimate for a series with 50538 localizations. Removing the density estimate and just running the spherical harmonic shell fit, I still hit maybe 1.1 GB. 

I found that my series are float32 in xy, but double in z (probably my fault in the multiview recipes somewhere). This PR forces float32 in a couple key places in the spherical harmonics code, and keeps RAM at about 800 GB now for the full recipe (note, 275 MB just opening the thing in PYMEVis). This may be a big enough difference to keep us from stalling out. PR'ing to share, but will be giving it a try.

<details> <summary> recipe </summary>

```
- localisations.DensityMapping:
    colours:
    - chan0
    inputLocalizations: _mapped
    jitterScale: 30.0
    outputImage: jt2d
    pixelSize: 40.0
    renderingModule: Gaussian
- processing.Threshold:
    dimensionality: XYZ
    inputName: smoothed
    method: otsu
    outputName: thresholded
- processing.Label:
    inputName: thresholded
    minRegionPixels: 1
    outputName: labeled
- processing.SelectLargestLabel:
    inputName: labeled
    outputName: nucleus
- processing.BinaryDilation:
    dimensionality: XYZ
    inputName: nucleus
    iterations: 10
    outputName: mask
- localisations.LabelsFromImage:
    inputImage: mask
    inputName: _mapped
    outputName: labeled_points
- tablefilters.FilterTable:
    filters:
      objectID:
      - 0.5
      - 1.5
    inputName: labeled_points
    outputName: inside_nucleus_mask
- localisations.ExtractTableChannel:
    channel: chan0
    inputName: inside_nucleus_mask
    outputName: lamin
- surface_fitting.SphericalHarmonicShell:
    input_name: lamin
    output_name: harmonic_shell
    output_name_mapped: lamin_shell_mapped
- surface_fitting.AddSphericalHarmonicShellMappedCoords:
    input_localizations: inside_nucleus_mask
    input_shell: harmonic_shell
    output_mapped: shell_mapped
- filters.MaxFilter:
    dimensionality: XYZ
    inputName: jt2d
    outputName: smoothed
    sizeX: 15
    sizeY: 15
    sizeZ: 1
- localisations.ExtractTableChannel:
    channel: chan1
    inputName: shell_mapped
    outputName: mapped_coilin
- surface_fitting.SHShellRadiusDensityEstimate:
    input_shell: harmonic_shell
    output: r_uniform_kde
    r_bin_spacing: 0.02
    sampling_nm:
    - 200.0
    - 200.0
    - 200.0
```

</details>

enhancement
**Proposed changes:**
- force float in several places in spharm code, follow type of inputs in others
- add volume to unit test directly

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
@David-Baddeley just realized this hasn't been merged. Chance you can take a look before our meeting?

David-Baddeley:
Hi Andrew,

I'll try and have a look - I've been punting it as we've got two competing
modifications to the same code and I think it's going to be slightly tricky
to work out which to go with for a long term merge:

yours:
--------
- pragmatic solution for the paper (brings memory in low enough to allow
analysis to run on your data)
- still somewhat memory hungry and not guaranteed to be OK on all data in
future.
- potential minor issues from using a regular grid (can't recall exactly
what, but think it related to noise/aliasing when radial bin sizes become <
around 10x the grid spacing)

mine:
-------
- guaranteed upper bound on memory usage
- no grid/bin aliasing
- noiser on radial bin 0 ???
- maybe a volume issue (but need to double check / remind myself)

Cheers,
David

PYME/Analysis/points/coordinate_tools.py
```diff
@@ -82,7 +82,7 @@ def find_principle_axes(x, y, z, sample_fraction=None):
 def scaled_projection(x, y, z, scaling_factors, scaling_axes):
     # make sure everything is 1D
     points = np.stack([x.ravel(), y.ravel(), z.ravel()], axis=0)
-    xp, yp, zp = [np.dot(scaling_factors[ind] * scaling_axes[ind], points) for ind in range(3)]
+    xp, yp, zp = [np.dot((scaling_factors[ind] * scaling_axes[ind]).astype(points.dtype), points) for ind in range(3)]
```
David-Baddeley, line 85: is there any reason we would want to cast to `points.dtype`, rather than just, e.g. `'f4'`? `points.dtype` will result in somewhat pathological behaviour if x, y, z are integers, whereas the worst case for casting to float is loss of precision.

PYME/Analysis/points/spherical_harmonics.py
```diff
@@ -293,8 +294,10 @@ class ScaledShell(object):
     ]
 
     def __init__(self, sampling_fraction=1.):
-        self.sampling_fraction = sampling_fraction
+        from PYME.IO import MetaDataHandler
+        self.mdh = MetaDataHandler.NestedClassMDHandler()
```
David-Baddeley: I'm not really happy with a class having a `.mdh` attribute which is not metadata in the classical, PYME, sense. The only real overlap with the canonical metadata is that it's a key-value set of properties / settings. They are somewhat orthogonal to standard PYME metadata, and do not follow standard metadata formatting rules. 

I think I overlooked it in the original PR, where it was constrained to the .hdf output out of expediency (you effectively have a single use, custom file format), but when you might end up with an object sitting in a recipe namespace with a `.mdh` attribute, it should really be canonical metadata, not something else.

David-Baddeley:
Note - I'l try and work through the above comments myself - mostly just there for reference.

PR #1125, opened 2021-09-27
rename traits.py -> pyme_traits.py to avoid name collision with traits itself during import

barentine:
@David-Baddeley , this is your least favorite kind of PR, but hear me out

Say you want to run a recipe module in a script, e.g.
```
from PYME.recipes import surface_fitting
from PYME.IO.tabular import HDFSource
t = HDFSource('something.hdf', 'mapped')
surface_fitting.SphericalHarmonicShell().apply(t)
```
because you want to point your IDE's debugger at it, because print statements are great but not painless. You think you've got a good leg to stand on, since you can run all of that in an interactive shell, but for some reason your IDE debugger goes to import traits and throws a fit. This is independent of the IDE debugger, but happens when you try to run your script too

```
Andrew$ /opt/miniconda3/envs/pdev/python.app/Contents/MacOS/python /Users/Andrew/code/python-microscopy/PYME/recipes/testspharmdtype.py
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/testspharmdtype.py", line 2, in <module>
    from PYME.recipes import surface_fitting
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/__init__.py", line 6, in <module>
    from .recipe import Recipe
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/recipe.py", line 1, in <module>
    from traits.api import HasTraits, List, Bool
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/traits.py", line 11, in <module>
    from traits.api import *
ModuleNotFoundError: No module named 'traits.api'; 'traits' is not a package
```

Drumroll, `sys.path`:
```
(pdev) vpn172022146138:python-microscopy Andrew$ python
Python 3.7.11 (default, Jul 27 2021, 07:03:16) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
>>> from PYME.recipes import surface_fitting
>>> quit()
(pdev) vpn172022146138:python-microscopy Andrew$ cd PYME/recipes
(pdev) vpn172022146138:recipes Andrew$ python
Python 3.7.11 (default, Jul 27 2021, 07:03:16) 
[Clang 10.0.0 ] :: Anaconda, Inc. on darwin
>>> from PYME.recipes import surface_fitting
  File "<stdin>", line 1, in <module>
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/__init__.py", line 6, in <module>
    from .recipe import Recipe
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/recipe.py", line 1, in <module>
    from traits.api import HasTraits, List, Bool
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/traits.py", line 11, in <module>
    from traits.api import *
ModuleNotFoundError: No module named 'traits.api'; 'traits' is not a package
```

So depending on what directory you're in, or what directory your script is in, Python goes looking for `traits` and if `PYME/recipes` directory is above wherever enthought's traits is in the `sys.path` list you find `PYME/recipes/traits.py` which is the wrong droid. We can avoid this name collision by changing the name of traits.py. In this PR, I changed it to pyme_traits.py. Now I can use my IDEs debugger without any weird configurations.

Given that writing recipe modules is one of the key ways to extend PYME, I think we want to make it as painless as possible, and while you should be able to sidestep it if you know it's there and have some sys.path knowledge, it's annoying and probably more than we can expect most people to know about how python imports work, so I think we should change it here.


David-Baddeley:
The reason you are seeing this error is that you are running code with your working directory set to the recipes subdirectory in the PYME tree. This is not a good idea, and is not PYME specific - it applies more generally to all python module development - modules will usually only work when interfaced as a module, not as a collection of scripts. It will cause all sorts of other errors and no end of pain. E.g. running with a module subdirectory as the working directory will mess up all relative imports.  As a result,  I think your error is actually a pretty niche case which shouldn't impact development more broadly, and one where the steps taken to mitigate (not running with a module sub-directory as the working directory) will probably be necessary for other reasons. 

The traceback has however turned up a somewhat embarrassing error which crept into recipe.py, which should be importing from `PYME.recipes.traits` rather than `traits` itself. The convention is always to import traits through `PYME.recipes.traits` in all PYME code, rather than directly. The reasoning behind this convention is threefold:

- `traits` appears at different locations on different installs (most often it's just `traits`, but sometimes it's `enthought.traits`). To be portable code needs to try one location, catch an import error, and then try the other. By directing all access through  `PYME.recipes.traits` we abstract this out and avoid the need to try both imports everywhere (which is hard to remember to do).
- Importing the base traits and our custom ones (e.g. `Input` and `Output` traits) from one place is more consistent and easier. It also protects us from future name collisions if the traits library were to define a trait that collides with one of our custom trait names.
- It gives us the freedom to hack-around some of the less useful bits of traits / traitsui in the future by, e.g. subclassing the traits with ones which work better with a web ui or similar.

Given the above, I think the current naming is both logical and consistent.

Your proposed fix  **WILL BREAK ANY EXISTING 3rd PARTY RECIPE MODULE**.  As a result, it would need an extremely compelling reason to incorporate, and a well thought out and sign-posted deprecation path. I think we're a long way from this threshold.

I'll fix the import in recipe.py which will make this specific error go away, but it would still be sensible to set your working directory to something else (e.g. your home directory) to avoid other problems.

Issue #415, opened 2020-08-15
[ENHANCEMENT] io plugins for pipeline construction


csoeller:
**Is your feature request related to a problem? Please describe.**
There are different file formats out there for event storage and I was hoping it be feasible to come up with an 'import' plugin infrastructure. This is also along the lines of "becoming the ImageJ of SMLM viewers".

**Describe the solution you'd like**
Similar to the way plugins for VisGUI already exist an io import structure would be useful (e.g. for csv with particular column name mapping; picasso hdf5; etc). The plugin could register an import function that provides the ingredients to construct a pipeline + metadata. The user could explicitly use that import function from the menu. No fancy auto-detection required but could be considered if not too much 'opening a can of worms'.

**Describe alternatives you've considered**
One can convert, e.g. I have a converter for picasso to pyme friendly CSV but would really like a visgui IO plugin API.

**Are you prepared to work on it?**
Once we settle on a way how it would and should work, sure.

David-Baddeley:
Definitely a worthwhile objective. Can think of a few wrinkles, particularly when it comes to metadata and name translation etc ...  Something to keep in mind when we incorporate #284 (or a modified variant thereof).

Is likely going to involve:

- refactoring `pipeline._ds_from_file()`
- writing a new `tabular` source for the desired file type.

Initially, binding it through something like `File->Import` in VisGUI would probably be desirable (lets the user choose the filetype rather than having to do detection magic), but ultimately some form of detection stuff would also be good.

Have you got any other formats in addition to picasso in mind? Where are they documented?

Unclear at this stage if they should be plugins, or just have things refactored and documented so that it's easy for someone to contribute a loader without wading through the complicated historical mess that is the current data import process. I'd imagine most loaders would only be a single file or so.

csoeller:
_Have you got any other formats in addition to picasso in mind? Where are they documented?_

It seems like Thunderstorm at least has a set of known CSV columns that should be considered, though not sure if a proper format description (see this post [here](http://www.neurocytolab.org/tscolumns/)).

This general idea also relates to another thought I had and will get started on RSN, i.e. starting a repo of SMLM example data, using our own and any available external data that is interesting. It would serve as a convenient set of test data to validate IO against, try some methods with, see if old PYME formats can still be properly digested, etc. It's something that has bugged me for a while, always having to search for good example data in the depth of my file system or elsewhere.

[...]

PR #1128, opened 2021-10-02
PYME - import CSVs and similar external SMLM event files

csoeller:
Addresses issue #415 .

Enhancement.

**Proposed changes:**

This is a work in progress. The goal is to start a discussion about the implementation and refine this jointly until ready for merging at some point in the future.

**Test suite:**

I have crerated a repository with example data sets and a pytest file, see https://github.com/csoeller/PYME-SMLM-import-tests. This should be regularly run against updated code in this PR.

**Current limitations:**

1. It currently does load files with extension `.csv` and `.tsv` only.
2. currently completely bypasses the `ImportTextDialog` - mostly for proof of principle. Eventual proper integration should likely integrate going through a suitably modified `ImportTextDialog`, at least if any issues are encountered upon parsing the file and/or a suitable chosen config option is set.
3. currently plugged via a kludge into pipeline.py/viscore.py that checks for the `.csv` and `.tsv` extensions
4. however, that kludge is only used if the `VisGUI-use_csv_reader` config option is set
5. module naming, method/function naming may be subject to debate
6. uses a modified version of the `csv_flavours` dict approach suggested by @David-Baddeley 
6. probably more, some comments in the code try to point out current compromises

**Checklist:**

- [x ] Tested with numpy=1.14
- [ x] Tested on python 3.7
- [ N/A] Tested with wx=3.x and wx=4.x [if UI code]

- [ x] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)?
- [ will do eventually] Does this change how users interact with the software? How will these changes be communicated?
- [ should do] Does this maintain backwards compatibility with old data?
- [- ] Does this change the required dependencies?
- [to be discussed] Are there any other side effects of the change?


David-Baddeley:
Thanks! And sorry for taking so long to get around to looking at this.

I think it's a great start, but have a few minor reservations:

- I'm a little concerned by the duplication between this and existing .csv import.  I feel that this will increase both maintenance burden and complexity for the end user.
- Skipping the `TextImportDialog` on .csv import breaks loading for existing supported .csv formats (this is somewhat mitigated by requiring the command line option,  but this trades it off against significant pain for an end user).

I think the safest path is to use the flavours to pre-populate the text import dialog rather than having them as an alternative path. I'm happy to work on it from here ...

PYME/IO/csv_flavours.py
```diff
@@ -0,0 +1,229 @@
...
+    def read_csv_data(self):
+        return np.genfromtxt(self.filename,
+                             comments=self.flavour_value_or_default('comment_char','#'),
+                             delimiter=self.flavour_value_or_default('delimiter',','),
+                             skip_header=self.nHeaderLines,
+                             skip_footer=self.flavour_value_or_default('skip_footer',0),
+                             names=self.translatedNames, dtype='f4', replace_space='_',
+                             missing_values=None, filling_values=np.nan, # use NaN to flag missing values
+                             invalid_raise=not self.flavour_value_or_default('must_ignore_errors', False),
+                             encoding='latin-1') # Zeiss Elyra bombs unless we go for latin-1 encoding, maybe make flavour specific?
```
David-Baddeley: @csoeller - could you please provide an example Elyra file? It would be nice to double check if any alternate encodings work (UTF8 would be ideal).

csoeller: For that purpose I have created a githuib repo with test files and also a small test suite, see https://github.com/csoeller/PYME-SMLM-import-tests.

This repo includes example files for all the flavours I have tried to cover so far, including [elyra examples](https://github.com/csoeller/PYME-SMLM-import-tests/tree/main/elyra). The test rely on the ability of the programmatic opening of these files without gui dialog via a `pipeline` constructor command, e.g.

```python
    pl = pipeline.Pipeline(filename=fname)
```

It would be good to preserve this ability, possibly with a `noGUI` or similar keyword if required.

csoeller:
Thanks @David-Baddeley, I appreciate you looking into it. This was always meant to be a proof of principle to work from, not a ready-made PR.

Specific responses to your points:

- it would make sense to make this one codebase for all textual file imports, csv and otherwise, and amalgamate the existing `.csv` import code into this (or vice versa). This was always meant merely as a proof-of-principle that allows a user/tester to fall back to the original `.csv` support by setting a config flag (via `VisGUI-use_csv_reader` flag). This should all eventually go.
- Very happy to see the integration happen using the `TextImportDialog`. So far I shrank back from that and focused on getting the import logic itself in place. If you could lead the way on the integration that would be great. As I mentioned in a reply to your request for test files, it would be good to preserve the ability to open such files programmatically without GUI dialog (if possible with the info given, otherwise throw an error), e.g. `pl = pipeline.Pipeline(filename='myelyra.txt', noGUI=True)` or similar.

PYME/IO/csv_flavours.py
David-Baddeley: yeah, I realised shortly after I'd written that that you'd already mentioned the example data earlier - sorry about that.

David-Baddeley: Have just looked at the example elyra file -  is the kind of corruption (see below) that requires the `invalid_raise=False` option common, or is it limited to this one file?

```
40146	1486	1	0	2195.5	35172.0	23.3	885	47.42	0.65	173.4	1	1
40147	1486	1	0	35413.4	40391.3	30.0	446	71.18	0.68	131.2	1	1
40148	1486	1	0	29306.8	50081.7	29.6	395	43.82	0.59	134.3	1	1
40149	1486	1	0	31793.5	42414.7	28.6	569	58.19	0.74	148.0	1	1
40150	1486	1	0	31947.4	55524.0	25.3	549	40.44	0.80	141.8	1	1
40151	1486	1	0	31913.2	40789.0	25.6	815	45.05	0.63	176.4	1	1
40152	1486

												
40155	1486	1	0	35894.0	35253.5	37.2	289					
40156	1486	1	0	14841.3	41790.8	35.8	254	52.38	0.68	114.6	1	1
40157	1486	1	0	32679.2	34909.4	37.0	583	75.16	0.68	166.0	1	1
40158	1486	1	0	30516.5	54053.7	34.2	449	68.38	0.69	141.7	1	1
40159	1486	1	0	37139.9	31912.4	28.1	75	54.00	0.69	45.2	1	1
40160	1487	1	0	16720.8	35844.3	9.2	2268	64.80	0.89	150.2	1	1
```

If it's a one-off problem with a specific corrupted file, then we should possibly not just ignore it. If it's common occurrence for elyra files, I'd be happier ignoring.

csoeller:
In our limited tests it seemed pretty common...

David-Baddeley:
Still needs a bit of tidying, but flavours are now supported through the import UI, should also work for code-based pipeline usage.

Known remaining issues:

- asthetics of dialog (flavour choice appears on far right, separated from label)
-  X, Y, Z -> x, y, z "default" behaviour currently missing (I kind of want this to be opt-in, not sure how to go about that)
- loading in recipes

There's also definitely scope for some consolidation / further refactoring

csoeller:
Did you test the latest commit with the data I supplied. On all files I get, when trying to open with dh5view:

```
RuntimeError: Cannot load file /Users/csoe002/Documents/src/PYME-src/SMLM-CSV-data-format-samples/ThunderSTORM/ThunderSTORM_TS3D.csv, no native handler and failed to import bioformats
Finished main loop
```

Not sure if I am doing anything silly?

David-Baddeley:
PYMEVis vs dh5view???

csoeller:
I see, this seems to do it. I had (apparently naively) assumed that one is merely an alias for the other.

David-Baddeley:
PYMEVis/VisGUI and PYMEImage/dh5view have always been separate programs ... although a common entry point has been on the wishlist for a while.

csoeller:
PYMEVis/VisGUI and PYMEImage/dh5view mixup indicates I need some rest after the big move.

Thanks for folding this feature into master.

Issue #1083, opened 2021-07-09
[ENHANCEMENT] Change how we pass camera noise properties


zacsimile:
Reproducing @David-Baddeley 's suggestion from closed PRs #144 and #1079. We can finish the discussion here and tick this off.

> My proposal:
> - we modify the base camera class to specify that all cameras should implement a .noise_properties property that returns a dictionary of the current properties
> - we modify the __init__() signature of the camera class to accept a noise_info parameter which can either be a path to a .yaml file (preferred), or a dictionary. The resulting noise_info dictionary must be keyed on the camera serial number (this should help avoid problems when swapping cameras), but what sits under that is up to the individual camera.
> - we make the above parameter optional, with cameras using either a PYME.config option to specify the location of the .yaml file, or a fixed location within the pyme config tree by default (this allows, amongst other things, a smooth upgrade path by transferring the current hard-coded properties to a .yaml that we distribute with PYME).
> - The use of a path to a .yaml allows a master file to be held on shared storage, if desired, making camera swaps etc ... easy.

It looks like `.noise_properties` is already included in the camera base class, and the return value suggests we implement this property in a derived class. I think everything described in @David-Baddeley's suggestion is common to all cameras, and I suggest we put all of the filename passing/YAML loading logic for noise properties in the camera base class, rather than re-implement it for each camera. We can simply default to `noise_info=None` and pass that information in specific cameras when necessary. Does this sound reasonable?

David-Baddeley:
Can probably close this now with #1131

PR #1131, opened 2021-10-17
Cam noiseprops - user addable camera config via PYME.config mechanism

csoeller:
Addresses issue #1083 .

Enhacement for long standing issue.

**Proposed changes:**
Along lines suggested earlier by adding camera spec yaml files to the `cameras` sub directory of config directories. Details see `PYME.config` docs.

Implemented via a `CameraProperties` base class that derived camera classes should inherit from plus uses functionality from `PYME.config.get_cam_props()`.

So far tested with fakecam and will soon test with Sona I have for testing. 

Example use see [this notebook](https://github.com/csoeller/PYME-camprops-mock/blob/master/camera_properties_test.ipynb).

Will need further testing by other with other cams, e.g. Hamamtsu Orca.

**Checklist:**

- [ x] Tested with numpy>1.14
- [ x] Tested on Python 3.7
- [ NA] Tested with wx=3.x and wx=4.x [if UI code]

- [ x] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)?
- [x] Does this change how users interact with the software? How will these changes be communicated? ->Via docs.
- [x ] Does this maintain backwards compatibility with old data?
- [no] Does this change the required dependencies?
- [not known] Are there any other side effects of the change?


csoeller:
Now successfully tested with a Sona using yaml definition as below.
May want to merge `CameraProperties` functionality directly into base class `Camera` if we are certain that all backends inherit from it.

```yaml
CSC-00597:
  comment: a Sona entry where the noise property names are unicode strings; not sure
    if this matters
  model: Andor Sona
  noiseProperties:
    12-bit (low noise):
      ADOffset: 100
      ElectronsPerCount: 0.46
      ReadNoise: 1.22
      SaturationThreshold: 1720
    16-bit (high dynamic range):
      ADOffset: 100
      ElectronsPerCount: 1.03
      ReadNoise: 1.73
      SaturationThreshold: 42769
```

csoeller:
Now all merged into the `Camera` base class.

zacsimile:
Thanks, @csoeller! This worked for me. I'm excited to use this in the future! 

Going to post a few notes here so we keep them in mind for this switch. Most of this comes down to making sure we write YAMLs properly.

- I wonder if `_hardcoded_properties` should supersede the YAML files instead of the other way around (see `PYME.Acquire.Hardware.Camera.get_cam_prop_or_default()`. When I think "hardcoded" I think unchangeable, rather than a fallback. Either way, it would be nice for the logger to record where the `noise_properties` get loaded from (YAML vs. hardcoded).
- `get_cam_prop_or_default()` re-implements dictionary logic. I think it'd be fine to use the standard python dictionary `get(key, default_value)` calls here.
- `noise_properties` and `_get_cam_noise_props_from_config()` re-implement some of the logic in `get_cam_prop_or_default()`. I think these could be combined into `self.noise_properties = self.get_cam_prop_or_default('noiseProperties', None)[self._preamp_mode_repr()]` (or similar, if you wanted a specific error).
- In the example YAML you provided, you don't put quotes around anything, because you don't need to. However, for Hamamatsu serial numbers, which are integers, they must be wrapped in quotes in order to be used. That is
```
'301777':
  comment: Hamamatsu ORCA for SMLM arm of SMLM/OIDIC scope
  model: Hamamatsu ORCA Flash 4.0 V4
  noiseProperties:
    fixed:
      ReadNoise: 1.63
      ElectronsPerCount: 0.47
      NGainStages: 0
      ADOffset: 100
      DefaultEMGain: 1
      SaturationThreshold: 65535
```
works, but this fails if I enter `301777` instead of `'301777'`.
- This is no big deal, but expressions are not evaluated, so if I enter `(2**16-1)` in the YAML for `SaturationThreshold`, I get `{'ReadNoise': 1.63, 'ElectronsPerCount': 0.47, 'NGainStages': 0, 'ADOffset': 100, 'DefaultEMGain': 1, 'SaturationThreshold': '(2**16 - 1)'}` for noise properties. Basically, I can't copy/paste the `_hardcoded_properties` into a YAML without tweaking this.

csoeller:
Thanks @zacsimile for feedback, very useful!

Code improvement suggestions all taken on board, I'll look into it.

My existing use case for yaml over `_hardcoded_properties`: got a repaired Zyla back and needed to input the new chip properties - serial number is still the same. In that case I wanted the user changable data to override the `_hardcoded_properties` without having to submit a PR for `_hardcoded_properties`. The reason `_hardcoded_properties` exists is mainly for backward compatibility with cameras that people entered in the past so that nothing breaks when they upgrade their PYME.

The string vs numeric issue for serial nos without letters and symbols is a real one, we should have a way to avoid this (try both?). Also the eval question is a good one, would be nice to numercially evaluate the entries while staying clear of any security issues if that is hopefully easy enough.

zacsimile:
That's fair, it's probably fine to default to the YAMLs, and I'll likely want to change them more frequently than I want to PR as well.

For the string vs. numerical serial number issue, I think it's fine to always wrap the serial number in single quotes. This will work no matter what. 

I don't think we want to eval when we load YAMLs, I just wanted to point out that we need to watch for that if we copy+paste over from the previous properties.

csoeller:
@zacsimile's suggestions for code clean-up have been largely incorporated.

Remaining issues I'm aware of:

  - clean up examples in docs (remove expressions that would need eval)
  - consistent handling of serial numbers as stringified values?

Eventually nice to have:

  - plugin to edit camera info in GUI and save to local config file

David-Baddeley:
Sorry I'm only just getting around to having a look ...

A few initial thoughts:

- **naming:** - This should still include noise (or potentially readout) as cameras have a bunch of other properties which are not captured here (and many of the camera APIs are described in terms of getting and setting properties). A generic `GetCamPropOrDefault()` or `get_cam_prop()` will be confusing in this context. Ditto, for, e.g., `_hardcoded_properties` 
- `_hardcoded_properties` (or it's more appropriately named - e.g. `_legacy_noise_properties` equivalent) should not be an attribute of the camera class. Acceptable options would be as a module-level variable, or better yet, pooling all the legacy data into a single file and using it to pre-populate e.g. the `config.camera_noise_properties` dictionary.
- having a function (`get_cam_props()`) in `config` only makes sense if the function actually does something - e.g. does the dictionary lookup by serial number etc ... Otherwise, we should just access the `config.camera_noise_properties` dictionary directly.
- We should probably also think of a way of registering noise properties from an init script (as this is the default way of configuring new hardware.

David-Baddeley:
I'll have a bit of a crack ...

csoeller:
Looking good. I just fixed a couple of small niggles in latest commit.

Could possibly do with a few more tests using different camera models. I just tested with a repaired Zyla and it now correctly uses the updated noise properties after repair, overriding the legacy coded properties.

Still possibly worth looking at: serial numbers that are currently treated as numerical (e.g. Andor Ixon)? Should we possibly use strings everywhere and force a `repr()` or similar on returned serial nos?

PR #1134, opened 2021-10-26
Name layer by mesh filename if loaded

zacsimile:
Loaded meshes now display the filename in the layer, rather than just `surf0`, `surf1`, etc. Helpful for loading & comparing meshes.

### DIFF
--- a/PYME/LMVis/Extras/extra_layers.py
+++ b/PYME/LMVis/Extras/extra_layers.py
@@ -1,5 +1,6 @@
 import numpy as np
 import wx
+import os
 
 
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I suspect we're fine, but are we sure that this will give a valid name?

- most obvious potential problem (dots in name) is dealt with by taking `name.split('.')[0]`, although there is a potential edge case here if people use odd file naming conventions - e.g. `mysurface.channel0.stl`, `mysurface.channel1.stl` would create an error if attempting to load both surfaces. I'm somewhat tempted to just note this in the comments and leave as "won't fix" though.
- Related to the above, this will error if loading the same surface twice (or reloading a modified version of an already loaded surface).
- Are there any other characters that could cause problems?
- Will unicode / international filenames cause problems? I have a sneaky feeling that we might at least assume that recipe  input/output names are ASCII

David-Baddeley:
We could fix issues 1 & 2 by simply re-instating the name update logic (or a modified version thereof).

PR #1135, opened 2021-10-26
Add fan controls to pco_sdk/pco_sdk_cam

zacsimile:
Added these to test if my camera's fan was broken (it was), and they could be generally useful.

### DIFF
--- a/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
@@ -365,6 +365,7 @@ class PcoSdkCam(Camera):
         return mode, speed
 
     def _set_fan_mode_speed(self, mode, speed):
+        # TODO - unified fan control interface for all cameras (see also AndorIxon `SetFan()` and comments on PR1135)
         pco_sdk.set_fan_control_parameters(self._handle, mode, speed)
 
     def _get_temps(self):
--- a/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
@@ -360,11 +360,11 @@ class PcoSdkCam(Camera):
         self._ccd_temp_set_point = temp
         self._get_temps()
 
-    def GetFanModeSpeed(self):
+    def _get_fan_mode_speed(self):
         mode, speed = pco_sdk.get_fan_control_parameters(self._handle)
         return mode, speed
 
-    def SetFanModeSpeed(self, mode, speed):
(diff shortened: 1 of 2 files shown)


PYME/Acquire/Hardware/pco/pco_sdk_cam.py
```diff
@@ -360,6 +360,13 @@ def SetCCDTemp(self, temp):
         self._ccd_temp_set_point = temp
         self._get_temps()
 
+    def GetFanModeSpeed(self):
```
David-Baddeley: There's already a `SetFan(self, state)` method on the IXon camera - we should probably try adopt a consistent signature between cameras. In this case `state` is either 0, 1, or 2 (effectively an enum, meaning off, low, high). The simplest would probably be to adopt `SetFan(self, state)` across all cameras where `state` had a camera-specific meaning (as long as it was easily serialisable as, e.g. json). A little more work would be to define consistent states across cameras. For the PCO cameras, does the speed take a fixed set of values (effectively an enum), or is it variable across a range)?

David-Baddeley: or to make the low level camera-api `_set_fan_mode_speed()` a private instance of the PCO class.

David-Baddeley: Cancel above question re what the PCO SDK takes - you've covered it in the docstring

David-Baddeley: Short term solution is probably to just make the methods "private" - i.e. not part of the camera API with a comment to revisit if/when we want API control. As I understand it, the main rational for adding them at the moment is testing.

zacsimile: Yeah, they're just for testing. I'll just convert them to private.

PYME/Acquire/Hardware/pco/pco_sdk_cam.py
```diff
@@ -360,6 +360,13 @@ def SetCCDTemp(self, temp):
         self._ccd_temp_set_point = temp
         self._get_temps()
 
+    def _get_fan_mode_speed(self):
+        mode, speed = pco_sdk.get_fan_control_parameters(self._handle)
+        return mode, speed
+
+    def _set_fan_mode_speed(self, mode, speed):
```
David-Baddeley, line 367:
```suggestion
    def _set_fan_mode_speed(self, mode, speed):
        # TODO - unified fan control interface for all cameras (see also AndorIxon `SetFan()` and comments on PR1135)
```

PR #1142, opened 2021-11-06
Pass at parallelizing DMC

zacsimile:
Profiling revealed that the `face_proc_` and `edge_proc_` functions are the biggest bottle neck in mesh generation. There's no reason they cannot be run in parallel at each level of recursion, since they do not depend on one another and the resulting `self.vertices` is an unsorted set of cubes to march. Presumably pushing a bunch of `apply_async` calls to a multiprocessing pool should take care of this, but it seems to not work. Removing the `apply_async` calls from `node_proc` works better, but it's still missing quite a few cubes. It's unclear to me why this is the case. Pushing this as a draft to see if any of the other devs have ideas.

### DIFF
--- a/PYME/experimental/dual_marching_cubes.py
+++ b/PYME/experimental/dual_marching_cubes.py
@@ -10,7 +10,7 @@ import multiprocessing
 
 class DualMarchingCubes(ModifiedMarchingCubes):
     
-    def __init__(self, isolevel=0, parallel=True):
+    def __init__(self, isolevel=0, parallel=False):
         super(DualMarchingCubes, self).__init__(isolevel)
         self._ot = None
         self._parallel = parallel
@@ -43,11 +43,12 @@ class DualMarchingCubes(ModifiedMarchingCubes):
         
         if self._parallel:
             self._pool = multiprocessing.Pool(multiprocessing.cpu_count())
-        self.node_proc(self._ot._nodes[0])  # Create the dual grid
-        self._pool.apply_async(time.sleep, (10,))
-        if self._parallel:
+            ret = self._pool.apply_async(self.node_proc, (self._ot._nodes[0],))
+            ret.get()
             self._pool.close()
             self._pool.join()
+        else:
+            self.node_proc(self._ot._nodes[0])  # Create the dual grid
 
         # Make vertices/values np arrays
         self.vertices = np.vstack(self.vertices).astype('float64')
--- a/PYME/experimental/dual_marching_cubes.py
+++ b/PYME/experimental/dual_marching_cubes.py
@@ -6,12 +6,14 @@ from PYME.experimental.modified_marching_cubes import ModifiedMarchingCubes
 from PYME.experimental._octree import has_children
 
 import time
+import multiprocessing
 
 class DualMarchingCubes(ModifiedMarchingCubes):
     
-    def __init__(self, isolevel=0):
+    def __init__(self, isolevel=0, parallel=True):
         super(DualMarchingCubes, self).__init__(isolevel)
         self._ot = None
+        self._parallel = parallel
 
     def set_octree(self, ot):
         t_ = time.time()
@@ -24,6 +26,11 @@ class DualMarchingCubes(ModifiedMarchingCubes):
         self.vertices = []
         self.values = []
         self.depths = []
+        if self._parallel:
+            self._manager = multiprocessing.Manager()
+            self.vertices = self._manager.list()
+            self.values = self._manager.list()
+            self.depths = self._manager.list()
         
         #precalculate shift scales for empty boxes
         max_depth = self._ot._nodes['depth'].max()
@@ -34,8 +41,13 @@ class DualMarchingCubes(ModifiedMarchingCubes):
         #TODO - get this from the octree
         self._octant_sign = np.array([[2 * (n & 1) - 1, (n & 2) - 1, (n & 4) / 2 - 1] for n in range(8)])
         
-
+        if self._parallel:
+            self._pool = multiprocessing.Pool(multiprocessing.cpu_count())
         self.node_proc(self._ot._nodes[0])  # Create the dual grid
+        self._pool.apply_async(time.sleep, (10,))
+        if self._parallel:
+            self._pool.close()
+            self._pool.join()
 
         # Make vertices/values np arrays
         self.vertices = np.vstack(self.vertices).astype('float64')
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Hmm, a bunch of potential issues:

- multiprocessing worker processes generally get a **copy** of the data they are working on, rather than a reference, meaning that any modification of the data gets lost when they return. 
- This means that the only thing that gets saved in an `apply_async(...)` is the return value. As written, all the functions work via side effects (they modify the underlying datastructure) rather than returning results - this makes it difficult to apply a multiprocessing paradigm to parallelisation.
- You can potentially work around this by using `PYME.util.shmarray` (see the parallelised jittered triangulation code) so that each process is looking at the same memory, but this can be fiddly / un-natural/error prone.
- I'm doubtful that multiprocessing is the correct idiom for parallelisation in this instance - I think you want something that is a) in-process (to allow modification of a common memory structure) and b) can scale to > n_cpu_cores. To me this feels like a good candidate for CUDA/OpenCL. Another option is threaded, GIL-releasing c code, but that's still limited to n_cpu_cores. 

Some more mundane issues:

`self._pool.apply_async(self.face_proc_xy, (n0, n4))` more or less translates to:

process 0: `self.face_proc_xy(*n0)`
process 1: `self.face_proc_xy(*n4)`

Where the `self` in each case is a different copy of the initial datastructure.

David-Baddeley:
PS - due to various quirks, `shmarray`s can only be passed at process *creation*, not to an already existing subprocess. This makes them difficult to use with multiprocessing `Pool`s - I think I've always had to manually create the processes (see, e.g. `visHelpers.py` line 513).

zacsimile:
Gave this a go with some minimal examples. I think `self._pool.apply_async(self.face_proc_xy, (n0, n4))` is equivalent to `self.face_proc_xy(n0,n4)`.  Here's a minimal example showcasing this:

```
def f(x, y=None):
    if y is None:
        y = x
    return x*y

cpu_count = multiprocessing.cpu_count()
print(f'cpu_count is {cpu_count}')
with multiprocessing.Pool(cpu_count) as pool:
    for i in range(2*cpu_count):
        ret = pool.apply_async(f, (i,i+1))
        print(f'f({i},{i+1}) = {ret.get()}')
```

prints

```
cpu_count is 4
f(0,1) = 0
f(1,2) = 2
f(2,3) = 6
f(3,4) = 12
f(4,5) = 20
f(5,6) = 30
f(6,7) = 42
f(7,8) = 56
```

`multiprocessing.Manager().list()` does let us create a list in shared memory. I think what I'm missing is a `.get()` call at the end of everything, although I'm not sure why this matters. The following minimal example works for asynchronous adding of data to a list from different processes.

```
import time

manager = multiprocessing.Manager()
shared_list = []
shared_list = manager.list()

def f2(x, y=None):
    time.sleep(np.random.rand())
    if y is None:
        y = x
    shared_list.append(x*y)

with multiprocessing.Pool(cpu_count) as pool:
    for i in range(2*cpu_count):
        ret = pool.apply_async(f2, (i,i+1))
    ret.get()
print(shared_list)
```

yields

```
[2, 0, 6, 12, 20, 42, 30, 56]
```

As such, as I think what I have here should be OK modulo a `.get()` call. Whether or not this is the best way to approach this is still very much up for debate.

zacsimile:
No, a `get()` call doesn't solve it. I get a pickling issue...

```
TypeError                                 Traceback (most recent call last)
/var/folders/g9/tj_thbrx23qcft96nq5j5d4sx7_5b_/T/ipykernel_47988/1110754126.py in <module>
     89                        threshold_density=density, n_points_min=n_points_min, remesh=remesh,
     90                        repair=repair, cull_inner_surfaces=cull_inner_surfaces, smooth_curvature=smooth_curvature)
---> 91 recipe.add_modules_and_execute([dmc,])
     92 mProfile.profileOff()
     93 mProfile.report()

~/Code/python-microscopy/PYME/recipes/recipe.py in add_modules_and_execute(self, modules, rollback_on_failure)
    455                 self.modules.append(m)
    456 
--> 457             self.execute()
    458         except:
    459             if rollback_on_failure:

~/Code/python-microscopy/PYME/recipes/recipe.py in execute(self, progress_callback, **kwargs)
    201                     print('Executing %s' % m)
    202                     m.check_inputs(self.namespace)
--> 203                     m.execute(self.namespace)
    204                     m._last_error = None
    205                     m._success = True

~/Code/python-microscopy/PYME/recipes/surface_fitting.py in execute(self, namespace)
    151 
    152         dmc = dual_marching_cubes.PiecewiseDualMarchingCubes(self.threshold_density)
--> 153         dmc.set_octree(namespace[self.input].truncate_at_n_points(int(self.n_points_min)))
    154         tris = dmc.march(dual_march=False)
    155 

~/Code/python-microscopy/PYME/experimental/dual_marching_cubes.py in set_octree(self, ot)
     45             self._pool = multiprocessing.Pool(multiprocessing.cpu_count())
     46             ret = self._pool.apply_async(self.node_proc, (self._ot._nodes[0],))
---> 47             ret.get()
     48             self._pool.close()
     49             self._pool.join()

/opt/miniconda3/envs/pyme/lib/python3.7/multiprocessing/pool.py in get(self, timeout)
    655             return self._value
    656         else:
--> 657             raise self._value
    658 
    659     def _set(self, i, obj):

/opt/miniconda3/envs/pyme/lib/python3.7/multiprocessing/pool.py in _handle_tasks(taskqueue, put, outqueue, pool, cache)
    429                         break
    430                     try:
--> 431                         put(task)
    432                     except Exception as e:
    433                         job, idx = task[:2]

/opt/miniconda3/envs/pyme/lib/python3.7/multiprocessing/connection.py in send(self, obj)
    204         self._check_closed()
    205         self._check_writable()
--> 206         self._send_bytes(_ForkingPickler.dumps(obj))
    207 
    208     def recv_bytes(self, maxlength=None):

/opt/miniconda3/envs/pyme/lib/python3.7/multiprocessing/reduction.py in dumps(cls, obj, protocol)
     49     def dumps(cls, obj, protocol=None):
     50         buf = io.BytesIO()
---> 51         cls(buf, protocol).dump(obj)
     52         return buf.getbuffer()
     53 

~/Code/python-microscopy/PYME/experimental/_octree.cpython-37m-darwin.so in PYME.experimental._octree.Octree.__reduce_cython__()

TypeError: self._cnodes cannot be converted to a Python object for pickling
```

PR #1143, opened 2021-11-07
dual marching cubes

David-Baddeley:
minor optimisations to dual marching cubes (hopefully better memory)


zacsimile:
Works much faster for me. I say merge.

Issue #1151, opened 2021-11-24
[BUG] listWrapper image output conflict (affecting Filters e.g. projection module)


barentine:
**Describe the bug**

```
 Error whilst running Batch Analysis
==============================================
python-microscopy=21.09.23
python=3.7.6, platform=darwin
numpy=1.19.2, wx=4.0.4 osx-cocoa (phoenix) wxWidgets 3.0.5

Traceback
=========
Traceback (most recent call last):
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/recipeGui.py", line 913, in OnBake
    self.last_run = batchProcess.bake(self.rm.activeRecipe, {'input':inputs[0]}, out_dir, num_procs=num_procs, metadata_defaults=self._default_md)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/batchProcess.py", line 80, in bake
    last_rec = runRec(task)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/batchProcess.py", line 27, in runRec
    r = runRecipe.runRecipe(*args)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/runRecipe.py", line 105, in runRecipe
    recipe.save(context)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/recipe.py", line 515, in save
    mod.save(self.namespace, context)
  File "/Users/Andrew/code/python-microscopy/PYME/recipes/output.py", line 219, in save
    v.Save(out_filename)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/image.py", line 1326, in Save
    self.filename = dataExporter.ExportData(self.data_xyztc, self.mdh, self.events, filename=filename, progressCallback=progressCallback)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/dataExporter.py", line 504, in ExportData
    exp.Export(ds, filename, slice(0, ds.shape[0], 1), slice(0, ds.shape[1], 1), slice(0, ds.shape[2], 1),mdh, events, origName, progressCallback=progressCallback)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/dataExporter.py", line 119, in Export
    data, tslice, xSize, ySize, nZ, nT, nChans, nframes = self._prepare(data, xslice, yslice, zslice, tslice)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/dataExporter.py", line 84, in _prepare
[...]

PR #1152, opened 2021-11-24
force listwrap to xyztc array

barentine:
Addresses issue #1151 . Not clean. @David-Baddeley I haven't been following the change to XYZTC closely enough to know what you want done with listwrapping / recipes.base.Filter outputs, but I'll put my hack-around here for now.

bugfix
**Proposed changes:**
sloppily force 5d listwrap (e.g. len 2 list wrapping xyz) into xyztc for output

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/IO/dataExporter.py
+++ b/PYME/IO/dataExporter.py
@@ -77,9 +77,11 @@ class Exporter:
         from PYME.IO.DataSources import BaseDataSource, CropDataSource
         data = dataWrap.Wrap(data) #make sure we can index along a colour dimension
     
-        if data.ndim != 5 or isinstance(data, dataWrap.ListWrapper):
+        if data.ndim != 5:
             # promote to xyztc and force to array datasource
             data = BaseDataSource.XYZTCWrapper.auto_promote(data)
+        elif isinstance(data, dataWrap.ListWrapper):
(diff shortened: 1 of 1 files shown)


barentine:
Note this does not get us both colors if saving a 2 channel to h5...

PYME/IO/dataExporter.py
```diff
@@ -73,13 +73,15 @@ def Export(self, data, outFile, xslice, yslice, zslice, metadata=None, events =
...
     def _prepare(self, data, xslice, yslice, zslice, tslice):
-        from PYME.IO.dataWrap import Wrap
+        from PYME.IO import dataWrap
         from PYME.IO.DataSources import BaseDataSource, CropDataSource
-        data = Wrap(data) #make sure we can index along a colour dimension
+        data = dataWrap.Wrap(data) #make sure we can index along a colour dimension
     
         if data.ndim != 5:
-            # promote to xyztc
+            # promote to xyztc and force to array datasource
```
David-Baddeley: Note: data source is not forced to an array datasource - the `XYZTCWrapper` is just a pass-through wrapper which changes the indexing - there is no copy  or array coercion.

David-Baddeley:
Alternative fix (making CropDataSource less fussy) now merged.

PR #1156, opened 2021-11-26
issue 1130

### LINKED ISSUE #1130: [BUG/ENHACEMENT] Frame doesn't change when I click on `dh5view` playback bar
Not sure if this is a bug or an enhancement. If I click anywhere on the dh5view playback bar, the frame number doesn't change, even though the slider position does. If I hit play, the slider goes back to where it was before and keeps running. GOTO changes the frame number just fine. I expect clicking somewhere on the play bar to have the same behavior as GOTO.

David-Baddeley:
Candidate to fix #1130. As yet untested on windows


David-Baddeley:
@zacsimile - could you give this a try?

zacsimile:
Nice! This fixes everything I mentioned #1130. There's one weird thing: when I click on the scrollbar, the position it goes to is not the same as where I click (it's a bit further back). Illustrated below, red circle is where I clicked. This is consistent across the scale bar. This is not a problem for click-and-drag.

image

David-Baddeley:
That's super odd ... goes where you click on mac ... Was initially thinking it could be a rounding issue (rounding down to the closest tick to where you clicked), but you've got enough frames that that shouldn't be an issue. Going to blame it on windows weirdness for now.

PR #1163, opened 2021-12-07
Some compatibility fixes

### LINKED ISSUE #99: Linux, wx4, points layer not automatically created in VisGUI on data load
**[Original report](https://bitbucket.org/david_baddeley/python-microscopy/issue/99) by David Baddeley (Bitbucket: [david_baddeley](https://bitbucket.org/david_baddeley), ).**

----------------------------------------

Has to do with the view window not having focus and the openGL context not having been created when `add_layer` is called \(error is glCreateProgram failed\). 

‌

‌

David-Baddeley:
Currently just making missing mpld3 a soft error to make installing in custom evironments a bit easier, will likely grow

### DIFF
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -781,8 +781,14 @@ class Camera(object):
             self.GetStatus()
 
             # Personal identification
-            mdh.setEntry('Camera.Name', self.GetName())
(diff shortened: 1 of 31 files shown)


David-Baddeley:
deferred shaders stuff is mostly an un-successful attempt to get things working on linux, py38, but should be cleaner than the status quo, and useful when moving to common layers/overlays between pymevis and pymeimage

PR #1164, opened 2021-12-07
ueye updates

barentine:
Addresses issue was no GetSerialNumber, raised error if you tried to snapshot

both

**Proposed changes:**

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/ueye.py
+++ b/PYME/Acquire/Hardware/ueye.py
@@ -63,6 +63,12 @@ class UEyeCamera(Camera):
         self.check_success(ueye.is_InitCamera(self.h, None))
         self.initialized = True
 
+        # get serial number
+        cam_info = ueye.CAMINFO()
+        self.check_success(ueye.is_GetCameraInfo(self.h, cam_info))
+        # see PYME.Acquire.Hardware.Camera, want this forced to str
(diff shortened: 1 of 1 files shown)


csoeller:
Hi @barentine, I have a number of changes in a yet to be submitted PR, that goes a bit further with getting `ueye.py` up to the level of `uc480` module, see the https://github.com/csoeller/python-microscopy/tree/ueye-tweaking branch. Shall we perhaps look into merging our efforts?

David-Baddeley:
@barentine - if Christians changes in #1166 work for you then I'll mark this as a duplicate and close. The only substantive difference is the `IS_RECT()` handling. I'm not really sure what the defining differences here are.

csoeller:
Unless I overlooked something, the difference in the `IS_RECT()` assignments is an additional use of an `ueye.int()` explicit conversion. This may not be needed, I think I picked it up from an example in the `pyueye` docs. Certainly these make explicit that these will be ultimately converted to appropriate ints and performance impact is negligible.

David-Baddeley:
@csoeller - you are correct, I just glanced at the PRs side by side and somehow though there was more of a difference

David-Baddeley:
Duplicate of #1166

barentine:
Yep, sorry for not replying. Thanks both

PR #1170, opened 2021-12-13
issue762

### LINKED ISSUE #762: [ENHANCEMENT] Ability to set layer display defaults in visgui
When I visualize data in visgui, I pretty much always use the same LUTs, point sizes, alphas , pointsprites, etc. It would be really nice to be able to choose my own default visualization settings (instead of having it always be points, gist_rainbow, 30 nm points,  colored in time, etc). Additionally, it would be even better to have the option to do this for multiple layers. So for example, when I first load up my data, it will display a layer colored in time with pointsprites, 10 nm points, 0.1 alpha, and Jet LUT and another layer colored in Z with pointsprites, 10 nm points, 0.1 alpha, and gist_rainbow LUT. Then I can easily toggle on/off which layer I want to see without having to spend time setting each of them up.

David-Baddeley:
An initial attempt at allowing configurable layer defaults. Fixes #762


David-Baddeley:
image

PR #1171, opened 2021-12-15
first attempt at adding new clumpEdge property in findClumps

csoeller:
Enhancement to add clump edgeness to chaining functionality.

Enhancement as discussed with @David-Baddeley.

**Proposed changes:**

Chaining now returns via the `with_clumps` datasource a property `clumpEdge` which is nonzero for all edges in a clump. Edgeness is in the temporal sense, i.e. first and last event in a clump.

This can be useful when one wants to filter out those events for certain procedures, such as the SIMPLER procedure that we are currently testing.

### DIFF
--- a/PYME/Analysis/Tracking/trackUtils.py
+++ b/PYME/Analysis/Tracking/trackUtils.py
@@ -386,13 +386,18 @@ def findTracks2(datasource, rad_var='error_x', multiplier='2.0', nFrames=20, min
     
     numPerClump, b = np.histogram(clumpIndices, np.arange(clumpIndices.max() + 1.5) + .5)
     
+    v, edges = calcTrackVelocity(x[I], y[I], clumpIndices[I], t.astype('f')[I])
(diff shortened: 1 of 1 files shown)


PYME/Analysis/Tracking/trackUtils.py
```diff
@@ -452,7 +458,16 @@ def calcTrackVelocity(x, y, ci, t):
     #reorder
     v[I] = (1.0*v.copy())
 
-    return v
+    #edges
+    edges[1:] += (1.0-mask) # here we are looking for positive steps (i.e. np.diff(ci) >= .5)
+    edges[:-1] += (1.0-mask)
```
csoeller, line 463: possibly 'edges = (1.0 - w) > 0.5' would be even more compact but I am a little concerned that the meaning may be too obscure...

David-Baddeley:
Many thanks!

PR #1175, opened 2021-12-25
fix colourmap indexing issue in displayPane

csoeller:
Addresses issue where the updated colourmap infrastructure breaks the older colourmap selection code in `displayPane`.

Bugfix, we noticed it when running analysis with the latest code, the colourmap panel did not come up properly.

**Proposed changes:**

Adapt display panel code to be compatible with current colourmap infrastructure.

**Checklist:**

- [x ] Tested with latest PYME from git and python 3.7

### DIFF
--- a/PYME/LMVis/displayPane.py
+++ b/PYME/LMVis/displayPane.py
@@ -61,7 +61,7 @@ class DisplayPane(afp.foldingPane):
         self._pc_clim_change = False
 
         #Colourmap
-        self.cmapnames = list(colormaps.cm.cmapnames)
+        self._cmapnames = list(colormaps.cm.cmapnames)
         
         #print((cmapnames, self.glCanvas.cmap.name))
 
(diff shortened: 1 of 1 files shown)


PYME/LMVis/displayPane.py
```diff
@@ -61,7 +61,7 @@ def __init__(self, panel, glCanvas, visFr):
         self._pc_clim_change = False
 
         #Colourmap
-        cmapnames = colormaps.cm.cmapnames
```
David-Baddeley, line 64: I assume that making `cmapnames` a property of the class is to avoid an additional `list()` cast in `OnCmapChange()`??

It's a very minor point, but in the spirit of moving towards more readable code, it would be preferable to either make it private, or just do the additional cast (which should be cheap) - it would be a mistake to infer that it was safe to access the `.cmapnames` property from other classes.

csoeller: Well, I was thinking of the (possibly very unlikely but possible) case that some code injects new colour maps into the list during the lifetime of the displayPane. If we then re-evaluate the cmapnames the index could get off by one or two, unless I got the logic wrong when I thought about this (this is an ordered dict I guess, so perhaps not really an issue). Perhaps exceedingly unlikely but probably hard to debug if it occurred? So a private method would be preferable possibly, i.e. `_cmapnames`?

David-Baddeley: That seems like sound logic, making a private variable name the best choice.

Issue #1179, opened 2022-01-06
[BUG] IDS uEye camera stall-outs when changing settings/acquiring images


barentine:
**Describe the bug**
I have an IDS UI327x which runs 'fine' until I try to do something in the PYMEAcquire GUI - basically anything: changing integration time sometimes works, but toggling an ROI or acquiring a snapshot is ~gaurunteed to freeze up the camera similar to previous issue (#313). Note that with my previous IDS usage, the HTSMS focus lock, I never had to do anything with the GUI/change camera settings once loaded - I took care of setting my ROI / integration time in the init script and never needed to acquire images with it.

@csoeller is this all working fine in your hands?

**To Reproduce**
1. IDS UI327x
2. ![image](https://user-images.githubusercontent.com/31105780/148466784-3ff3ec1b-b637-4c71-9f5f-bace2b78a8b3.png)
3.
```
@init_hardware('Camera')
def cam(scope):
    from PYME.Acquire.Hardware.ueye import UEyeCamera
    cam = UEyeCamera(0, 12)
    scope.register_camera(cam, 'IDS')
```
4. acquire snapshot
5. snapshot is acquired, but camera silently freezes.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: Win10
 - Python version: 3.78
 - numpy version: 1.19
 - wxpython version [if a UI bug]: 4.0.4
 - pyueye                    4.95.0                   pypi_0    pypi

csoeller:
Hi @barentine, what you describe sounds pretty much like what I encountered. I figured out a workaround which basically involved downgrading to IDS software version 4.93. The nuts and bolts are described in my [installation writeup](https://github.com/csoeller/pyme-install-docs/blob/master/PYMEAcquire/IDS-camera-setup.md). Would be good to know if this works for you as well.

Note that (1) `pyueye` can stay on version `4.95.X` just fine (we are not using any of the new 4.95 features at the moment) and (2) it would be good to eventually tackle the question if upgrading to the 'new' 4.95 API functions for acqusition would allow us to use the latest software.

barentine:
Very helpful, @csoeller! Thank you. 

I had a crack at using the new functions for the queue handling. It's behaving a little goofy at the moment - the ImageQueue version of waitfornext is to pass a pointer to an `ImageQueueWaitBuffer` instance, which I do, but the nmemid field which we need to pass to the memcopy function never gets set - it remains a Null pointer.

[...]

PR #1180, opened 2022-01-10
ueye 4.96 driver updates, bugfixes, and output trigger options

barentine:
Addresses issue #1179 .

bugfix, but not there yet
**Proposed changes:**
-use new uEye ImageQueue methods

@David-Baddeley & @csoeller  - I think I should be instantiating an imagequeuewaitbuffer and passing it (by reference*) as shon below, however, the pointers that get instantiated do not seem to ever be filled and I just get Null pointer access violations. IDS doc shows (void*)&waitBuffer, but this appears to be handled by pyueye. Any ideas for what I'm doing wrong are appreciated!

If I include a `IS_IMAGE_QUEUE_CMD_GET_PENDING` call (and don't hard-throw the valueerror by checking contents on the pointers explicitly) it usually returns zero unless I do something to gum things up for a minute (change integration time or something). I haven't checked if in these non-zero cases the pointers get set.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/ueye.py
+++ b/PYME/Acquire/Hardware/ueye.py
@@ -602,6 +602,64 @@ class UEyeCamera(Camera):
         gain = self.GetGain()
         ret = ueye.is_SetHWGainFactor(self.h, ueye.IS_INQUIRE_MASTER_GAIN_FACTOR, gain)
         return 0.01*ret
+    
+    def SetOutputTrigger(self, mode, delay=0, width=0.0001, positive=True):
+        """
+        Set output trigger of the camera. Currently hard-coded to use GPIO1
+
+        Parameters
+        ----------
+        mode : str
+            Currently supported modes include:
+                low: constant TTL low
+                high: constant TTL high
+                flash: goes TTL high during exposure. If 0 is passed, the flash
+                    output will be active until the end of the exposure time.
+                    With global start shutter this is the time until the end of
+                    exposure for the first row.
+        delay : float, optional
+            delay after trigger event, in seconds, to emit TTL high (assuming 
+            posiive polarity), by default 0 s.
+        width : float, optional
+            TTL high pulse width, in seconds, by default 0.0001 s, or 0.1 ms
+        positive : bool, optional
+            CURRENTLY IGNORED Sets polarity of the output trigger to positive 
+            (True) or negative (False). True, by default.
+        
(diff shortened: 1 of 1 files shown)


csoeller:
Hi @barentine, thanks for trying. I have no immediate suggestion what to change and right now have been happy to work with 4.93 and the existing interface. We may get back to it at some stage, perhaps somebody has figured things out by then, but will also try again when I have another PC to play with. Low and behold, we could also ask IDS for an example that uses the new queue handling API from pyueye.

csoeller:
Just a brief update about a possibly fragile state of ueye backend behaviour. On one PC I have encountered timeouts in acquisition that became progressively more frequent (over minutes) until the camera stopped providing any new frames. I have since then swapped the cameras I have between PCs and the other camera now works fine on the previously problematic PC. The camera that froze works fine on the PC I moved it to and shows none of the timeout behaviour any more.

Not sure what the explanation is and if related to the other issues that we have seen with freezing upon ROI changing etc. I keep also wondering if this could be related to USB connecions, which USB bus on etc. I will experiment a little more if we keep running into issues.

I suppose it suggests that it would be good to see if we can make the new 4.95 queue interface work and if it resolves other instabilities.

barentine:
@csoeller it needs some clean up/doc/and testing, but I just pushed changes that let me run the camera with the new 4.96.1 ueye driver, and the new 4.96.952 pyueye package. This uses the `ImageQueue` method and `IMAGEQUEUEWAITBUFFER` business.

There's some more I'd like to do to the ueye cam (see #1282), but wanted to be working on the current API

barentine:
Will push updates on Monday, when I'm at the microscope, but I have things running much more robustly. Biggest fix, as mentioned in #1282, is using `self.n_full >0` instead of `not full_buffers.empty()` in the `ExpReady` method. Combined with setting `n_full = 0` when clearing the buffers, this avoids a messy race condition where the `frameWrangler` polling thread thinks there's a frame when there isn't and can take out the current frame lock. While the poll thread has the lock (which we also need to stop/start the frameWrangler), it eventually hits the `ueye` `ExtractColor` where we call `full_buffers.get()`, a blocking call with no timeout! So we freeze up the display and provide no error message.

csoeller:
Very cool, we will aim to check it out in due course.

PR #1149, opened 2021-11-19
Annotation tidy up

David-Baddeley:
Some progress on #252, adds deletion of annotations. Incidentally fixes #1148

PR #1183, opened 2022-01-12
set scalebar size in PYMEAcquire

barentine:
Addresses issue: after #1149 (which in general I'm really appreciating) the array view overlays are now drawn in PYMEAcquire (they would have to be toggled off by something like `MainFrame.vp.overlays[0].visible = False`). I actually like this, and would eventually enjoy seeing an overlay option panel (or really just a scalebar toggle added to the current 'display') but the voxelsize never gets set, so the scalebar is currently drawn at 2000 pixels = 1 um. 

temporary bugfix I guess
**Proposed changes:**
- grab pixelsize from the scope and set the view panel voxelsize with it

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/acquiremainframe.py
+++ b/PYME/Acquire/acquiremainframe.py
@@ -190,9 +190,10 @@ class PYMEMainFrame(AUIFrame):
             if 'vp' in dir(self):
                     self.vp.SetDataStack(self.scope.frameWrangler.currentFrame)
             else:
-                self.vp = arrayViewPanel.ArrayViewPanel(self, self.scope.frameWrangler.currentFrame)
-                self.vp.crosshairs = False
-                self.vp.showScaleBar = False
+                vs_x_um, vs_y_um = self.scope.GetPixelSize()
+                vs = MetaDataHandler.VoxelSize(1000 * vs_x_um, 1000* vs_y_um, z=1)
+                self.vp = arrayViewPanel.ArrayViewPanel(self, self.scope.frameWrangler.currentFrame,
+                                                        voxelsize=vs)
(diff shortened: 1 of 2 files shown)


barentine:
@David-Baddeley feel free to shamelessly reject this. I realize the cleaner thing would be to just fix the regression by looping trough the overlays list and setting the `visible` attribute to `False`, but it just so happens that I'm doing some alignment today and the scalebar is kind of handy ;)

David-Baddeley:
Agree that it would be nice to have the pixel size set (even if scale bar not displayed by default in PYMEAcquire), but changes as envisaged will cause issues in a couple of places:

1) 2-camera set ups (voxel size is specific to the camera, so if you are switching between cameras on different sideports you need to change the voxel size in the view as well)

2) initial start up on a new install (before you've set the voxel size) - this will prevent the display from showing. The initial calibration procedure (image an object micrometer and use that to calibrate the voxel size) requires you to be able run the acquisition without having a voxel size set.

David-Baddeley:
In the medium term, it might make sense to move the pixel size info out of the view panel and into either display options or just refer to the underlying data source ... as it is there are a few other places where setting pixel size on panel creation is broken (most notably loading data with an incorrect pixel size, and then correcting it in the metadata once open - in this case the scale bar will not update correctly).

David-Baddeley:
I think the best solution which avoids the above issues is probably to just disable the scalebar in PYMEAcquire for now (although #1185 should allow this to be added back in more easily whilst addressing the multiple camera issue). 

Ultimately I can see a calibration overlay as something which could be useful, potentially displaying a grid rather than a scalebar and/or multiview ROIs.

PR #1191, opened 2022-01-25
int + str fix for sliding window background estimation module

barentine:
Addresses issue cannot add int + str, which sometimes happens after configuring the recipe module in PYMEImage recipes tab (windows 10, wx 4.0.4)

bugfix
**Proposed changes:**
-hard type window as listint

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -6,7 +6,7 @@ Created on Mon May 25 17:15:01 2015
 """
 
 from .base import ModuleBase, register_module, Filter
-from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI, ListInt
+from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI, CInt
 
 #try:
 #    from traitsui.api import View, Item, Group
@@ -1942,7 +1942,7 @@ class BackgroundSubtractionMovingAverage(ModuleBase):
     """
 
     input_name = Input('input')
-    window = ListInt([-32, 0, 1])
+    window = List(type=CInt, value=[-32, 0, 1])
     output_name = Output('background_subtracted')
 
     percentile = 0
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -6,7 +6,7 @@ Created on Mon May 25 17:15:01 2015
 """
 
 from .base import ModuleBase, register_module, Filter
-from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI
+from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI, ListInt
 
 #try:
 #    from traitsui.api import View, Item, Group
@@ -1942,7 +1942,7 @@ class BackgroundSubtractionMovingAverage(ModuleBase):
     """
 
     input_name = Input('input')
-    window = List([-32, 0, 1])
+    window = ListInt([-32, 0, 1])
     output_name = Output('background_subtracted')
 
     percentile = 0


David-Baddeley:
Could this cause issues with previously saved recipes? Would be worth checking (there are backwards compatible workarounds if this is the case - i.e. leaving as a list and explicitly casting to int on access, but the approach in this PR would be cleaner if it works)

barentine:
@David-Baddeley good thought but I think we're fine
```
In [1]: from PYME.recipes.traits import List, ListInt

In [2]: List([-32.0, 0.0, 1.0])
Out[2]: <traits.trait_types.List at 0x26702d72cc8>

In [3]: ListInt([-32.0, 0.0, 1.0])
Out[3]: <traits.ctrait.CTrait at 0x2677f6a81a8>
```

David-Baddeley:
think that's the incorrect test ... The below is more relevant if things are getting parsed as string from a saved recipe:

```python
In [5]: from PYME.recipes.traits import List, ListInt, HasTraits

In [6]: class Foo(HasTraits):
   ...:     l = ListInt()

In [7]: f = Foo(l=["1", "2", "3"])
TraitError                                Traceback (most recent call last)
Input In [7], in <module>
----> 1 f = Foo(l=["1", "2", "3"])

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/trait_types.py:2699, in List.validate(self, object, name, value)
   2696     if object is None:
   2697         return value
-> 2699     return TraitListObject(self, object, name, value)
   2701 self.error(object, name, value)

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/trait_list_object.py:582, in TraitListObject.__init__(self, trait, object, name, value)
    579 value = list(value)
    580 self._validate_length(len(value))
--> 582 super().__init__(
    583     value,
    584     item_validator=self._item_validator,
    585     notifiers=[self.notifier],
    586 )

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/trait_list_object.py:213, in TraitList.__init__(self, iterable, item_validator, notifiers)
    211 if item_validator is not None:
    212     self.item_validator = item_validator
--> 213 super().__init__(self.item_validator(item) for item in iterable)
    214 if notifiers is not None:
    215     self.notifiers = list(notifiers)

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/trait_list_object.py:213, in <genexpr>(.0)
    211 if item_validator is not None:
    212     self.item_validator = item_validator
--> 213 super().__init__(self.item_validator(item) for item in iterable)
    214 if notifiers is not None:
    215     self.notifiers = list(notifiers)

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/trait_list_object.py:865, in TraitListObject._item_validator(self, value)
    862     return value
    864 try:
--> 865     return trait_validator(object, self.name, value)
    866 except TraitError as excp:
    867     excp.set_prefix("Each element of the")

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/trait_handlers.py:146, in TraitCoerceType.validate(self, object, name, value)
    143         return fv[1](value)
    145 # Otherwise, raise an exception:
--> 146 self.error(object, name, value)

File ~/opt/miniconda3/envs/pyme_as/lib/python3.8/site-packages/traits/base_trait_handler.py:74, in BaseTraitHandler.error(self, object, name, value)
     57 def error(self, object, name, value):
     58     """Raises a TraitError exception.
     59 
     60     This method is called by the validate() method when an assigned value
   (...)
     72         The proposed new value for the attribute.
     73     """
---> 74     raise TraitError(
     75         object, name, self.full_info(object, name, value), value
     76     )

TraitError: Each element of the 'l' trait of a Foo instance must be a value of class 'int', but a value of '1' <class 'str'> was specified.

```

David-Baddeley:
Turns out `ListInt` is deprecated, and you are supposed to do `List(Int)` instead. This means we can do `List(CInt)` e.g.

```python
In [12]: class Foo(HasTraits):
    ...:     l = List(CInt)

In [13]: f = Foo(l=["1", "2", "3"])

In [14]: 

```

PYME/recipes/processing.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from .base import ModuleBase, register_module, Filter
-from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI
+from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI, ListInt
```
```suggestion
from PYME.recipes.traits import Input, Output, Float, Enum, CStr, Bool, Int, List, FileOrURI, CInt
```

PYME/recipes/processing.py
```diff
@@ -1942,7 +1942,7 @@ class BackgroundSubtractionMovingAverage(ModuleBase):
     """
 
     input_name = Input('input')
-    window = List([-32, 0, 1])
+    window = ListInt([-32, 0, 1])
```
```suggestion
    window = List(type=CInt, value=[-32, 0, 1])
```

barentine:
Thanks for the fix, David!

PR #1192, opened 2022-01-26
some experimental mesh optimisations

David-Baddeley:
still not quite working (occaisional segfaults) ...  but concept is to move allocation outside the loop over the edges so we can eventually parallise it


zacsimile:
This works for me and is faster. I recommend a merge.

PR #1193, opened 2022-01-28
add multiview classes for Hammy Fusion and uEye cameras

barentine:
Addresses issue # .

enhancement
**Proposed changes:**
- add multiview class for hamamatsu fusion
- add multiview class for ueye cameras

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
Any feedback on this @David-Baddeley ? I've got more uEye stuff coming in and being lazy am just tempted to pile it on.

David-Baddeley:
Gut feeling is that we should find a way of re-factoring this so that we don't need camera type specific multiview classes - e.g create a multiview camera that sits as a separate layer on top of an existing camera instance. i.e.

```python
_cam = HamamatsuOrca(...)
cam = MultiviewCamera(_cam)
```

Could accept on a temporary basis with comments to the above

PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
```diff
@@ -442,3 +442,11 @@ def __init__(self, camNum, multiview_info):
         # default to the whole chip
         default_roi = dict(xi=0, xf=2048, yi=0, yf=2048)
         MultiviewCameraMixin.__init__(self, multiview_info, default_roi, HamamatsuORCA)
+
+
+class MultiviewFusion(MultiviewCameraMixin, Fusion):
```
David-Baddeley, line 448:
```suggestion
#TODO - replace MultiviewCameraMixin with a Multiview wrapper so that we don't need to have explicit multiview versions of all cameras.
class MultiviewFusion(MultiviewCameraMixin, Fusion):
```

PYME/Acquire/Hardware/ueye.py
```diff
@@ -534,3 +534,10 @@ def _GetDeviceInfo(self):
         self.check_success(ueye.is_DeviceInfo(self.h,ueye.IS_DEVICE_INFO_CMD_GET_DEVICE_INFO,
                                               dev_info,ueye.sizeof(dev_info)))
         return dev_info
+
+class MultiviewUEye(MultiviewCameraMixin, UEyeCamera):
```
David-Baddeley, line 539:
```suggestion
#TODO - replace MultiviewCameraMixin with a Multiview wrapper so that we don't need to have explicit multiview versions of all cameras.
class MultiviewUEye(MultiviewCameraMixin, UEyeCamera):
```

Issue #1198, opened 2022-02-10
[ENHANCEMENT] some work towards removing wx deprecation warnings


csoeller:
**Please describe.**
Currently, I am still getting tons of messages like this:

```
  img = wx.ImageFromData(ima.shape[1], ima.shape[0], ima.ravel())
c:\python-support-files\pyme-py37\PYME\DSView\scrolledImagePanel.py:52: wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  MemBitmap = wx.EmptyBitmap(s.GetWidth(), s.GetHeight())
c:\python-support-files\pyme-py37\PYME\DSView\arrayViewPanel.py:432: wxPyDeprecationWarning: Call to deprecated item BitmapFromImage. Use :class:`wx.Bitmap` instead
  im2 = wx.BitmapFromImage(im)
c:\python-support-files\pyme-py37\PYME\DSView\arrayViewPanel.py:947: wxPyDeprecationWarning: Call to deprecated item ImageFromData. Use :class:`Image` instead.
  img = wx.ImageFromData(ima.shape[1], ima.shape[0], ima.ravel())
c:\python-support-files\pyme-py37\PYME\DSView\scrolledImagePanel.py:52: wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  MemBitmap = wx.EmptyBitmap(s.GetWidth(), s.GetHeight())
c:\python-support-files\pyme-py37\PYME\DSView\arrayViewPanel.py:947: wxPyDeprecationWarning: Call to deprecated item ImageFromData. Use :class:`Image` instead.
  img = wx.ImageFromData(ima.shape[1], ima.shape[0], ima.ravel())
c:\python-support-files\pyme-py37\PYME\DSView\arrayViewPanel.py:432: wxPyDeprecationWarning: Call to deprecated item BitmapFromImage. Use :class:`wx.Bitmap` instead
  im2 = wx.BitmapFromImage(im)
c:\python-support-files\pyme-py37\PYME\DSView\scrolledImagePanel.py:52: wxPyDeprecationWarning: Call to deprecated item EmptyBitmap. Use :class:`wx.Bitmap` instead
  MemBitmap = wx.EmptyBitmap(s.GetWidth(), s.GetHeight())
[...]

PR #1199, opened 2022-02-11
Wx deprecation fixes

csoeller:
Addresses issue #1198.

bugfix for a bug waiting to happen (which is what deprecation warnings eventually morph into).

**Proposed changes:**
Replace all deprecated `EmptyBitmap` and similar calls so that newer interfaces are used.

**State of this PR:**
This is submitted basically as a proof-of-principle. Primary goal is to elicit some feedback and agree on the best way to solve the issue.

Things that may still need doing:

1. agree if we need a backwards compatible solution or not.
2. if backwards compatibility is a thing the compatibility layer adopted here seems like an ok approach. The naming and placement of such a module can obviously be adjusted (currently `PYME.ui.wx_compat`).
3. not all occurences of the deprecated statements have been exhaustively changed. So far I did the ones that came up in my tests but presumably we should go through the whole dist.
4. probably more but first of all hoping to kick off a brief discussion.

For completeness, the approach used here I came across in [OSGeo/grass](https://github.com/OSGeo/grass) which itself is GPL2 (I think).

**Checklist:**

This has been tested on a production system with `wx` 4.x.


csoeller:
All appearances of `EmptyBitmap`, `ImageFromData` and `BitmapFromImage` etc have now been replaced with `wx_compat` calls. If we eventually adopt a different way to fix these calls we can find and replace them in all files with one query-replace through the whole directory tree.

PR #1206, opened 2022-02-17
Piezo C867 improvements - servo toggling and error messages

csoeller:
Enhancement to allow servo toggling by the user so that no actuation occurs during imaging. Also uses textual error messages if the user has installed the [`PIPython`](https://pypi.org/project/PIPython/) package.

**Proposed changes:**
The main changes are not user visible to any large degree. Essentially, we have moved the actual servo mode switching into the thread loop and just message into the thread loop using flags, similar to the `stopMove` flag.

Note that the original `SetServo` and thus the `joystick` implementation hung PYME in our tests, prompting us to adopt the  thread loop messaging approach. Also tidied the thread loop to only issue stage commands compatible with the current servo state. Initial testing suggests this now works fine.

We also support proper full text error messages if the [`PIPython`](https://pypi.org/project/PIPython/) has been installed by the user which is available via [PyPI](https://pypi.org). This was very helpful during debugging and testing.

Testing and small tweaks are still ongoing, but I wanted to raise the PR to elicit potential comments and feedback.

**Checklist:**

Ongoing testing with latest PYME git, on Win 10.

- [x] Does this change how users interact with the software? How will these changes be communicated? No user level changes.
- [x] Does this maintain backwards compatibility with old data? Yes.
- [x] Does this change the required dependencies? No, should all work as before.
- [x] Are there any other side effects of the change? Not to our knowledge.

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_c867.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_c867.py
@@ -304,8 +304,10 @@ class piezo_c867T(PiezoBase):
                         logger.debug('Setting stage target pos: %s' % pos)
                         time.sleep(.01)
                     
-                # check to see if we're on target - SERVOCHECK: CS: seems to me this only makes sense when servo is on - check manual
-                # from the manual on 'ONT?' command: "The detection of the on-target state is only possible in closed-loop operation (servo mode ON)"
+                # check to see if we're on target
+                # from the manual on 'ONT?' command:
+                #   "The detection of the on-target state is only possible in closed-loop operation (servo mode ON)"
+                # note that the cleaned up on-target check does not use the 'ONT?' command any more
                 if self.servo:
                     self.onTarget = np.allclose(self.position, self.targetPosition, atol=self.ptol)
                 else: # servo is off
--- a/PYME/Acquire/Hardware/Piezos/piezo_c867.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_c867.py
@@ -216,6 +216,9 @@ class piezo_c867T(PiezoBase):
         self.targetVelocity = self.velocity.copy()
         
         self.lastTargetPosition = self.position.copy()
+
+        self.disableServo = False
+        self.enableServo = False
         
         self.lock = threading.Lock()
         self.tloop = threading.Thread(target=self._Loop)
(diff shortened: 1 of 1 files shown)


PYME/Acquire/Hardware/Piezos/piezo_c867.py
```diff
@@ -275,20 +304,15 @@ def _Loop(self):
...
-                res1 = self.ser_port.readline()
-                ont2 = int(res1.split(b'=')[1]) == 1
+                # check to see if we're on target
+                # from the manual on 'ONT?' command:
+                #   "The detection of the on-target state is only possible in closed-loop operation (servo mode ON)"
+                # note that the cleaned up on-target check does not use the 'ONT?' command any more
+                if self.servo:
+                    self.onTarget = np.allclose(self.position, self.targetPosition, atol=self.ptol)
+                else: # servo is off
+                    self.onTarget = False
```
David-Baddeley: Seeing as we're effectively doing soft servoing, I don't think we should be reporting `onTarget=False` if we are on target, even if the servo is disabled (there is no reason to).

PYME/Acquire/Hardware/Piezos/piezo_c867.py
```diff
@@ -310,16 +334,18 @@ def close(self):
...
-            self.servo = state == 1
-        finally:
-            self.lock.release()
-            
+    # directly enabling/disabling the servo outside the thread loop
+    #      seems to hang PYME
+    # the implementation below instead 'messages' the thread loop by setting flags
+    #      that get picked up in thread loop processing
+    # I dislike state being an int and rather have opted treating as bool which reflects the on/off state better IMHO
+    def SetServo(self, state=True):
```
David-Baddeley: We would need to check with other piezos / stages that might use more than 1 state. 

Although a boolean might ultimately make sense, the `SetServo` method is part of more general piezo/stage interface, so we can't change it arbitrarily here - we would need to check all `SetServo` usages across other piezos / stages and all calling functions.

PYME/Acquire/Hardware/Piezos/piezo_c867.py
```diff
@@ -310,16 +334,18 @@ def close(self):
...
+        if self.servo and not state: # need to toggle servo off
```
David-Baddeley: I don't like setting two separate flags (which then get modified in the polling thread) for turning the servo on or off. This is neither thread safe, nor particularly clear.

a better approach would be to adopt the approach used for the positions, and just set a `._target_servo_state` variable when called outside the communications thread. Inside the thread, we could check if the current servo state (or a cached local variable recording the last commanded state) is the same as `_target_servo_state` and change the servo state if it is not. This is thread-safe, as the value only ever gets written from outside the communications thread.

PYME/Acquire/Hardware/Piezos/piezo_c867.py
```diff
@@ -237,11 +247,17 @@ def _Loop(self):
...
+                    else:
+                        logger.error('Stage Error: %d' % self.errCode)
                 
                 #print self.targetPosition, self.stopMove
                 
-                if self.stopMove:
+                if self.stopMove: # SERVOCHECK: check if this is ok to process when servo is off!!!
+                    # note that issueing the HLT command sets an error condition, from the manual: "Error code 10 is set."
+                    # question: should we check the error status to unset the error code from such a HLT command?
+                    logger.debug('piezo_c867T: stopMove issuing HLT command, expect error 10 to be set')
```
David-Baddeley, line 260: This will spam the console with warning messages whenever a soft-joystick (e.g. 3D mouse etc ...) is used.

David-Baddeley: Although I think some of the existing debug messages will as well ...

Issue #1204, opened 2022-02-16
[SUPPORT] further rationalising logging in PYME?


csoeller:
**Issue:**

While there is now good support for configurable logging from `PYMEAcquire`, some issues remain, that are at least a little obscure to me in (a) if they be solved by configuration and (b) remaining potential inconsistencies.

**Issue a:**

`matplotlib` leaves substantial footprints in the log file when at debug level (happens with the default logging as set in the repo). I cannot suss out any easy way to get rid of these using just a logging config file. What does work is explicitly forcing matplotlib loggers to a higher debug level, i.e. the following one-line patch to `PYMEAcquire.py`:

```python
    else:
        logging.basicConfig(level=default_level)

    # PATCH line below: we want to somehow supress the level of matplotlib messages below debug level
    logging.getLogger('matplotlib').setLevel(logging.WARN)


class BoaApp(wx.App):
    def __init__(self, options, *args):

```

Can we add something like this, or via some other route, to supress `matplotlib` debug messages which are really noise within `PYMEAcquire` logs?

**Part b in next comment...**

csoeller:
**Issue b:**

While PYME now has greatly improved logging there are still a few inconsistencies where I am uncertain if this is by design or remnants from earlier itereations of the code. Case in point, in `HTTPTaskPusher.py` we have a mixture of `logging.XXX` and `logger.XXX` calls, e.g.

```python
...
    def fileTasksForInputs(self, **kwargs):
        from PYME.IO import clusterIO
        input_names = kwargs.keys()
        inputs = {k : kwargs[k] if isinstance(kwargs[k], list) else clusterIO.cglob(kwargs[k], include_scheme=True) for k in input_names}

        numTotalFrames = len(list(inputs.values())[0])
        self.currentFrameNum = 0

        logger.debug('numTotalFrames = %d' % numTotalFrames)
        logger.debug('inputs = %s' % inputs)

        while numTotalFrames > (self.currentFrameNum + 1):
            logging.debug('we have unpublished frames - push them')

...
```

The effect of this is that `logging.XXX` generally goes to stdout/stderr and bypasses any settings from the logging config file. This is a little annoying as when I consult the log file these statements are all missing and I am unclear if this is desired. I.e. should we loose these `logging.XXX` statements completely or convert them to `logger.XXX` calls?

**Issue c:**

This is pretty much a continuation of issue b) above, it refers to bunches of remaining `print()` statements in the code that pollute, for example

[...]

PR #1208, opened 2022-03-03
Pymeacquire logging tweaks

csoeller:
Addresses issue #1204 (partially, i.e. only for `PYMEAcquire` and `matplotlib` noise).

more enhancement

**Proposed changes:**
Tell the `matplotlib` loggers to stay at higher logging level (here chosen `logging.ERROR`).

**Checklist:**

- [x] Tested on python 3.7 and latest git

### DIFF
--- a/PYME/Acquire/PYMEAcquire.py
+++ b/PYME/Acquire/PYMEAcquire.py
@@ -66,6 +66,9 @@ def setup_logging(default_level=logging.DEBUG):
     else:
         logging.basicConfig(level=default_level)
 
+    # we want to somehow supress the level of matplotlib messages below debug level
+    logging.getLogger('matplotlib').setLevel(logging.WARN)
+
 
 class BoaApp(wx.App):
     def __init__(self, options, *args):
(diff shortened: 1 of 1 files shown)


PYME/Acquire/PYMEAcquire.py
```diff
@@ -66,6 +66,9 @@ def setup_logging(default_level=logging.DEBUG):
     else:
         logging.basicConfig(level=default_level)
 
+    # we want to somehow supress the level of matplotlib messages below debug level
+    logging.getLogger('matplotlib').setLevel(logging.ERROR) #clobber unhelpful matplotlib debug messages
```
csoeller, line 70: Note that I tried the `dsviewer` approach of just silencing selected `matplotlib` packages, i.e.

```python
logging.getLogger('matplotlib.font_manager').setLevel(logging.ERROR) #clobber unhelpful matplotlib debug messages
logging.getLogger('matplotlib.backends.backend_wx').setLevel(logging.ERROR)
```
but this still left a lot of debug garbage, so resorted to silencing at the top level. This works now fine, one can discuss if setting to `logging.ERROR` or `logging.WARN`.

David-Baddeley: hmm - I think the targeted silencing was initially because the top-level silencing didn't work  - maybe that's now fixed in matplotlib.

Issue #1207, opened 2022-03-02
[ENHANCEMENT] photoelectrons/second PYMEImage hist option


barentine:
**Is your feature request related to a problem? Please describe.**
arguably better unit than ADU, of course, though I can see how it spirals quickly - maybe someone really wants to look at (photoelectrons/s)/(kW/cm^2) or whatever. Still, I think e-/s is straightforward if you've got basic metadata present, camera maps or not, and is something I want to semi-automate to save myself from manually typing my calc in the shell every time I go to slap an image into a presentation / do a calculation.

**Describe the solution you'd like**
drop down to select 'ADU', 'e-/s' in PYMEImage display settings. If e-/s is selected, could consider having a camera map entry box pre-populated from metadata like we do in LMAnalyser2 gui panel.

**Describe alternatives you've considered**
doing it in the shell, manually

**Are you prepared to work on it?**
absolutely, going to do it regardless, but would be great to do it in a pre-approved manner in case others would like it (like my labmates, who would like it).

David-Baddeley:
Can definitely see how it could snowball with metadata / camera map dependencies, but not fundamentally opposed

To clarify - do you just want to display a histogram of photo-electrons (would the need be served by, e.g. a menu function for "photoelectron histogram", or do you want to adjust scaling in photoelectrons etc ...?). 

The 3 ways I can see of addressing this would be ...

a) a menu function to display a histogram of image photoelectrons/s, leaving everything else in ADUs (this is probably the simplest, but least satisfactory way)
b) changing the display scaling histogram (I think this is what you are hinting at) with a drop down 
c) a "convert this image to photo-electrons/second" menu function which opens a new (potentially shallow - i.e. a datasource that re-scales the original data as you access it) copy of the image with the data re-scaled to photoelectrons/s

I think b) has the potential to get really messy quite quickly:
- changing the units in the display settings would not change the units in, e.g. line profiles, or any subsequent analysis (region detection etc) unless a lot of additional special case logic was added. This could lead to some pretty counter-intuitive user experiences.
- I can see a lot of people being interested in slightly different rescalings (e.g. photoelectrons/frame rather than /s which might be more relevant when looking at noise performance, even or ADUs/s), which means that you could end up with several different options and associated spaghetti

[...]

PR #1212, opened 2022-03-28
electrons/second datasource and ADU to e/s recipe module

barentine:
Addresses issue #1207 .

enhancement
**Proposed changes:**
- thinly wrap FlatFieldDataSource in a new datasource to additionally convert to electrons/second
- add a recipe module interface for this new datasource

Thanks for the conversation in #1207 @David-Baddeley and @csoeller. I think electrons/second is sort of the holy as far as Poisson is concerned, and being what I need at the moment I just wrote that, but it would be simple to add an option to choose whether we output was just photoelectrons instead of electrons/second. It just starts to get bigger, and maybe uglier unless some real time is spent on it (I'm sure we can all envision a 'units' datasource snowballing a bit).

Still some metadata questions - thanks @csoeller for pointing these out in your mapTools.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/IO/DataSources/ElectronsPerSecondDataSource.py
+++ b/PYME/IO/DataSources/ElectronsPerSecondDataSource.py
@@ -1,9 +1,9 @@
 #!/usr/bin/python
 
 ##################
-# HDFDataSource.py
+# ElectronsPerSecondDataSource.py
 #
-# Copyright David Baddeley, 2009
+# Copyright Andrew Barentine, David Baddeley, 2022
 # d.baddeley@auckland.ac.nz
 #
 # This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


barentine:
I'm finding myself putting this on multiple computers, not all of them mine. Mind taking a look at it @David-Baddeley ?

David-Baddeley:
Probably doesn't hurt to merge, but leaves a few things which still need resolving

- The `Units` metadata entry seems like a reasonable way of flagging "Don't do another conversion on this"
- To be really useful, however, we would need to look at units in a bunch of other places (localisation fitting, deconvolution, NPhotons conversion from localisation data, potentially also elsewhere). This could be pretty tricky to find and do properly.
- I'm not quite ready to commit to the `Units` metadata entry globally (at present it seems like a good idea, but it would impact so many areas of the code I think it warrants a little more thought).

Will add a couple of comments / expand the docstring and then merge.

David-Baddeley:
I've added a bunch of stuff to the docstrings, but am wondering if we shouldn't just settle on `Units.Intensity` rather than `Units` for the metadata entry, as something I could support as a backwards compatible convention.

barentine:
I'd say go for it! `Units.Intensity` is fine by me. Alternatively, maybe even just `IntensityUnits`

PR #1218, opened 2022-04-08
Deprecate multiview mixin

David-Baddeley:
Change multiview code to be a wrapper, rather than a mixin, removing the need for explicit multiview classes for different camera types


David-Baddeley:
This is an attempt to reduce the maintenance burden from having to have multiview versions of all the cameras. See also discussion on #1193.

@barentine - do you see any obvious issues with this? For now it's backwards compatible (i.e. I haven't set a date for removing the mixin), but it should make things a bit easier/cleaner going forwards.

PR #1219, opened 2022-04-08
Docker containers for PYME's distributed servers

zacsimile:
This is a Docker file and GitHub Action that creates a container for PYME distributed servers and uploads it to https://hub.docker.com/. Tested and working in Docker on Ubuntu 20.04.4 LTS. 

One uploaded (or built locally with `docker build -t pyme-server .`), a server node can be brought up with

`docker run -dp 9000:9000 pyme-server`

and the main  node can be brought up with 

`docker run -dp 9000:9000 pyme-server --main`.

Note the `-d` flag runs in detached mode and `-p 9000:9000` maps port 9000 of the Docker image to port 9000 of the host machine (more info at https://docs.docker.com/get-started/02_our_app/).

Still on the to-do list...

- [x] Enable GPU capability by building with https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/overview.html, and add https://github.com/python-microscopy/pyme-warp-drive.
- [x] PZF is currently disabled due to whatever is causing https://github.com/python-microscopy/pymecompress/issues/9.
- [ ] Properly package `pymecompress` and convert the Docker to conda package installs only.

This is a WIP, and thoughts on how to do this better are most welcome. Feel free to try on your home machine :).

### DIFF
--- a/.github/workflows/make_release.yml
+++ b/.github/workflows/make_release.yml
@@ -47,10 +47,6 @@ jobs:
         -H "Accept: application/vnd.github.v3+json" \
         -u ${{ secrets.REC_REPO_TOKEN }}\
         -d '{"ref":"ref"}'
-    
-  docker-server:
-    runs-on: ubuntu-latest
-    steps:
     - name: Checkout
       uses: actions/checkout@v2
     - name: Set up QEMU
@@ -67,4 +63,4 @@ jobs:
       with:
         context: .
         push: true
-        tags: user/app:latest
+        tags: python-microscopy/pyme-server:latest
--- a/.github/workflows/make_release.yml
+++ b/.github/workflows/make_release.yml
@@ -48,3 +48,23 @@ jobs:
         -u ${{ secrets.REC_REPO_TOKEN }}\
         -d '{"ref":"ref"}'
     
+  docker-server:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
+    - name: Set up QEMU
+      uses: docker/setup-qemu-action@v1
+    - name: Set up Docker Buildx
+      uses: docker/setup-buildx-action@v1
+    - name: Login to DockerHub
+      uses: docker/login-action@v1
+      with:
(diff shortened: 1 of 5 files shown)


zacsimile:
`nvidia-container-toolkit` (https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/overview.html) must be installed on the Docker host machine for this to work. Launch the GPU-enabled Docker server (main node uses `--main` flag, other nodes do not have this flag) with

`docker run --rm -dp 9000:9000 --gpus all pyme-server --main`

barentine:
Looks like we might also be able to run the multi-process service. Wild. Would be neat from a performance standpoint, but if its a headache its a headache
https://github.com/NVIDIA/nvidia-docker/issues/419

barentine:
Also big thanks for doing this Zach!

zacsimile:
Would be cool. And no problem. Once this is merged/built on our own Docker Hub account (presumably `python-microscopy`? or maybe `david_baddeley` for continuity?) we can write up the instructions for install, which will be something like.

### Before you begin

Note that this could create some big security holes in your network. We recommend running this on a separate, private network connected only to your microscope computer.

### Requirements

- GNU/Linux x86_64 with kernel version > 3.10
- Docker >= 19.03
- NVIDIA GPU with Architecture >= Kepler (or compute capability 3.0)
- NVIDIA Linux drivers

### Installation

1. Install NVIDIA container toolkit (https://docs.nvidia.com/datacenter/cloud-native/container-toolkit/overview.html).
2. Run `docker pull python-microscopy/pyme-server:latest`.

### Run
1. Enter `docker run --rm -d --gpus all pyme-server --main` to launch the main node of a PYME server array.
2. For each additional machine you wish to bring into the server array, enter `docker run --rm -d --gpus all pyme-server`.

zacsimile:
NOTE: There is currently no `python manage.py runserver 9000` call (https://python-microscopy.org/doc/cluster/cluster_install.html#cluster-install). Maybe we want to add this?

PR #1221, opened 2022-04-18
GUI panel for manually logging an acquisition event

barentine:
Addresses issue I am waiting for a piece of equipment, and in the meantime I am flipping a physical switch instead of sending a serial command, however I still need a rough time point for these switch-throws (writing frame number does not cut it when imaging with two unsynchronized cameras, at high frame rates). Clicking my mouse and throwing the switch at "the same" time works for now.

kind of a cludge, but maybe someone finds it useful (e.g. 'lost immersion oil')
**Proposed changes:**
- gui panel to allow an acquisition event to be logged manually

image

image

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/ui/manual_event_panel.py
+++ b/PYME/Acquire/ui/manual_event_panel.py
@@ -1,0 +1,48 @@
+
+import wx
+
+class ManualEventPanel(wx.Panel):
+    def __init__(self, parent):
+        """
+        Panel to allow manually logging an acquisition event in PYMEAcquire.
+
+        Parameters
+        ----------
+        parent : wx parent, i.e. MainFrame in your init script
+
+        example usage:
+            @init_gui('events')
+            def events(MainFrame, scope):
+                from PYME.Acquire.ui.manual_event_panel import ManualEventPanel
+                event_panel = ManualEventPanel(MainFrame)
+                MainFrame.camPanels.append((event_panel, 'manual event panel', False))
+        """
+        wx.Panel.__init__(self, parent)
+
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Not sure that manual events are ever something I'd recommend using, due to obvious limitations in how well synchronised they can be as well as potential reliability issues (it's not going to be generally feasible to base any analysis on these events as a) people may forget to log them and b) there is no way of enforcing consistent naming).

That said, having some way of recording ancillary "happenings" /  events (e.g. immersion oil loss, someone turns room lights on, etc ...) would be nice. My gut says that cobbling them into the existing events framework is probably the wrong thing to do (due to synchronisation issues). It's also hard to get enthusiastic about devising an entirely new mechanism. A compromise I could potentially entertain would be as follows:  

- Make the event type for manually triggered events constant (e.g. `'UserEvent'`) and put any information in the description (this avoids polluting the event namespace and lets these be easily filtered in analysis if desired). It also reduces latency somewhat by having only one field to complete for the event (although likely not by enough)
- Optionally, allow pre-defined per-microscope manual event classes - ie the manual event panel `__init__()` would accept a list of event classes and display a trigger button for each - e.g. `['UserLostOil', 'UserToggleWidget']`.

Would probably still like to put a discouragement in the documentation (i.e. please, please, find a way to emit events automatically rather than using this), as I suspect it might be functionality you would regret using due to the inherent problems with synchronisation and reproducibility.

Other options would be to leave as a site-specific hack, or put in an extras or similar repository (which has been on the TODO list for a while).

Issue #1215, opened 2022-04-01
[ENHANCEMENT] gain calibration / more flatfield map support


barentine:
**Is your feature request related to a problem? Please describe.**
Picked up a particularly junk camera (UI-3240 NIR - not one of the ones others have shown working for single-molecules. dark and variance maps are horrifically non-uniform compared to e.g. a UI-3270), looking rather unusable with just a dark-map calibration. Further, the company would not like to disclose their "proprietary" gain conversion, so I'm going to calibrate it using the sqrt(var)/mean relation, but didn't know if we already had code to fit that curve, or if I should just write my own. I see `PYME.DSView.modules.dark_and_flatfield.flat` for calculating a flatfield from a series, but not something to fill `Camera.ElectronsPerCount` from a couple series at different intensities. 

Further, I wonder if we want to extend support for flatmaps so they can get installed similarly to dark and var maps in `gen_sCMOS_maps.py`?

@csoeller how are you calibrating the gain on your IDS cameras? I see some notes in the comments about emailing the company, and I can see I'll want to account for the minimum hardware gain being non-unity.

**Describe the solution you'd like**
recipe module with a couple three `FileOrURI` inputs for series acquired at the same intensity

**Describe alternatives you've considered**
writing my own script

**Are you prepared to work on it?**
sure

David-Baddeley:
I feel that it's something we should have, although not 100% sure. Will have a little bit of a dig ... As far as I recall, there are a couple of caveats which make the data collection side a bit non-trivial, especially if you want to get per-pixel gains not overall gains.

- estimating the gain needs a very accurate estimate of offset, and read noise (i.e. the dark and read noise measurements need to be taken with a cover on the camera (or similar), not just room lights off.
- You need quite a large series (1000s of frames)
- a flatfield map and a per-pixel gain map are not exactly equivalent (the flatfield map includes dust on the sensor, apodisation etc ..., the per-pixel gain map doesn't). I'm of the opinion that you usually want a flat field rather than a per-pixel gain map (i.e. measure the response to uniform illumination, rather than each pixels mean vs var behaviour). This might change somewhat (you might even need both) when you have high read noise (for low read noise, anything between sample and sensor just looks like a straight scaling of sensor sensitivity - and hence detection probability / Poisson noise).
[...]

PR #1223, opened 2022-04-27
use camera_noise in ueye, falling back on current base properties

barentine:
Addresses issue default of 10 e-/ADU being wrong for my camera.

enhancement
**Proposed changes:**
- note the variance maps are in e-^2 in `remFitBuf.CameraInfoManager.getVarianceMap`
- add a 327x camera calibration to the ueye base props dict.
- switch `UEyeCamera.noise_properties` to default to standard base-class `Camera.noise_properties` yaml lookup by serial number and fall-back on baseprops dict loudly

Don't want to interrupt anyone using an IDS camera for e.g. a focus lock who doesn't care, so figured best not to hard-switch to the current convention.

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
Note that the read noise value for the 327x is currently wrong, following my comment in #1215 . I'll update that shortly.

PR #1227, opened 2022-05-03
update fitpoints docstring to include some unit information

barentine:
Addresses issue try and communicate what's going on with units in this module to someone who might not have read the code.

enhancement
**Proposed changes:**
- note that intensity units are not touched by this module (despite throwing in camera noise metadata)
- note that position units might be in nanometers or might be in pixels depending on the weather (but really whether your input ImageStack has voxelsize metadata)

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/recipes/measurement.py
+++ b/PYME/recipes/measurement.py
@@ -178,8 +178,29 @@ class DetectPoints2D(ModuleBase):
 
 @register_module('FitPoints')
 class FitPoints(ModuleBase):
-    """ Apply one of the fit modules from PYME.localization.FitFactories to each of the points in the provided
-    in inputPositions
+    """ 
+    Apply one of the fit modules from PYME.localization.FitFactories to each of
+    the points in the provided in inputPositions. 
+
+    Parameters
+    ----------
+    inputImage: PYME.IO.image.ImageStack
+        FitPoints does not do unit conversions normally done during localization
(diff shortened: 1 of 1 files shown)


PYME/recipes/measurement.py
```diff
@@ -178,8 +178,29 @@ def execute(self, namespace):
...
-    in inputPositions
+    """ 
+    Apply one of the fit modules from PYME.localization.FitFactories to each of
+    the points in the provided in inputPositions. 
+
+    Parameters
+    ----------
+    inputImage: PYME.IO.image.ImageStack
+        FitPoints does not do unit conversions normally done during localization
+        analysis in PYME (where we convert from raw analog-digital units [ADU]
```
David-Baddeley: All our fitting is done in ADUs! we don't ever convert, so this is no different. We only convert to electrons in PYMEVis/Pipelineify where we generate the `nPhotons` derived variable. There are also inconsistencies between modules (in some cases A is the amplitude - i.e. max, and in others it is the sum.

barentine: Yep, sorry, but in remFitBuf fitTask we define `self.data = cameraMaps.correctImage(...)`, which dark/flat corrects. So we probably want to change this to instruct folks to feed through `processing.FlatfieldAndDarkCorrect` first.

PR #1232, opened 2022-05-06
hardware support for micro photon devices picosecond delayer

barentine:
enhancement
**Proposed changes:**
- add a [picosecond delayer](http://www.micro-photon-devices.com/Products/Instrumentation/Picosend-Delayer) [distributed by picoquant](https://www.picoquant.com/products/category/accessories/psd-mod-picosecond-delayer) useful for synchronizing pulsed diode lasers
- add a GUI panel to control/visualize the delay set-point
- fix trivial typo in a microscopy.py docstring
- add the picosecond delayer to the list of supported hardware under miscellaneous. 

Note: will likely follow up at some point with a class to support the [Picoquant Taiko](https://www.picoquant.com/products/category/picosecond-pulsed-driver/taiko-pdl-m1) pulsed/cw driver.

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/mpd_picosecond_delayer.py
+++ b/PYME/Acquire/Hardware/mpd_picosecond_delayer.py
@@ -43,9 +43,9 @@ class PicosecondDelayer(object):
     cached values, while GetXXXX calls will query the board over serial.
 
     Echo mode means the board replies back the command it received, the string
-    terminator '#' and then its response. Turning off echo mode while their
-    labview software communicates with the board is problematic because it can
-    interupt communication.
+    terminator '#' and then its response (and anoher #). Somewhat nebulous 
+    message in manual that turning off echo mode can interupt communication 
+    with he board.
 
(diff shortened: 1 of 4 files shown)


David-Baddeley:
Will merge as is ... but probably need to think a bit more about the conventions for new hardware going forward. As much as there is a current convention it is to use functional setters/getters rather than attribute/property access for hardware parameters that you might change. A lot of this is historical, and in some respects property-based access is more "pythonic", but there is definitely a case for being consistent across different hardware types. The potential arguments for functional getters/setters are:
- what we currently use (i.e. less work). Not sure this is a great argument
- generally closer to the APIs/SDKs of the hardware itself (also not necessarily a clincher)
- property access is generally assumed to be instantaneous and zero cost. A lot of hardware operations take some time (although we often hide this with threading). A programmer is more likely to expect some latency / to have to wait with a functional API.

PR #1234, opened 2022-05-08
Hc single shot

Yujin-Bao:
Addresses issue # .

**This is an enhancement**

**Proposed changes:**
1. Now we can swtich between imaging modes in GUI for Hamamatsu
2. Single shot mode is added to HamamatsuORCA

**Checklist:**

- [ ] Tested with numpy=1.14

### DIFF
--- a/PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
+++ b/PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
@@ -76,7 +76,7 @@ DCAMPROP_TRIGGERSOURCE_EXTERNAL = 2
 DCAMPROP_TRIGGERSOURCE_SOFTWARE = 3
 
 DCAMCAP_START_SEQUENCE = ctypes.c_int32(int("-1",0))
-
+#DCAMCAP_START_SNAP = ctypes.c_int32(int("0",0))
 
 class DCAMZeroBufferedException(Exception):
     pass
@@ -122,6 +122,7 @@ class HamamatsuORCA(HamamatsuDCAM, CameraMapMixin):
             self.setDefectCorrectMode(False)
             self.enable_cooling(True)
             self._mode = self.MODE_CONTINUOUS
+            #self._mode = self.MODE_SINGLE_SHOT
             self.initialized = True
             logger.debug('Hamamatsu Orca initialized')
 
@@ -161,12 +162,12 @@ class HamamatsuORCA(HamamatsuDCAM, CameraMapMixin):
 
         
     def GetAcquisitionMode(self):
-        #FIXME - actually support both continuous and single shot modes
+        # Support both continuous and single shot modes
         #return self.MODE_CONTINUOUS
         return self._mode
 
     def SetAcquisitionMode(self, mode):
-        if mode in [self.MODE_CONTINUOUS, self.MODE_SOFTWARE_TRIGGER]:
+        if mode in [self.MODE_CONTINUOUS, self.MODE_SOFTWARE_TRIGGER, self.MODE_SINGLE_SHOT]:
             self._mode = mode
         else:
             raise RuntimeError('Mode %d not supported' % mode)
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Many thanks!

It looks like this is a spoofed single-shot using software triggering. I'm trying to think if there are any major issues with this spoofing approach (there are definitely some performance/memory usage caveats, as you'll still allocate buffers as though you are doing continuous imaging, although these are unlikely to be critical).

In general, single shot mode has has some potential performance downsides when compared with just using software triggering, including buffers etc ... getting torn down and recreated. This will add a latency to each frame and slow things down unnecessarily. It should be possible to accomplish everything which currently needs single shot mode using software triggering instead, and a medium term goal is to effectively deprecate single-shot for cameras which support software triggering, although we're not quite there yet.

Could I ask what motivated the addition of single shot mode to the orca (i.e. where do we need to improve software triggering support)?

Best wishes,
David

Yujin-Bao:
Hi David,

Thanks for you response! Currently we are using Hamamatsu camera for the fluorescent channel, and when we are z stacking, the camera takes images so fast that it finishes all the frames before the piezo moves on target. We realize that is because it's set to continuous mode and using internal trigger. So the idea is to use software triggering and every time the piezo is on target, the software fires a trigger. I understand we do have the software triggering mode originally, but it somehow does not have "fire a trigger" part so using the original code does not work (actually this confuses me because I think we need to have that "fire a trigger" part in the code for the software triggering). We are using single shot mode for PCO camera in OIDIC channel to do z stacking all the time, so I'd like to solve the problem in the fluorescent channel by also calling this single shot mode but using software triggering to mimic it. By doing this, the previous software triggering code can be remained unchanged. That is why I added a part named "single shot mode" but still using software tirggering.

Currently it runs well for z stacking. The only two things come to me are that, firstly, the integration time cannot be larger than 250 ms, or it gives me a buffer overflowing warning; secondly, when I swtich from continous mode to single shot mode, and switch the integration time on single shot mode, sometimes there is a message saying "lost frame detected". However, in order to prevent saturating, the integration time is often set at 10 or 25 ms, and if I'm not changing the imaging modes or integration time, no "lost frame detected" messages showing even when I do z stacking. So I guess this is good for now?

Thanks again!

Best,
Yujin

David-Baddeley:
I think I'm ahead in my head of where I think we should be - it's currently only tiling where we use the software trigger. I had intended to add that to z-stacking as well, but looking at the code it seems I haven't yet. 

This PR should be fine in the interim (until z-stepping etc works with triggering), but it's probably worth making a note (if just a mental one) that we should really be using software triggering for z-stacks as well.

PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
```diff
@@ -192,20 +193,29 @@ def StartExposure(self):
...
+
+        elif self._mode == self.MODE_CONTINUOUS:
+            # Continuous mode, internal trigger
             self.setCamPropValue('TRIGGER SOURCE', DCAMPROP_TRIGGERSOURCE_INTERNAL)
+            self.checkStatus(dcam.dcamcap_start(self.handle,
+                                            DCAMCAP_START_SEQUENCE),
+                         "dcamcap_start")
 
+        elif self._mode == self.MODE_SINGLE_SHOT:
+            # Single shot mode, software trigger with firetrigger
```
David-Baddeley:
```suggestion
            # Spoofed single shot mode, using the software trigger
            # NOTE: this should no longer be needed when we add software trigger support to z-stepping etc ...
```

Yujin-Bao:
Sounds good! Thank you so much!

David-Baddeley:
I suspect the buffer warning with long integration times can probably be ignored, but not 100% sure - we should look into the cause. I warn pretty aggressively even before the buffer actually overflows, and I suspect it might be that we just aren't allocating enough buffer frames. See Line 186 in HamamatsuOrca.py

```python
self.bs = int(max(int(2.0*self._frameRate), 1))
```

As the integration time approaches 2 seconds, we only allocate 1 frame. It might be safer to set, e.g., a minimum allocation of 10 frames at all times (changing the 1 in the `max()` call to 10).

barentine:
re: the Hammamatsu buffer overflowing at long exposures, I can confirm this behavior on a fusion. To get around it I made the `2.0` in the code @David-Baddeley quoted an attribute so if I run into it, I increase the buffer from the shell and reset the integration time so the buffer gets refreshed

Issue #1237, opened 2022-05-11
[ENHANCEMENT] HTSMS simulator / demo


barentine:
**Is your feature request related to a problem? Please describe.**
We're trying to walk people through a semi-automated acquisition / chained analysis using the simulator

**Describe the solution you'd like**
overview on simulated cells, manually click, load from tile viewer, run protocol acq at each ROI with an associated rule chain to localize and then render a histogram image. 

Ideally this would work on the executable installer, but that raises several questions, such as whether we want to show clusterUI / need to include django in the packaged installer

Outstanding issues:
- [ ] simul642.py protocol has an OnBAnalyse call which annoys me since I'd rather just let it localize and look at after. For sure inconvenient if you set the rule chain to post on series start (cool feature) and then all of a sudden it's trying to open something (failed for me)
- [ ] with django 2.1, clusterUI 'Selected Files > Localize molecules' brings up the menu as it should, but clicking Launch takes you to the distributor page without actually posting any rule tasks.
- [ ] fix package builds. Something has changed with the actions, and making a release no longer works

David-Baddeley:
- Note that posting on series start has potential issues with incomplete/incorrect metadata (in practice this probably just means that we need to hard-code ADOffset in the camera metadata). This should not be too much of a problem in the simulator, and might already happen - will double check, but is not portable to all cameras and protocols.

barentine:
adding while we talk:
- [ ] make sure fakeCam adds ADOffset to metadata

barentine:
Running into an issue with launching localization on spool stop when spooling to h5
<details/>

```
Traceback (most recent call last):
----------------------------------------
Exception happened during processing of request from ('127.0.0.1', 56934)
  File "<string>", line 3, in raise_from
  File "C:\Users\aesb\Miniconda3\lib\site-packages\urllib3\connectionpool.py", line 440, in _make_request
    httplib_response = conn.getresponse()
  File "C:\Users\aesb\Miniconda3\lib\http\client.py", line 1369, in getresponse
    response.begin()
  File "C:\Users\aesb\Miniconda3\lib\http\client.py", line 310, in begin
    version, status, reason = self._read_status()
  File "C:\Users\aesb\Miniconda3\lib\http\client.py", line 279, in _read_status
    raise RemoteDisconnected("Remote end closed connection without"
http.client.RemoteDisconnected: Remote end closed connection without response

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
[...]

PR #1239, opened 2022-05-18
HTSMS sim tweaks

barentine:
Addresses issue #1237 .

enhancement
**Proposed changes:**
note: still need to re-check spool-start localization launch and whether fakecam ADOffset needs hardcoding


barentine:
Looks like fakeCam adds ADOffset, should be OK.

barentine:
Localization launched on spool-start works fine. (modulo currently requires spooling to pzf rather than h5, same error noted in #1237

Issue #1242, opened 2022-05-27
[BUG] When creating a new 'layer' in Visgui, there is no option for meshes


lukasfue:
**Describe the bug**
If you make a new layer in Visgui, it is currently not possible to set it as a mesh. So, if you set the data to a mesh (isosurface or shrink-wrapped), it is treated as a point-cloud and gives you the display options for points rather than a mesh.

**To Reproduce**
1. Click 'New' under layers
2. Select a mesh under 'data'
3. Look through the 'method' options and notice they are all for point-clouds
Note: it seems like the vertices of the mesh are treated as points as you can display them still using the point methods

**Expected behavior**
The mesh data is treated as a mesh and gives display options used for meshes (wireframe, shaded, etc).

**Screenshots**
Isosurface:
![image](https://user-images.githubusercontent.com/36608397/170735352-6c9f0b97-b903-411d-9787-e9281c835d6b.png)

Shrink-wrap:
![image](https://user-images.githubusercontent.com/36608397/170735438-bfe37d68-646a-4899-8782-c660655c8556.png)

David-Baddeley:
Hi Lucas,

The "new layer" button should really read "new pointcloud layer", as that
is how it's coded. It's been on the todo list for a while to support manual
creation of different layer types (there's also currently no way to
manually add an image layer, an octree layer, or a tracks layer), but
hasn't been pressing, as the ways of generating surfaces, octrees, etc ...
all add layers for them by default. I'd be curious to know how you got a
surface without a surface layer being added at generation time (we should
still fix the layer adding).

In terms of user interface, there are a few of options:

- "new" pops up simple layer type menu/dropdown and then creates the
appropriate type
- "new" pops up a dialog that has notebook pages for each layer type and
allows you to configure the layer (setting data source etc ...) before
adding it
- "new" pops up a dialog that asks you which data you want to create the
layer with, then allows you to choose (between compatible layers) which
type of layer you want to use for it
- separate "new" buttons for each layer type.

The first and last options are probably easiest to implement, although the
last one could get visually messy.

Cheers,
David

lukasfue:
David,

I see. It came up after I attempted to shrink-wrap an isosurface, which failed giving the common 'singular matrix' error. I adjusted the parameters of the initial isosurface in an attempt to get rid of the error. That solution worked, and the shrink-wrapp completed error-free, but it did not create a new layer with the new surface. So, I tried to make one manually and that's how I discovered this.

[...]

PR #1247, opened 2022-06-28
Neck removal

David-Baddeley:
Fixes to mesh vertex deletion and combination with high energy vertext deletion to remove topological necks


David-Baddeley:
closes #1242 

as an aside

PR #1249, opened 2022-06-28
Add note about PiezoOnTarget event in PiezoBase docstring

barentine:
enhancement
**Proposed changes:**
- let devs know how PYME logs actual stage position after settling

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
Have added a bit more context so as to make the "why" clear.  Also described a couple of caveats to change the tone from "implement this if you can" to "implement this if you really need it". I think at this point it's probably fair to say that `PiezoOnTarget` is a useful hack, rather than a recommended feature. Philosophically it would be nicer to have it in the later category, but we need to address the events from multiple piezos and thread safety issues first (with the threading, I think you might have issues when spooling directly to HDF).

PR #1254, opened 2022-07-09
Create sensible default values for new mesh layer

zacsimile:
It was bothering me that no mesh came up when I added a mesh layer. This is strictly a UI thing.

### DIFF
--- a/PYME/LMVis/visCore.py
+++ b/PYME/LMVis/visCore.py
@@ -422,7 +422,7 @@ class VisGUICore(object):
         if ds_name is None:
             from PYME.experimental._triangle_mesh import TriangleMesh
             mesh_ds = [k for k, d in self.pipeline.dataSources.items() if isinstance(d, TriangleMesh)]
-            ds_name = mesh_ds[0]
+            ds_name = mesh_ds[-1]
         ds_stub = ds_name.rstrip('0123456789')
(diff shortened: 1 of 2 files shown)


PYME/LMVis/visCore.py
```diff
@@ -416,10 +416,16 @@ def add_pointcloud_layer(self, method='points', ds_name='output', **kwargs):
...
 
-    def add_mesh_layer(self, method='shaded', ds_name='mesh0', **kwargs):
+    def add_mesh_layer(self, method='shaded', ds_name=None, **kwargs):
         from PYME.LMVis.layers.mesh import TriangleRenderLayer
         from PYME.misc.colormaps import cm
-        surf_count=0 #FIXME
+        if ds_name is None:
+            from PYME.experimental._triangle_mesh import TriangleMesh
+            mesh_ds = [k for k, d in self.pipeline.dataSources.items() if isinstance(d, TriangleMesh)]
+            ds_name = mesh_ds[0]
```
```suggestion
            ds_name = mesh_ds[-1]
```

It's quite likely that you will want to add the most recently created mesh, rather than the first one you created (this should work on python>3.7, as dicts are ordered by default).

David-Baddeley:
Thanks! This has been on my TODO ever since adding the option to add meshes ...

Issue #1263, opened 2022-07-19
[ENHANCEMENT] Save Time for events as float or string


d-perez-1:
**Is your feature request related to a problem? Please describe.**
PYMEAcquire saves time for events in the hdf5 file in a format not supported by various python packages or matlab. Python sees the data type as Time64Col, and trying to view the data in matlab throws an error, and displaying the contents shows the data type as H5T_TIME (unsupported).

**Describe the solution you'd like**
Would be great if the data could be saved as a different type (i.e. float or string) so PYMEAqcuire data can more easily be used in existing analysis workflows using other software packages.

**Describe alternatives you've considered**
Using PYME to save events as a separate text file which, then loading that along with ImageData instead of getting everything straight from the .h5

**Additional context**
For matlab:
```
h5disp('8_7_series_E.h5');
Warning: H5T_TIME is not a supported HDF5 class. 
In h5info (line 125)
In h5disp>display_hdf5 (line 144)
In h5disp (line 122) 
HDF5 8_7_series_E.h5 
Group '/' 
    Attributes:
        'TITLE':  ''
        'CLASS':  'GROUP'
        'VERSION':  '1.0'
        'PYTABLES_FORMAT_VERSION':  '2.1'
    Dataset 'Events' 
........
Member 'Time':  H5T_TIME (unsupported)
```

For python 3.7 (for a different file):
```
In [9]: t.root.Events
Out[9]:
/Events (Table(1,)) ''
  description := {
  "EventDescr": StringCol(itemsize=256, shape=(), dflt=b'', pos=0),
  "EventName": StringCol(itemsize=32, shape=(), dflt=b'', pos=1),
  "Time": Time64Col(shape=(), dflt=0.0, pos=2)}
  byteorder := 'little'
  chunkshape := (221,)
```

**Are you prepared to work on it?**
If I implement my own workaround I'll share it.

David-Baddeley:
Definitely supported in python as `Time64Col` (which casts to a double, or potentially a datetime object as soon as you access the events table, by, e.g. `t.root.Events[:]`), and part of the H5 standard (https://support.hdfgroup.org/HDF5/doc1.6/UG/11_Datatypes.html), so basically means that Matlab has a broken H5 implementation, and that this is a Matlab bug. 

[...]

PR #1265, opened 2022-07-20
change events 'Time' field dtype from time to float

barentine:
Addresses issue #1263 .

enhancement
**Proposed changes:**
- change Time field dtype in SpoolEvent from Time64Col to Float64Col
- import events.SpoolEvent where we need it rather than redefining it each time

Appears to work fine with PYMEAcquire simulator and PYMEImage, though I haven't done anything too intense with it
```
>>> image.events
array([(b'0', b'StartAq', 1.65834111e+09)],
      dtype=[('EventDescr', 'S256'), ('EventName', 'S32'), ('Time', '<f8')])
>>> image.events[0]['Time']
1658341107.9741135
```

### DIFF
--- a/PYME/IO/FileUtils/RawToHdf5.py
+++ b/PYME/IO/FileUtils/RawToHdf5.py
@@ -28,13 +28,8 @@ import tables
 import os
 import sys
 import numpy
-
+from PYME.IO.events import SpoolEvent
 from PYME.Analysis import MetaData
-
-class SpoolEvent(tables.IsDescription):
-   EventName = tables.StringCol(32)
-   Time = tables.Time64Col()
-   EventDescr = tables.StringCol(256)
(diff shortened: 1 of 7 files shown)


David-Baddeley:
Some of the files touched here (in fileutils) are pretty old and potential candidates for removal. Probably really not important for considering the PR though.

Just to confirm, you tested with an acquisition protocol with events (e.g. z-stepping) and saw those properly displayed in the PYMEImage event viewer? If that's the case I'd be tempted to just merge and then revert/fix if it causes any issues down the road.

d-perez-1:
From my limited testing in PYMEAcquire (with actual hardware) and PYMEImage the events display the same as before. And the .h5 file now plays (relatively) nicely with matlab.

Issue #1270, opened 2022-07-25
[ENHANCEMENT] PYMEAcquire histogram


d-perez-1:
**Is your feature request related to a problem? Please describe.**
I want to be able to quickly tell if I'm saturating the camera on the fly, so I can adjust settings appropriately. Generally more histogram control would also be nice.

Is the histogram calculated the same way as in PYMEImage, i.e. using subset of pixels and not the entire frame? If I have one bright emitter in an otherwise dark field of view, and I check auto on the histogram scaling, then read off the top bound, can I use that to determine if I'm bellow saturation? Or is it possible the brightest pixels aren't included in the the histogram, or auto scaling does not set bounds to max and min?

**Describe the solution you'd like**
An option to use the full frame in the histogram, if this isn't already how its done. Also live scale image to max and min, unless that's what auto already does. Additionally, display some sort of non-obtrusive warning if some pixels are above threshold. Would also be nice to highlight a region and use just that for setting bounds.

David-Baddeley:
The "auto" scaling in PYMEAcquire does min-max scaling, but still on a subsampled image (1000 pixels). The whole subsampling thing actually started with PYMEAcquire as the histogram calculation etc ... was slowing down data acquisition. This is probably around 10 years ago though, so might not be as big an issue. The way it's written it would also be fairly easy to calculate the min and max on the full image and still do the histogram on the subsampled data.

Will do a quick check on how long min and max actually take on a typical 4MP sCMOS frame and see if we still need to worry. If they are fast enough, it should be a simple change. I think the histograming was expensive enough however, that it might make sense to still do that on a subset even if we set the scaling using min/max of the full frame.

As to actually detecting saturation, there have been various thoughts in the past about, e.g., turning the histogram red if saturation is detected, (could also look into using an overflow LUT, although this gets complex when the scaling is not maxed out). There is also some saturation checking code (which now seems to be commented out, although is potentially activated in specific init scripts) which automatically turned EMCCD gain down if saturation was detected, and a corresponding camera property `SaturationThreshold` (which is effectively the bit depth of the camera).

David-Baddeley:
Timing results (4MP uint16 camera frame, mac M1):

`.min()`  : 3.8 ms
`.max()` : 3.8 ms
`np.histogram(...)` : 183 ms

[...]

PR #1271, opened 2022-07-26
C coded minmax

David-Baddeley:
An attempt to speed up min-max calculation to allow realtime scaling on full rather than subsampled data (see #1270)


David-Baddeley:
peak intensity > half saturation threshold ...
image

saturated ...
image

PR #1272, opened 2022-07-27
SaturationThreshold property for camera base class

barentine:
Addresses issue only fakeCam.py and specCam.py having SaturationThreshold attributes, now required for #1271

**Proposed changes:**
- give all cameras a SaturationThreshold attribute which leans on the camera noise property unless SaturationThreshold as an attribute is overwritten during init (as in fakeCam)
- add ueye gain_mode to alias the bitdepth for noise_properties key access (@csoeller  - look OK?)

**Checklist:**

- [ ] Tested with numpy=1.14


David-Baddeley:
Thanks! I somehow missed the fact that most cameras didn't have `SaturationThreshold` - it certainly used to work with the IXons, but I have a suspicion there used to be an over-ridden `__getattr__` which looked into the noise properties.

Issue #1273, opened 2022-08-18
[BUG] dh5view -t generates missing voxelsize errors


csoeller:
**Describe the bug**
Try `dh5view -t` and see missing voxelsize errors

**To Reproduce**

1. Try `dh5view -t` and see missing voxelsize errors

**Expected behavior**
Should have a voxel size.

PR #1274, opened 2022-08-18
add pixel size for dh5view -t generated image

csoeller:
Addresses issue #1273.

bugfix

**Proposed changes:**
provide a voxel size

**Checklist:**

- tested with python 3.8 and conda-forge build on mac ARM64


David-Baddeley:
seems reasonable, thanks!

Issue #1275, opened 2022-08-23
[BUG] newer skimage 0.19.x has new dtype option which trips up skfilters.py recipe autowrapper


csoeller:
**Describe the bug**
Latest `skimage.filters` (0.19.x) has a change in the gabor_kernel optional arguments that adds a `numpy`dtype which trips up the current autowrapper code.

**To Reproduce**
1. Open one of the PYME apps that import the `skimage` recipes, e.g. `dh5view`, `PYMEVis`etc
4. See error on import of skimage

```python
Could not import skimage

Traceback (most recent call last):

  File "/Users/csoe002/Documents/src/PYMEsrc/pyme38/PYME/recipes/modules.py", line 26, in <module>

    from . import skfilters

  File "/Users/csoe002/Documents/src/PYMEsrc/pyme38/PYME/recipes/skfilters.py", line 128, in <module>

    exec(cd)

  File "<string>", line 81

    dtype = _IntFloat(<class 'numpy.complex128'>)

                      ^

SyntaxError: invalid syntax
```

**Expected behavior**
Should import cleanly.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
```
python-microscopy=22.05.25.post0.dev[git]80684a40bcad
python=3.8.13, platform=darwin
numpy=1.23.2, wx=4.2.0 osx-cocoa (phoenix) wxWidgets 3.2.0
```

David-Baddeley:
should now be able to close

PR #1276, opened 2022-08-23
fix for issue #1275 - skfilters wrapper bug with skimage 0.19.x

csoeller:
Addresses issue #1275 .

Bugfix.

**Proposed changes:**
Candidate fix. Also has a little more error catching and reporting to fail more gracefully.
Happy to see edits of this PR if it helps achieve more robust wrapping.

**Checklist:**

Tested with

```
python-microscopy=22.05.25.post0.dev[git]80684a40bcad
python=3.8.13, platform=darwin
numpy=1.23.2, wx=4.2.0 osx-cocoa (phoenix) wxWidgets 3.2.0
```

### DIFF
--- a/PYME/recipes/skfilters.py
+++ b/PYME/recipes/skfilters.py
@@ -13,6 +13,7 @@ import numpy as np
 import six
 import skimage.filters as skf
 import inspect
+import sys
 
 """Automagically generate filter objects for all skimage filters"""
 
@@ -100,6 +101,9 @@ for filtName in skFilterNames:
                         argTypes[a] = 'bool'
                     elif isinstance(ad, int) and (ad != 0):
                         argTypes[a] = 'int'
+                    # below is a hack for gabor filter dtype argument - new option in skimage 0.19.x
+                    elif ad is np.complex128 or ad is np.complex64:
+                        argTypes[a] = 'dtype'
             
             #disregard parameters which need another image for now        
             args = [a for a in args if not argTypes[a] in ['image', 'dict']]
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I think we'd need to so something a bit more to make this fully work - (ie have a Trait type and editor for dtypes) - I wonder if just ignoring it (and other un-matchable parameters) and running with the default double precision might be a a better approach.

David-Baddeley:
Option 1) - add to ignore (like dict and image)

Option 2) - cast to a string (if skimage accepts string dtypes)

csoeller:
I think option 1 (ignore) is entirely sufficient for now. I have tried to change the code according to that idea.

PYME/recipes/skfilters.py
```diff
@@ -100,9 +101,12 @@ class %(FilterID)s(SKFilter):
                         argTypes[a] = 'bool'
                     elif isinstance(ad, int) and (ad != 0):
                         argTypes[a] = 'int'
+                    # below is a hack for gabor filter dtype argument - new option in skimage 0.19.x
+                    elif ad is np.complex128 or ad is np.complex64:
+                        argTypes[a] = 'dtype'
             
             #disregard parameters which need another image for now        
-            args = [a for a in args if not argTypes[a] in ['image', 'dict']]
+            args = [a for a in args if not argTypes[a] in ['image', 'dict', 'dtype']]
```
David-Baddeley:
```suggestion
            #TODO - log/warn when we ignore parameters
```

PR #1277, opened 2022-08-30
Add output trigger options for Hamamatsu Orca cameras

barentine:
Addresses issue I need some analog frame counting

enhancement
**Proposed changes:**
- add DCAM methods for checking property value text (for text options. Note all args passed to dcam are as float)
- add SetOutputTrigger method to ORCA class, including low, readout start, and readout end triggers for now.

Tested on a BT Fusion, looking at the output trigger 1 SMA on a scope

**Checklist:**

- [ ] Tested with numpy=1.14


barentine:
@David-Baddeley  now expanded to include a hardware start trigger mode which involves touching the base Camera class.

PYME/Acquire/Hardware/Camera.py
```diff
@@ -609,15 +623,29 @@ def GetAcquisitionMode(self):
 
     def SetAcquisitionMode(self, mode):
         """
-        Set the readout mode of the Camera object. PYME currently supports two
-        modes: single shot, where the camera takes one image, and then a new
+        Set the readout mode of the Camera object. PYME currently supports
+        several modes; different options can be useful for timing. The primary
+        modes are single shot, where the camera takes one image, and then a new
         exposure has to be manually triggered, or continuous / free running,
         where the camera runs as fast as it can until we tell it to stop.
```
David-Baddeley:
```suggestion
        where the camera runs as fast as it can until we tell it to stop. Not all cameras support all modes. All cameras should support `MODE_CONTINUOUS` and `MODE_SINGLE_SHOT`, but care should be taken when using the other modes (i.e. usage probably need to be restricted to code which knows what type of camera it's dealing with). TODO - add a mechanism for cameras to report which modes they support.

```

PYME/Acquire/Hardware/Camera.py
```diff
@@ -599,7 +600,20 @@ def GetAcquisitionMode(self):
...
+                    captures single frame on acquisition start
+                2: SOFTWARE_TRIGGER
+                    captures a single frame every time a software trigger is
+                    sent. See `FireSoftwareTrigger`
+                3: HARDWARE_TRIGGER
+                    captures a single frame every time an external hardware
+                    trigger is received by the camera
+                4: HARDWARE_START_TRIGGER
+                    camera waits for an external hardware trigger and enters 
+                    free-running internal timing mode once received
```
David-Baddeley, line 616: We make **no** assurances that the integer values will remain consistent, while unlikely, these might change in the future. Calling code should use (and where needed test against) the enum defines (`Camera.MODE_CONTINUOUS` etc) rather than their integer equivalents. This is standard best practice, and our code should reflect this

PYME/Acquire/Hardware/HamamatsuDCAM/HamamatsuORCA.py
```diff
@@ -423,6 +449,97 @@ def SetShutter(self, mode):
         """
         if self.external_shutter is not None:
             self.external_shutter.SetShutter(mode)
+    
+    def SetOutputTrigger(self, mode, delay=0, width=0.0001, positive=True):
+        """
+        Set output trigger of the camera. For now, only sets output trigger 0, even if
+        the camera supports multiple output triggers.
```
David-Baddeley:
```suggestion
        the camera supports multiple output triggers. 
        
        TODO: have a look at Andor and PCO SDKs to see if this is sufficiently similar in those, and potentially move/adapt to main camera spec.
```

PR #1281, opened 2022-09-13
Revision fixes

David-Baddeley:
A bunch of fixes to facilitate simpler 3rd party usage of the code (and a couple of bugfixes). Notably, this contains a breaking change in how recipe modules from plugins are accessed, which will require modification of recipes using these plugins (although not the plugin code itself). In short, recipe modules from plugins are now registered under the top-level plugin package name, NOT the most proximal python module name.


docs/cluster/programatic_usage.rst
```diff
@@ -0,0 +1,166 @@
...
+
+    streamer = cluster_streaming.Streamer()
+
+    # put a single file to the cluster. The data is written exactly as provided
+    # Actual IO, however, is asynchronous with the file being placed on a queue 
+    # and the function returning immediately
+    streamer.put(filename : str, data : bytes)
+
+
+An more complete example of the low-level streaming interface, including the uses of a custom distribution function to
```
barentine: A* more complete

docs/cluster/programatic_usage.rst
```diff
@@ -0,0 +1,166 @@
...
+enure data-locality when creating an image pyramid be found in :py:mod:`PYME.Analysis.distributed_pyramid`
```
barentine: can* be found

docs/cluster/programatic_usage.rst
David-Baddeley: Thanks! obviously written fast without much proof-reading. Still very much a work in progress as well, just pushed as a backup.

PR #1283, opened 2022-09-18
Add some `RecipeRule` documentation

zacsimile:
Some documentation for `RecipeRule`s, based on recent experiences. Probably this belongs in `docs/cluster/programatic_usage.rst` from #1281, rather than in its own `rules.rst`.

### DIFF
--- a/docs/cluster/rules.rst
+++ b/docs/cluster/rules.rst
@@ -82,7 +82,7 @@ following.
 
     rule.push()
 
-Note that in this case, only a single file is passed as the input shape. 
-However, the server expects a list and therefore treats the input name 
-``shape`` as a stub. In the event only a single input is in the list,
-the input is named ``shape_shape`` (stub followed by stub).
+Note that in this case, an HDF file is passed as input. This is opened in the
+pipeline as a data source with name ``<input_name>_<table_name>``. In this 
+case, the table name is also named ``shape`` and so ``input_localizations`` in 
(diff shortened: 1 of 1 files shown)


docs/cluster/rules.rst
```diff
@@ -0,0 +1,88 @@
...
+
+    rule = RecipeRule(recipe=recipe_text, output_dir=output_dir, 
+                      inputs={'shape': [f'pyme-cluster:///{output_dir}/shape.hdf']})
+
+    rule.push()
+
+Note that in this case, only a single file is passed as the input shape. 
+However, the server expects a list and therefore treats the input name 
+``shape`` as a stub. In the event only a single input is in the list,
+the input is named ``shape_shape`` (stub followed by stub).
```
David-Baddeley: This is incorrect - the `<input_name>_<table_name>` pattern only applies to HDF files which can contain multiple data tables (ie you can't assume that an HDF only contains a single table), and has nothing to do with whether the inputs is a list or not.

zacsimile: Good to know. I took a guess on this one because I couldn't find where it maps the name. Where does the `<input_name>_<table_name>` bit happen?

David-Baddeley: `Recipe._inject_tables_from_hdf5()` (line 558 of recipe.py). Looking at the code there, if you use the `'input'` as the key to the rule etc ... it will just inject with the table name.

PR #1284, opened 2022-09-22
Add metadata to Octree and TriangleMesh and associated recipes

zacsimile:
Metadata handlers for `Octree` and `TriangleMesh` allow us to add recipe metadata as well.

### DIFF
--- a/PYME/experimental/_octree.pyx
+++ b/PYME/experimental/_octree.pyx
@@ -2,6 +2,8 @@ cimport numpy as np
 import numpy as np
 cimport cython
 
+from PYME.IO.MetaDataHandler import DictMDHandler
+
 #size to initialize the storage to
 INITIAL_NODES = 1000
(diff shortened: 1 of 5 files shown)


David-Baddeley:
- should really bring sample metadata through from the image / points used to create the surfaces (basically just inherit the metadata of the input data)

- I think all the metadata added here should be under the `Processing.` top level key (it's not quite as clear cut as for most recipe modules, as there is a pronounced change in data format, but still probably best).

zacsimile:
Sounds good. Could you update or rewrite this to pass the metadata as you prefer?

David-Baddeley:
I've taken a very different tack in #1288 which tries to achieve the same thing, but with much less boilerplate. With any luck, this will form the basis of all the metadata stuff being handled more or less automatically.

David-Baddeley:
superseeded by https://github.com/python-microscopy/python-microscopy/pull/1288

PR #1286, opened 2022-09-28
remove deprecated has_key with in operator

boneyag:
**Enhancement**

**Proposed changes:**
Remove deprecated `has_key` function and use `in` operator instead.

**Checklist:**

- [ ] Tested with numpy=1.14

- [ ] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)? No
- [ ] Does this change how users interact with the software? How will these changes be communicated? NA
- [*] Does this maintain backwards compatibility with old data? Yes
- [*] Does this change the required dependencies? No
- [*] Are there any other side effects of the change? No


David-Baddeley:
Many thanks!

David

PR #1287, opened 2022-09-29
Docs and sat threshold regression

David-Baddeley:
Some more cluster docs, fix regression noted in #1237


David-Baddeley:
@barentine

PR #1288, opened 2022-09-29
Recipe metadata

David-Baddeley:
Replacement for #1284 and hopefully the basis for generally improved metadata handling in recipes


David-Baddeley:
NB, this is on top of #1287, so has a few extra commits

PR #1306, opened 2022-10-18
add mac installer troubleshooting note and PYMEClusterOfOne to install verification table

barentine:
Addresses issue # .

**Proposed changes:**

**Checklist:**

- [ ] Tested with numpy=1.14


docs/Installation/Installation.rst
```diff
@@ -134,6 +136,8 @@ installed as part of PYME.
 +-------------------------+----------------------------------------------------------------------------------------------------------------------+
 | ``PYMEVis``             | This is for viewing point data sets. It shows a blank canvas when launched without any parameters.                   |
 +-------------------------+----------------------------------------------------------------------------------------------------------------------+
+| ``PYMEClusterOfOne``    | This is for parallelized analysis, e.g. localization. It will open a browser window to a simple task status monitor. |
```
David-Baddeley: I don't know that we need to include this in our install testing options.

PR #1309, opened 2022-10-25
Fixed size strings when saving to hdf tables

David-Baddeley:
Use fixed size strings when saving metadata to table columns so that pytables type inference works when aggregating

### DIFF
--- a/PYME/recipes/measurement.py
+++ b/PYME/recipes/measurement.py
@@ -869,7 +869,9 @@ class AddMetadataToMeasurements(ModuleBase):
                 v = img.mdh[mdk]
 
             if isinstance(v, str):
-                res[k] = np.array([v]*nEntries, dtype='U%d'%self.string_length)
+                # use bytes/cstring dtype (rather than U) so that 
+                # pytables doesn't bork on us
+                res[k] = np.array([v]*nEntries, dtype='S%d'%self.string_length)
             else:
                 res[k] = np.array([v]*nEntries)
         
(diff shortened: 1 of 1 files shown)


zacsimile:
Yes, this works. Minor observation: it seems like the string becomes a bit longer when it gets converted to characters in the HDF. So, for a string of length 36, I have to specify a dtype of `U200`.

```
from PYME.IO import tabular
import numpy as np
import pandas as pd

res = {'mse_rms': np.array([23.975166]), 
       'test_pointcloud_id': np.array(['4774262f-e59f-4dcb-b3ec-e3691d3d1b64'], dtype='U200')}

ds = tabular.DictSource(res)

print(ds['test_pointcloud_id'][0])
```

At this point, the output is

```
4774262f-e59f-4dcb-b3ec-e3691d3d1b64
```

```
fn = 'test.hdf'
ds.to_hdf(fn, tablename='measurements')

ds2 = pd.read_hdf(fn, key='measurements')
print(ds2['test_pointcloud_id'][0])
```

The output is

```
4774262f-e59f-4dcb-b3ec-e3691d3d1b64
```

If I instead use `U100`, the output is

```
4774262f-e59f-4dcb-b3ec-e
```

David-Baddeley:
That's super odd ... Per numpy docs (https://numpy.org/doc/stable/reference/arrays.dtypes.html#specifying-and-constructing-data-types), the length specifier on `'U'` **should** be the number of characters.

David-Baddeley:
Can't reproduce on mac, py3.8, numpy 1.22 - `dtype='Uxx'` generates a fixed length string with exactly the right number of characters for me.

David-Baddeley:
Hmm, crops when you round trip to disk. Would seem to be a pytables bug - probably need to explicitly `.encode()` and save as `bytes`.

David-Baddeley:
If we change the type to 'S40' pytables handles it OK (albeit converting to bytes).

zacsimile:
Confirmed, converting `U200` to `S36` in the example I posted above works.

PR #1313, opened 2022-11-21
Task accounting

David-Baddeley:
Candidate fix for long-running issue with incorrect reporting of task numbers when retries have been triggered.


David-Baddeley:
should (hopefully) finally tie up #490

David-Baddeley:
Now with a bunch of additional fixes for running rules from a notebook ...

PR #1314, opened 2022-11-21
PYMEImage histogram bounds: use all pixels to calculate, add choice-box for method

barentine:
Addresses issue @d-perez-1 appreciates PYMEAcquire histogram tweaks #1270 / #1271 but then opens data in PYMEImage and is frustrated again. Speed should be even less of an issue here, when the data is already acquired.

enhancement
**Proposed changes:**
- use all pixels to calculate histogram bounds (currently using a subset if we have 256 x 256 or larger image)
- allow users to see what 'stretch' button does, and to toggle mode with a choice box

image

image

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/DSView/DisplayOptionsPanel.py
+++ b/PYME/DSView/DisplayOptionsPanel.py
@@ -144,6 +144,11 @@ class OptionsPanel(wx.Panel):
 
         self.bOptimise = wx.Button(pan_opt, -1, "Stretch", style=wx.BU_EXACTFIT)
 
+        self.cb_stretch_method = wx.Choice(pan_opt, -1, choices=["min-99th", "min-max"])
+        self._stretch_settings = ['percentile', 'min-max']  # translate to displayOptions terms
+        self.cb_stretch_method.SetSelection(0)
+        self.cb_stretch_method.Bind(wx.EVT_CHOICE, self.OnStretchMethodChanged)
+
         self.cbScale = wx.Choice(pan_opt, -1, choices=["1:16", "1:8", "1:4", "1:2", "1:1", "2:1", "4:1", "8:1", "16:1"])
         self.cbScale.SetSelection(4)
         self.scale_11 = 4
@@ -153,6 +158,7 @@ class OptionsPanel(wx.Panel):
             
             hsizer = wx.BoxSizer(wx.HORIZONTAL)
             hsizer.Add(self.bOptimise, 0, wx.ALL|wx.ALIGN_CENTER, 5)
+            hsizer.Add(self.cb_stretch_method, 0, wx.ALL|wx.EXPAND, 5)
             hsizer.Add(self.cbScale, 0, wx.ALL|wx.ALIGN_CENTER, 5)
             
             pan_opt.SetSizerAndFit(hsizer)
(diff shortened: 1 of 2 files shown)


David-Baddeley:
A couple of thoughts:

- The subsampling is still absolutely critical for histogram calculation on some images, so we can't completely remove it (large volume imaging is of considerable interest to some users, and at, e.g. 50k x 50k pixels you'd be looking at about 80s per channel to calculate the histogram). As the histogram is recalculated every time you change slice, the 200ms for calculating the histogram of a 2k x 2k sCMOS frame is already going to slow things down appreciably when you are scrolling through a video series (you are effectively limiting your playback rate to < 5 frames/sec). I think the current subsampling is about right in terms of speed/fidelity for histogram calculation (the histogram doesn't change appreciably when you increase the sampling).
- That's not to say we can't look at using something different in `Optimize()` which can reasonably be a bit slower (I'd say anything under a couple of seconds is fine here) - potentially the c-coded min-max if we are doing min-max scaling, or bringing the subsampling threshold parameter to `get_hist_data(subsample_threshold=1e4)` and passing something just over the full sCMOS frame size when `get_hist_data()` is called from `Optimise()`
- as to UI I'm not sure I'm quite sold on the combo-box, as it's not immediately clear that it's a setting for the optimise button, and a combo-box implies state - if you manually drag the histogram bounds, or move in z/t the implied state is no longer valid. I'd probably lean towards having a Shift-click or long click on the Optimise button display a pop-up menu which let you select how optimise was going to work, although this has the downside of being less discoverable. This is less critical as it doesn't break things and can shake out later.

barentine:
Thanks for having a look, David! Please modify as you like.

I think the ideal would be to use c-coded min/max and min/99th, single-call bound functions that worked for float since we appear to be casting to that at the moment. As you found with uint16 in #1271, that's plenty fast for real-time acquisitions, should be plenty fast for float post-facto.

Meanwhile for truly diabolical data sizes (not 256 x 256, or even full frame cmos), I didn't know if the data.chunks (cruft at the moment?) code above would come into play with something in pymenf or so.

I wouldn't say the UI is ideal either, but personally I like that there's at least text to tell you what the stretch button does (particularly frustrating at the moment to push stretch and if your data is sparse, look at zero and definitely not have stretched anything - hurts worse I expect if you don't know what's going on behind the scenes). One alternative could be what you suggested with right click for selecting what the button did (does wx have a drop-down button class?) but change the text from stretch to whichever bounds method was being used. Maybe minimal cludge would be to drawn a static box around that hsizer to show the choicebox is linked to the stretch button.

David-Baddeley:
`data.chunks` definitely comes into play if you have chunked data (which you should, but don't always have for diabolical sizes). Chunks are a fairly new addition to the code, so it's a bit clunky in places, but doesn't need anything other than the main code base. They are an added feature of the main ArrayDataSource, and come into play if you are viewing as zarr or dask array. You don't really see much on the surface, but if you open a .zarr file `image.ImageStack(filename='some_zarr_file.zarr')` or run `View3D(...)` on a a dask array, the display (and slicing) is chunk-aware. There is also some limited support for pyramidal .zarr (i.e. OME-NGFF).

David-Baddeley:
mostly covered by #1315 (potentially modulo discoverability)

Issue #1326, opened 2023-01-26
[PROPOSAL] remove scipy.random calls


barentine:
**Is your feature request related to a problem? Please describe.**
they're deprecated. Not sure exactly where they disappear, but they're certainly gone by scipy 1.9.3

```
FAILED PYME/localization/test_dumbell.py::test_Dumbell3DFitR - AttributeError: Module 'scipy' has no attribute 'random'
FAILED PYME/localization/test_localization_algorithms.py::test_AstigGaussFitFR - AttributeError: Module 'scipy' has no attribute 'random'
FAILED PYME/localization/test_localization_algorithms.py::test_LatGaussFitFR - AttributeError: Module 'scipy' has no attribute 'random'
FAILED PYME/localization/test_localization_algorithms.py::test_InterpFitR_astigmatism - AttributeError: Module 'scipy' has no attribute 'random'
FAILED PYME/localization/test_localization_algorithms.py::test_InterpFitR_PRI - AttributeError: Module 'scipy' has no attribute 'random'
FAILED PYME/localization/test_localization_algorithms.py::test_SplitterFitInterpNR_astigmatism - AttributeError: Module 'scipy' has no attribute 'random'
```

**Describe the solution you'd like to implement**
change them to numpy.random calls

**Describe alternatives you've considered**
let sleeping dogs lie

barentine:
fixed with #1327

PR #1327, opened 2023-01-26
change scipy.random to numpy.random calls

barentine:
Addresses issue #1326 

enhancement
**Proposed changes:**
change remaining 3 scipy.random calls to numpy.random calls

**Checklist:**
lets tests pass on later scipy versions (e.g. 1.9.3)
note, tested with numpy 1.23.5

### DIFF
--- a/PYME/Acquire/Hardware/Simulator/fakeCam.py
+++ b/PYME/Acquire/Hardware/Simulator/fakeCam.py
@@ -148,10 +148,10 @@ class NoiseMaker:
         if self.approximate_read_noise:
             o = self._read_approx(im.shape)
         else:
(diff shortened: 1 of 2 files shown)


David-Baddeley:
We should probably change the `scipy.rand()` ones while we're at it.

Won't stop this, but might influence how zealous we want to be with `scipy.rand()` etc ... -  I need to check if we are actually using visHelpersMin anywhere - it might just be cruft.

barentine:
Fair enough. Cruft or not, have added the visHelpersMin changes to the PR.

`<ipython-input-3-13d645ef4dd7>:1: DeprecationWarning: scipy.randn is deprecated and will be removed in SciPy 2.0.0, use numpy.random.randn instead`

`<ipython-input-4-3bcfc0fd1b85>:1: DeprecationWarning: scipy.rand is deprecated and will be removed in SciPy 2.0.0, use numpy.random.rand instead`

Issue #1328, opened 2023-01-26
[BUG] test_cluster_spooler.py waits for finished() method on HTTPSpoolerv2, which is incomplete


barentine:
**Describe the bug**
test_cluster_spooler.py currently hangs waiting for the spooler to finish, even though it's already finished. My hunch is this is why we have a 
```
if os.environ.get('AZURE_TEST', False):
    pytestmark = pytest.mark.skip(reason="Doesn't seem to work on azure")
```
on it.

We switched from testing `PYME.IO.HTTPSpooler` to `PYME.IO.HTTPSpooler_v2`. The former has a `finished` method which checks that the queue is empty and the polling thread is exited and the spooling threads are done. The _v2 `finished` method has a `#fixme` and just checks `self._stopping`, which only gets set after a `PYME.IO.Spooler.finalise` call. 

In the test it's easy enough to add a `finalise` call, but we might also want to fix the `finished` method of the _v2.

**To Reproduce**
1. Go to python-microscopy/tests/PYME/IO
2. run pytest test_cluster_spooler.py
3. wait a very long time

If you want some output, you can:

1.  import test_cluster_spooler as tcs
2. manually run `setup_module()`
3. run `test_spooler()`
5. watch it count to 50 and then not return

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] OSX 12.6.2
 - Python version: 3.8
 - numpy version: 1.23
 - wxpython version [if a UI bug]: 4.2.0

**Additional context**

barentine:
closing with merge of #1329 and #1345

PR #1329, opened 2023-01-26
stop spool in test_cluster_spooling

barentine:
Addresses issue #1328 

bugfix

**Proposed changes:**
move stopspool call to be in front of finished check as new spooler only gets marked as finished in the `finalise` call in `stopspool`

This fixes the test on my machine. @David-Baddeley would be interested to hear if this also lets it run on Azure.


David-Baddeley:
hmm ... will need to double check the logic - at one point calling stopspool effectively aborted (i.e. you would truncate spooling so get a variable amount spooled if you call it directly).

PR #1334, opened 2023-02-27
Mesh tidying

David-Baddeley:
Some tidying of the meshing code. To date, just removed some confusing implicit updates that had me thinking we had a bug which wasnt there


David-Baddeley:
Now including some translation to c /  moving to a second file. It would be nice to drop triangle_mesh.pyx to a more reasonable size.

David-Baddeley:
and remove one of the versions of `edge_split`, along with a bunch of allocation stuff we no longer need. Next step changing vacancies arrays from lists to arrays (so as to be usable from vanilla c code).

PR #1336, opened 2023-03-09
Recipe .run() refactoring

David-Baddeley:
This is a somewhat major refactor of recipe modules, which replaces the `.execute()` method with a `.run()` method and takes the namespace access stuff out of user code.

There are several goals here:

1) it should reduce recipe boilerplate (fetching inputs and writing outputs etc results in an awful lot of overhead)
2) it should automate metadata propagation for all modules (no need to do this manually anymore)
3) hopefully easier learning curve (inputs come in as parameters and outputs get returned as one would expect, injection into and reading from the namespace are all done automatically)

This should also hopefully make writing extension modules easier .... there are, however, some more technical (and speculative) objectives behind this.

4) this should hopefully help us do chunk-parallel stuff for very large datasets (e.g. either through the use of dask or our own backend)
5) removing injection shifts us away from an "all intermediate results are written" (even if only to memory)  paradigm, and makes it conceptually easier to e.g. compile recipes such that consecutive modules execute together. Potentially even letting, e.g. two modules which use the GPU to leave their data on there. 

Everything is backwards compatible for now (defining execute still works).

@zacsimile , @barentine - Would appreciate any thoughts/ opinions.

### DIFF
--- a/PYME/IO/DataSources/ArrayDataSource.py
+++ b/PYME/IO/DataSources/ArrayDataSource.py
@@ -177,10 +177,15 @@ class XYZTCArrayDataSource(BaseDataSource): #permit indexing with more dimension
         # zarr/dask support
         if hasattr(data, 'chuncks'):
             self.chunks =  np.ones(5, 'i')
+            self.chunksize = np.ones(5, 'i')
             for j, ax in enumerate('XYZTC'):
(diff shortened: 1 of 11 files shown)


zacsimile:
Yeah, this seems much easier to use. I like it.

PR #1337, opened 2023-03-09
listctrlMixins fix - used to use a now changed wxWindows API access

csoeller:
this fix seems needed with more recent wxWindows; previously I either got a warning that the original use is now deprecated or even a bug was raised when a PYME module used the listctrlMixins...

### DIFF
--- a/PYME/contrib/listctrlMixins.py
+++ b/PYME/contrib/listctrlMixins.py
@@ -301,7 +301,10 @@ class ListCtrlAutoWidthMixin(object):
         listWidth = self.GetClientSize().width
         if wx.Platform != '__WXMSW__':
             if self.GetItemCount() > self.GetCountPerPage():
-                scrollWidth = wx.SystemSettings_GetMetric(wx.SYS_VSCROLL_X)
+                try:
+                    scrollWidth = wx.SystemSettings_GetMetric(wx.SYS_VSCROLL_X)
+                except AttributeError:
+                    scrollWidth = wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Thanks! Only suggestion would be to reverse the order of trying things (try the new version first), although I'm not sure it's worth the effort to re-write that way.

PYME/contrib/listctrlMixins.py
```diff
@@ -301,7 +301,10 @@ def _doResize(self):
         listWidth = self.GetClientSize().width
         if wx.Platform != '__WXMSW__':
             if self.GetItemCount() > self.GetCountPerPage():
-                scrollWidth = wx.SystemSettings_GetMetric(wx.SYS_VSCROLL_X)
+                try:
+                    scrollWidth = wx.SystemSettings_GetMetric(wx.SYS_VSCROLL_X)
```
David-Baddeley:
```suggestion
                    scrollWidth = wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)
```

PYME/contrib/listctrlMixins.py
```diff
@@ -301,7 +301,10 @@ def _doResize(self):
...
+                except AttributeError:
+                    scrollWidth = wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)
```
David-Baddeley, line 305:
```suggestion
                    # fallback for old wx - TODO - do we still actually need this?
                    scrollWidth = wx.SystemSettings_GetMetric(wx.SYS_VSCROLL_X)
```

PR #1338, opened 2023-04-06
recipe execution - remove indiscriminate print() statements

csoeller:
`recipe.execute` still retains some indiscriminate print statements that were presumably useful in debugging. As I use recipe updates in Jupyter notebooks these show up as somewhat annoying 'noise' in outputs, e.g.

```
Opening %s using pandas (set TextFileSource-use_pandas: False in config.yaml to use legacy np.genfromtxt instead)
recipe.execute()
recipe.execute()
recipe.execute()
recipe.execute()
```

No bug per se, rather a cleanup.

**Proposed changes:**
In the PR I have so far just commented the print statements out but if they do serve a purpose perhaps conversion into logging/debug statements could be advisable?

### DIFF
--- a/PYME/recipes/recipe.py
+++ b/PYME/recipes/recipe.py
@@ -198,7 +198,7 @@ class Recipe(HasTraits):
         #remove anything which is downstream from changed inputs
         
         try:
-            #print('recipe.execute()')  # possibly change these into logging/debug messages
+            logger.debug('recipe.execute()')  
             #print self.namespace.keys()
             for k, v in kwargs.items():
                 #print k, v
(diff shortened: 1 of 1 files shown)


PYME/recipes/recipe.py
```diff
@@ -198,18 +198,18 @@ def execute(self, progress_callback=None, **kwargs):
         #remove anything which is downstream from changed inputs
         
         try:
-            print('recipe.execute()')
+            #print('recipe.execute()')  # possibly change these into logging/debug messages
```
David-Baddeley: This one in particular is quite useful when working out why a recipe doesn't run in PYMEVis/PYMEImage

David-Baddeley:
```suggestion
            logger.debug('recipe.execute()')  
```

PR #1339, opened 2023-04-06
add "discard trivial" switch to MergeClumps module

csoeller:
The `MergeClumps` recipe module uses in its implementation the `multiview.coalesce_dict_sorted` function which has a useful `discard_trivial` keyword option which is currently not exposed through the `MergeClumps` interface.

I have recently encountered several cases where I need to remove "trivial" clumps that are retained (identified by `clumpSize=0`), such as external clumpIndices that are not contiguous etc. Simply switching on discard trivial is then very convenient and gets saved with recipes when stored to disk.

This PR proposes to expose the `discard_trivial` option via a `discardTrivial` input of `MergeClumps`.

Enhancement.

The enhancement is backwards compatible as it is set to `False` by default, perserving previous behaviour and correct interpretation of older saved recipes (e.g. in yaml form).

### DIFF
--- a/PYME/Analysis/points/DeClump/pyDeClump.py
+++ b/PYME/Analysis/points/DeClump/pyDeClump.py
@@ -234,7 +234,7 @@ def coalesceClumps(fitResults, assigned, nphotons=None):
     return fres
 
 
-def mergeClumps(datasource, labelKey='clumpIndex'):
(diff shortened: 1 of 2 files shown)


David-Baddeley:
I guess this is OK / reasonable as off by default. It might need some caveats in actual use.

PR #1340, opened 2023-04-10
Topology functions + cleanup

zacsimile:
Adds Euler characteristic/genus to mesh, removes extraneous commented old functions and switches a couple of `object` flags to `bint` flags for clarity.

### DIFF
--- a/PYME/experimental/_triangle_mesh.pyx
+++ b/PYME/experimental/_triangle_mesh.pyx
@@ -512,6 +512,10 @@ cdef class TriangleMesh(TrianglesBase):
         return chi
 
     @property
+    def genus(self):
+        return (2-self.euler_characteristic)/2
+
+    @property
     def bbox(self):
         """
         Mesh bounding box. xl, yl, zl, xu, yu, zu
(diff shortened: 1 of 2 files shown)


PYME/experimental/_triangle_mesh.pyx
```diff
@@ -797,11 +837,6 @@ cdef class TriangleMesh(TrianglesBase):
         #self._E = None
         #self._pE = None
 
-    def _clear_flags(self):
```
David-Baddeley, line 800: This is / was used in ch_shrinkwrap (hence the deprecation rather than deletion

David-Baddeley: OK - just seen that it's changed in ch-shrwinkwrap#22 - do we want to leave the deprecated version here for a bit in case people are on differing versions?

zacsimile: I think it calls a function that no longer exists?

zacsimile: Fine to leave it, though.

David-Baddeley: ahh, it was supposed to call `_invalidate_cached_properties()` as a fallback for the renaming, but I obviously got the naming wrong (not sure how my ch-shrinkwrap still works if that's the case ...)

David-Baddeley: At this point it's probably easier to just delete than fix

PR #1345, opened 2023-04-14
try spooler test on azure again

barentine:
Addresses issue #1328 now that #1329 is merged.

enhancement
**Proposed changes:**
run the cluster spool test on azure now that we fixed what ought to be the issue


barentine:
Looks like it ran fine

PR #1348, opened 2023-04-14
FFBase.getROIAtPoint fixes (axialHalfROI arg and actually cropping background)

barentine:
bugfix + small doc
**Proposed changes:**
- axialHalfSize was being clobbered by roiHalfSize, and this undoes that. Should only effect 3D fit factories where FFBase.data.shape[2] > 1 to begin with.
- remove a superfluos `not` which was causing subtractBackground: True to signal returning 0 for background

### DIFF
--- a/PYME/localization/FitFactories/FFBase.py
+++ b/PYME/localization/FitFactories/FFBase.py
@@ -81,7 +81,7 @@ class FFBase(object):
             z = self.data[x, y, :].argmax()
     
         roiHalfSize = int(roiHalfSize)
-        axialHalfSize = int(roiHalfSize)
+        axialHalfSize = int(axialHalfSize)
     
         xslice = slice(int(max((x - roiHalfSize), 0)), int(min((x + roiHalfSize + 1), self.data.shape[0])))
         yslice = slice(int(max((y - roiHalfSize), 0)), int(min((y + roiHalfSize + 1), self.data.shape[1])))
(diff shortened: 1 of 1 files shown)


David-Baddeley:
The `not` is not superfluous - if `Analysis.subtractBackground` is `True`, the background is subtracted from the entire frame before we get here (extracting ROIs), making the effective ROI background 0.

barentine:
@David-Baddeley you're right for `MULTIFIT=False`.

What's the preferred tool for grabbing the ROI, and the background, if `MULTIFIT=True`? I assumed it also would be `getROIAtPoint`. Either way, some doc on what you just said would be appreciated. I'm not sure that it is always the case, e.g. for `getMultiviewROIAtPoint`, which is marked as experimental, sub background as True would flag actually returning an ROI background.

David-Baddeley:
If `MULTIFIT=True` it's generally assumed that you'll be handling your own ROI extraction (and won't need to use the ROI extraction logic), although there is nothing to stop doing the frame-based background-subtraction approach (basically just copying what happens in remFitBuff after the `MULTIFIT` branch) here too. If you just want to substitute object detection, we could look at making that pluggable (at the moment there is a hard-coded switch at line 490 for PRI, but this could be made more generic).

I'll have a look at `getMultiviewROIAtPoint` which is stalled WIP (the ultimate goal of which is to allow 4Pi fitting and similar with an interpolated model) for several years now. There is a good chance that the code here is incorrect as it's never been tested / used.

David-Baddeley:
Actually, `getSplitROIAtPoint` (which is widely used) would also seem to get this wrong.

David-Baddeley:
Looking through again, I think the you might be right with the superfluous `not` after all ... the FitFactory gets the unsubtracted data ....

David

barentine:
Cool, thanks for having another look @David-Baddeley! I ran into the issue while testing the double helix `MULTIFIT=True` fit I'm writing, and I agree that in that instance the background subtracted is zero, which I hadn't expected. 

For now I'd be happy to see the `not` removed, but going forward I think a framework that lets folks plug detection would be neat.

PR #1349, opened 2023-04-17
Automatic thresholding for dual marching cubes

David-Baddeley:
Add automatic thresholding based on the median distance from the surface to the localisations

### DIFF
--- a/PYME/LMVis/Extras/extra_layers.py
+++ b/PYME/LMVis/Extras/extra_layers.py
@@ -94,6 +94,24 @@ def gen_isosurface(visFr):
         dmc._invalidate_parent = True
         print('Isosurface layer added')
 
+def gen_point_dis_isosurface(visFr):
+    from PYME.LMVis.layers.mesh import TriangleRenderLayer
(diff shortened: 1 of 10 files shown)


David-Baddeley:
Also added a new octree class to allow efficient generation of meshes from SDFs. Usage as below. This is mostly a prelude to allowing CSG operations and, e.g. dilation and erosion on meshes. Usage as below:

```python
from PYME.experimental import func_octree
from PYME.experimental import dual_marching_cubes
from PYME.experimental import _triangle_mesh as triangle_mesh

def f(x, y, z):
    return np.sqrt(x**2 + y**2 + z**2) - 3

o = func_octree.FOctree([0, 10, 0, 10, 0, 10], f, maxdepth=6)
mc = dual_marching_cubes.PiecewiseDualMarchingCubes()
mc.set_octree(o)
tris = mc.march()
surf = triangle_mesh.TriangleMesh.from_np_stl(tris, smooth_curvature=True)
surf.to_stl('~/test_sdf_iso.stl')
```

Issue #1350, opened 2023-04-19
[BUG] # typo in updated MergeClump module


csoeller:
**Describe the bug**
There is a typo in the updated `run()` interface of the recipe module `MergeClump`

**To Reproduce**
1. Use `MergeClump` module in recipe
4. See error `unknown parameter Inp`

David-Baddeley:
I'm guessing we can close this now #1351 is merged

PR #1351, opened 2023-04-19
fix typo in mergeClumps argument

csoeller:
Addresses issue #1350 .

bugfix

**Proposed changes:**
fix argument type to correct use of `inputName`


David-Baddeley:
Thanks!

PR #1352, opened 2023-04-26
supress debug printing of metadatahandler contents when recipes are executed

csoeller:
Since the recipe refactor a sizable list of mdh printouts are generated when recipes are executed.
I suspect this reflects a debug leftover...

I'd consider it a bugfix (or rather cleanup)

**Proposed changes:**
remove left over `print()` statement

**Related question:**
The use of `DictMDHandler` in `execute` causes a number of warnings of the type

```
warnings.warn('DictMDHandler is not yet fully supported, and will likely cause failures for anything related to localisation fitting')
```

when recipes are excuted. This is a little distracting in Jupyter Notebooks. I can override as a user with

```
import warnings
# supress warnings from the DictMDHandler about inability to handle localisations
warnings.filterwarnings("ignore",message=r'DictMDHandler')
```

Just wondering it it makes sense that these warnings are always generated now that `DictMDHandler` is generally used in `recipes.base.execute()`? Is `DictMDHandler` still unsafe to use as part of standard modules?

### DIFF
--- a/PYME/recipes/base.py
+++ b/PYME/recipes/base.py
@@ -281,7 +281,7 @@ class ModuleBase(HasStrictTraits):
                 v.mdh = MetaDataHandler.DictMDHandler()
 
             v.mdh.mergeEntriesFrom(mdhin) #merge, to allow e.g. voxel size overrides due to downsampling
-            print(v.mdh, mdh)
+            #print(v.mdh, mdh)
             v.mdh.copyEntriesFrom(mdh) # copy / overwrite with module processing parameters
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Great minds think alike ... Also in #1349 ...

csoeller:
Ah, great, feel free to bin this one then.

What's your take on the `DictMDHandler` warning business?

David-Baddeley:
Warning is likely no longer required, will double check.

csoeller:
change set already merged with #1349

Issue #1187, opened 2022-01-20
[BUG] Mac M1 support


David-Baddeley:
This is a bit of a super-issue to track anything we come up with (note testing on M1Pro, running MacOS Montery 12.1)

# Rosetta (i386)

Need to make a copy of `Applications/Utilities/Terminal.app`, and set terminal to use Rosetta (see https://www.courier.com/blog/tips-and-tricks-to-setup-your-apple-m1-for-development/)

## existing conda development install (python=3.7)
from time machine backup, and likely indicative of the executable installer 

works with following caveats :

- sporadic "Illegal Insruction 4" message and hard crash (most likely the result of compilation with SSE/AVX support enabled). Reliably triggered by:
    - isosurfaces in dh5view
    - pointsprites in PYMEVis
    - running test suite

## Development install (python=3.7)

- Resolves visualisation related "Illegal Instruction 4" messages
- ~~Still get "Illegal Instruction 4"  on test suite~~ (fixed by building pymecompress from source, with latest changes [version 0.2.2], will need work on pymecompress packaging / building for multiple platforms.

## Executable installer (latest from python-microscopy.org/downloads/automated_builds)

- Need to enable installing of unsigned packages (try installing package, when it complains about being unsigned, click OK and then go to "System Preferences" -> "Security and Privacy" and click the allow / install button beside the the python-microscopy package.
- To run, you need to use a Rosetta terminal (as outlined above)
- Sporadic "Illegal Instruction 4" errors triggered as for the time machine backup case above

# Native

These instructions are starting from an i386 (Rosetta) miniconda install. If starting from scratch it might be simpler to use
a miniforge install (https://github.com/conda-forge/miniforge)

1) Create a new, **empty**, conda environment:
```conda create -n pyme_aarm64```
2) Activate the new environment:
```conda activate pyme_aarm64```
3) Setup so that this environment pulls arm64 packages:
```
conda env config vars set CONDA_SUBDIR=osx-arm64
conda deactivate pyme_aarm64
conda activate pyme_aarm64
```
4) Install (base) dependencies. Note, you currently need to use conda-forge and at least python=3.8:
```
conda install -c conda-forge python=3.8 numpy scipy matplotlib pytables pyopengl jinja2 cython pip requests pyyaml psutil pandas scikit-image scikit-learn sphinx
conda install -c conda-forge traits traitsui==7.1.0 pyface==7.1.0
```
5) build wxpython from source (the wxpython package on conda-forge is broken):

[...]

PR #1357, opened 2023-05-03
open OBF and MSR files without java/bioformats

barentine:
Addresses issue me wanting to open some files from an Aberrior instrument without messing with bio formats / java (see #1187 , though note that bioformats/java/M1 should hopefully be possible at this point)

enhancement
**Proposed changes:**
add an OBF datasource

note also: could probably spruce up our docs with a 'supported image formats' list somewhere. `DataFormats.rst` and `AnalysingForeignData.rst` are close, but someone wondering if they could open something other than tiff or a PYMEAcquire'd series wouldn't know from those pages about bioformats / some of the other support we have (dbl, numpy, dcimg, etc.)

### DIFF
--- a/PYME/IO/image.py
+++ b/PYME/IO/image.py
@@ -762,6 +762,34 @@ class ImageStack(object):
         self.seriesName = getRelFilename(filename)
 
         self.mode = 'default'
+    
+    def _loadOBF(self, filename):
+        from PYME.IO.DataSources import OBFDataSource
+        import numpy as np
+
+        stack_number = None
+        
+        if '?' in filename:
+            #we have a query string to pick the stack number
+            from six.moves import urllib
+            filename, query = filename.split('?')
+            
+            try:
+                stack_number = int(urllib.parse.parse_qs(query)['stack'][0])
+            except KeyError:
+                pass
+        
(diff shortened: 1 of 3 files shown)


PYME/IO/image.py
```diff
@@ -1270,6 +1314,11 @@ def Load(self, filename=None, prompt=None, haveGUI = False):
                 self._loadNPY(filename)
             elif filename.endswith('.dbl'): #treat this as being an image series
                 self._loadDBL(filename)
+            elif '.obf' in os.path.splitext(filename)[1] or '.msr' in os.path.splitext(filename)[1]:
+                self._loadOBF(filename)
+                # hack to get around self.filename = filename below, otherwise if we open multiple
```
David-Baddeley, line 1319: Won't that work with the query string?

barentine: Wasn't working for me. The query string gets parsed inside the load sub methods (e.g. loadOBF or the bioformats one), but at the bottom of `Load` proper, there's a `self.filename=filename`, so whatever went in originally is the stored one. Note that I'm not so sure what `self.seriesName` is really up to - I thought changing seriesName would do it, but it looks like we use filename for the window title / grabbing images for making composites.

PR #1359, opened 2023-05-05
allow FitFactories (custom localization routines) from plugins 

barentine:
Addresses issue on our roadmap (official or not, I'm not sure, but certainly discussed!)

enhancement
**Proposed changes:**
- allow fitfactories to be plugged using the new yaml plugin registration (NOTE - need to update `pyme-plugin`)
- small update to include reports folder in old-style plugin reg.

**Checklist:**
of the TODO list).

Note that I changed the fit factory import to use import_module, which the docs say to use instead of __import__ these days

We really ought to update the docs accordingly
- [x] Are there any other side effects of the change?


barentine:
This should now allow most fit factory imports to work for plugin fitfactories, with the exception of imports that are more specific, e.g. `self.interpolator = __import__('PYME.localization.FitFactories.Interpolators.' + interpModule , fromlist=['PYME', 'localization', 'FitFactories', 'Interpolators']).interpolator`

For full functionality throughout PYME we would have to do a bit more fit factory metadata collection. But I'd be happy for now with what's here, so I'm going to ~stop.

barentine:
Some notes on documentation I think we'll want to include:
- fithelpers need doc, including e.g. that first derivatives if provided need to be raveled
- Need to tell people about fitresult key name oddities, like x0 and y0 being special (called mine x and y the first time), and underscores being problematic

barentine:
Any thoughts on this @David-Baddeley ?

PYME/localization/FitFactories/__init__.py
```diff
@@ -22,8 +22,13 @@
 ################
 import glob
 import os
+from PYME import config
+import importlib
 
-__all__ = [os.path.splitext(os.path.split(p)[-1])[0] for p in glob.glob(__path__[0] + '/[a-zA-Z]*.py')]
+# discover PYME-internal fitfactories and fully qualify them
+__all__ = ['PYME.localization.FitFactories.' + os.path.splitext(os.path.split(p)[-1])[0] for p in glob.glob(__path__[0] + '/[a-zA-Z]*.py')]
```
David-Baddeley: `__all__` is a special variable (although this looks like it's my mistake, just propagated)

PYME/localization/FitFactories/__init__.py
```diff
@@ -52,3 +61,20 @@
...
+    ----------
+    fit_factory : str
+        fit factory module name, e.g. LatGaussFitFR
+
+    Returns
+    -------
+    module
+        loaded fit factory module
+    """
+    pkg = package[fit_factory]
```
David-Baddeley: Leaves the possibility for name collisions in plugins - maybe better to:

simple name -> PYME internal (check first)
fully resolved (i.e. `"my_plugin.my_factory"`) -> from plugin

PYME/localization/remFitBuf.py
```diff
@@ -346,7 +348,8 @@ def __init__(self, dataSourceID, frameIndex, metadata, dataSourceModule=None, re
     @property
     def fitMod(self):
         if self._fitMod is None:
-            self._fitMod = __import__('PYME.localization.FitFactories.' + self.fitModule, fromlist=['PYME', 'localization', 'FitFactories']) #import our fitting module
+            pkg = PYME.localization.FitFactories.package[self.fitModule]
+            self._fitMod = importlib.import_module('.'.join([pkg, self.fitModule]), pkg)  # import our fitting module
```
David-Baddeley: is there any reason not to use `import_fit_factory()` here?

David-Baddeley:
Sorry I've taken so long to have a look at this ... No major issues, but:

- I'd like to use fully resolved names (rather than just shortened names) for plugin modules as it a) makes it clear that it's coming from a plugin (which is good for reproducibility, and directing support queries to the right place) and b) reduces the chance of collisions .
- I'll probably change the `__all__` think to something non-broken.

d-perez-1:
image
- AESB & DDP working a fix on opening PYMEImage

```
  File "C:\ProgramData\Miniconda3\lib\site-packages\wx\core.py", line 3259, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "c:\userfiles\code\python-microscopy\PYME\DSView\dsviewer.py", line 470, in LoadData
    vframe = DSViewFrame(im, None, im.filename, mode = mode)
  File "c:\userfiles\code\python-microscopy\PYME\DSView\dsviewer.py", line 181, in __init__
    self.CreateFoldPanel()
  File "c:\userfiles\code\python-microscopy\PYME\ui\AUIFrame.py", line 139, in CreateFoldPanel
    genFcn(self.sidePanel)
  File "c:\userfiles\code\python-microscopy\PYME\DSView\modules\LMAnalysis.py", line 258, in GenAnalysisPanel
    self._populateCustomAnalysisPanel(self.customOptionsPan, self.customOptionsSizer)
  File "c:\userfiles\code\python-microscopy\PYME\DSView\modules\LMAnalysis.py", line 171, in _populateCustomAnalysisPanel
    pkg = PYME.localization.FitFactories.package[fitMod]
KeyError: 'LatGaussFitFR'
```

PR #1362, opened 2023-05-12
Label astigmatic CRB curves during PSF extraction

barentine:
Addresses issue #no one knows what those are unless they read the code

enhancement
**Proposed changes:**
label the astigmatic PSF curves
plot the dashed astig curves in the same color as their corresponding user PSF curves

Note - could also just not plot them? Because if you know what they are, your next question is for what NA objective, wavelength, etc?


David-Baddeley:
Thanks!

PR #1368, opened 2023-05-22
Add class to wrap multiaxis controllers to mimic single-axis piezo object

barentine:
Some existing code e.g. focus lock code is currently configured for a single-axis PiezoBase object. This `SingleAxisOnMultiaxisController` is essentially a shim to wrap a given axis of an initialized multiaxis controller as a single axis PiezoBase. This lets one use the focus lock code as normal even if your z axis is controlled by a 3-axis stage.

enhancement

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/base_piezo.py
+++ b/PYME/Acquire/Hardware/Piezos/base_piezo.py
@@ -63,7 +63,7 @@ class PiezoBase(object):
         raise NotImplementedError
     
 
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Seems reasonable, only potential critique is that class name is a bit verbose - I wonder if it makes sense to change to, e.g. `SingleAxisWrapper` or `SingleAxisPassthrough`

barentine:
I'm good with that. Changed to `SingleAxisWrapper`. Thanks @David-Baddeley

PR #1369, opened 2023-05-22
add pipython-based piezo class, note that e727 controller is supported

barentine:
enhancement
**Proposed changes:**
add a pipython-based GCS piezo class. I used this for an e727 stage, but hope would be we can remove some GCS boiler plate/ maitenance using their[ pipython package](https://github.com/PI-PhysikInstrumente/PIPython)

** note **
Would be really nice to set controller units to um. I didn't immediately get the encoding of um the way the DLL wanted it, but my controller was already in um so I left this alone for now.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
@@ -1,0 +1,144 @@
+
+# Defines a GCSPiezo class which uses the pipython package distributed 
+# by Physik Instrumente to initialize and run their stages.
+# Tested with a E-727 controller, on pipython 2.9.0.4 (pypi)
+# see https://github.com/PI-PhysikInstrumente/PIPython
+
+
+from PYME.Acquire.Hardware.Piezos.base_piezo import PiezoBase
+from pipython import GCSDevice
+from pipython import PILogger, WARNING
+import numpy as np
+import logging
+PILogger.setLevel(WARNING)
+
+logger = logging.getLogger(__name__)
+
+
+def get_gcs_usb():
+    """
+    returns list of PI devices connected by USB and their serial numbers. Use
+    to get str description suitable for initializing GCSPiezo, e.g.
+    'PI E-727 Controller SN 0118035989'
+    """
+    with GCSDevice() as pidev:
+        return pidev.EnumerateUSB()
+
+
+class GCSPiezo(PiezoBase):
+    units_um = 1  # assumes controllers is configured in units of um.
+    def __init__(self, description=None, axes=None):
+        """
+        Parameters
+        ----------
+        descriptions: str
+            description as found by GCS enumerate, e.g. 
+            PI E-727 Controller SN 0118035989.
+        axes : list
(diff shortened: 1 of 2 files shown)


PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
```diff
@@ -0,0 +1,144 @@
...
+            self.axes = axes
+        
+        units = self.GetUnits(axes)
+        logger.debug('stage units: %s' % units)
+        # if any([r'\u03BCm' != unit for unit in units]):
+        #     raise AssertionError('PYME expects stage to be in units of um, found: %s' % units)
+        # self.SetUnits(self.axes, None)  # None default sets all axes to um
+        
+    
+    def GetUnits(self, axes=None):
```
David-Baddeley: These have the potential to break spec. The implicit positioning API is that `GetPos()`, `SetPos()` etc ... work in the units specified by `class.units_um` (i.e. if the underlying unit is in um, this should be one, if mm this should be 1000). If we allow the stage to change units on the fly, we would have to make `units_um` a property and update the scaling factor whenever we called `SetUnits()`. 

More generally, I don't think there is a use case for exposing `SetUnits()` and `GetUnits()`  on the piezo class, as in general the only thing calling the class directly should be the `microscope` class /  the state manager. Everything else should really go through `scope.SetPos()` or `scope.SetState('Positioning.XXX', ...)` which are in um by definition.

David-Baddeley: Proxy piezos as are used in focus locks are an exception here, but I also can't see them needing to set units.

David-Baddeley: I could be persuaded to either:

- remove the unit setting functions completely (but still potentially setting units to um in `__init__` for safety)
- make them private to the class ie `_set_unit()`, and make `units_um` change accordingly (I don't really like this option)

David-Baddeley: One of the reasons I don't like the idea of supporting unit changing is that it opens up a can of worms for anything which is talking to the piezos directly, rather than through `scope.GetPos()` etc ... (which is still some of the code). If you, e.g. created a slider to set the position and then changed the units in a protocol, you'd need to find some way of triggering an update to the slider bounds and position. Potentially more destructively, if you had set up a z-stack or tiling in one set of units and then changed them while the code was running you could end up crashing the objective into something.

barentine: Because of how important it is to be in um, I generally like to set the stage to um in the init if it's an option in the API. That said, in this case, I wasn't getting the set units call to work, so I'm more than happy to remove it. 

I see your point though that rather than adding a 'private' method, just calling the set units to um hardcoded in the init would be ideal. For now, I'll just delete the methods and log the current units in the init (where we would ideally change this to a set um in the future).

barentine:
Should be set for now with last commit, @David-Baddeley

PR #1370, opened 2023-05-22
Pull fitfactory-specific default parameters in recipes.measurement.FitPoints

barentine:
Addresses issue Currently FitPoints does not fill fitfactory-specific metadata (i.e. the PARAMETERS list is ignored), which will fail a fit factory if it hard-pulls fit factory-specific metadata. 

in between
**Proposed changes:**
- take default fit factory parameters for measurement.FitPoints. Doing so allows hard-pulls on fit module specific metadata, and also propagates this metadata into the results.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/recipes/measurement.py
+++ b/PYME/recipes/measurement.py
@@ -193,6 +193,10 @@ class DetectPoints2D(ModuleBase):
 class FitPoints(ModuleBase):
     """ Apply one of the fit modules from PYME.localization.FitFactories to each of the points in the provided
     in inputPositions
+
+    Fit-specific metadata will be set to defaults specified in fitModule.PARAMETERS
+
+    Background subtraction is not currently supported in this module.
     """
     inputImage = Input('input')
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I think we should probably go a bit further here by adding a `DictStrAny` (or similar) trait for additional metadata to the `FitPoints` class so we can set non-default parameters. As defaults are usually handled within the fit factories, just manually adding the defaults feels like it shouldn't do anything. Although by no means 100% implemented this way, the choice not to use `mdh.get(key, default)` (or equivalent)  in the fit factory code has an implication that a suitable default might not be easy to decide on. The `PARAMETERS` section forces a default, but under the implicit assumption that the user will see an be able to modify this - these defaults might not be that suitable for blind use.

barentine:
I think we would actually benefit from a `DictStrAny` to add metadata in general, not just for this class!

PR #1382, opened 2023-06-12
Add PYMEAcquire intensity trace page / panel

barentine:
Addresses issue several of us in the lab have been using a fastgraph based intensity plot, where the average value within a user-selected ROI is plotted in time (and can be reset with a clear button on a control panel).

My writing it (based on the focus lock profile fastgraph code) predated the somewhat-recent display opts / selection tweaks. Having now updated it so it works on other more-recent installs in the lab, I would like to PR it so it gets maintained through further displayOpts/selection tweaks, etc, and also as discussed in our recent chat - we've already used this avg intensity trace on 3 setups, and as more features are discussed, pooling development in one place seems like a good idea.

Here is the current PR added to the init_sim_min.py script, with some simple modulation done by deselecting the ROI (avg set to zero) then selecting a new one:
image

**Proposed changes:**
- add a fastGraph based intensity in time plot for use in PYMEAcquire
- add a simple control panel for it.

Features that have been discussed: ability to change the extent of the time interval, label the axes (with units), potentially even save output to file

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/Acquire/ui/intensity_trace.py
+++ b/PYME/Acquire/ui/intensity_trace.py
@@ -11,12 +11,19 @@ logger = logging.getLogger(__name__)
 
 class IntensityTracePanel(FastGraphPanel):
     def __init__(self, parent, frame_wrangler, winid=-1, n_frames=1000):
-        """
-        If the square region select tool is active, will plot the average value within it
-        over time
+        """If the square region select tool is active, will plot the average value within it
+        over time. Note that the update rate is set by the GUI timer, not the camera frame rate.
+
         Parameters
         ----------
-        
+        parent : wx.Window
+            PYMEAcquire MainFrame / AUIFrame
+        frame_wrangler : PYME.Acquire.FrameWrangler
+            frame_wrangler object, which is grabbing new frames from the camera
+        winid : int, optional
+            direct pass-through to FastGraphPanel, by default -1
+        n_frames : int, optional
+            Number of frame-updates to store and display, by default 1000
         """
         self.frame_vals = np.arange(n_frames)
         self.intensity_avg = np.zeros_like(self.frame_vals, dtype=float)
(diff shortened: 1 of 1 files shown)


PYME/Acquire/ui/intensity_trace.py
```diff
@@ -0,0 +1,111 @@
...
+        self.intensity_avg = np.zeros_like(self.frame_vals, dtype=float)
+        FastGraphPanel.__init__(self, parent, winid, self.frame_vals, 
+                                self.intensity_avg)
+        self.wrangler = frame_wrangler
+        # differ display opts assignment because it might not have been created yet
+        self.do = None  # PYME.UI.displayOps.DisplayOptions gets assigned later
+        self._mf = weakref.ref(parent)  # weak ref to MainFrame
+        self._relative_val = 1
+
+    def assign_do(self):
```
David-Baddeley: this is cheap - what about making it a property instead of the testing and assign if not present logic (which is missing from `clear()`)

PYME/Acquire/ui/intensity_trace.py
```diff
@@ -0,0 +1,111 @@
...
+        if self.do is None:
+            try:
+                self.assign_do()
+            except:
+                return
+        if self.do.selection.mode != SELECTION_RECTANGLE:
+            return
+        x0, x1, y0, y1, _, _ = self.do.sorted_selection
+
+        self.intensity_avg[:-1] = self.intensity_avg[1:]
```
David-Baddeley: This is a relatively expensive way to do a rolling buffer. Might not matter here, as only 1000 values, so the simplicity could outweigh the cost, but worth keeping in mind if you want to increase the length/frequency (e.g. hooking the mean calculation and append to the `onFrame` signal, and display to `onFrameGroup` (or similar).

PYME/Acquire/ui/intensity_trace.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+# example init script plug:
+# @init_gui('intensity trace')
+# def intensity_trace(MainFrame, scope):
+#     from PYME.Acquire.ui.intensity_trace import IntensityTracePanel, TraceROISelectPanel
+
+#     intensity_trace = IntensityTracePanel(MainFrame, scope.frameWrangler)
+#     MainFrame.AddPage(page=intensity_trace, select=False, caption='Trace')
+#     MainFrame.time1.WantNotification.append(intensity_trace.refr)
```
David-Baddeley: **Note on choice of timers**  Timers in general should probably be reworked. 

`time1` is not really "API", and it's behaviour is not guaranteed (it has a 50ms period during startup, then drops to 500ms once everything is running). It might end up being called more than once per frame if the integration time is longer than timer period. It might go away / and or be made private at some point in the future.

The `FrameWrangler.onFrameGroup` signal is API. It has a default 100ms period (although there are no hard guarantees). It currently runs in the GUI thread, although this is not a hard guarantee (for a potential future hybrid web and wx version of PYME it will likely not run in the GUI thread), so it would be prudent to wrap GUI code in `wx.CallAfter` if using this signal.

David-Baddeley:
Have made a few modifications to allow recording of the frame-by-frame signal if desired (which to me feels a bit more generally useful)

PR #1390, opened 2023-06-26
powerControlable to mirror power_controllable in OBIS and MPBCW

barentine:
Addresses issue I was using AOTFControlled Lasers in grad school, where we mirror power_controllable to powerControlable. If you use the lasers standalone at the moment you won't get sliders even though you might want them/ should get them

bugfix
**Proposed changes:**

**Checklist:**
of the TODO list).


David-Baddeley:
Is there a reason to mirror this, rather than to just standardise on one spelling/format?

David-Baddeley:
have changed to use the canonical `powerControlable` spelling

David-Baddeley:
note we should also do this for aotf.py

PR #1392, opened 2023-06-27
add picoquant taiko m1 support

barentine:
Addresses issue I have a PicoQuant Taiko M1 that I need to mess with in a protocol, so I've added an interface for it. 

It's a pretty flexible laser, which is quite cool / something I take advantage of, but it means there's a lot to interface (pulsed int / ext, CW operation).

There's a good chance its worth writing a threaded version that uses it's "tag" functionality to query a whole list of things at once, so we can get everything we need to construct a fairly complicated scope state with a single serial command, but I figure let me run into a performance issue before I try and fix it

enhancement
**Proposed changes:**


David-Baddeley:
@barentine - Looking at the license on pdlm.py, It's a bit unclear whether we can merge/ distribute it. The first thing I saw was

"You may not distribute the software to third parties" 

but it then goes on to say that you can modify and distribute the demo code, which makes me think the first statement applied to the driver code, not the file in which it was found. In either case the combined statement is extremely unclear. I wonder if it might be worth reaching out to Picoquant to clarify and/or offering an alternative along the lines of:

"This code is modified from the demo code provided by PicoQuant GmbH for interfacing with the PDLM driver. 
You may use and modify demo code to create your own software. Original or modified demo code 
may be re-distributed, provided that the original disclaimer and notes are not removed from it.

Possession of this modified demo code does not confer any rights to the PDLM driver
software itself. The PDLM driver software is protected by copyright and intellectual property laws.
You may not distribute the PDLM driver software to third parties or reverse engineer, decompile or disassemble 
the software or any part thereof. Copyright of all product documentation belongs to PicoQuant GmbH. 
No parts of it may be reproduced or translated into other languages without written consent of PicoQuant GmbH. 
"

barentine:
thanks @David-Baddeley . I'll send an email to PicoQuant and keep you posted

Issue #1393, opened 2023-06-27
[BUG] countdir relative import fails


barentine:
**Describe the bug**
and another xfail for no neuroglancer
```
Using c:\programdata\miniconda3\envs\pyme\lib\site-packages
Finished processing dependencies for python-microscopy==23.6.15.post0.dev0

(pyme) C:\Userfiles\code\python-microscopy>pytest tests\PYME\test_imports.py
================================================= test session starts =================================================
platform win32 -- Python 3.7.11, pytest-6.2.4, py-1.10.0, pluggy-0.13.1
rootdir: C:\Userfiles\code\python-microscopy
collected 5 items

tests\PYME\test_imports.py F.F..                                                                                 [100%]

====================================================== FAILURES =======================================================
___________________________________________________ test_IO_imports ___________________________________________________

    @pytest.mark.xfail(not HAVE_WX, reason="Depends on wx, which is not installed on this platform")
    def test_IO_imports():
        from PYME.IO import buffer_helpers, buffers
        from PYME.IO import clusterExport, clusterGlob
        from PYME.IO import clusterIO, clusterListing, clusterResults, clusterUpload
        from PYME.IO import dataExporter, dataWrap, dcimg, h5rFile
        from PYME.IO import image, load_psf, MetaDataHandler, PZFFormat
>       from PYME.IO import ragged, tabular, unifiedIO, countdir

tests\PYME\test_imports.py:20:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

    from PYME import config
    if config.get('cluster-listing-no-countdir', False):
        raise ImportError('Not using countdir')

>   from .countdir import *
E   ModuleNotFoundError: No module named 'PYME.IO.countdir.countdir'

PYME\IO\countdir\__init__.py:5: ModuleNotFoundError
-------------------------------------------------- Captured log call --------------------------------------------------
WARNING  PYME.misc.pyme_zeroconf:pyme_zeroconf.py:45 using an old version of zeroconf (0.24.0), consider upgrading
_____________________________________________ test_DSView_modules_imports _____________________________________________

    @pytest.mark.xfail(not HAVE_WX, reason="Depends on wx, which is not installed on this platform")
    def test_DSView_modules_imports():
[...]

PR #1394, opened 2023-06-27
test coundir import separate, expect fail on windows

barentine:
Addresses issue we don't compile coundir.c on non-posix systems, but we fail the test on all currently
issue #1393 

bugfix
**Proposed changes:**
- test countdir import separate of other io imports
- mark xfail on win32

**Checklist:**
of the TODO list).

### DIFF
--- a/tests/PYME/IO/test_countdir.py
+++ b/tests/PYME/IO/test_countdir.py
@@ -1,4 +1,8 @@
 
+import pytest
+import sys
+
+@pytest.mark.xfail(sys.platform=='win32', reason='countdir.c not compiled on non-posix platforms')
 def test_dirsize():
     from PYME.IO import countdir
     import os
--- a/tests/PYME/IO/test_countdir.py
+++ b/tests/PYME/IO/test_countdir.py
@@ -2,7 +2,7 @@
 import pytest
 import sys
 
-@pytest.mark.xfail(sys.platform=='win32', reason='countdir.c not compiled on non-posix platforms')
+@pytest.mark.skipif(sys.platform=='win32', reason='countdir.c not compiled on non-posix platforms')
 def test_dirsize():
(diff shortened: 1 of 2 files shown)


tests/PYME/IO/test_countdir.py
```diff
@@ -1,4 +1,8 @@
 
+import pytest
+import sys
+
+@pytest.mark.xfail(sys.platform=='win32', reason='countdir.c not compiled on non-posix platforms')
```
David-Baddeley:
```suggestion
@pytest.mark.skipif(sys.platform=='win32', reason='countdir.c not compiled on non-posix platforms')
```

There's no point in trying on windows - I'd usually use xfail for something which *should* work but has some reason (e.g. lack of libraries on platform x) which is is making it fail in the test environment. If the test is not relevant, we should just skip it.

tests/PYME/test_imports.py
```diff
@@ -17,9 +17,13 @@ def test_IO_imports():
     from PYME.IO import clusterIO, clusterListing, clusterResults, clusterUpload
     from PYME.IO import dataExporter, dataWrap, dcimg, h5rFile
     from PYME.IO import image, load_psf, MetaDataHandler, PZFFormat
-    from PYME.IO import ragged, tabular, unifiedIO, countdir
+    from PYME.IO import ragged, tabular, unifiedIO
     
     #from PYME.IO import clusterDuplication #Known failure due to dependence on pyro
+
+@pytest.mark.xfail(sys.platform == 'win32', reason='countdir.c not compiled on non-posix systems')
```
David-Baddeley:
```suggestion
@pytest.mark.skipif(sys.platform == 'win32', reason='countdir.c not compiled on non-posix systems')
```

PR #1398, opened 2023-06-29
Mesh visualisation in notebooks

David-Baddeley:
Use pythreejs to render meshes in jupyter notebooks

image


zacsimile:
Gave it a go, didn't work for me. Do I need to call something else?

image

David-Baddeley:
The magic is in _triangle_mesh.pyx, so needs to be built. Did you force a rebuild by running `python setup.py build_ext -i`. Your output would suggest that the `_repr_mimebundle_()` method is not being found and called. Also, pythreejs needs to be installed in the same environment that the jupyter server is running in, as well as the environment you are using for your notebook (although this should have led to an error message which indicated this).

zacsimile:
Ah I totally forgot to rebuild. It works great!

image

PR #1402, opened 2023-07-06
Enable opening of headerless ViSP files

zacsimile:
Addresses issue #415.

### DIFF
--- a/PYME/IO/csv_flavours.py
+++ b/PYME/IO/csv_flavours.py
@@ -1,4 +1,4 @@
-import numpy as np
+import os
 from PYME.IO import MetaDataHandler
 
 import logging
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Seems reasonable ... suspect we can address this if/when it becomes necessary, but:

- we might need to be able to specify more than one extension per flavour
- we might want to do something a little more complex when doing the guessing. e.g. 1) take the subset of flavours which matches the given extension, 2) do column-matching within this subset.

David-Baddeley:
Also, very minor point, but I wonder if  `guess_flavour()` should take the extension, rather than the filename (slightly safer for, e.g. URLs on windows, where os.path parsing doesn't work properly).

Issue #1400, opened 2023-07-05
[ENHANCEMENT] Export ImageStack to series of images suitable for video-generation by ffmpeg


barentine:
**Is your feature request related to a problem? Please describe.**
I'm frustrated when I don't know how to quickly get PYMEImage rendering into a video other than doing a screen-recording (which I'm not particularly great at)

**Describe the solution you'd like**
Similar to PYMEVis animation, I'd like to export every frame from a stack as a png or so such that I can run FFMPEG and get a movie suitable for presentation

**Describe alternatives you've considered**
screen recording
imageJ

David-Baddeley:
Definitely worth considering, but potentially a bit more involved than it might first seem. The rendering code is pretty well embedded in the GUI and the current snapshot functionality (`View->Save as PNG`) is essentially a screen-grab. To automatically render every frame in a series, in a smooth exporter you'd want to either abstract or reproduce the rendering code in a form which didn't involve drawing things to a `wx.DC`. Reproduction will be bad for maintenance, and too much abstraction might not be great for performance (operating system graphics routines are generally pretty efficient - so pulling all rendering out into code which doesn't use the windowing libraries might not be ideal). 

With the current code structure it wouldn't be too hard to abstract the colour mapping code, but if we move to an openGL renderer in the future (somewhat on the cards) this logic will likely move into a shader, so we'd end up maintaining two versions. If you want any of the overlays (scale bar, annotations, etc ...) to be present in the exported video then abstracting the rendering code becomes quite tricky.

This leads me to the conclusion that, despite being a bit clunky, a screen capture approach (as we do for animation in PYMEVis) might be the most pragmatic. Basically just call the logic behind `View->Save as PNG` in a loop (there would be some fiddling here to make sure the display updated between frames - you might need to hook wx.EVT_IDLE to do your saving and incrementing).

barentine:
Closed with #1403

PR #1403, opened 2023-07-07
PYMEImage menu item to export series of PNG suitable for making a movie

barentine:
Addresses issue #1400 

enhancement
**Proposed changes:**
- menu item to iterate through display opts z and save the display to png
- docs to describe the View exports

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/DSView/arrayViewPanel.py
+++ b/PYME/DSView/arrayViewPanel.py
@@ -467,6 +467,28 @@ class ArrayViewPanel(scrolledImagePanel.ScrolledImagePanel):
         MemBitmap = self.GrabImage(fullImage)
         img = MemBitmap.ConvertToImage()
         img.SaveFile(filename, wx.BITMAP_TYPE_PNG)
+    
+    def ExportStackToPNG(self, filename, fullImage=True):
+        """Save current view to a series of PNG files with z (or t) index as suffix, suitable for use in making a movie
+        via ffmpeg or similar tools
+
+        Parameters
+        ----------
+        filename : str
+            fully qualified path, with extension. Note that _%d will be appended to the filename to generate the
+            individual files
+        fullImage : bool, optional
+            whether to export the full image even if it is clipped in the GUI, by default True
+        """
+        import os
+        filestub, ext = os.path.splitext(filename)
+        for ind in range(self.do.ds.shape[2]):
+            self.do.zp = ind
+            if ('update' in dir(self.GetParent())):
+                self.GetParent().update()
+            else:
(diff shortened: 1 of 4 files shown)


PYME/DSView/arrayViewPanel.py
```diff
@@ -467,6 +467,28 @@ def GrabPNG(self, filename, fullImage=True):
...
+        via ffmpeg or similar tools
+
+        Parameters
+        ----------
+        filename : str
+            fully qualified path, with extension. Note that _%d will be appended to the filename to generate the
+            individual files
+        fullImage : bool, optional
+            whether to export the full image even if it is clipped in the GUI, by default True
+        """
```
David-Baddeley, line 483:
```suggestion
        FIXME - make this work with time series / 5D image data model.
        """
```

PYME/DSView/modules/arrayView.py
```diff
@@ -29,13 +29,32 @@ def save_as_png(view):
     if filename.strip():
         view.GrabPNG(filename)
 
+def save_as_series_of_png(view, filename=None):
+    """Save current view to a series of PNG files with z (or t) index as suffix, suitable for use in making a movie
```
David-Baddeley, line 33: Currently only does z

docs/Analysis.rst
```diff
@@ -39,5 +39,7 @@ A particular mode can be forced by calling ``dh5view`` with the ``-m`` option, e
 
     dh5view -m LM filename
 
-
+Rendering
```
David-Baddeley:
```suggestion
Exporting colour-mapped images
```

I think the term "rendering" is a bit vague - especially for a biological reader. What we are doing here is exporting images with colour -mapping and scale bars etc ... This applies to all uses of rendering in the docs.

docs/PYMEImage_rendering.rst
```diff
@@ -0,0 +1,33 @@
+.. _pymeimage_rendering:
+
+
+Rendering Images and Movies with PYMEImage
+===========================================
+
+PYMEImage facilities exporting image data, and rendering, two similar but distinct tasks.
+Exporting images via `File > Save As` and `File > Export Cropped` will save the image data to disk,
+while here we will cover exporting and saving rendered views of these data.
```
David-Baddeley: See comments above - I don't really like the rendering term - we need to be a little more explicit that this is exporting images with LUT etc ... applied. It's probably also worth putting a sentence in about when you would (for presentations etc ) and wouldn't (for analysis) use data exported in this way.

PR #1417, opened 2023-08-23
allow test frame to handle not subtracting background in fit

barentine:
Addresses issue `FFBase._get_roi` assigns 0 to the `background` attribute if subtract background in fit is unchecked, which then breaks `LMAnalyser2.testFrame` because it checks for an array type, which scalar 0 is not. np.isscalar isn't robust to e.g. str, so I propose a fix with `numbers.Number` 

bugfix
**Proposed changes:**

### DIFF
--- a/PYME/DSView/modules/LMAnalysis.py
+++ b/PYME/DSView/modules/LMAnalysis.py
@@ -977,8 +977,9 @@ class LMAnalyser2(Plugin):
                 #figure()
                 #imshow()
             except AttributeError:
+                from numbers import Number
                 #d = self.image.data[:,:,zp].squeeze().T
-                if isinstance(ft.bg, np.ndarray):
+                if isinstance(ft.bg, np.ndarray) or isinstance(ft.bg, Number):
                     # We expect our background estimate to take the form of a numpy array, correct our data by subtracting the background
                     d = (ft.data - ft.bg).squeeze().T
                 elif hasattr(ft.bg, 'get_background'):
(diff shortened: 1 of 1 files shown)


David-Baddeley:
just checked and the `numbers` module is part of the standard library since 3.0 (didn't want to add a dependency).

That said, I wonder if we should fix `FFBase` instead? The proposed fix is specific to a python scalar and somewhat sets a pattern of allowing certain things and iteratively expanding the test (rather than duck-typing which is what it was before the GPU stuff was hacked in. If we are not going to duck-type (ie accept anything that you can perform subtraction with an array - i.e. arrays, scalars, dask arrays, array like objects etc ...) it might be preferable to enforce a single convention/api (i.e. `background must be an ndarray') instead. If we do want to duck-type, it's probably better to shift the logic and test for the GPU special case first and then just default to subtracting as though it's an array.

I'll probably just end up merging with an additional comment.

PYME/DSView/modules/LMAnalysis.py
```diff
@@ -977,8 +977,9 @@ def testFrame(self, gui=True):
                 #figure()
                 #imshow()
             except AttributeError:
+                from numbers import Number
                 #d = self.image.data[:,:,zp].squeeze().T
-                if isinstance(ft.bg, np.ndarray):
+                if isinstance(ft.bg, np.ndarray) or isinstance(ft.bg, Number):
                     # We expect our background estimate to take the form of a numpy array, correct our data by subtracting the background
```
David-Baddeley, line 984:
```suggestion
                    # TODO - revisit special cases - should we check for the GPU stuff first and then just duck-type everything else?
                    # We expect our background estimate to take the form of a numpy array, correct our data by subtracting the background
```

PR #1418, opened 2023-08-23
Allow visualization of localizations colored by cluster measurements

barentine:
Addresses issue at the moment folks can't run a clustering, map a density, then filter on this density, and see what that actually looks like applied to the localizations themselves. 

enhancement
**Proposed changes:**
- record the labelKey each cluster corresponds to in `MeasureClusters3D`. Not strictly necessary, but feels like a safer option than assuming everything will stay sorted later
- add a module to map per-cluster quantities (scalar) onto the localizations used to generate them, such that they can be visualized in the GUI. 

Example recipe for something that's been clustered with a label key `gmm_label`:

```
- localisations.AddPipelineDerivedVars:
    inputEvents: ''
    inputFitResults: FitResults
    outputLocalizations: Localizations
- localisations.ProcessColour:
    input: Localizations
    output: colour_mapped
- tablefilters.FilterTable:
    inputName: colour_mapped
    outputName: filtered_localizations
- localisations.MeasureClusters3D:
    inputName: filtered_localizations
    labelKey: gmm_label
    outputName: clusterMeasures
- localisations.AddMeasurementsByLabel:
    input_measurements: mapped
    input_points: filtered_localizations
    label_key: gmm_label
    output_points: annotated_points
- tablefilters.Mapping:
    inputName: clusterMeasures
    mappings:
      circular_density: count / (np.pi * (gyrationRadius ** 2))
    outputName: mapped
- tablefilters.FilterTable:
    filters:
      gmm_label_circular_density:
      - 0.005
      - 9000000000.0
    inputName: annotated_points
    outputName: filtered_and_annotated
```

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/recipes/localisations.py
+++ b/PYME/recipes/localisations.py
@@ -1030,8 +1030,11 @@ class MeasureClusters3D(ModuleBase):
             cmorph.measure_3d(x, y, z, output=measurements[cluster_index])
 
             indi = indf
-
-        return tabular.RecArraySource(measurements)
+        
+        # propagate the labelKey used to generate these measurements
+        output = tabular.MappingFilter(tabular.RecArraySource(measurements))
+        output.addColumn(self.labelKey, unique_labels)
+        return output
 
 
 @register_module('FiducialCorrection')
--- a/PYME/recipes/localisations.py
+++ b/PYME/recipes/localisations.py
@@ -1069,7 +1069,7 @@ class AddMeasurementsByLabel(ModuleBase):
         # only propagate 1D measurements
         annotations = dict()
         for k in input_measurements.keys():
-            if input_measurements[k].ndim == 1:
+            if input_measurements[k].ndim == 1: #TODO - also check for object dtype?
                 annotations[k] = np.zeros(len(input_points), dtype=input_measurements[k].dtype)
         
         try:
--- a/PYME/recipes/localisations.py
+++ b/PYME/recipes/localisations.py
@@ -1037,6 +1037,69 @@ class MeasureClusters3D(ModuleBase):
         return output
 
 
+@register_module('AddMeasurementsByLabel')
+class AddMeasurementsByLabel(ModuleBase):
+    input_points = Input('input')
+    input_measurements = Input('clusterMeasures')
+    label_key = CStr('label')
+    output_points = Output('annotated_points')
+
+    def run(self, input_points, input_measurements):
+        """
+        Propagate measurements from e.g. MeasureClusters3D back to points they were calcualted from.
+        This is particularly useful for visualizing e.g. the gyration radius of a cluster in PYMEVis
+        while looking at the original localization data.
+
+        Parameters
+        ----------
+        input_points : PYME.IO.tabular.TabularBase
+            points used to generate the measurements
+        input_measurements : PYME.IO.tabular.TabularBase
+            measurements to propagate back to the points
+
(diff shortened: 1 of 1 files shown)


PYME/recipes/localisations.py
```diff
@@ -1030,8 +1030,74 @@ def run(self, inputName):
             cmorph.measure_3d(x, y, z, output=measurements[cluster_index])
 
             indi = indf
+        
+        # propagate the labelKey used to generate these measurements
+        output = tabular.MappingFilter(tabular.RecArraySource(measurements))
+        output.addColumn(self.labelKey, unique_labels)
```
David-Baddeley: This looks wrong ... should be `np.arange(maxLabel)` rather than `unique_labels`

PYME/recipes/localisations.py
```diff
@@ -1030,8 +1030,74 @@ def run(self, inputName):
...
+            point data with new columns added, one for each scalar measurement in input_measurements, which
+            can be accessed by '<label_key>_<measurement_key>', e.g. 'clumpIndex_gyrationRadius'.
+        
+        """
+        from PYME.IO.tabular import MappingFilter
+        
+        # only propagate 1D measurements
+        annotations = dict()
+        for k in input_measurements.keys():
+            if input_measurements[k].ndim == 1:
```
David-Baddeley: should we also check for `Object` dtype??

David-Baddeley: Can probably just leave this as a TODO comment.

PYME/recipes/localisations.py
David-Baddeley: @barentine - any thoughts of this

barentine: Sorry for the delay! I think `unique_labels` is considerably safer here than a `np.arange` call. `unique_labels` matches what we actually iterate over, and is chosen to be robust against cases where you've filtered/merged/somehow ended up with non-contiguous labels, e.g. 0, 1, 2, 4 which would cause us to propagate the wrong labels if we used `np.arange`. Obviously nice if labels are contiguous, but would be a nasty place to silently break things if they aren't.

David-Baddeley: Note comments in code above, which would imply `arange` is the correct choice:

``` python
# Note that missing labels result in zeroed entries (i.e. the initial values are not changed).
# Missing values can be filtered out later, if desired, by filtering on the 'counts' column, but having a dense
# array where index == label number makes any postprocessing in which we might want to find the data
# corresponding to a particular label MUCH easier and faster.

```

and the assignment:
 `output=measurements[cluster_index])`

We iterate over the set of unique labels, but assign into the array using the label, not it's position.

barentine: Yep, looks good! Sorry I missed that and was thinking more aligns of the new `AddMeasurementsByLabel`. Please feel free to change to arange here!

David-Baddeley:
```suggestion
        output.addColumn(self.labelKey, np.arange(maxLabel))
```

PYME/recipes/localisations.py
```diff
@@ -1030,8 +1030,74 @@ def run(self, inputName):
...
```
David-Baddeley:
```suggestion
            if input_measurements[k].ndim == 1: #TODO - also check for object dtype?
```

PR #1419, opened 2023-08-23
Fix for #1416

### LINKED ISSUE #1416: [BUG] intensity_trace GUI broken
**Describe the bug**
Initializing PYMEAcquire with intensity_trace GUI makes it so the whole GUI does not update, and repeatedly print this error.

```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
  File "c:\userfiles\code\python-microscopy\PYME\ui\mytimer.py", line 42, in Notify
    a()
  File "c:\userfiles\code\python-microscopy\PYME\Acquire\ui\intensity_trace.py", line 96, in update_display
    if self._buf_idx > self.frame_vals:
```

self.frame_vals is called as if it was a value but its and array

```
if self._buf_idx > self.frame_vals:
            #move data backwards in the buffer so that the last data point is at the right hand side of the displayed interval.

            offset = self._buf_idx - self.frame_vals
            self.intensity_avg[:self.frame_vals] = self.intensity_avg[offset:(offset+ self.frame_vals)]
            self._buf_idx = self.frame_vals
        
        print(self.intensity_avg[self.frame_vals-1])
        self.SetData(self.frame_vals, self.intensity_avg[:self.frame_vals])
```

David-Baddeley:
Fixes #1416 (at least hopefully). Several of the `self.frame_vals` references should have been `n_frames` aka `len(self.frame_vals)`

### DIFF
--- a/PYME/Acquire/ui/intensity_trace.py
+++ b/PYME/Acquire/ui/intensity_trace.py
@@ -69,7 +69,9 @@ class IntensityTracePanel(FastGraphPanel):
             
         # try to get data from the signal argument (if we bound to the onFrame signal), otherwise use
         # the current frame (if bound to onFrameGroup, time1)
-        data = kwargs.get('frame_data',self.wrangler.currentFrame) 
+        data = kwargs.get('frameData',self.wrangler.currentFrame).squeeze() 
+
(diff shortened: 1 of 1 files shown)


d-perez-1:
works, thank you!

PR #1422, opened 2023-09-14
Thorlabs motorized filter flipper

barentine:
enhancement
**Proposed changes:**
- add @zacsimile 's [kinesis ctypes interface](https://github.com/Thorlabs/Motion_Control_Examples/pull/14) , and implements the MFF101 two-position motorized filter flip-mount.

I'm currently using one to flip a QWP into / out of a beam path, hence the example. Includes a GUI panel which a drop-down to toggle positions by custom names (e.g. 'QWP' or 'Blank' in my example).

**Checklist:**
of the TODO list).


David-Baddeley:
Thanks.

This looks like it could be a candidate for the `FilterWheel` interface (`PYME.Acquire.Hardware.FilterWheel.FilterWheelBase`), which would save on creating a new interface, along with some UI duplication.

David-Baddeley:
Many thanks!

Issue #1423, opened 2023-09-15
[BUG] "test this frame" always tests on first frame of sequence regardless of current frame position


csoeller:
**Describe the bug**
Whenever one tries "Test this frame" with the LMAnalyser in dh5view, only the first frame is tested, the logic to find the current frame positions seems broken (presumably still from pre-5D data times):

![Screenshot 2023-09-15 at 11 50 52](https://github.com/python-microscopy/python-microscopy/assets/3750860/46c4c721-a0ff-4bfe-a54b-505b77a13ead)

and the shown event locations all are from frame 0:

![Screenshot 2023-09-15 at 11 51 12](https://github.com/python-microscopy/python-microscopy/assets/3750860/99471bac-0d86-4b11-aea2-d85a73d7b64b)

**To Reproduce**
1. Go to 'Test this frame' in `dh5view`
2. Be sure to be on a later frame number in the view
3. Press 'Test this frame'
4. See that always frame 0 is tested

**Expected behavior**
It should test on the current frame.

**Screenshots**
Attached above.

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 Checked on latest github `Master` on macos.

David-Baddeley:
An embarrassing regression, thanks for spotting it!

PR #1424, opened 2023-09-15
fix "Test This Frame" behaviour to properly work on current frame in dh5view

csoeller:
Addresses issue #1423 .

Bugfix.

**Proposed changes:**

Tries to fix the code that figures out the current frame to work on to be compatible with 5D data model.

Note that the fix has been tested with single channel data only and is probably still a hack in that it does currently **not** properly fix `visible_bounds` in `arrayViewPanel.py` where the fix should probably go (see comment in code).

_This is only a work in progress_ and hopefully the proper fix in `visible_bounds` can be made by working on this PR, I am still not sure about the 5D logic fully at this stage...

**Screenshots:**

Works as expected with this fix:

Screenshot 2023-09-15 at 12 13 22

Screenshot 2023-09-15 at 12 13 39

**Checklist:**

Tested for basic functionality with single channel data on macos with latest code from master and this fix in.

### DIFF
--- a/PYME/DSView/fitInfo.py
+++ b/PYME/DSView/fitInfo.py
@@ -164,14 +164,16 @@ class FitInfoPanel(wx.Panel):
     def DrawOverlays(self, vp, dc):
         do = vp.do
 
-        if do.ds.shape[2] > 1:
-            # stack has z
-            zp = do.zp
-        else:
+        if do.ds.shape[3] > 1:
             # stack is a time series
-            zp = do.tp
+            t = do.tp
+        else:
+            # stack is a formatted as a z-stack - pretend it's a time series
+            # this can occur due to limitations in the backwards compatibility code that guesses
+            # whether a stack is a time series or not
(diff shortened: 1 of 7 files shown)


PYME/DSView/fitInfo.py
```diff
@@ -163,7 +163,15 @@ def UpdateDisp(self, index):
         
     def DrawOverlays(self, vp, dc):
         do = vp.do
-        frameResults = self.fitResults[self.fitResults['tIndex'] == do.zp]
```
David-Baddeley, line 166: with the 5D model, we should really be checking for t first, will modify accordingly.

David-Baddeley:
@csoeller - Many thanks for picking up on this. Could you verify it still works with my modifications to bring everything inline with xyztc model?

csoeller:
@David-Baddeley, many thanks for looking into it! I can confirm the fix worked well in my testing - see you have merged already so should be all good

Issue #1231, opened 2022-05-04
[ENHANCEMENT] Context aware warnings


David-Baddeley:
**Is your feature request related to a problem? Please describe.**
Sometimes it is appropriate to warn the user with a popup, sometimes it isn't, see, e.g. ADOffset estimation warning in #1226. If you are interactively setting up to do localisation analysis in dh5view, a pop-up warning that you might be using a brain-dead estimation of the ADOffset is probably useful.  Much less so when running a recipe, in which case you probably just want it logged.

**Describe the solution you'd like**
A `pyme.warnings` or similar module with, e.g., and `automated` context manager and a `warn()` function that always emits a `logger.warning()` (and/or `warnings.warn()`), but shows a popup when a ui is present and you are not in an automated context.

Could also be used/integrated with the existing UI error context manager.

**Describe alternatives you've considered**
- hooking `logging.warn`
- doing everything with logging, and having a visible display of log warnings and errors

PR #1429, opened 2023-10-02
warnings module, bugfixes

David-Baddeley:
Adds context aware warnings, fixes some regressions / bugs


David-Baddeley:
Partial solution to #1231

Issue #1433, opened 2023-10-09
[BUG] (mostly harmless) error when cancelling .csv and .mat file opening in PYMEVis


csoeller:
**Describe the bug**
An error is raised when the tabular dialogs for `.csv` and `.mat` files are canceled

Note: error is largely inconsequential but should ideally be avoided. Fix is quite simple, PR to follow...

**To Reproduce**
1. Ty opening `.cvs` file in PYMEVis
2. Cancel tabular dialogs that show field mapping 

**Error Message**

```
Traceback (most recent call last):

  File "/Users/csoe002/miniconda3/envs/pyme38_2/lib/python3.8/site-packages/wx/core.py", line 2346, in Notify

    self.notify()

  File "/Users/csoe002/miniconda3/envs/pyme38_2/lib/python3.8/site-packages/wx/core.py", line 3552, in Notify

    self.result = self.callable(*self.args, **self.kwargs)

  File "/Users/csoe002/Documents/src/PYMEsrc/pyme38/PYME/LMVis/visCore.py", line 800, in OpenFile

    self.pipeline.OpenFile(filename, **args)

TypeError: OpenFile() argument after ** must be a mapping, not NoneType
```

**Expected behavior**
No error.

David-Baddeley:
@csoeller - can we close this now #1434 has been merged?

csoeller:
fixed with #1434

PR #1434, opened 2023-10-09
Log and deprecation cleanup

csoeller:
Addresses issue #1433 .

Bugfix **and** logging/depreciation cleanup.

**Proposed changes:**

- fix bug #1433 by checking return value from `_populate_open_args` for being `None` as indication of cancelation
- fix some wx deprecation warnings by moving to newer method names (hopefully compatible with min wx version we require)
- as part of general warning/message cleanup turn print statements into debug messages

Note that I am trying to cleanup messages to turn remaining print statements into log messages and fix deprecation warnings so that messages from PYMEVis are more readable and can be filtered as needed etc. I have not yet looked at dh5view for anything similar.

**Checks:**
Checked with latest master on macOS

### DIFF
--- a/PYME/LMVis/visCore.py
+++ b/PYME/LMVis/visCore.py
@@ -135,7 +135,7 @@ class VisGUICore(object):
         #logger.debug('Ev Idle')
         if self.glCanvas._is_initialized and not self.refv:
             self.refv = True
-            logger.debug((self.viewMode, self.pointDisplaySettings.colourDataKey))
+            #logger.debug((self.viewMode, self.pointDisplaySettings.colourDataKey))
             self.SetFit()
            
             if self._new_layers:
--- a/PYME/LMVis/visCore.py
+++ b/PYME/LMVis/visCore.py
@@ -127,25 +127,25 @@ class VisGUICore(object):
         
     
     def OnIdle(self, event=None):
-       """ Refresh the glDisplay *after* all windows have been created and data loaded.
+        """ Refresh the glDisplay *after* all windows have been created and data loaded.
        
-       TODO - rename, as the OnIdle name is a historical artifact (was originally called using an wx.EVT_IDLE binding, 
-       now uses wx.CallLater with a delay.
-       """
+        TODO - rename, as the OnIdle name is a historical artifact (was originally called using an wx.EVT_IDLE binding, 
+        now uses wx.CallLater with a delay.
+        """
         #logger.debug('Ev Idle')
         if self.glCanvas._is_initialized and not self.refv:
             self.refv = True
             logger.debug((self.viewMode, self.pointDisplaySettings.colourDataKey))
             self.SetFit()
-            
+           
             if self._new_layers:
                 pass
-                # if self.pipeline.ready and not len(self.layers) > 0:
-                #     l = self.add_layer(method='points')
-                #     if 't' in self.pipeline.keys():
-                #         l.engine.set(vertexColour='t')
-                #     elif 'z' in self.pipeline.keys():
-                #         l.engine.set(vertexColour='t')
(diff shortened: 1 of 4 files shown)


PYME/ui/AUIFrame.py
```diff
@@ -61,7 +64,7 @@ def AddPage(self, page=None, select=True,caption='Dummy', update=True):
             self._mgr.Update()
             pn = self._mgr.GetPaneByName(self.pane0)
             if pn.IsNotebookPage():
-                print((pn.notebook_id))
+                logger.debug((pn.notebook_id))
```
David-Baddeley:
```suggestion
                #logger.debug((pn.notebook_id))
```
I think this one can just be commented out (potentially also the creating fold panel one below, although that probably doesn't hurt to log). In general, when we are replacing print statements with logging calls, we should check if they still serve a purpose and if the provide enough context to be useful (just printing the value of a handle clearly doesn't). If they don't have enough context, then I think it's best to just comment them out.

PYME/LMVis/visCore.py
```diff
@@ -127,10 +127,10 @@ def __init__(self, use_shaders=False):
         
     
     def OnIdle(self, event=None):
-        print('Ev Idle')
+        logger.debug('Ev Idle')
         if self.glCanvas._is_initialized and not self.refv:
             self.refv = True
-            print((self.viewMode, self.pointDisplaySettings.colourDataKey))
+            logger.debug((self.viewMode, self.pointDisplaySettings.colourDataKey))
```
```suggestion
            #logger.debug((self.viewMode, self.pointDisplaySettings.colourDataKey))
```

PYME/LMVis/visCore.py
```diff
@@ -127,10 +127,10 @@ def __init__(self, use_shaders=False):
         
     
     def OnIdle(self, event=None):
-        print('Ev Idle')
+        logger.debug('Ev Idle')
```
```suggestion
       """ Refresh the glDisplay *after* all windows have been created and data loaded.
       
       TODO - rename, as the OnIdle name is a historical artifact (was originally called using an wx.EVT_IDLE binding, 
       now uses wx.CallLater with a delay.
       """
        #logger.debug('Ev Idle')
```

Don't really need this one either

PYME/LMVis/visCore.py
```diff
@@ -147,10 +147,10 @@ def OnIdle(self, event=None):
                 
             self.glCanvas.Refresh()
             self.glCanvas.Update()
-            print('refreshed')
+            logger.debug('refreshed')
```
```suggestion
            logger.debug('Refreshed glCanvas after load')
```

PYME/LMVis/visCore.py
```diff
@@ -792,13 +792,16 @@ def OpenFile(self, filename, recipe_callback=None):
...
             self.layers.pop()
         
-        print('Creating Pipeline')
+        logger.debug('Creating Pipeline')
         if filename is None and not ds is None:
             self.pipeline.OpenFile(ds=ds)
         else:
             args = self._populate_open_args(filename)
+            if args is None:
+                logger.info("OpenFile was canceled or not a valid file format")
```
```suggestion
                #TODO - should we raise an exception in _populate_open_args() instead? Do we need a hard-fail here?
                logger.info("OpenFile was canceled or not a valid file format")
```

csoeller: I have added a commit to do all logging and checking within _populate_open_args() since it knows the details what failed and why. I would not generally raise an exception if a user canceled but have added a warning example that brings up a GUI msg for the user when in VisGUI. Please have a look and edit as required.

In other cases `_populate_open_args(`) can in future raise exception if warranted (for example, if we add new checks or file formats to this function).

`OpenFile` now just gracefully returns without creating a new pipeline if a None return value is received, assuming that all logging/warning is done already.

Issue #1435, opened 2023-10-10
[BUG] some issues in unpickling older metadata


csoeller:
**Describe the bug**
On occasion there are issues unpickling older metadata, error being:

```
Traceback (most recent call last):

  File "/Users/csoe002/Documents/src/PYMEsrc/pyme38/PYME/IO/MetaDataHandler.py", line 871, in getEntry

    val = pickle.loads(base64.b64decode(val.encode('ascii')))

UnicodeDecodeError: 'ascii' codec can't decode byte 0x94 in position 0: ordinal not in range(128)
```

This seems fixable, see PR that is to follow.

**To Reproduce**
1. Open older image (possibly written with python 2.X)
2. Inspect `dh5view` logs and metadata
3. See that some values of metadata could not be decoded (screenshots below)

**Expected behavior**
Ideally still able to read olter metadata.

**Screenshots**

![Screenshot 2023-10-10 at 13 32 17](https://github.com/python-microscopy/python-microscopy/assets/3750860/7e7af9a3-59ef-4f98-8c25-652495796e8c)
![Screenshot 2023-10-10 at 13 32 32](https://github.com/python-microscopy/python-microscopy/assets/3750860/5ffd995c-1cda-43cf-bf73-ad1110b48e97)

csoeller:
fixed with #1436

PR #1436, opened 2023-10-10
MetaData unpickling of some values, add 'latin1' encoding option

csoeller:
Addresses issue #1435 .

bugfix

**Proposed changes:**
Add `enconding='latin1'` option to relevant `pickle.loads` command

**Checks:**
Tested with old and current data, seems to work for old data and not clobber newer data reading.

**Screenshots:**
Before
Screenshot 2023-10-10 at 13 32 17

After

Screenshot 2023-10-10 at 13 42 58


David-Baddeley:
Thanks!

PR #1438, opened 2023-10-13
add function to change obis cw or modulated

d-perez-1:
Enhancement

**Proposed changes:**
Add function to select operation mode for obis lasers. Choose between CW power, digital, and analogue modulation. Tested and works for USB connection to controller for 594 (LS) and 405 (LX) Obis.

### DIFF
--- a/PYME/Acquire/Hardware/Coherent/OBIS.py
+++ b/PYME/Acquire/Hardware/Coherent/OBIS.py
@@ -61,12 +61,15 @@ class CoherentOBISLaser(Laser):
 
 
         time.sleep(1)
-
+        
+        self.is_on = False
         self.power = 0
         self.SetPower(init_power)
+        self.mode=0
+        self.SetOpMode(self.mode)
         self.MIN_POWER = 1e3 * float(self.query(b'SOUR:POW:LIM:LOW?\r\n', lines_expected=1)[0])
         self.MAX_POWER = 1e3 * float(self.query(b'SOUR:POW:LIM:HIGH?\r\n', lines_expected=1)[0])
-        self.is_on = False
+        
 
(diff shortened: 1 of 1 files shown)


PYME/Acquire/Hardware/Coherent/OBIS.py
```diff
@@ -138,6 +141,35 @@ def SetPower(self, power):
     def GetPower(self):
         return self.power
 
+    def SetOpMode(self,mode):
```
David-Baddeley, line 144: Is this something we might encounter on other lasers? Do we need operating mode as part of the general laser interface? I don't think we need to answer those yet, but should probably put a TODO with those questions in the comments. In the interim, I think leaving this as an OBIS specific command should be fine, but it might be worth signalling that it is OBIS specific in the method name - e.g. 
```suggestion
    def set_obis_operating_mode(self,mode):
```

PYME/Acquire/Hardware/Coherent/OBIS.py
```diff
@@ -138,6 +141,35 @@ def SetPower(self, power):
...
+        """
+        mode = 0: CWP, CW with constant power
```
David-Baddeley, line 146: Arbitrary integer constants doesn't make for very readable code (where this is being called). Having class level constants, e.g. `CoherentOBISLaser.OBIS_MODE_CW` is better style / more readable in calling code ...

compare:

```python
my_laser.SetOpMode(2) # Doesn't convey what this is doing without looking at the laser code
```
with
```python
m_laser.SetOpMode(my_laser.OBIS_MODE_ANALOG_MOD)
```
The other pythonic way of doing it would be to take a string, e.g. 'CW', 'ANALOG_MOD', etc ... and do the translation into mode numbers in the method.

PR #1446, opened 2023-12-13
InterpFitR-3D fit info visualization and two-modality acquisition

Yujin-Bao:
Bugfix: plots in the fit info not showing up
Enhancement: OI-DIC acquisition coordinate with transmitted-light drift tracking

**Proposed changes:**
- When using InterpFitR-3D in LMAnalysis, some plots in the 'fit info' dialog do not show up after clicking 'test this frame'. The first commit fixes this issue.
- I realize z-stacking PAINT makes cCoeff smaller but the focus lock still works. The change in the second commit will not de-active the focus lock when cCoeff is small.
- The third change coordinated the OI-DIC acquisition with enabling/disabling the drift tracking. I feel I am using a very aggressive way and there may be better ways to do this.

### DIFF
--- a/PYME/Acquire/Hardware/driftTracking.py
+++ b/PYME/Acquire/Hardware/driftTracking.py
@@ -343,7 +343,8 @@ class Correlator(object):
             self.history.append((time.time(), dx, dy, dz, cCoeff, self.corrRef, self.piezo.GetOffset(), self.piezo.GetPos(0)))
             eventLog.logEvent('PYME2ShiftMeasure', '%3.4f, %3.4f, %3.4f' % (dx, dy, dz))
             
-            self.lockActive = self.lockFocus and (cCoeff > .5*self.corrRef)
+            #self.lockActive = self.lockFocus and (cCoeff > .5*self.corrRef)
+            self.lockActive = self.lockFocus
             if self.lockActive:
                 if abs(self.piezo.GetOffset()) > 20.0:
                     self.lockFocus = False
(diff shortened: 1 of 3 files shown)


PYME/Acquire/Hardware/driftTracking.py
```diff
@@ -343,7 +343,8 @@ def tick(self, **kwargs):
             self.history.append((time.time(), dx, dy, dz, cCoeff, self.corrRef, self.piezo.GetOffset(), self.piezo.GetPos(0)))
             eventLog.logEvent('PYME2ShiftMeasure', '%3.4f, %3.4f, %3.4f' % (dx, dy, dz))
             
-            self.lockActive = self.lockFocus and (cCoeff > .5*self.corrRef)
```
David-Baddeley, line 346: What motivated this change?

Some form of lock checking code is definitely required to prevent runaway when the system falls out of lock (otherwise any error condition can end up with the drift correction driving the piezo right to one end of it's travel and, at the very least, completely dropping the focus, but potentially also loosing immersion oil etc ...). 

Lock can be lost when:
- there is too much lateral drift (causing calibration and true transmitted images to loose correlation)
- there are glitches in the drift tracking camera and/or illumination (the current logic seemlessly recoverers from these - i.e. when the correlation comes back, the lock continues).
- the system is bumped (or similar) resulting in a sudden position change to outside the calibration region
- your z-stack goes outside the calibration region.

It's imperative to detect and handle these - The current correlation coefficient way is fairly crude, but can't be removed without an effective and tested replacement.

David-Baddeley:
As mentioned above, drift tracking change is problematic.

As suspected, changes to xyztc.py likely not really appropriate, but we should discuss how best to achieve this in a setup-specific way.

The `InterpFitR` changes look fine, and look like they address regressions (not sure why our tests didn't pick these up)

PR #1447, opened 2023-12-14
add basic acquire server usage doc

barentine:
Addresses issue potential PYME user interested in different UI than our wx-based UI. They are interested in software development, so I wanted to show them and others what we already have in the space of an alternative UI.

enhancement
**Proposed changes:**
- add page on acquire server / web ui
- link from PYMEAcquire basic usage page

**Checklist:**
of the TODO list).


David-Baddeley:
I realise this is dear to your heart, but I wonder if we are jumping the gun a bit here (particularly with linking the web UI from the main PYMEAcquire docs). The Web UI as it stands is not really functional, and is likely to be a source of immense frustration for anyone who tries to use it. 

I'll have a look to see how we can incorporate it  - maybe, e.g. under a `Future directions` heading rather than the current `Changing the UI` heading.

barentine:
Sounds good, @David-Baddeley . All fair points.  `Future directions` would make more sense, probably, particularly for folks excited by it who want to get involved

David-Baddeley:
also just a note that once #1461 is merged, `PYMEAcquire -s` will launch the PYMEAcquire normally, but with the server running at the same time.  `PYMEAcquire -s -b` will launch the browser interface as well.

PR #1450, opened 2023-12-18
expose andor cam setemgainmode from the ixon class

barentine:
Addresses issue our lab members tend to run their setups with gain mode 3, and have each individually hacked it in

enhancement
**Proposed changes:**
- expose SetEMGainMode from the iXon class

**Checklist:**
of the TODO list).


David-Baddeley:
Added a bit of info to the help warning against using anything other than mode 0, as this is assumed elsewhere in the software (in the gain calibration routines and in the conversion to `RealEMGain`. In practice the gain calibration code might run correctly for modes 2 and 3 as long as you stick to one mode and don't change it, but it will likely have a reduced range.

barentine:
sounds good - thanks David!

PR #1451, opened 2023-12-19
Recipe modules added for 180 degree image rotation and spline fitting background subtraction

Yujin-Bao:
Enhancement

**Proposed changes:**
I added two modules under 'processing', for 180 image rotation and spline fitting background subtraction.

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2853,7 +2853,7 @@ class RawADUToElectronsPerSecond(ModuleBase):
 @register_module('Rotate180')
 class Rotate180(Filter):
     """
-    Rotate the image 180 degrees
+    Rotate the image 180 degrees in x-y 
 
     Parameters
     ----------
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2868,7 +2868,7 @@ class Rotate180(Filter):
     
 
 @register_module('SplineFitBgSub')
-class SplineFitBgSub(Filter):
+class SplineFitBackgroundSubtract(Filter):
     """
     Background subtraction using spline fit
 
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2867,7 +2867,7 @@ class Rotate180(Filter):
         return data[::-1,::-1]
     
 
-@register_module('SplineFitBgSub')
+@register_module('SplineFitBackgroundSubtract')
 class SplineFitBackgroundSubtract(Filter):
     """
     Background subtraction using spline fit
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2877,6 +2877,7 @@ class SplineFitBgSub(Filter):
     smooth : Positive smoothing factor defined for estimation
     """
     smooth = Float(1e7)
+    dimensionality = Enum('XY', desc='Which image dimensions should the filter be applied to?')
     
     def applyFilter(self, data, chanNum, i, image0):
         from scipy.interpolate import RectBivariateSpline
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2848,3 +2848,36 @@ class RawADUToElectronsPerSecond(ModuleBase):
         series_epers.mdh['Camera.ADOffset'] = 0
 
         return series_epers
+    
+
+@register_module('Rotate180')
+class Rotate180(Filter):
+    """
+    Rotate the image 180 degrees
+
+    Parameters
+    ----------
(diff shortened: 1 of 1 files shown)


PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+@register_module('Rotate180')
+class Rotate180(Filter):
+    """
+    Rotate the image 180 degrees
+
+    Parameters
+    ----------
+    input_name : PYME.IO.ImageStack
+    output_name : PYME.IO.ImageStack
+    """
```
David-Baddeley, line 2862:
```suggestion
    """
    
    dimensionality = Enum('XY', desc='Which image dimensions should the filter be applied to?') # our code only works in 2D, so restrict dimensionality choices
    
```
Because the code in `applyFilter` expects a 2D array, we have to redefine the dimensionality trait to ensure it doesn't get 3D (or 4D) data.

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+    def applyFilter(self, data, chanNum, i, image0):
```
```suggestion
    def apply_filter(self, data, voxelsize):
```

`applyFilter()` is deprecated, please define `apply_filter()` instead

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+@register_module('SplineFitBgSub')
+class SplineFitBgSub(Filter):
+    """
+    Background subtraction using spline fit
+
+    Parameters
+    ----------
+    smooth : Positive smoothing factor defined for estimation
+    """
+    smooth = Float(1e7)
```
David-Baddeley, line 2879:
```suggestion
    smooth = Float(1e7)
    dimensionality = Enum('XY', desc='Which image dimensions should the filter be applied to?')
```

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+    
+    def applyFilter(self, data, chanNum, i, image0):
```
```suggestion
    def apply_filter(self, data, voxelsize):
```

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+    ----------
+    input_name : PYME.IO.ImageStack
+    output_name : PYME.IO.ImageStack
+    """
+    def applyFilter(self, data, chanNum, i, image0):
+        return data[::-1,::-1]
+    
+
+@register_module('SplineFitBgSub')
+class SplineFitBgSub(Filter):
```
```suggestion
class SplineFitBackgroundSubtract(Filter):
```

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+    Parameters
+    ----------
+    input_name : PYME.IO.ImageStack
+    output_name : PYME.IO.ImageStack
+    """
+    def applyFilter(self, data, chanNum, i, image0):
+        return data[::-1,::-1]
+    
+
+@register_module('SplineFitBgSub')
```
```suggestion
@register_module('SplineFitBackgroundSubtract')
```

Module names should be as descriptive as possible to increase the chance that people will hit on them when searching in the recipe GUI.

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
         series_epers.mdh['Camera.ADOffset'] = 0
 
         return series_epers
+    
+
+@register_module('Rotate180')
+class Rotate180(Filter):
+    """
+    Rotate the image 180 degrees
```
```suggestion
    Rotate the image 180 degrees in x-y 
```

Just pedantry - as recipe modules are often used with 3D data it's otherwise slightly ambiguous

PYME/recipes/processing.py
```diff
@@ -2848,3 +2848,36 @@ def run(self, input_name):
...
+    output_name : PYME.IO.ImageStack
+    """
+    def applyFilter(self, data, chanNum, i, image0):
+        return data[::-1,::-1]
+    
+
+@register_module('SplineFitBgSub')
+class SplineFitBgSub(Filter):
+    """
+    Background subtraction using spline fit
```
```suggestion
    Background subtraction using a 2D smoothing spline fit
```
pedantry again ...

David-Baddeley:
Generally looks good - see suggested changes (all relatively minor).

PYME/recipes/processing.py
David-Baddeley: If we wanted to be super safe, we could also add

`assert(data.ndim ==2)` to the beginning of `apply_filter`. This would only really be helpful in terms of giving a useful error message if somehow it did end up being called with 3D data.

PR #1453, opened 2023-12-20
add to the acquisition protocols documentation

barentine:
Addresses issue sparse documentation on what protocols really are/do

enhancement
**Proposed changes:**

**Checklist:**
of the TODO list).


David-Baddeley:
Thanks! Have re-jigged the timing documentation a bit, hopefully to make it clearer (I think it's important to lead in with the fact that they are not strictly synchronous).

barentine:
Looks like a good change - thanks @David-Baddeley !

Issue #1321, opened 2022-12-19
[PROPOSAL] IDS Peak API version of ueye / ucam camera classes


barentine:
**Is your feature request related to a problem? Please describe.**
IDS has deprecated their UI cameras, e.g. IDS-UI-3240ML-NIR-GL, and their new U3 cameras run on IDSPeak. More info [here](https://www.1stvision.com/cameras/IDS-imaging-software).

Note that uEye (UI) cameras appear to also be supported by IDS Peak (see [here](https://www.1stvision.com/cameras/IDS/IDS-manuals/en/ueye-intro.html))

**Describe the solution you'd like to implement**
A camera class using IDS Peak API

**Describe alternatives you've considered**
buying old cameras

csoeller:
We also find purchasing any of the older IDS camera models virtually impossible. I would support the idea moving to their newer cams and building an IDS Peak based interface. Have you yet looked at how different the Peak based interface would have to be?

David-Baddeley:
Sounds good to me ...

barentine:
Just an update that we have an IDS Peak only camera inbound - will give this a try at some point

barentine:
addressed with merge of #1454

PR #1454, opened 2023-12-20
add IDS Peak camera class

barentine:
Addresses issue #1321.

enhancement
**Proposed changes:**
- add a camera class using the ids peak api

**Couple things missing:**
- [x] SetROI
- [x] camera noise_properties stuff (do we want the same as ueye.py here?)
- [x] GetCCDTemp (if we care for these cmos)
- [x] possibly change GetNumImsBuffered to be the number of full camera buffers, rather than PYME-side frame buffer


David-Baddeley:
Sorry I took soo long to get to this ...

PR #1460, opened 2023-12-26
Remove deprecated np.bool calls

zacsimile:
Fixes

```
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipe.py", line 237, in execute
    m.execute(self.namespace)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 259, in execute
    ret = self.run(**inputs)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/localisations.py", line 427, in run
    cached_output = tabular.CachingResultsFilter(output)
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 955, in __init__
    self.Index = np.ones(len(self.resultsSource), dtype=np.bool)
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
DEBUG:PYME.recipes.recipe:removing failed module dependencies
DEBUG:PYME.recipes.recipe:notifying failure
Traceback (most recent call last):
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/recipe.py", line 237, in execute
    m.execute(self.namespace)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/base.py", line 259, in execute
    ret = self.run(**inputs)
  File "/Users/zachcm/Code/python-microscopy/PYME/recipes/localisations.py", line 427, in run
    cached_output = tabular.CachingResultsFilter(output)
  File "/Users/zachcm/Code/python-microscopy/PYME/IO/tabular.py", line 955, in __init__
    self.Index = np.ones(len(self.resultsSource), dtype=np.bool)
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/numpy/__init__.py", line 305, in __getattr__
    raise AttributeError(__former_attrs__[attr])
[...]

Issue #1449, opened 2023-12-15
[PROPOSAL] update np.int / other dtype aliases for compatibility with numpy 1.24+


barentine:
**Is your feature request related to a problem? Please describe.**
I did a pip install, whihc pulled numpy 1.24. Running the tests, I found some errors along the lines of:
```
AttributeError: module 'numpy' has no attribute 'int'.
`np.int` was a deprecated alias for the builtin `int`. To avoid this error in existing code, use `int` by itself. Doing this will not modify any behavior and is safe. When replacing `np.int`, you may wish to use e.g. `np.int64` or `np.int32` to specify the precision. If you wish to review your current use, check the release note link for additional information.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

the deprecation is described at the link [here](https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations), and expired with 1.24 ([see here](https://numpy.org/devdocs/release/1.24.0-notes.html)).

**Describe the solution you'd like to implement**
First:
- PR to change requirements.txt to require numpy <1.24
- PR to change conda recipes to build with <1.24
Second:
- change the numpy aliases to their native types and then unpin numpy again

barentine:
@David-Baddeley  looks like current pyme-depends (windows, py38 anyway) is pinned for numpy>1.22, so depending on what's in the anaconda channel for numpy folks might pull something disfunctional

barentine:
more or less completed

PR #1467, opened 2024-02-10
fix remaining numpy dtype expiries

csoeller:
Fixes issues on numpy >= 1.24 which I now hit with more recent python based installs (notably python 3.10). See also https://numpy.org/devdocs/release/1.24.0-notes.html#expired-deprecations.

See also already merged #1460 which already fixed a couple of `np.bool` deprecations. I hit a few more when running with numpy >= 1.24 and did a quick grep

        grep -r '\(np\.bool\|np\.float\|np\.int\|np\.str\|np\.object\)[^a-z,0-9,_(]' . --include \*.py

which revealed the ones fixed in this PR.

Replacements are generally `np.X` to just `X` with `X` in `int`, `float`, `bool` which I think should be ok. Some of these are in rarely exercised code so not really explictly tested by me but should just work.

### DIFF
--- a/PYME/Analysis/Tracking/trackUtils.py
+++ b/PYME/Analysis/Tracking/trackUtils.py
@@ -95,7 +95,7 @@ class Clump(object):
         self.pipeline = pipeline
         self.clumpID = clumpID
         
-        self.index = np.array(pipeline['clumpIndex'] == clumpID, dtype=np.bool)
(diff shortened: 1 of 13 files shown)


David-Baddeley:
See also #1449

I do wonder if we shouldn't actually go to the fixed width types for some of these instead (e.g. `np.int32`).

David-Baddeley:
Merging as obviously not an issue with non-fixed width types so far, so this obviously doesn't make things worse

PR #1344, opened 2023-04-14
reinstantiate point display overlay if we don't have a filter

barentine:
Addresses issue #1343 .

**Is this a bugfix or an enhancement?**
bugfix
**Proposed changes:**
if we have a `PointsDisplayOverlay` without a filter attribute when we run test frame, just go ahead and make a new `PointsDisplayOverlay` that does.

**Checklist:**
The below is a list of things what will be considered when reviewing PRs. It is not prescriptive, and does not
imply that PRs which touch any of these will be rejected but gives a rough indication of where there is a potential 
for hangups (i.e. factors which could turn a 5 min review into a half hour or longer and shunt it to the bottom
of the TODO list).

- [ ] Does the PR avoid variable renaming in existing code, whitespace changes, and other forms of tidying? [There is a place for code tidying, but it makes reviewing 
much simpler if this is kept separate from functional changes. The auto-formatting performed by some editors is particulaly egregious and can lead to files with thousands
of non-functional changes with a few functional changes scattered amoungst them]

If an enhancement (or non-trivial bugfix):

- [ ] Has this been discussed in advance (feature request, PR proposal, email, or direct conversation)?
- [ ] Does this change how users interact with the software? How will these changes be communicated?
- [ ] Does this maintain backwards compatibility with old data?
- [ ] Does this change the required dependencies?
- [ ] Are there any other side effects of the change?

PR #1468, opened 2024-02-10
fix for AttributeError: 'LMAnalyser2' object has no attribute '_ovl'

csoeller:
We regularly get (inconsequential) errors

      'LMAnalyser2' object has no attribute '_ovl'

when running PYMEAcquire.

This PR fixes that but should be looked at if the `_ovl` is an overhang from something that should be changed upstream from this fix.


David-Baddeley:
This is related to #1344, but we should try and find the root cause and address that. If the errors are inconsequential I wonder if it's worth hanging on until we fix the root cause.

David-Baddeley:
`_ovl` is created by `SetFitInfo`, which in turn is only run when you hit `TestFrame`. We need to make sure this runs with `Go` as well, but might need to defer until we actually have some results to show ...

David-Baddeley:
Many thanks! Despite taking a slightly different tack, this saved me a lot of time.

PR #1469, opened 2024-02-10
sCMOS map integration time fix - change to rounding

csoeller:
We see very annoying rounding issues in the naming of sCMOS map matching. PYME uses the frame integration time for map naming and depending on ROI choice and camera type, the frame integration time can change very subtly, so that a simple int conversion as is in the code now gives sometimes, for example, 99 ms versus 100 ms leading to a map mismatch.

This fix introduces rounding and should be entirely inconsequential for anybody who has not seen this rounding issue before and help those who have seen it.

### DIFF
--- a/PYME/Analysis/gen_sCMOS_maps.py
+++ b/PYME/Analysis/gen_sCMOS_maps.py
@@ -44,7 +44,7 @@ def _meanvards(dataSource, start=0, end=-1):
 
 def map_filename(mdh, type):
     if type != 'flatfield':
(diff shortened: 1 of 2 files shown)


David-Baddeley:
As `int()` is equivalent to `floor()` it just changes where the break is - i.e. we could still have the problem for small variations in integration time around YY.5. Because we are more likely to set integral integration times, this is probably an improvement, however. Will merge, and we can re-visit if needed.

PR #1470, opened 2024-02-10
Python 3.10 compatibilty fixes

csoeller:
This is a PR in progress and testing for further as yet missed issues is ongoing. We have run experimental PYME installs with python 3.10 including PYMEacquire acquisitions which expose a number of incompatabilities. This PR aims at compatibility while remaining backwards compatible with earlier 3.x Python versions.

### abolished float to int conversions

Python 3.10 triggers an issue in that implicit float to int conversions in pre 3.10 are abolished and now raises errors, see also [discussion here](https://discuss.wxpython.org/t/float-conversion-differences-between-python-3-9-and-3-10/36718).

We have run experimental PYME installs with python 3.10 including acquisition and with the changes in this PR things seem fine. This does not exclude the possibility that we have not yet exercised some parts of the code that require further int conversion fixes.

### searching for DLLs changed on windows

We also came across the observation reported here when trying to load camera DLLs: [DLL not found after switching from Python 3.7.6 to Python 3.10.3 [duplicate]](https://stackoverflow.com/questions/71645822/dll-not-found-after-switching-from-python-3-7-6-to-python-3-10-3). Apparently the change is py3.8 onwards but we include it here. The PR contains backwards compatible fixes for the DLL loading most relevant to us but there are other instances in the code base. Perhaps a more general solution with a dedicated `compat` function could be warranted.

### wx slider issues

These are not really py3.10 issues but sizer property issues in `PYME/LMVis/colourPanel.py` conflicting with more recent wx (that is installed when using py3.10). Not sure what's up with the diff which shows the whole file for `PYME/LMVis/colourPanel.py`, there are only two actual changes. Maybe windows vs unix CR/LF...


PYME/DSView/eventLogViewer.py
```diff
@@ -32,6 +32,16 @@
 from matplotlib import cm
 from PYME.Analysis.piecewiseMapping import times_to_frames, frames_to_times
 
+# python 3.10 compat definitions
```
David-Baddeley: I sort of understand why you've done this (to avoid boilerplate changes), but think it would be preferable in terms of long-term readability and maintainability if the `int()` casts were made in the original dc.XXX function calls, rather than defining new helper functions. This is probably fixable with a regex search and replace ...

PYME/LMVis/colourPanel.py
```diff
@@ -1,484 +1,484 @@
...
-
-        btSizer.Realize()
-
-        sizer1.Add(btSizer, 0, wx.ALIGN_RIGHT|wx.ALIGN_CENTER_VERTICAL|wx.ALL, 5)
-
-        self.SetSizer(sizer1)
-        sizer1.Fit(self)
-
-
+#!/usr/bin/python
```
David-Baddeley: I'm not sure what has happened here to imply that the whole file has changed ...

Quite hard to review as is.

David-Baddeley: Ahh - it's a whitespace change (maybe changing line endings???). These are best avoided if at all possible.

David-Baddeley:
Many thanks!

will try and remove the compat definitions in eventLogViewer and then merge ...

csoeller:
Ok, I tried to fix the whitespace changes in colourPanel.py, probably annoying autoconversion on windows when editing.

Also expanded compat definitions into explicit int conversions with the help of perl grep/sed functionality; I'll do some testing if all converted ok as I use a py3.10 based working install on my mac now. Hopefully will expose any remaining glitches...

csoeller:
and eventLogViewer now expanded into explict int conversions in code; some of these will be superfluous, e.g. some int(0) etc, perhaps prefereable to walk through manually now, but I am out of steam...

testing with visgui and events looks good under 3.10 now.

David-Baddeley:
Many thanks!

PR #1472, opened 2024-02-18
Fix AAOptoMDS(AOTF) initiation error

Yujin-Bao:
**This is a bugfix for an AOTF initiation error in PYMEAcquire initiation script**

**Checklist:**
of the TODO list).

- [√] Does the PR avoid variable renaming in existing code, whitespace changes, and other forms of tidying? [There is a place for code tidying, but it makes reviewing much simpler if this is kept separate from functional changes. The auto-formatting performed by some editors is particulaly egregious and can lead to files with thousands of non-functional changes with a few functional changes scattered amoungst them]

### DIFF
--- a/PYME/Acquire/Hardware/AAOptoelectronics/MDS.py
+++ b/PYME/Acquire/Hardware/AAOptoelectronics/MDS.py
@@ -58,7 +58,7 @@ class AAOptoMDS(AOTF):
         # initialize serial
         self.timeout = serial_timeout
         # initialize and configure the serial port without opening it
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I moved the baud rate setting to `__init__()` with the new value as a default (rather than hard-coding) as this is quite likely to change from one system to the next, and should really be settable in the configuration-specific init file rather than having to alter the driver code.

David-Baddeley:
There is still a chance this will break things on the HTSMS system which obviously worked well with the default baud rate - at least doing it this way should make it a simple fix and allow using a different baud rate on different setups.

PR #1473, opened 2024-02-18
Drift tracking bugfix and enhancement to better access the calibration parameters in GUI

Yujin-Bao:
**It contains a bugfix and an enhancement**

**Proposed changes:**
1. The first commit proposes a bugfix of a contribute error of `Correlator._setRefN` caused by a missing underline `_`.
2. The second commit proposes an enhancement to better access two parameters in drift tracking GUI: the calibration stack z increment and stack halfsize. This would be helpful when we do z-stacking localization microscopy acquisition with different step size and live cell imaging.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/Acquire/Hardware/driftTracking.py
+++ b/PYME/Acquire/Hardware/driftTracking.py
@@ -349,7 +349,7 @@ class Correlator(object):
             
         elif (self.calibState == self.NCalibStates):
             # print "cal finishing"
-            self.setRefN(frameData, int(self.calibState - 1))
+            self._setRefN(frameData, int(self.calibState - 1))
             
             #perform final bit of calibration - calcuate gradient between steps
             #self.dz = (self.refC - self.refB).ravel()
--- a/PYME/Acquire/Hardware/driftTracking.py
+++ b/PYME/Acquire/Hardware/driftTracking.py
@@ -119,15 +119,15 @@ class OIDICFrameSource(StandardFrameSource):
             # clobber all frames coming from camera when not in the correct DIC orientation
             pass
 class Correlator(object):
-    def __init__(self, scope, piezo=None, frame_source=None):
+    def __init__(self, scope, piezo=None, frame_source=None, focusTolerance=.05, deltaZ=0.2, stackHalfSize=35):
         self.piezo = piezo
 
         if frame_source is None:
             self.frame_source = StandardFrameSource(scope.frameWrangler)
         
-        self.focusTolerance = .01 #how far focus can drift before we correct
-        self.deltaZ = 0.3 #z increment used for calibration
-        self.stackHalfSize = 20
+        self.focusTolerance = focusTolerance #how far focus can drift before we correct
+        self.deltaZ = deltaZ #z increment used for calibration
+        self.stackHalfSize = stackHalfSize
         self.NCalibStates = 2*self.stackHalfSize + 1
         self.calibState = 0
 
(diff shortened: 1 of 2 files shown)


PYME/Acquire/Hardware/driftTracking.py
```diff
@@ -125,9 +125,9 @@ def __init__(self, scope, piezo=None, frame_source=None):
         if frame_source is None:
             self.frame_source = StandardFrameSource(scope.frameWrangler)
         
-        self.focusTolerance = .05 #how far focus can drift before we correct
```
David-Baddeley, line 128: These defaults were carefully chosen through a lot of trial and error on a couple of systems. I would be reluctant to make hard-coded changes to them, although would entertain/support shifting them to defaults in `__init__()` so that setup-specific alternative defaults could be specified.

Notably:

focusTolerance - setting this too low results in instability on systems with weak illumination for the tracking camera / a noisy signal. When you have a strong focus tracking signal, 0.01 is generally achievable, but it's safer to have a somewhat more permissive default.

deltaZ - the value of 0.2 is based on Nyquist sampling of the PSF in z for a ~1.4NA objective. It is already a bit optimistic (especially if you are going to decrease the focus tolerance), as we assume that signal intensity varies linearly with z between calibration steps. I really wouldn't want to increase it much unless you were using a low NA objective - especially if also decreasing focusTolerance.

stackHalfSize - this is based on being able to do z-stepped imaging of an entire cell nucleus (around 10-12 um) and have enough leeway at each end so that you can still correct. I can see an argument for decreasing this for some use cases to make things faster, but think this is a pretty good default for having something which mostly works out of the box.

PYME/Acquire/Hardware/driftTracking.py
```diff
@@ -202,6 +202,38 @@ def set_focus_tolerance(self, tolerance):
     def get_focus_tolerance(self):
         return self.focusTolerance
 
+
+    def set_delta_Z(self, delta):
```
David-Baddeley, line 206: Very minor note:

This is presumably just copying/ adapting the code for focus_tolerance, but I'm not sure we gain anything from having setter and getter functions over just setting the attribute directly (setters and getters are not really considered idiomatic in python). The best argument would probably be that it allows documentation of the parameters.

PYME/Acquire/Hardware/driftTracking.py
```diff
@@ -202,6 +202,38 @@ def set_focus_tolerance(self, tolerance):
...
 
+
+    def set_delta_Z(self, delta):
+        """ Set the Z increment for calibration stack
+
+        Parameters
+        ----------
+
+        delta : float
+            The delta in nm
```
David-Baddeley:
```suggestion
            The delta in um. This should be the distance over which changes in PSF intensity with depth can be approximated as being linear, with an upper bound of the Nyquist sampling in Z. At Nyquist sampling, the linearity assumption is already getting a bit tenuous. Default = 0.2 um, which is approximately Nyquist sampled at 1.4NA.

```

PYME/Acquire/Hardware/driftTracking.py
```diff
@@ -202,6 +202,38 @@ def set_focus_tolerance(self, tolerance):
...
+
+        self.deltaZ = delta
+
+    def get_delta_Z(self):
+        return self.deltaZ
+
+
+    def set_stack_halfsize(self, halfsize):
+        """ Set the calibration stack half size
+
```
David-Baddeley, line 227:
```suggestion


This dictates the maximum size of z-stack you can record whilst retaining focus lock. The resulting calibration range can be calculated as deltaZ*(2*halfsize), and should extend about 1 micron above and below the size of the the largest z-stack to ensure that lock can be maintained at the edges of the stack. The default of 35 gives about 12 um of axial range.
```

PR #1475, opened 2024-02-22
Make ROI half size in InterpFitR accessible and tunable in GUI

Yujin-Bao:
Enhancement

**Proposed changes:**
The ROI half size can be adjusted in GUI when using InterpFitR fitfactory, similar to AstigGaussFitFR. I think this would be helpful because, for example, the astigmatic PSF is usually larger than non-astigmatic PSF, but the default ROI half size value is 5, which may not be ideal. Now I set the initial/default value to 7, and could be easily further adjusted.

**Checklist:**
of the TODO list).

- [√] Does this change how users interact with the software? How will these changes be communicated?
This adds an interaction between users and ROI half size, but does not change any previously existing interactions.
- [√] Does this maintain backwards compatibility with old data?


David-Baddeley:
Thanks! We should really expose this more broadly (I thought it was, but doesn't seem like it) in other fit factories.

David-Baddeley:
I'm pretty sure the fact that it didn't show up in the GUI was an (old) regression from when we moved to the per-fit module parameters, have now exposed it more broadly.

PR #1479, opened 2024-02-28
Enable sub-roi based drift tracking

David-Baddeley:
Initial steps towards implementing sub-roi awareness in the transmitted light drift tracking. This will enable the use of drift tracking on a small part of the image when the rest of the image contains useful information which we still want to record. This is most applicable when we do not have a dedicated tracking channel, but are rather re-using a transmitted or fluorescence channel. The principle application is to restrict tracking to fiducial(s) during live cell imaging when the cells themselves might be moving.

So far, just the GUI front-end has been implemented.

TODOS:
---------
- backend to record and apply an ROI in the Tracker object
- Display of the selected ROI as an overlay (expanding on the stub provided here)

### DIFF
--- a/PYME/Acquire/Hardware/driftTrackGUI.py
+++ b/PYME/Acquire/Hardware/driftTrackGUI.py
@@ -205,15 +205,30 @@ class ZFactorPlotPanel(PlotPanel):
 
 
 
-# add controls for lastAdjustment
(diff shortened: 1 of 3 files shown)


David-Baddeley:
@Yujin-Bao

PR #1481, opened 2024-03-01
PIPython based Piezo class tweaks

csoeller:
This is an enhancement of the PIPython based generic PI interface that @barentine had kindly created. It was superuseful in getting our new stages to go with PYME (a C-867 based PIline x-y stage and a E-727 fine 3D stage).

The additions in this PR add a few enhancements for simpler and more complete generic use from init files:

  - startup and referencing options added to `GCSPiezoThreaded` constructor
  - addition of a generic joystick class (we use the C-867 with a digital joystick)
  - a small fix to the threadloop to take the sleep command out of the locked code section
  - implemented the firmware method (potentially useful when issues encountered)
 
No equivalent enhancements have been added to the (non-threaded) `GCSPiezo` class. Perhaps its use should be deprecated or altogether removed?

Enhancement, see details above.

**Example usage**

Here is example usage from a working init file which starts and exercises the stages we have just fine with PYME, including switching joystick on and off fine from the GUI etc.

```python
@init_hardware('XY and 3D Fine Stage')
def xy_stage(scope):
    from PYME.Acquire.Hardware.Piezos.piezo_pipython_gcs import GCSPiezoThreaded
    from PYMEcs.Acquire.Hardware.Piezos.joystick_c867_digital import DigitalJoystick
    scope.stage = GCSPiezoThreaded('PI C-867 Piezomotor Controller SN 0122013807', axes=['1', '2'],
                                   refmodes='FRF',joystick=DigitalJoystick())
    scope.stage.units_um = 1000
    # the stage should match the camera reference frame -
    # i.e. the 'x' channel should be the one which results in lateral
    # movement on the camera, and the y channel should result in vertical movement on the camera
    # multipliers should be set (+1 or -1) so that the direction also matches.
    # NOTE: need to check channel and multiplier for our system
    scope.register_piezo(scope.stage, 'x', needCamRestart=False, channel=0, multiplier=-1)
    scope.register_piezo(scope.stage, 'y', needCamRestart=False, channel=1, multiplier=1)
    scope.joystick = scope.stage.joystick
    scope.CleanupFunctions.append(scope.stage.close)

    # note that both stages need to be started in the same thread as the pitools.startup calls seem not thread safe!!!
    scope.fine_stage = GCSPiezoThreaded('PI E-727 Controller SN 0123022828', axes = ['1', '2', '3'],
                                        update_rate=0.01)
    scope.register_piezo(scope.fine_stage, 'x_fine', needCamRestart=False, channel=0, multiplier=1)
    scope.register_piezo(scope.fine_stage, 'y_fine', needCamRestart=False, channel=1, multiplier=1)
    scope.register_piezo(scope.fine_stage, 'z_fine', needCamRestart=False, channel=2, multiplier=1)
    scope.CleanupFunctions.append(scope.fine_stage.close)
```

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
@@ -169,7 +169,7 @@ from PYME.Acquire.eventLog import logEvent
 class GCSPiezoThreaded(PiezoBase):
     units_um = 1  # assumes controllers is configured in units of um.
     def __init__(self, description=None, axes=None, update_rate=0.01, startup=True,
-                 stages=None, refmodes=None, servostates=True, controlmodes=None):
+                 stages=None, refmodes=None, servostates=True, controlmodes=None, joystick=None):
         """
         Parameters
         ----------
@@ -187,7 +187,7 @@ class GCSPiezoThreaded(PiezoBase):
             on-targets
         startup: boolean to use pitools.startup function to startup device
 
-        remaining keyword parameters are 'inherited' from pitools.startup command:
+        next keyword parameters below are passed to the pitools.startup command:
         
         stages: Name of stage(s) to initialize as string or list (not tuple!) or 'None' to skip.
         refmodes: Referencing command(s) as string (for all stages) or list (not tuple!) or 'None' to skip.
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Sorry I took so long to look at this.

PR #1483, opened 2024-03-07
add a recipe module to modify the origin when needed

Yujin-Bao:
Enchancement

**Proposed changes:**
With this module we could modify the xy origin in PYMEImage. This could serve as the initial coarse offset before calculating the shiftmap of two cameras/modalities. I did not try do to it in PYMEAcquire because practically there would be some drift and the shiftmap should be calibrated regularly (once every 2 to 3 weeks). Every time the initial offset is a little bit different.

**Checklist:**
of the TODO list).

- [√] Does this maintain backwards compatibility with old data?

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2906,3 +2906,23 @@ class Offset(Filter):
             return data - self.offset_constant
         elif self.offset_selection == 'offset by minimum':
             return data - np.min(data)
+        
+
+@register_module('ModifyOrigin')
+class ModifyOrigin(Filter):
+    """
+    Modify the origin metadata 
+
+    Parameters
+    ----------
+    ROI_OriginX : the origin of x that the image will set to, int
+    ROI_OriginY : the origin of y that the image will set to, int
+    """
+    ROI_OriginX = Int(578)
+    ROI_OriginY = Int(519)
+    dimensionality = Enum('XY', desc='Which image dimensions should the filter be applied to?')
+
+    def applyFilter(self, data, chanNum, i, im):
+        im.mdh.Camera.ROIOriginX = self.ROI_OriginX
+        im.mdh.Camera.ROIOriginY = self.ROI_OriginY
+        return data
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2908,21 +2908,37 @@ class Offset(Filter):
             return data - np.min(data)
         
 
-@register_module('ModifyOrigin')
-class ModifyOrigin(Filter):
+@register_module('SetOriginMetadata')
+class SetOriginMetadata(ModuleBase):
     """
-    Modify the origin metadata 
+        Set the origin metadata
 
-    Parameters
-    ----------
-    ROI_OriginX : the origin of x that the image will set to, int
-    ROI_OriginY : the origin of y that the image will set to, int
+        Parameters
+        ----------
+        Origin_x : the origin of x that the image will be set to, float in nm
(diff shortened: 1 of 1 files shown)


David-Baddeley:
A number of thoughts / considerations here:

- This feels very special case for the general PYME repo, and almost certainly doesn't belong in `processing`. Maybe this should be in a plugin, or we should write a more generic metadata modification recipe,  or potentially add a metadata module to group different metadata modifications together.

- This is also the wrong way to do it - `applyFilter()` is deprecated and should not be used. More importantly, the code as written modifies the metadata of the input. It is a pretty important convention that the input is not modified, moreover, it is currently only by accident that modifications to the input data propagate to the output - there is no guarantee that the output metadata is not created and before the call to applyFilter. The correct way of doing it would be to implement `completeMetadata(im)` (which takes a copy of the **output** image as a parameter and, by convention executes at the end of module exeution) and make the changes there.

- While using `Filter` as a base, and implementing `completeMetadata()` (along with a dummy `apply_filter()` which simply returns the data un-modified) would work, it would not be ideal from a performance stand-point, as it will force a copy of the data in memory. A more efficient implementation would be something along the lines of the following:

```python
class SetOriginMetadata(ModuleBase):
    inputImage = Input('input')
    outputImage = Output('output')

    def run(self, inputImage):
        # create a new image, using the data from the old image
        # this re-uses the reference to the unmodified image data so we don't need to allocate more memory
        # but creates a new, empty, metadata
        im = ImageStack(inputImage.data, titleStub=self.outputImage)
        
        # copy the metadata entries from the original image
        im.mdh.copyEntriesFrom(inputImage.mdh)
        
        # add any module-specific metadata to the output image
        self.completeMetadata(im)

        # return the new image
        return im

    def completeMetadata(self, im):
        '''Do any required modifications to the output metadata"
        # Set origin etc ...
```

David-Baddeley:
Actually, on second thoughts, don't implement `completeMetadata` - just make the changes in `run()` method on the new image metadata.  ie ...

```python
class SetOriginMetadata(ModuleBase):
    inputImage = Input('input')
    outputImage = Output('output')

    def run(self, inputImage):
        # create a new image, using the data from the old image
        # this re-uses the reference to the unmodified image data so we don't need to allocate more memory
        # but creates a new, empty, metadata
        im = ImageStack(inputImage.data, titleStub=self.outputImage)
        
        # copy the metadata entries from the original image (note, this is not technically required, as the .execute()
        # method will automatically copy (merge) the input metadata to the output, but it doesn't hurt to be explicit)
        im.mdh.copyEntriesFrom(inputImage.mdh)
        
        # add any module-specific metadata to the output image
        im.mdh['Origin.x'] = #New origin value
        ....

        # return the new image.
        return im
```

The rationale here is the `completeMetadata()` is a bit of a relic whose function has mostly been moved to `ModuleBase.execute()`  and should be retired.

David-Baddeley:
Also note that `"Origin.x"` etc are in units of nm, not pixels, and would need to be `Float()`s rather than `Int()`

Yujin-Bao:
Thanks, David! I am also not sure either if this is a long-term solution, but making the origin changable does make it easier to overlay two imaging modalities on different cameras, although the origin is actually an 'effective origin' based on users and instrument instead of 'real metadata origin'. Previously I used `im.mdh.Camera.ROIOriginX` because `Origin.x` metadata in OI-DIC images are missing (have not looked into why). But I made the suggested changes and adding `Origin.x` metadata also did work.

PR #1489, opened 2024-03-21
More xyztc stuff

Yujin-Bao:
This is a follow up of an enhancement

**Proposed changes:**
A small bugfix for xyztc sequence. I also realize when I capture smlm time/blinking series, it saves to T when using cluster but saves to Z when not using cluster. I am not sure what causes this but I changed it to saving to T all the time. However, all z steps will also be saved to T but not Z. This may need some further adjustment.


David-Baddeley:
The bugfixes look correct - not sure how the errors crept in there ...

PR #1492, opened 2024-03-21
Spooling refactoring

David-Baddeley:
This is a long overdue change that lays the foundation for putting synchronous acquisitions (e.g. z stacks and related acquisition types) on an equal footing with asynchronous acquisitions (i.e. protocol driven SMLM acquisitions) when it comes to automated acquisition and spooling to a cluster. 

Currently a work in progress, it entails some not insignificant changes to the user interface and underlying code.
Comments welcome @barentine @csoeller. 

image

## Main proposed changes:

- Separation of the settings that are specific to protocol acquisitions from the controls for starting and stopping acquisition.
- End goal is to show settings for one acquisition type at a time (i.e. only stack, or only SMLM, or only tiling), clicking start then starts that acquisition type using the same spooling control. This means that to start a z stack, you would expand the z-stack tab, and then hit start in the spooling tab. The start button in the z-stack panel would be removed.
- This should make things more consistent and let all acquisition types be queued and spooled to the cluster.

## Open questions

- Should we change the selected spooling type when the user changes acquisition types? (currently z stacks are 'Memory', whereas SMLM acquisitions are 'Cluster'). I could see this being annoying both if we change and if we don't.


csoeller:
@David-Baddeley, what is the best way to test the user experience at this early stage, e.g. run it in the simulator?

David-Baddeley:
The best way is likely to be the simulator, but it might still be slightly
premature - there is still no GUI way of selecting which acquisition mode
to use.

David-Baddeley:
@csoeller - it should now be in a reasonable shape for a first try in the simulator. Still a bit rough around the edges, but it would seem to be mostly working (I've somehow broken the time-spoofing for protocol events in the simulator, but that should be reasonably easy to fix.

image

Mode of operation is now to choose an acquisition type, and then always use the same buttons to start the acquisition.

image

# Remaining major TODOs ...

- Fix up QueuedActions interface for non SMLM acquisiitons
- Bring tiling etc under the SpoolController
- some UI tidying (e.g. disable memory acquisition radio button for SMLM acquisitions, potentially change appearance of mode toggle buttons)
- Documentation

Yujin-Bao:
Thanks very much for spending a lot of effort on this! I tested this PR branch together with the OIDIC side. I think it would be better to fix some minor issues:
 - When capturing and spooling a z-stack for the first time after changing an acquisition type, the piezo does not go back to the origin after finishing.
 - For `Z-Stack` and `OIDIC ` acquisition types, there is an error that no attribute `spool_complete` for `oidic_acquisition` and `protocol_acquisition`.

While I believe it may not be so hard to fix for me, I do not know if what you are currently working on includes those issues. Please let me know if I should go ahead and try to fix them and update the PR. Thanks!

David-Baddeley:
Thanks for the testing! I'm happy to have a look at the 2 issues  ... I
think the StackSettings object needs a bit of attention anyway ....

David

Yujin-Bao:
Thanks David! With those changes made, the OI-DIC side does not have the dark frames during capturing any more. And z stacks can be acquired with the piezo returning to the origin after finishing. However, I am wondering which button I need to click to start the queue actions, since now all actions will just queue there without starting.

David-Baddeley:
@Yujin-Bao - The action queue should now be fixed.

David-Baddeley:
@csoeller, @barentine - Just a heads up that I'm going to merge this soon - it's getting unwieldy and there is a bunch of stuff which I want to build on top of this. I think the worst of the issues have fallen out by now, but there is obviously still some scope for minor regressions / niggles.

PR #1495, opened 2024-04-09
Queued actions improvements

David-Baddeley:
Adds the ability to set a delay for queued acquisitions to allow timelapse type acquisitions. Also contains:

- a lot of tidying to ensure that multiple different acquisition types can be queued with different settings (i.e. remembering the settings at time of queueing). This enables queueing of both ZStacks and blinking series (and any other acquisition types we might want to add) 
- Some long-overdue UI improvements.

image

### Remaining TODOs
- ui functions for queueing repetitive acquisitions (rather than queueuing them each separately with a different delay)
- fix queue from txt and TileViewer so that they also keep settings.
- Find a more robust way of setting timeout for queue from file and queue from tileviewer (we can no longer do anything sensible based on frames per second). 

@Yujin-Bao

### DIFF
--- a/PYME/Acquire/ActionManager.py
+++ b/PYME/Acquire/ActionManager.py
@@ -52,7 +52,8 @@ class ActionManager(object):
             calling eval('scope.functionName')
         
         """
-        self.actionQueue = Queue.PriorityQueue()
+        self.actionQueue = Queue.PriorityQueue() # queue for immediate execution
+        self.scheduledQueue = Queue.PriorityQueue() # queue for scheduled execution
         self.scope = weakref.ref(scope)
         
         #this will be assigned to a callback to indicate if the last task has completed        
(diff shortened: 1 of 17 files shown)


David-Baddeley:
Hopefully final (and more flexible) QueueActions UI

image

image

- Unifies the single-ROI and multi-ROI interface so that there are not 2 similar but different interfaces to the same thing.
- Makes the default action combination a move then an acquisition (likely the most common)
- Still lets you do more complex things like CentreOn -> SetState (for lasers etc) ->SpoolSeries
- Adds *repetitions* for repeatedly imaging the same cell (or set of cells) at a time interval given by *period*. *repetitions*=0 gives old behaviour (period is ignored).

David-Baddeley:
One remaining TODO is to make the spool_series text update to tell you what type of acquisition (z-stack, protocol, etc ...) you are adding.

PR #1497, opened 2024-04-18
Small bugfix and follow-up on PR#1495

Yujin-Bao:
Here is a bugfix I mentioned earlier - piezo does not return to the origin when finishing the acquisition using queue actions. I am not sure I use the best way, since the piezo does return to the origin when clicking the 'Stop' button in the spool panel. I must have missed something...

### DIFF
--- a/PYME/Acquire/ActionManager.py
+++ b/PYME/Acquire/ActionManager.py
@@ -52,7 +52,8 @@ class ActionManager(object):
             calling eval('scope.functionName')
         
         """
-        self.actionQueue = Queue.PriorityQueue()
+        self.actionQueue = Queue.PriorityQueue() # queue for immediate execution
+        self.scheduledQueue = Queue.PriorityQueue() # queue for scheduled execution
         self.scope = weakref.ref(scope)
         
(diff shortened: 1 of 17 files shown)


David-Baddeley:
I'd thought I'd already addressed this - but maybe not. It's probably better to handle this within the individual acquisition types and/or by a general push and pop of the state.

This will work as long as you don't do anything too funky (e.g. stepping along a non-default axis), but there are edge cases where this might not have the desired effect.

Yujin-Bao:
I thought so, too. Unfortunately when I used queued actions for time/blinking series with z-stacking, the piezo did not return (I found this when I tried interleaved fluorescence and OI-DIC imaging, but the issue also existed when only queuing fluorescence acquisition). I have not found out a better way of doing this, but will give an update if I do. Thanks!

David-Baddeley:
Another issue with having `piezoGoHome()` in `SpoolingStopped` is that it will execute on abort. It's generally desirable (when considering the health of objectives etc ...) that there is a way of aborting without any motion.

PR #1498, opened 2024-04-18
Changes to text import behavior.

zacsimile:
These are just usability changes I propose to make it easier to open unknown file types that overlap with some of our existing flavors. It is in response to a weird CSV I have to open.

- Make the import dialog display even if ambiguous flavours are found, rather than raise an error. This allows the user to still map the columns.
- Make the GUI comboboxes update on flavour changes so we can see which columns are mapped to what in the GUI.

### DIFF
--- a/PYME/IO/csv_flavours.py
+++ b/PYME/IO/csv_flavours.py
@@ -203,7 +203,7 @@ def guess_flavour(colNames, delim=None, ext=None):
     for flavour in csv_flavours:
         if (not flavour == 'default') and all(idn in colNames for idn in csv_flavours[flavour]['idnames']):
             if fl is not None:
-                warnings.warn('Ambiguous flavour database: file matches both %s and %s' % (fl, flavour))
+                warnings.warn('Ambiguous flavour database: file matches both %s and %s. Using default flavour.' % (fl, flavour))
                 fl = 'default'
                 break
             fl = flavour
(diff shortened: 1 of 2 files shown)


PYME/IO/csv_flavours.py
```diff
@@ -202,7 +202,9 @@ def guess_flavour(colNames, delim=None, ext=None):
     for flavour in csv_flavours:
         if (not flavour == 'default') and all(idn in colNames for idn in csv_flavours[flavour]['idnames']):
             if fl is not None:
-                raise RuntimeError('Ambiguous flavour database: file matches both %s and %s' % (fl, flavour))
+                logger.info('Ambiguous flavour database: file matches both %s and %s' % (fl, flavour))
```
David-Baddeley: I'm pretty sure that this runs within the error handling context manager, so the current `RuntimeError` results in the error message being visible to the user in a dialog. Demoting this to a `logger.info` seems unwise as there will now be no GUI alert. I think that in most cases this is an error, and we should inform the user. Maybe use `PYME.warnings`? 

If we match more than one flavour, I'm not sure the most obvious thing is to ignore both the possible flavours and just go with the default - if we are doing this we should also explicitly tell the user that flavour detection failed and we are falling back on the default flavour.

zacsimile: I agree it would be better to have a GUI response. What I find frustrating about the current approach is if flavor detection fails, PYMEVis raises this error and then refuses to open the file. Is there an easy way to throw this error in the context manager but then still display the `ImportTextDialog`?

zacsimile: Yeah `warnings` is a good idea. Thanks!

zacsimile: We could instead match the flavour with the larger number of matching columns. That may be a better default choice.

PR #1501, opened 2024-04-30
Modification to #1497

David-Baddeley:
contains logging fix from #1497, but a modified piezo-return fix as this should be acquisition-type specific. 

We might need to further modify this to ensure that 

a) return happens on the correct channel
b) we only return if we are actually controlling the piezo (i.e. z-stacking) so that we don't inadvertently mess with a simultaeneous acquisition that is controlling the piezo.


David-Baddeley:
a) and b) should now be addressed.

This closes #1497

PR #1502, opened 2024-05-01
PCO index out of bounds fix

David-Baddeley:
putative fix for index out of bounds error encountered at long integration times. This eliminates a potential race condition.

### DIFF
--- a/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
@@ -117,6 +117,8 @@ class PcoSdkCam(Camera):
         self._cam_type = pco_sdk.get_camera_type(self._handle)
         self.SetHotPixelCorrectionMode(pco_sdk.PCO_HOTPIXELCORRECTION_OFF)
 
+        self._buffer_lock = threading.Lock()
+        
         self._polling = True
         self._poll_thread = threading.Thread(target=self._poll_loop)
         self._poll_thread.start()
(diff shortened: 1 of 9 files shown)


David-Baddeley:
@Yujin-Bao - give this a try ... no promises, but might do the trick.

Yujin-Bao:
Hi David, thanks for the fix! Since I have an error running the updated PYME after PR#1500, I tested the code using the old version. The problem is still there with the same error: index out of bounds, and this probably comes from the same issue - the current buffer number exceeds the buffer size when we change to a higher exposure time. Maybe the current buffer number should return to 0 upon changing the exposure time and initiating buffers? This error could be got around temporarily by set a fixed buffer size, for example, 200. Then I found another error with exposure time more than 90ms: the wait time `self._n_timeouts` is increasing and exceeds the `MAX_TIMEOUTS` in the end, giving me another exception as follows.
image

David-Baddeley:
The current buffer number should already return to zero ... (done in the `StopAq()`) method, and the added lock should ensure that this happens at the right time. Will look into the timeout.

David-Baddeley:
timeout hopefully fixed ...

Yujin-Bao:
Thanks for looking into it! However the problem still exist. I am sorry for not being clear earlier. I took a more careful look by printing out some parameters. I checked `StopAq()` and agreed it should already set the current buffer number to zero. However, the problem is still there in line 146 of `pco_sdk_cam.py.` that index `i` of `self._buf_event[i]` is out of bounds. This is because the `i = self._buffers_to_queue.get()` does not return to zero upon changing the integration time. As for the timeout, I realized `self._n_timeouts` is increasing (line 158) and exceeding the `MAX_TIMEOUTS`.

David-Baddeley:
Changing the integration time does not stop and start the camera (which is potentially problematic if you go from a long integration time to a really short one, but shouldn't be too bad), which means that the buffers are not reset and we don't expect the index to go to zero. The timeout patch should make the length of the timeout change when changing integration time - you could still run into `MAX_TIMEOUTS`, but this would imply that something else is happening, not just the length of the timeout being wrong (the "100ms" bit should be longer).

David-Baddeley:
@Yujin-Bao - I'm going to merge this as I've added a couple of important bugfixes (most notably the fact that we are sometimes missing events on cluster spool) on top of the camera related ones.

PR #1500, opened 2024-04-20
Dark mode traits

David-Baddeley:
Very partial solution for issue #1331 .

Introduces a way of monkey-patching traitsui in order to define suitable text editor background colours so that text is visible in dark mode.

PR #1503, opened 2024-05-01
fix code object namespace error in traits patch, note minimum wxpython verison in pip requirements.txt

barentine:
Addresses issue see conversation in #1500 

bugfix
**Proposed changes:**
- import aui_utilities in StepBaseColour. Why is this not needed in Py3.8 on mac? I don't know, but the code object doesn't generally get packaged with context from the module (e.g. the previous import?) and does not have aui_utilities within the namespace when called by wxpython on python 3.7 on windows.
- note the minimum wxpython verison is now 4.1. This will also need to be updated in the PYME conda-recipes repository. 

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/ui/patch_traitsui.py
+++ b/PYME/ui/patch_traitsui.py
@@ -72,6 +72,9 @@ def StepColour(c, ialpha):
 aui_utilities.StepColour.__code__ = StepColour.__code__
 
 def GetBaseColour():
+    # this import may not be needed on all wxpython / python versions (but is for 4.1 / 3.7)
+    from wx.lib.agw.aui import aui_utilities
+    
     base_colour = wx.SystemSettings.GetColour(wx.SYS_COLOUR_3DFACE)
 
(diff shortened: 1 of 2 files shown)


Yujin-Bao:
I just checked and confirmed it works on my system! Thanks @barentine !

PYME/ui/patch_traitsui.py
```diff
@@ -79,7 +79,8 @@ def GetBaseColour():
     if ((255-base_colour.Red()) +
         (255-base_colour.Green()) +
         (255-base_colour.Blue()) < 60):
-
+        # this import may not be needed on all wxpython / python versions (but is for 4.1 / 3.7)
```
David-Baddeley: can we move this up to line 75?

barentine: of course! Just made the change

David-Baddeley:
Many thanks!

PR #1506, opened 2024-05-07
More py310 compatability fixes

csoeller:
Continuation of #1470 related fixes.

Some further compatibility fixes, mostly explicit float to int conversions for a few more GUI components. I discover these as I go along and thus this may not be the last of these...


David-Baddeley:
Thanks!

PR #1508, opened 2024-05-21
Fix dialog windows in dark mode

zacsimile:
Follow up to #1500. Fix an issue I ran into when adding a `MappingFilter` to the pipeline:

image

image

### DIFF
--- a/PYME/ui/traitsui_constants.py
+++ b/PYME/ui/traitsui_constants.py
@@ -50,7 +50,7 @@ EditableColor = OKColor
 
 #: Color for background of windows (like dialog background color)
 if is_mac:
-    WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_INFOBK)
+    WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_3DFACE)
     BorderedGroupColor = wx.Colour(224, 224, 224)
 else:
     WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_MENUBAR)
--- a/PYME/ui/traitsui_constants.py
+++ b/PYME/ui/traitsui_constants.py
@@ -40,10 +40,7 @@ else:
 ReadonlyColor = wx.Colour(244, 243, 238)
 
 #: Color for background of fields where objects can be dropped
-if is_dark:
-    DropColor = wx.Colour(94, 131, 167)
-else:
-    DropColor = wx.Colour(215, 242, 255)
+DropColor = wx.Colour(215, 242, 255)
 
 #: Color for an editable field
 EditableColor = OKColor
--- a/PYME/ui/traitsui_constants.py
+++ b/PYME/ui/traitsui_constants.py
@@ -50,7 +50,7 @@ EditableColor = OKColor
 
 #: Color for background of windows (like dialog background color)
 if is_mac:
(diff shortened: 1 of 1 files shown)


PYME/ui/traitsui_constants.py
```diff
@@ -43,8 +47,7 @@
 
 #: Color for background of windows (like dialog background color)
 if is_mac:
-    WindowColor = wx.Colour(232, 232, 232)
-    BorderedGroupColor = wx.Colour(224, 224, 224)
+    WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_WINDOW)
```
David-Baddeley, line 50: We might still need bordered group colour.

PYME/ui/traitsui_constants.py
```diff
@@ -20,6 +20,7 @@
 
 #: Define platform and wx version constants:
 is_mac = sys.platform == "darwin"
+is_dark = wx.SystemSettings.GetAppearance().IsDark()
```
David-Baddeley: We need to wrap this in a `try ... except AttributeError` block to maintain backwards compatibility with older versions of wx (necessary for now for package resolvability / installability)

PYME/ui/traitsui_constants.py
```diff
@@ -43,7 +47,7 @@
 
 #: Color for background of windows (like dialog background color)
 if is_mac:
-    WindowColor = wx.Colour(232, 232, 232)
+    WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_WINDOW)
```
David-Baddeley: This looks too dark - maybe `SYS_COLOUR_MENUBAR` as on windows ??

zacsimile: MENUBAR is a bit bright in my opinion:

image

zacsimile: `wx.SYS_COLOUR_INFOBK` looks good to me:

image

What do you think?

David-Baddeley: Looks OK, but might be slightly safer as `wx.SYS_COLOUR_3DFACE` (which is what wx uses for panels - so the rest of the dialog). It appears to be numerically the same colour on mac, but who knows if it will be portable ... 

I'll change it.

PYME/ui/traitsui_constants.py
```diff
@@ -43,7 +50,7 @@
 
 #: Color for background of windows (like dialog background color)
 if is_mac:
-    WindowColor = wx.Colour(232, 232, 232)
+    WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_INFOBK)
```
David-Baddeley:
```suggestion
    WindowColor = wx.SystemSettings.GetColour(wx.SYS_COLOUR_3DFACE)
```

PYME/ui/traitsui_constants.py
zacsimile: Sounds good. How do you know that `wx` uses `SYS_COLOUR_3DFACE` for panels? I've been looking at https://docs.wxpython.org/wx.SystemColour.enumeration.html to figure out these colorings. All I see there is that `SYS_COLOUR_3DFACE` is a synonym for `SYS_COLOUR_BTNFACE`, which is the face shading color on push buttons. Is there better documentation somewhere else?

PR #1509, opened 2024-05-28
Fix for performance issue with remote spooling

David-Baddeley:
Fixes issue where a performance degradation was noticed after triggering a remote acquisition due to the polling cost of checking whether it was complete. @Yujin-Bao - let me know if this helps

Also has a few minor tweaks related to .hdf flushing (the fairly paranoid flushing of events was causing looking issues in the hdf library and decresasing performance)

### TODOs.

- Change to using long polling for remote acquisitions to lower the overhead further (as it could be an issue for simultaeneous acquisitions).

### DIFF
--- a/PYME/Acquire/ActionManager.py
+++ b/PYME/Acquire/ActionManager.py
@@ -238,6 +238,7 @@ class ActionManager(object):
             try:
                 self.currentTask.finalise(self.scope())
                 self.currentTask = None
+                self.isLastTaskDone = None # prevent us from continuing to poll the last task if it's done
             except AttributeError:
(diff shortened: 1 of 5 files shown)


Yujin-Bao:
Thanks for fixing the issue! Unfortunately the issue remains the same and I am still not sure what causes that. By the way, I think it probably would be better to include the change made in PR#1503 to fix the `wxPython` issue: adding a line `from wx.lib.agw.aui import aui_utilities` to line 75 in `patch_traitsui.py` (or probably it would not matter since the change has been merged?).

PR #1511, opened 2024-05-29
Add a protocol for fluorescence capturing

Yujin-Bao:
This contains a bugfix and an enhancement. The bugfix is for an attribute error due to a naming issue when adding a protocol. The enhancement is a simple protocol that automatically turns the laser on/off before/after the fluorescence acquisition.


csoeller:
Hi @Yujin-Bao , @David-Baddeley can probably better judge if this protocol is worth adding to the general distribution. I just wanted to check if you are aware of the `PYME.config` support for adding local protocols, see `pydoc PYME.config`. We maintain our custom protocols in a github repo for site configuration and install that on all acquisition computers. This takes care of the fact that it is very difficult to build a large set of generally useful protocols, often one needs local tweaking for specific requirements.

David-Baddeley:
Sorry about not mentioning it yesterday - as @csoeller has alluded, protocols generally belong on the local computer (potentially in a separate repository, but could also be an untracked local file).

The default place to store machine specific protocols is `<your home directory>\.PYME\protocols` (if keeping protocols in a local repository so as to have version control, you can symlink the protocols into here).

The caveat is that protocols may be added to the main repository if they are likely to be a useful example for people setting up a new system. There are a few protocols in the main repository which are in the repo mainly as historical artifacts, and which should probably be removed (although I won't be aggressively removing as there is something to be said for having a wide range of protocols for people to draw from). As there are plenty of examples of turning lasers on and off in the protocols we already have, so I don't think the new protocol warrants inclusion.

Yujin-Bao:
Thanks for the reminder @csoeller @David-Baddeley ! I should have asked first before making this pull request - I had a feeling that this protocol might not worth an inclusion but I still made the change together with the bugfix. I will make some adjustment according your suggestions and have another look at the bugfix on my scope computer. I will update or close this PR and let you know!

Yujin-Bao:
I have removed the protocol out of the repo, but it still needs the bugfix to run. Shall we still merge this PR as a small bugfix? Thanks!

PR #1512, opened 2024-05-31
protocol termination - adjust OnFinish logic so that any non-termination tasks are skipped

csoeller:
Addresses issue that stopping a protocol run before all commands have been processed just runs through all remaining tasks.

A bit of both, opinions on this may differ.

**Proposed changes:**

The logic is changed so that only proper "termination tasks" are executed when stopping a run. All other tasks are discarded and this is mentioned in a warning that is logged.

For most protocols only those asks with a `when` property equal to `maxint` are typically the remaining tasks when a run is stopped (most tasks happen early in the run). As such, implicitly, tasks with a `when` property equal to `maxint` can be regarded as termination tasks. This patch makes this treatment more literally and also proposes an `on_finish` symbol to mark such tasks explicitly, currently just a synonym for `maxint` but the implementation could change in future if needed, thus putting more emphasis on semantics as compared to `maxint`.

Notably, with the explicit treatment of termination tasks, it is generally straightforward to include a few such tasks in every protocol to ensure the system is left in a defined state (lasers off, piezo returned, shutter closed etc) when stopped before all tasks have been excecuted.

**Checklist:**
Checked on a running acquisition system with a PALM type protocol that involves many laser switches...

### DIFF
--- a/PYME/Acquire/protocol.py
+++ b/PYME/Acquire/protocol.py
@@ -29,6 +29,10 @@ logger = logging.getLogger(__name__)
 from PYME import config
 
 from sys import maxsize as maxint
+# more semantically explicit synonym for maxint
+# implying that tasks with a `when` property of maxint are excecuted upon finishing the protocol
+# e.g. after the stop button is pressed or an explicit stop task was executed
+on_finish = maxint
 
 #minimal protocol which does nothing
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Thanks!

PR #1514, opened 2024-06-04
make scipy.stats.mode call compatible with scipy>=1.10

barentine:
Addresses issue [scipy.stats.mode output keepdims behavior change](https://github.com/scipy/scipy/issues/16418)

enhancement
**Proposed changes:**
make scipy.stats.mode calls (I only found one) compatible with scipy 1.10+

previously:
```
..\..\..\PYME\recipes\base.py:259: in execute
    ret = self.run(**inputs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <PYME.recipes.processing.StatisticsByFrame object at 0x000001A18913ADE0>, input_name = <PYME.IO.image.ImageStack object at 0x000001A188A4F3D0>
mask = <PYME.IO.DataSources.ArrayDataSource.ArrayDataSource object at 0x000001A188A4E800>

    def run(self, input_name, mask=None):
        from scipy import stats

        data = input_name.data
        if mask:
            # again, handle our mask being either 2D, 3D, or 4D. NB - no color (4D) handling implemented at this point
            mask = mask.data

        var = np.empty(data.shape[2], dtype=float)
        mean = np.empty_like(var)
        median = np.empty_like(var)
        mode = np.empty_like(var)

        for si in range(data.shape[2]):
            slice_data = data[:,:,si, 0]

            if mask is not None:
                if mask.shape[2] == 1:
                    slice_data = slice_data[mask[:,:,0,0].astype('bool')]
                elif mask.shape[2] == data.shape[2]:
                    slice_data = slice_data[mask[:,:,si,0].astype('bool')]
                else:
                    raise RuntimeError('Mask dimensions do not match data dimensions')

            var[si] = np.var(slice_data)
            mean[si] = np.mean(slice_data)
            median[si] = np.median(slice_data)
>           mode[si] = stats.mode(slice_data, axis=None)[0][0]
E           IndexError: invalid index to scalar variable.

..\..\..\PYME\recipes\processing.py:2677: IndexError
```

which passes with the PR'd change. 

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2674,7 +2674,7 @@ class StatisticsByFrame(ModuleBase):
             var[si] = np.var(slice_data)
             mean[si] = np.mean(slice_data)
             median[si] = np.median(slice_data)
-            mode[si] = stats.mode(slice_data, axis=None).mode.item()
+            mode[si] = stats.mode(slice_data, axis=None)[0].squeeze()
 
         # package up and ship-out results
         return tabular.DictSource({'variance': var, 'mean': mean, 'median': median, 'mode': mode})
--- a/PYME/recipes/processing.py
+++ b/PYME/recipes/processing.py
@@ -2674,7 +2674,7 @@ class StatisticsByFrame(ModuleBase):
             var[si] = np.var(slice_data)
             mean[si] = np.mean(slice_data)
             median[si] = np.median(slice_data)
-            mode[si] = stats.mode(slice_data, axis=None)[0][0]
+            mode[si] = stats.mode(slice_data, axis=None).mode.item()
 
         # package up and ship-out results
         return tabular.DictSource({'variance': var, 'mean': mean, 'median': median, 'mode': mode})


David-Baddeley:
Thanks!

1) Do we break backwards compatibility here (the proposed change seem to be fine on 1.9, but I'm not sure about earlier versions - or how far back we need to go)?

2) Should we also address the deprecation warning at the same time

```python
In [8]: stats.mode(x).mode
<ipython-input-8-a1e97f2209b3>:1: FutureWarning: Unlike other reduction functions (e.g. `skew`, `kurtosis`), the default behavior of `mode` typically preserves the axis it acts along. In SciPy 1.11.0, this behavior will change: the default value of `keepdims` will become False, the `axis` over which the statistic is taken will be eliminated, and the value None will no longer be accepted. Set `keepdims` to True or False to avoid this warning.
```

David-Baddeley:
Looking at things a bit more:

- the documentation indicates that `scipy.stats.mode` returns two separate values rather than the `ModeResult`  `namedtuple` that is actually returned, making the `.mode` attribute undocumented (although it is in some examples). 
- The change between returning two separate values and a `namedtuple` occurred in scipy 1.5 (so we probably don't really need to worry about backwards compatibility). That said, we could make it `stats.mode(...)[0].squeeze()` or `stats.mode(...)[0].item()` and match the documented function signature and not have to worry about it.

David-Baddeley:
Advantage of `.squeeze()` over `.item()` is that it avoids a copy and cast.

barentine:
applied your suggestion! Thanks, @David-Baddeley

Issue #1504, opened 2024-05-02
[ENHANCEMENT] Quiver rendering without vertices (e.g. for dipole orientation visualization)


barentine:
**Is your feature request related to a problem? Please describe.**
All of the heavy lifting has already been done (thanks @zacsimile and @David-Baddeley !), but the `QuiverRenderLayer` is typed for `triangle_mesh.TrianglesBase` only.
We are interested in plotting single-molecule orientation microscopy (SMOLM) data where we do not have a surface, but do have XYZ and 3D orientation vectors (mu_x, mu_y, mu_z) for each localization. 
We're happy to re-tool what's there a bit, but wanted to check in first to see a) whether I missed something and there is already a convenient way to visualize this in PYMEVis without modifying anything, or b) whether our approach would be the cleanest.

**Describe the solution you'd like**
- Subclass `QuiverRenderLayer` and rejig `update_from_datasource` so it accepts a normal localization datasource rather than TriangsBase.
- render each localization twice, with opposite orientations such that the point stays in the middle of the bar that is drawn

**Describe alternatives you've considered**
you having already written this somewhere ;)

**Additional context**
[active smolm work in our lab](https://doi.org/10.1016/j.optcom.2023.129589)
[style of visualizations we are aiming for (but interactively in 3D, of course!)](https://doi.org/10.1364/OPTICA.388157)

**Are you prepared to work on it?**
yes

David-Baddeley:
Sorry I've taken so long to get to this ... it came in at an awkward time and then got forgotten. 

- I don't think there is a solution already implemented.
- Subclassing `QuiverRenderLayer` seems reasonable, although we might want to rejig ever so slightly at some point to put the shared logic in a common base class instead.
- Instead of rendering each location twice, consider making another `Engine` which which subtracts from one end of the vector and adds to the other end instead. I think this would be a bit cleaner, and would be faster / less memory intensive. 

ie  duplicate the engine and replace

```python
    vec_buffer = np.empty((vertices.shape[0]+vecs.shape[0],3), dtype=vertices.dtype)
    vec_buffer[0::2,:] = vertices
    vec_buffer[1::2,:] = vertices
            
    vec_buffer[1::2,:] += layer.scaling*vecs
```  

with 

```python
    vec_buffer = np.empty((vertices.shape[0]+vecs.shape[0],3), dtype=vertices.dtype)
    vec_buffer[0::2,:] = vertices - layer.scaling*vecs
[...]

PR #1520, opened 2024-06-07
point quiver enginer/layer for single molecule orientation microscopy data

barentine:
Addresses issue #1504 .

enhancement
**Proposed changes:**

image

MWE, from a jupyter notebook:
```
from PYME.LMVis import VisGUI
%gui wx

pymevis = VisGUI.ipython_pymevisualize()
pipeline = pymevis.pipeline


import numpy as np
from PYME.IO import tabular, MetaDataHandler

n_points = 100
ds = tabular.DictSource({'x': np.linspace(0, 100, n_points),
                         'y': np.linspace(0, 100, n_points),
                         'z': np.zeros(n_points),
                         'azimuth': np.linspace(0, 2* np.pi, n_points),
          
                         'zenith': np.linspace(0, np.pi, n_points)})

ds.mdh = MetaDataHandler.DictMDHandler()


pipeline.addDataSource('points', ds)
pipeline.selectDataSource('points')

from PYME.recipes.localisations import LocalizationOrientationAngles
mod = LocalizationOrientationAngles(pipeline.recipe, input_name='points', output_name='oriented_points')

pipeline.recipe.add_modules_and_execute([mod])

pipeline.selectDataSource('oriented_points')

pymevis.add_point_quiver_layer('oriented_points')

```

**remaining:**
- [ ] refactor how orientation angles / unit vectors get identified. Hacking them into a potentially large array as an attribute is certainly not the way to do it.
- [ ] make this accessible from the 'add layer' menu (might be able to tie in the orientation/unit vector parameter identification to this part too)

@David-Baddeley , any thoughts on how best to identify what datasource columns correspond to azimuth/zenith or mu_x/mu_y_mu_z?

### DIFF
--- a/PYME/LMVis/layer_panel.py
+++ b/PYME/LMVis/layer_panel.py
@@ -162,7 +162,7 @@ class LayerPane(afp.foldingPane):
         #logger.debug('Layer panel update done')
 
     def add_layer(self, evt):
-        dlg = wx.SingleChoiceDialog(self, 'Choose type of layer to add:', 'Add Layer', ['points', 'mesh', 'image', 'tracks', 'quiver'])
+        dlg = wx.SingleChoiceDialog(self, 'Choose type of layer to add:', 'Add Layer', ['points', 'mesh', 'image', 'tracks', 'quiver', 'oriented points'])
         if dlg.ShowModal() == wx.ID_OK:
             type = dlg.GetStringSelection()
             if type == 'points':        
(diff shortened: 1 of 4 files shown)


barentine:
Couple outstanding issues:
- [x] color/alpha variation along vectors [unintentional] that then doesn't match orientation
- [x] not all points getting drawn, it seems

image

barentine:
OK, that's sorted. 
I was only drawing half the vertices, and was only providing colors for that half (so each half of the line was taking a different color value rather). Fixed:
image

PR #1522, opened 2024-06-17
OIDICFrameSource tweaks

David-Baddeley:
@Yujin-Bao - as discussed. Hopefully this fixes your issue.

### DIFF
--- a/PYME/Acquire/Hardware/driftTracking.py
+++ b/PYME/Acquire/Hardware/driftTracking.py
@@ -103,19 +103,42 @@ class OIDICFrameSource(StandardFrameSource):
     """
 
     def __init__(self, frameWrangler, oidic_controller, oidic_orientation=0):
-        super().__init__(frameWrangler)
+        #super().__init__(frameWrangler)
+        self._fw = frameWrangler
+        self._on_frame = dispatch.Signal(['frameData'])
+        
+        # connect to onFrame rather than onFrameGroup so we get the
+        # current frame which as opposed to an older one.
+        # this is important as the OIDIC orientation is expected to change between frames.
+        # NOTE: we still assume that this connection happens before any of the acquisition
+        # classes connect to onFrame, so that we get the frame data (and microscope state)
+        # before the acquisition classes have had a chance to modify the state (e.g. by changing the DIC orientation).
+        # TODO: This should be pretty safe, as the FrameSource is usually created in the init script, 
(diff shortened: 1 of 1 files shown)


Yujin-Bao:
Hi @David-Baddeley - Thanks for the quick fix! This actually works! Attached here are drift tracking curves when acquiring OI-DIC (first one shows 8-frame averge acquisition mode, second one shows no average acquisition mode). The 10 to 20 nm detected focus shift upon starting acquisition should be fine since this is usually within the focus lock tolerance. However, I am still not sure if it actually offsets the piezo because while the magenta curve updates, dz does not accordingly. 

8avg

noavg

David-Baddeley:
Great that it seems to work! I suspect it is actually moving, but would be
good to work out where the initial offset comes from ...

Yujin-Bao:
To try figuring out the initial offset, I manually set the liquid crystal channels. But at the moment it changes to home position from the other positions, the curve drops again like the previous OI-DIC acquisition before the fix. This makes sense actually, so I need to think of another way to check.

For the piezo offset, looking at the above attached images, it seems with the averaging acquisition mode it is actually moving but with the non-averaging mode it is not. Is it possible that the acquisition is too fast and the liquid crystal has already changed to another channel before the piezo actually moves? (I do not have the evidence, just making a wild guess here)

David-Baddeley:
As a temporary debugging measure, it might be worth commenting out the `scope.SetPos()` line in `set_z()` in `XYZTCAcquisition`, to see if that changes anything.

Yujin-Bao:
After a more careful test I believe the piezo is actually moving as expected. The initial offset comes from the `scope.SetPos()` in `XYZTCAcquisition`. Here `scope.SetPos()` calls `scope.state.setItems()` to update the piezo position `Positioning.z`. But as I check the state I realize the z position shown in `PYMEAcquire` is the value of `Positioning.z`, but the value of `Positioning.z_target` is always around 20 nm below. When I use `scope.state.setItems()` to update the `Positioning.z`, the `Positioning.z_target` will change to this new value but `Positioning.z` will be around 20 nm above this new value - this seems to be exactly the cause of the initial offset. I wonder if this is because we need to consider the piezo tolerance, or there are some other reasons that the current position and target position should be different? I just worry the piezo will move up too much when I take multiple OI-DIC timepoint sequences.

image

Issue #1521, opened 2024-06-17
[ENHANCEMENT] recipe input modules for some scenarios


csoeller:
**General issue:**

We find ourselves now semi-regularly with a use case where we would want to use an input module for an image type datasource in a PYMEVisualise recipe. This is typically an image that is used for masking localisations. While this can be achieved by `Open Raw/Prebeleach data` the main use case is that where you want to preserve the state of a PYMEVisualise session and quickly get back to that later - WITHOUT having to go through 3-5 GUI interactions to make it happen.

**Further background:**

Recipes seeme to be used in at least two quite distinct scenarios:

1. To run a type of computation on a number of datasets in kind of a loop (mostly with image data?), i.e. `bakeshop` et al.
2. To save the state of processing of a localisation file and allow reopening with that state loaded from a yaml.

The second case is the one we focus on mostly these days and here the ability to load a mask and associate it via recipe with the localisation dataset in question is the use case we frequently have.

**Current workaround solution:**

I made a fairly simple input module for something that can be loaded with the ImageStack class, as follows:

```python
from PYME.IO.image import ImageStack
        
@register_module('LoadMask')    
class LoadMask(ModuleBase):

    inputName = Input('FitResults') # we use this as a dummy input since we need at least one input
    outputImageStack = Output('image_mask')

    maskfile = FileOrURI('')

    def run(self,inputName):
        if self.maskfile != '':
            fp = Path(self.maskfile)
            if not fp.exists():
                warn("trying to load file '%s' that does not exist" % self.maskfile)
                return None
            mask = ImageStack(filename=self.maskfile)
            return mask
        else:
            return None # this will trigger an error as it tries to attach the mdh; may be better to have an empty container type
```

Some issues with that:

- it needs a dummy input (need at least one input per module with current code)
- it does not check extensions or similar
- it uses absolute paths (probably has to) and thus makes it more tricky to share with others

Nevertheless it serves a need reasoanbly well and serves as a proof-of-principle.

**Open questions:**

1. Is there space for a better way to implement this that acknowledges the two quite different use cases of recipes, see above?
2. WIth the idea of "preservation of state" is it conceivable to also save the layer arrangement with a recipe so that, for example, the layer layout could be restored with a recipe (or generalisation thereof)? This is interesting when you have several layers with specific datasource settings etc for a specific dataset.

**Caveats**

Apologies if I have overlooked existing functionality that already implements this.

David-Baddeley:
[...]

Issue #1465, opened 2024-02-08
[BUG] Bakeshop does not currently allow for localizations to be read from csv files


abalaji0827:
**Describe the bug**
When batch running a recipe with bakeshop, I am unable to use csv files for my input.

**To Reproduce**
1. open bakeshop
2. load recipe
3. load csv file to run recipe on

**Expected behavior**
I expect the recipe to run as it usually does for hdf input

David-Baddeley:
Could you please provide the recipe in question and the error message? 

There are subtle differences in how .csv and .h5r/.hdf load and what names they have when loaded into the namespace (hdf loads in as with a name based on the table name within the HDF file - e.g. `FitResults`, whereas .csv is likely to come in as just `input`. This is because you can have more than one table in an hdf file so you can't just use something generic like `input`, and .csv doesn't have any way of specifying a table name so you need to use something generic. I'd need to double check, but there might also be some assumptions and pre-processing made for .h5r (which can safely be assumed to be localisations) which can't be safely made for .csv which could be pretty much anything. You can get away with using .csv and .h5r almost interchangeably in PYMEVis because we assume everything is localisations. In recipes we make no assumption that the inputs are localisations, .csv could also be e.g. the sizes of segmented objects calculated by a previous recipe, or a histogram, or astigmatic calibration curves, or the output of some quantification performed in ImageJ.

The short term solution is likely going to be to modify your recipe to use input naming that is appropriate for .csv, but this is a bit ugly as it means that you need to maintain two copies of the recipe (one for .csv, one for .h5r/.hdf). In the medium term, we could look at making the input name that was used for .csv a settable option in the bakeshop so you could use the same recipe. We haven't done this so far because there are a lot of edge cases so it's unclear how often you would actually be able to use the one recipe - in addition to the localisations, the .h5r spec includes metadata, "events" (containing info about, e.g. focal steps etc), and an optional table of fiducial positions for use with fiducial based drift correction, so you would really need 3 separate .csvs and a metadata file to be equivalent for all use cases.

barentine:
[...]

PR #1523, opened 2024-06-18
PYMEVis extra datasources

David-Baddeley:
First steps towards a solution for #1521 

You can now load additional files into the PYMEVis recipe namespace on the command line e.g.

`PYMEVis myfile.h5r -l raw myfile_raw.h5`

or 

`PYMEVis myfile.h5r --load raw myfile_raw.h5`

Also includes some first steps towards #1465 and allowing .csv as input to bakeshop etc ... (note, this needs a `with unifiedIO.local_or_temp_filename(filename) as fn` before it will work on the cluster).

### DIFF
--- a/PYME/Acquire/Hardware/driftTracking.py
+++ b/PYME/Acquire/Hardware/driftTracking.py
@@ -103,19 +103,42 @@ class OIDICFrameSource(StandardFrameSource):
     """
 
     def __init__(self, frameWrangler, oidic_controller, oidic_orientation=0):
-        super().__init__(frameWrangler)
+        #super().__init__(frameWrangler)
+        self._fw = frameWrangler
+        self._on_frame = dispatch.Signal(['frameData'])
+        
+        # connect to onFrame rather than onFrameGroup so we get the
+        # current frame which as opposed to an older one.
+        # this is important as the OIDIC orientation is expected to change between frames.
+        # NOTE: we still assume that this connection happens before any of the acquisition
+        # classes connect to onFrame, so that we get the frame data (and microscope state)
+        # before the acquisition classes have had a chance to modify the state (e.g. by changing the DIC orientation).
+        # TODO: This should be pretty safe, as the FrameSource is usually created in the init script, 
+        # but can we make this robust agains the assumed ordering (execution of handlers in order of addition is an undocumented feature of dispatch)?
+        self._fw.onFrame.connect(self.tick)
 
         self._oidic = oidic_controller
+
+        self._last_tick_time = 0
+        # throttle to ensure 50 ms delay between completion of computation on one frame and start of computation on the next.
+        # the 50ms is emperical, but should (hopefully) be long enough to let anything else which is hooked to onFrame run. 
+        self._tick_throttle = 0.05 
         #self._target_orientation = oidic_orientation
 
-    def tick(self, *args, **kwargs):
-        # FIXME - check when onFrameGroup is emitted relative to when the OIDIC orientation is set.
-        # Is this predictable, or does it depend on the order in which OIDIC and drift tracking are
-        # registered with the frameWrangler?
+    def tick(self, frameData, **kwargs):
+        # Because we are connected to onFrame rather than onFrameGroup (which is inherently throttled by the GUI loop),
+        # we need to throttle the signal ourselves to avoid overwhelming things
+        # when the frameWrangler is running at high speed (i.e. drift computation time
+        # is slower than or on the order of the camera integration time). 
+        if (time.time() - self._last_tick_time) < self._tick_throttle:
+            return
+        
         #self._target_orientation = self._oidic.home_channel()
         #if self._oidic.current_channel == self._target_orientation:
         if self._oidic.current_channel == self._oidic.home_channel():
-            super().tick(*args, **kwargs)
+            # send with the frame data of the frame which triggered the signal, rather than frameWrangler.currentFrame, which lags.
+            self._on_frame.send(sender=self, frameData=frameData)
+            self._last_tick_time = time.time() 
         else:
             # clobber all frames coming from camera when not in the correct DIC orientation
             pass
--- a/PYME/IO/csv_flavours.py
+++ b/PYME/IO/csv_flavours.py
@@ -232,8 +232,8 @@ def guess_flavour(colNames, delim=None, ext=None):
 
     return fl
 
-
-
+text_extensions = ['.csv', '.txt'] + [e for e in [csv_flavours[flavour].get('ext',None) for flavour in csv_flavours] if e is not None]
+text_extensions = list(set(text_extensions))
 
 def gen_mdh(self): # generate some metaData that will be passed up the chain
                     # to record some bits of this import
--- a/PYME/IO/csv_flavours.py
+++ b/PYME/IO/csv_flavours.py
@@ -187,6 +187,9 @@ def guess_text_options(filename):
 
     return text_options
 
+def strict_options(text_options):
+    '''Remove all options from dictionary which aren't defined in our text_options definition'''
+    return {k: text_options[k] for k in text_options if k in ['columnnames', 'skiprows', 'delimiter', 'invalid_raise']}
 
 def check_required_names(self):
     reqNotDef = [name for name in self.requiredNames.keys() if not name in self.translatedNames]
--- a/PYME/IO/image.py
+++ b/PYME/IO/image.py
@@ -1356,6 +1356,10 @@ class ImageStack(object):
             self.saved = True
 
     def Save(self, filename=None, crop=False, roi=None, progressCallback=None):
+        warnings.warn(DeprecationWarning('The "Save" method is deprecated, please use "save" instead'))
+        self.save(filename=filename, crop=crop, roi=roi, progressCallback=progressCallback)
+
+    def save(self, filename=None, crop=False, roi=None, progressCallback=None):
         """
         Saves an image to file.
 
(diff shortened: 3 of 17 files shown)


csoeller:
Pretty cool, I had a quick try.

The following works:

```
PYMEVis '240608-103059_sA_3D_U2OS_NPC_&_HEK_RyR_MP_X2_0p8nM_optimal_beads.npy' -l mask 240608_sA_whole_cell_gau_10nm_NPC-z-clip_fitroi_MASK.tif
```

which includes in the diagnostic output

```
....
Input datasources:  ['FitResults', 'mask']
...
```

However, when I also supply a recipe as argument that tries to use this additional input datasource `mask` it does not work and the output does not mention the datasource `mask` any more:

```
PYMEVis '240608-103059_sA_3D_U2OS_NPC_&_HEK_RyR_MP_X2_0p8nM_optimal_beads.npy' -l mask 240608_sA_whole_cell_gau_10nm_NPC-z-clip_fitroi_MASK.tif -r 240608_sA_newstylerecipe_V4.yaml
```

with outputs and error:

```
...
PYME.recipes.base.MissingInputError: Input "mask" is missing from namespace; keys: ['FitResults', 'Localizations', 'colour_mapped', 'filtered_localizations', 'with_clumps', 'mapped', 'coalesced_nz', 'coalesced_nc', 'dbscanClusteredDyes', 'closemerged', 'coalesced_mbm', 'RyRonly']
...
Input datasources:  ['FitResults', 'filtered_localizations']
...
```

Sounds like the recipe processing happens before the mask datasource would be loaded and added to the namespace or similar?

David-Baddeley:
Have changed loading order, hopefully it works now ...

csoeller:
Very nice, this works now! Thanks!

csoeller:
Purely as a proof-of-concept I have implemented a quick `.pvms` format (based on yaml) that essentially collects arguments for `PYMEVis` and rewrites the args accordingly, as in diff below. The `.pvms` format could then become a type of container format and eventually be extended to become close to capturing the state or similar, or just have a single file to open localisation data + processing in one encapsulated step. The sample implementation works for simple scenarios as is.

Happy to discuss directions for such an approach and tweak formats, keywords, logic, implementation etc. In our hands even in this fairly primitive form it is already useful by cutting down on repetitive gui tasks, keeping a replayable record, etc.

**Sample .pvms container file:**

```yaml
pmvs_version: v1.0
localizations: 240131-215119_sC_3D_U2OS_NPC_MP_X2.npy
recipe: 240131-215119_sC_3D_NPC-recipe_masked.yaml
imageds:
    mask: 240131-134721_sC_3D_NPC-recipe_ROIzclipped-mask.tif
    raw: 240131-215119_sC_3D_NPC-recipe_ROIzclipped.tif
```

**Simplified proof-of-concept implementation:**

```diff
diff --git a/PYME/LMVis/VisGUI.py b/PYME/LMVis/VisGUI.py
index c89dad65b..ccacb927e 100755
--- a/PYME/LMVis/VisGUI.py
+++ b/PYME/LMVis/VisGUI.py
@@ -446,6 +446,31 @@ def main_(filename=None, use_shaders=False, args=None):
     check_for_updates.gui_prompt_once()
     application.MainLoop()
 
+def load_pmvs(filename):
+    import yaml
+    # TODO: needs to do some additional checks
+    # TODO: needs to consider relative and absolute filepaths
+    with open(filename, 'r') as file:
+        pvms_args = yaml.safe_load(file)
+    if pvms_args.get('pmvs_version') != 'v1.0':
+        raise RuntimeError("file %s is not a PVMS version 1.0 file" % filename)
+    if 'localizations' not in pvms_args:
+        raise RuntimeError("fatal: no localization file in PVMS file named")
+    return pvms_args
+    
+def load_and_parse_pmvs(args):
+    pmvs_args = load_pmvs(args.file)
+    if len(args.load) > 0: # check if this restriction is really needed
+        raise RuntimeError("loading additional files from the command line not allowed when using pmvs file")
+    if pmvs_args.get('recipe',None) is not None:
+        if args.recipe is not None: # check if restriction sensible
+            raise RuntimeError("recipe in pmvs file but also recipe supplied on command line - conflict")
+        args.recipe = pmvs_args['recipe']
+    for name in pmvs_args.get('imageds',{}):
+        args.load.append((name,pmvs_args['imageds'][name]))
+    args.file = pmvs_args['localizations']
+    return args
+    
 def parse():
     parser = argparse.ArgumentParser()
     parser.add_argument('file', help="file that should be used", default=None, nargs='?')
@@ -458,8 +483,11 @@ def parse():
     parser.add_argument('--no-layers', dest='new_layers', action='store_false', default=True)
     parser.add_argument('-l', '--load', nargs=2, action='append', default=[], dest='load', metavar=('KEY', 'FILENAME'), help='Load one (or more) additional files into the recipe namespace.')
     args = parser.parse_args()
+    if args.file.endswith('.pmvs'):
+        args = load_and_parse_pmvs(args)
+        #print(args)
     return args
-    
+
 def main():
     from multiprocessing import freeze_support
     import PYME.config
```

David-Baddeley:
A couple of things to consider:

- should we put the entire recipe text in the session (saves having to save a separate file and worry about it going out of sync) rather than just a filename?
- the way it's coded extra data sources can be anything (e.g. a second localisation channel etc ...) and not just images. The session key should reflect this - maybe `datasources`. 

This might give us something like the following 
```yaml
# PYMEVis saved session
format_version: 0.1
recipe:
    ...
datasources:
    ...
layers:
    ...
view:
    ... 
```

and raises the question of whether the `Localisations` data should be specified separately, or simply as a key in datasources. In many ways the latter would be cleaner (as we're saving the recipe anyway, we don't need much of the special case logic that, e.g., creates the `Pipelineify` and `ColourFilter` modules). The remaining problem is that if we used a standard recipe load, the table-name-munging logic `Localisations=somefile.h5r` would give us `Localisations_fitResults` and `Localisations_driftResults` rather than `Localisations` and `Fiducials`. Fixing this would also fix the long-standing issue of not being able to create a recipe in PYMEVis and run it in batch processing without modifying it (due to different input names).

David-Baddeley:
Maybe if we changed the logic so that we could specify a particular table on the command line (potentially using query strings - e.g.:
```yaml
datasources:
    Localizations: myfile.h5r?FitResults
```
or 
```yaml
datasources:
    Localizations: myfile.h5r/FitResults
```
or 
```yaml
datasources:
    Localizations: myfile.h5r?table=FitResults
```
Both using querystrings and just treating HDF as a directory have precedents in PYME - it's hard to know which to use. I'm leaning towards a full query string with `table=FitResults` etc ... as this interface would also allow for csv and .mat options to be specified / remembered.

David-Baddeley:
just pushed a change which should allow the `myfile.h5r?table=FitResults` syntax (along with a lot of tidying).

csoeller:
Cool, I'll try some tests to see how this works in practice. I like the various suggestions and will try a sample implementation how to read this info and make PYMEVis behave accordingly.

David-Baddeley:
Maybe don't invest too much time just yet - I've got a few more changes in
mind for the opening logic which should make it easier.

David-Baddeley:
@csoeller - I got a bit carried away, but we now have a semi-functional session framework, which saves and restores:

- input datasource(s)
- recipe
- selected datasource
- layers (currently the 4 main types)
- view

Save from the menu ('File->Save session'), load by specifying on the command line to PYMEVis (`PYMEVis mysession.pvs`). 

Still lots of rough edges and a lot of scope for making things cleaner by further reducing duplication between the recipe IO and pipeline (and refactoring a bunch of it out of the recipe base class).

### TODOs
- output filter settings (including pipeline ROI) (this could use some general attention, as it's a hangover from the less flexible pipeline)
- rework the open extra channel / open raw data stuff so that they get injected into pipeline.
- GUI open session support
- lots of testing!!!

David-Baddeley:
### More TODOs
- remember where to get events from (for event viewer)
- test with multicolour data
- additional layer types (e.g. quiver)

csoeller:
Cool, I'll do some initial testing probably in the coming week and will provide some feedback as I get to it. Really appreciated that you are looking into this feature set.

csoeller:
I have started testing this PR, works pretty well and amazing!

A couple of niggles, no deal-breakers but a bit annoying:

1. When you load a new recipe some old data sources can hang around. Example:

     pymevis 20_12_series_F.h5r -r rename-filtered_localizations-testrecipe.yaml

where rename-filtered_localizations-testrecipe.yaml has:

```yaml
- localisations.AddPipelineDerivedVars:
    inputEvents: ''
    inputFitResults: FitResults
    outputEventMaps: event_maps
    outputLocalizations: Localizations
- localisations.ProcessColour:
    input: Localizations
    output: colour_mapped
- tablefilters.FilterTable:
    filters:
      A:
      - 5
      - 100000
      error_x:
      - 0
      - 30
      error_y:
      - 0
      - 30
    inputName: colour_mapped
    outputName: new_filtered_localizations
```

Notice `new_filtered_localizations`; the previous `filtered_localizations` leads to a warning `datasource filtered_localizations has no filename...` etc.

I guess something is not quite right when previous datasource names should really be dropped uplon loading a different recipe?

csoeller:
2. when loading data without events (e.g. from `csv`) an event_map related warning shows up when saving... Perhaps needs testing if event_map should be added to namespace when no events?

csoeller:
PS - ignore the event_map business for now; I seem to actually have a different issue with csv but will need to check carefully beforer reporting...

David-Baddeley:
Going to merge this as is, and revisit serialisation for additional layer types later (after the glcore merge)

PR #1524, opened 2024-07-03
Plugin factory

David-Baddeley:
Adds a factory for templating user plugins. Should ultimately replace https://github.com/python-microscopy/pyme-plugin, which has suffered a bit from bitrot. Main differences to  https://github.com/python-microscopy/pyme-plugin:

- uses new (current) module config which puts all the module related config stuff in one .yaml file rather than spread across many text files
- avoids having to manually change package_name in umpteen places
- has example code for the various plugin types, rather than just empty files.

### Usage

`python -m PYME.util.plugin_factory output_directory package_name`

Defaults to creating all types of plugin (except reports), but that can be altered with optional command line flags:

```bash
usage: plugin_factory.py [-h] [--pymeimage PYMEIMAGE] [--pymevis PYMEVIS] [--recipe RECIPE] [--fit_factories FIT_FACTORIES] [--reports REPORTS] output_dir package_name

Create a template for a PYME plugin

positional arguments:
  output_dir            Directory to write the plugin template to
  package_name          Name of the package

optional arguments:
  -h, --help            show this help message and exit
  --pymeimage PYMEIMAGE
                        Create a PYMEImage plugin with the given name
  --pymevis PYMEVIS     Create a PYMEVisualize plugin with the given name
  --recipe RECIPE       Create a recipe module with the given name
  --fit_factories FIT_FACTORIES
                        Create a fit factory with the given name
  --reports REPORTS     Create a reports module with the given name
```

### DIFF
--- a/PYME/util/plugin_factory.py
+++ b/PYME/util/plugin_factory.py
@@ -1,0 +1,465 @@
+def create_pymeimage_plugin():
+    plugin_template = f'''\
+# This is a template for a PYMEImage UI plugin.
+from PYME.DSView.modules._base import Plugin
+
+class MyCoolPluginKlass(Plugin):
+    def __init__(self, dsviewer):
+        super().__init__(dsviewer)
+
+        # Add your code here
+
+        # Example: Add a menu item
+        dsviewer.AddMenuItem('Processing', 'Do something neat', self.OnDoSomethingNeat)
+
+    def OnDoSomethingNeat(self, event=None):
+        # Menu item callback
+        # Add your code here
+        pass
+    
+def Plug(dsviewer):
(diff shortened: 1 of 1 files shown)


csoeller:
Just checking the status of this PR. I had a couple of people asking me about plugin writing and was wondering how far this PR is from becoming the preferred approach to get into it.

PS - WIth setuptools phase out this may ultimately need updating to the replacement build system but presumably still some time away??

csoeller:
To answer my own question - the factory is 99% functional but needs some tweaks to fix typos and other small glitches. I'll try to some up with patches and raise a PR against the parent branch of this PR if I can figure this out...

Issue #1525, opened 2024-07-08
[BUG] FRC canvas set title error with matplotlib > 3.4


csoeller:
**Describe the bug**
With matplotlib > 3.4 you get with `Processing>FRC`:

```
 DEBUG:matplotlib.pyplot:Loaded backend WXAgg version unknown.
/Users/csoe002/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.10-conda_tmp/python-microscopy-master
Traceback (most recent call last):
  File "/Users/csoe002/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.10-conda_tmp/python-microscopy-master/PYME/ui/progress.py", line 153, in func
    return fcn(*args, **kwargs)
  File "/Users/csoe002/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.10-conda_tmp/python-microscopy-master/PYME/DSView/modules/coloc.py", line 404, in OnFRC
    correlationCoeffs.fourier_ring_correlation(imA, imB, voxelsize, True)
  File "/Users/csoe002/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.10-conda_tmp/python-microscopy-master/PYME/Analysis/Colocalisation/correlationCoeffs.py", line 143, in fourier_ring_correlation
    fig.canvas.set_window_title('Fourier Ring Correlation')
AttributeError: 'FigureCanvasWxAgg' object has no attribute 'set_window_title'. Did you mean: 'SetWindowStyle'?
```

**To Reproduce**
1. `Processing>FRC`

Reason, see https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.4.0.html#backend-deprecations. Probably error since 3.6.X or so. Relatively easy to fix, it seems, though, see separate PR.

**Expected behavior**
No error if (do we?) allow matplotlib >3.4. (we should, in any case).

csoeller:
fixed with merge of #1526

PR #1526, opened 2024-07-08
FRC canvas set title fix for compat with newer matplotlibs

csoeller:
Addresses issue #1525.

bugfix.

**Proposed changes:**
Refer to `manager` attribute if direct access to set_title from canvas fails. Should fix issue #1525 while staying backwards compatible for earlier matplotlib.

**Checked:** 
Checked with latest modules in fresh install.


David-Baddeley:
Thanks!

PR #1529, opened 2024-07-15
Add a labels layer type for annotated tabular data

David-Baddeley:
Work in progress ... This has been on the TODO for a while, adds a layer for text annotations - @csoeller 

image

Example test usage (within pymevis console):
```python
>>> from PYME.LMVis.layers.labels import LabelLayer
>>> from PYME.IO import tabular
>>> rs = tabular.RandomSource(xmax=10000, ymax=10000, nsamps=100)
>>> pipeline.addDataSource('foo', rs, False)
>>> ll = LabelLayer(pipeline, dsname='foo')
>>> pymevis.add_layer(ll)
>>> ll.update()
```

### Caveats and TODOs

- layer doesn't update properly when you change stuff in UI (need to call `.update()` on the layer manually, then resize window to force a repaint)
- Limited to relatively small datasets (you don't want to add a `LabelLayer` pointing to a standard localisation dataset)
- colour and font size settings are ignored
- should probably allow a programmable offset

### DIFF
--- a/PYME/DSView/modules/psfTools.py
+++ b/PYME/DSView/modules/psfTools.py
@@ -23,13 +23,19 @@
 import wx
 import wx.html2
 import wx.grid
+import wx.lib.agw.aui as aui
+
 
 from jinja2 import Environment, PackageLoader
 env = Environment(loader=PackageLoader('PYME.DSView.modules', 'templates'))
 
 from PYME.recipes.traits import HasTraits, Float, Int, Bool, Enum
+import numpy as np
 
-from .graphViewPanel import *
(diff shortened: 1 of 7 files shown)


csoeller:
Works pretty nicely in my initial tests, does exactly what I want and even without a programmable offset works ok for my initial purpose (although an optional offset would probably be nice, initially could be a fixed vector). Next on the wishlist may be to make it also work with session saving which as yet does not seem to work (not surprising probably) but would be very cool if it could in future. [sorry for wanting to mix two experimental features]

Screenshot 2024-07-17 at 21 04 58

PR #1531, opened 2024-07-29
Fix for TypeError when no Track Fiducials selected

Yujin-Bao:
**This is a bugfix**

**Proposed changes:**
In `testFrame`, I made displaying `driftResults` only available when  `Track Fiducials` is selected when doing localization. It is because when `Track Fiducials` is not selected, `res.driftResults` will be an empty list instead of an array and it gives `TypeError: list indices must be integers or slices, not str`. However, this bug only exists in `testFrame` and does not affect the functionality of the full localization analysis.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/DSView/modules/LMAnalysis.py
+++ b/PYME/DSView/modules/LMAnalysis.py
@@ -999,10 +999,11 @@ class LMAnalyser2(Plugin):
                     plt.yticks([])
                     plt.plot(res.results['fitResults']['x0']/vx, res.results['fitResults']['y0']/vy, '+b')
 
-                try:
-                    plt.plot(res.driftResults['fitResults']['x0']/vx, res.driftResults['fitResults']['y0']/vy, '*y', mew=2)
-                except AttributeError:
-                    pass
+                if ft.driftEst:
+                    try:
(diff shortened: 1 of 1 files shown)


csoeller:
Hi @Yujin-Bao, I came across the same error and fixed it ever so slightly differently, as below - hadn't got around to raise a PR. I imagine both are ok and just a heads up that this error occurs with more recent python versions, so should ideally be merged relatively swiftly.

```diff
diff --git a/PYME/DSView/modules/LMAnalysis.py b/PYME/DSView/modules/LMAnalysis.py
index 22e6579..1a96edd 100644
--- a/PYME/DSView/modules/LMAnalysis.py
+++ b/PYME/DSView/modules/LMAnalysis.py
@@ -1001,7 +1001,7 @@ class LMAnalyser2(Plugin):
 
                 try:
                     plt.plot(res.driftResults['fitResults']['x0']/vx, res.driftResults['fitResults']['y0']/vy, '*y', mew=2)
-                except AttributeError:
+                except (AttributeError, TypeError):
                     pass
                     
                 #figure()
```

David-Baddeley:
Thanks @Yujin-Bao and @csoeller!

Yujin-Bao:
Thanks @csoeller and @David-Baddeley ! I tried a similar way that includng `TypeError` in, but it seemed I did the syntax wrong. Thanks for letting me know the other way for it!

PR #1532, opened 2024-08-01
fix point sprite regression

David-Baddeley:
Fixes regression on large point sprites as noted by @csoeller


csoeller:
Fantastic, thanks for the quick turnaround. Indeed, works now well. 

I will keep using and also exercise the session I/o code and report on any findings I come across. 

In any case, for me already has resulted in a much nicer production environment.

PR #1536, opened 2024-08-09
Olympus IX81 control code with command set - currently halogen lamp only

Yujin-Bao:
Addresses issue # .

**This is an enhancement**

**Proposed changes:**
Control code for Olympus IX81 microscope stand - halogen lamp only for now.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/Acquire/Hardware/olympusix81.py
+++ b/PYME/Acquire/Hardware/olympusix81.py
@@ -1,18 +1,20 @@
 import serial
 import time
 import threading
+from PYME.Acquire.Hardware.lasers import Laser
 
-class halogenlamp:
-    def __init__(self, name, portname='COM17', **kwargs):
+class OlympusIX81HalogenLamp(Laser):
+    def __init__(self, name, portname='COM17', turn_on=False, **kwargs):
         self.ser_port = serial.Serial(portname, 19200, parity='E',
                                       timeout=2, writeTimeout=2)
         self.lock = threading.Lock()
-        self.name= name
+        self.name = name
         self.powerControlable = False
-        self.isOn=True
-        self.TurnOn()
+        self.isOn = False
+        #self.TurnOn()
+        Laser.__init__(self, name, turn_on, **kwargs)
 
-    def query(self, command, lines_expected=1):
+    def _query(self, command, lines_expected=1):
         with self.lock:
             self.ser_port.reset_input_buffer()
             self.ser_port.write(command)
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Hi Yujin,

a few thoughts:

- Naming convention is to use upper case for class names (we are trying to keep new code PEP8 compliant where reasonable - that doesn't extend to renaming interfaces - i.e. the methods from the `Laser` class, but does mean that class names, utility functions, etc ... should try and follow PEP8)
- Good practice for hardware is also to include manufacturer name (or other identifying product name in class name). E.G. `OlympusIX81HalogenLamp` or potentially just `IX81HalogenLamp` (makes it a bit harder to inadvertently use the wrong class).
- Shutters and lamps etc (anything where you are turning light on and off) ... should inherit from the `Laser` base class in `PYME.Acquire.Hardware.lasers`.
- I can see potential for issues with the `login, turn on`, `turn off, logout` flow - it assumes that things they will always execute in that order and that you are not going to want to e.g. interact with anything else in bank 1 of the stand during an acquisition. I can easily see scope for problems with e.g. the standard `turn_all_lasers_off()` call at the end of protocols, which will (deliberately) try and turn things off even if they are already off. It would likely be safer to do something like `login, turn on, logout` and `login, turn off, logout` (if the performance penalty is not too high).
- You just discard the responses without trying to interpret them and check for errors - this is likely to make debugging hard if / when something goes wrong - e.g. the microscope stand crashes, or you get the login sequence wrong etc ...
- Extremely minor, but while we are on PEP8,  I would probably make `query()` "private" - ie `_query()` as it's only expected to be used from within the class. 
- Probably something to worry about later, if and only if it looks like being a problem, but the docs say that replies can come out of sequence (which means that you'd need something rather a lot more complex than the current query handling).

Best wishes,
David

Yujin-Bao:
Hi David, thanks for the comments! The code has been modified. The only thing I do not quite follow is interpreting and checking for errors - I wonder if you could explain more since I mostly copied the `TurnOn` and `TurnOff` from `priorarclampshutter`.

PYME/Acquire/Hardware/olympusix81.py
```diff
@@ -0,0 +1,49 @@
+import serial
```
David-Baddeley, line 11:
```suggestion
"""
Control the halogen lamp (transmitted light) on an Olympus IX81 stand.

Uses the serial command set described at https://madhadron.com/science/olympus_ix81_commands.html

TODO - move logic to an IX81 class which also allows control of other stand features and support more of the command set.
TODO - some form of error handling
TODO - consider re-writing the query function to try and match responses to commands (these are not guaranteed to be synchronous) 

"""
import serial
```

David-Baddeley:
Re the error checking - the microscope returns success or failure codes in response to what you ask it to do. In an ideal world we would parse these (and either log the error, or potentially raise an exception depending on severity).  I'm not sure if the lack of error handling in the Prior Lumen code is because it didn't provide useful error codes, or (more likely) laziness on my part. With a lot of the hardware you can get away with not handling the errors most of the time, even if it's not best practice (where it can come back to bite you is when, e.g., something goes wrong in the middle of a 24 hr recording and you have no idea what happened because you didn't log the error messages).

I've added a comment with a link to the command set documentation and a list of things that could (in an ideal world) be added - mostly for the sake of anyone who comes later and wants to debug of expand it.

PR #1537, opened 2024-08-09
without the missing wx import dh5view fails

csoeller:
Addresses issue where `dh5view` fails to start because of missing `wx` import in `shell.py`.

Bugfix, but bug not separately reported.

**Proposed changes:**
Add missing import.

Hopefully reproducable on other systems, at least saw it on mine...

### DIFF
--- a/PYME/DSView/modules/shell.py
+++ b/PYME/DSView/modules/shell.py
@@ -22,6 +22,7 @@
 
 import PYME.ui.shell
 from PYME import config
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Thanks!

Change is also in #1535 ... so will probably merge that instead.

David

csoeller:
Should be ok to close with #1535 merged.

PR #1538, opened 2024-08-11
Make rule-based analysis work with different acquisition types

David-Baddeley:
This is an attempt to expand the slightly special-case automated analysis we had for the HTSMS paper a bit more generic (i.e. not limited to protocol-driven acquisitions), and to make it a bit easier to navigate for people who are not already familiar with things. Main changes compared with HTSMS version / current status quo:

- Rule chains are no longer tied to a protocol name, but rather a text tag (or filename if the rule chain has been loaded from file).
- This means that you have to specify `analysis_rule_name` in the dictionary of settings when queueing an acquisition task. If you do this through the GUI, this will automatically be the currently selected analysis rule. If you queue through code or the REST endpoint, you are responsible for specifying it.
- The UI should hopefully be a little bit clearer (see below). The "save analysis settings to metadata" switch is gone, as is the analysis settings fold down in the spooling panel. Instead, there is a now a clear switch between 'Interactive' (which means launch PYMEImage, let you choose and test fit modules and settings there), and fully "automatic". For fully automatic, the settings now live in the "Chained Analysis" tab. 

image

- Within the "Chained Analysis" tab, there are a few subtle UI changes, the biggest of which is that you can now edit the blocks of an existing rule. Changes propagate instantaneously (no need to press apply, or similar), and "add localization rule" or "add recipe rule" now add the rule and edit it, rather than editing the settings then adding.

### TODOS

- wire up saving and loading of rule chains
- load rule chains stored in `.PYME/analysis_rules/` at startup. Potentially also provide a default *2D localisation* rule chain, and maybe a *localise and render* and/or *deconvolve* as examples.
- make a null rule an option (maybe the default)
- handle null / empty rule chains 
- make the spooler UI rule name dropdown choices update when a rule is added
- add a visual cue as to which rule within a chain is being edited
- add a visual cue about whether the rule chain you are editing is the currently selected / active rule chain

### DIFF
--- a/PYME/Acquire/ActionManager.py
+++ b/PYME/Acquire/ActionManager.py
@@ -23,6 +23,17 @@ logger = logging.getLogger(__name__)
 
 from .actions import *
 
+class MutablePriorityQueue(Queue.PriorityQueue):
+    def __init__(self, *args, **kwargs):
+        Queue.PriorityQueue.__init__(self, *args,**kwargs)
+
+    def remove(self, item):
+        with self.mutex:
+            if item in self.queue:
+                self.queue.remove(item)
+                #self.queue.sort()
+                self.not_full.notify()
+
 class ActionManager(object):
     """This implements a queue for actions which should be called sequentially.
     
@@ -52,8 +63,8 @@ class ActionManager(object):
             calling eval('scope.functionName')
         
         """
-        self.actionQueue = Queue.PriorityQueue() # queue for immediate execution
-        self.scheduledQueue = Queue.PriorityQueue() # queue for scheduled execution
+        self.actionQueue = MutablePriorityQueue() # queue for immediate execution
+        self.scheduledQueue = MutablePriorityQueue() # queue for scheduled execution
         self.scope = weakref.ref(scope)
         
         #this will be assigned to a callback to indicate if the last task has completed        
@@ -211,6 +222,30 @@ class ActionManager(object):
                     #logger.debug('Queuing action %s for delayed execution (%s, %s)' % (functionName, curTime, execute_after))
                     self.scheduledQueue.put_nowait((after, (nice_, action, expiry, max_duration)))
             
+        self.onQueueChange.send(self)
+
+    def remove_actions(self, actions):
+        """Remove a specific action from the queue
+        
+        Parameters
+        ----------
+        action : Action
+            The action to remove
+        """
+        
+        #TODO - do we need to lock here?
+        
+        for action in actions:
+            # try and remove from both queues
+            try:
+                self.actionQueue.remove(action)
+            except ValueError:
+                pass
+            try:
+                self.scheduledQueue.remove(action)
+            except ValueError:
+                pass
+
         self.onQueueChange.send(self)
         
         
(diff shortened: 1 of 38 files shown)


David-Baddeley:
Going to merge, and deal with any remaining issues as they show up ...

David-Baddeley:
@barentine, @soeller - just a heads up that I'm merging this, and it comes with quite a few changes (hopefully all for the better) - let me know if anything breaks and I'll try and get onto it promptly. Most visible changes are with automatic analysis ...

- there is now a toggle between automatic and interactive analysis 
- *Interactive* is a slightly curtailed version of the default / old *Launch PYMEImage, then start analysis from there* (the curtailment is that there is no longer support for the semi-automatic version which was *Launch PYMEImage with pre-set analysis settings and automatically click the Go button*
- *Automatic* uses rule-chains to launch the analysis directly from PYMEAcquire, with no GUI interaction
- For starting mode, *Trigger* is the old behaviour - launched by either clicking the 'Analyse' button, or from a `launch_analysis()` call in a protocol. *Series end* launches analysis when the current series completes (this was previously only shown in the rule UI), even if the *Analyse* button is not pushed.
- There is a new *View* button, which translates to *Open in PYMEImage* (if an image series), or *Open in the tile viewer* if a tiled series. Tiled series no longer automatically open (i.e. you need to manually hit `View` after series completion) - this enables multiple tiled series to be queued without issues.
- The function of the *Analyse* button stays the same if analysis mode is interactive (i.e. *Open in PYMEImage*), for *Automatic* (rule based) analysis, this now triggers analysis on the cluster, and changes to "View results" which will (not fully implemented) open PYMEVis with a live view of the analysis results 

Behind the scenes, there are a number of smaller changes to make tiling a first-class acquisition type which can be queued like any other acquisition, along with some bug fixes for the interactive mode real-time analysis display. Also some tweaks to the action UI.

@barentine -There is also a reasonably major change in how analysis rules are associated with acquisitions. In broadening rule-based analysis to z-stacks and tiled acquisitions, it was no longer possible to associate rules with protocols. The rule used is now the one which was selected when the acquisition was started *or queued* - if using the acquisition queue. It's a parameter to `queue_actions()`, but I'm not sure if it's made it to the REST API yet. Rule chains can now be saved and loaded, which should be welcome. On the down side, there is a reasonable chance that circular and well plate tiled acquisitions are broken for now (they might still work, but not as first class acquisitions).

barentine:
Thanks for the heads up, @David-Baddeley ! Sounds like a load of good stuff in this PR. I really like the switch to separate out interactive / full-auto. 

Your last comment is a pretty good start to a doc-page! Maybe worth adding the screenshot you have above and the text here to a page on automatic analysis from PYMEAcquire? And a small touch-up to `docs\LocalisationAnalysis.rst`, particularly the [Localizing directly from PYMEAcquire section](https://python-microscopy.org/doc/LocalisationAnalysis.html#localizing-directly-from-pymeacquire).

I'm pleased to see all of this improving! At first pass, I'm surprised to see ~pulling away from protocols, particularly since you can run a tiler acquisition as a protocol (as you flagged, this is how the multiwell jazz was jigged), as well as single snapshot and z-stacks. How aligned those are with the other modes of starting an acquisition is of course a part of what you're (thankfully!) addressing here, but it also seems solvable using vaguely null-patterned protocols. I think I'd gotten it in my head that the roadmap included working towards all acquisitions having associated protocols. Sounds like that's not the plan?

PR #1541, opened 2024-08-16
set new imagebounds when loading a session

csoeller:
Addresses issue that rendering is currently broken when a session was loaded in `PYMEVis`.

The issue seems to result from `pipeline.imageBounds` not set after loading a session (because we bypass `OpenFile` where this normally happens).

Bugfix.
 
**Proposed changes:**
The code uses some heuristics to estimate the image bounds, based on the code in `OpenFile` but slightly adapted to session loading.

The code needs brief checking if this all makes sense. My tests to date indicated it works with both PYME acquired data and external data (CSV, MINFLUX).

### DIFF
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -799,6 +799,18 @@ class Pipeline(object):
 
         self.Rebuild()
 
+        # we need some way to calculate image bounds when loading a session
+        # - presumably this needs to happen after the Rebuild()?
+        # - there could also be an issue if the selectedDataSource only spans a subset of the image extent!
+        # - we therefore select 'FitResults' as datasource in the first attempt and fall back to the selected ds otherwise
+        if ('scanx' not in self.selectedDataSource.keys() or
+            'scany' not in self.selectedDataSource.keys()) and 'Camera.ROIWidth' in self.mdh.getEntryNames():
+            self.imageBounds = ImageBounds.extractFromMetadata(self.mdh)
+        elif 'FitResults' in self.dataSources.keys():
+            self.imageBounds = ImageBounds.estimateFromSource(self.dataSources['FitResults'])
+        else:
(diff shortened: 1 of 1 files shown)


PYME/LMVis/pipeline.py
```diff
@@ -799,6 +799,18 @@ def load_session(self, session_info):
 
         self.Rebuild()
 
+        # we need some way to calculate image bounds when loading a session
+        # - presumably this needs to happen after the Rebuild()?
+        # - there could also be an issue if the selectedDataSource only spans a subset of the image extent!
+        # - we therefore select 'FitResults' as datasource in the first attempt and fall back to the selected ds otherwise
```
David-Baddeley, line 805: in most cases `extractFromMetadata` should still do it's thing, so it's unclear if we really need to have the FitResults fallback (if we ditched this, we could factor the bounds estimation out into a single function and avoid the duplication).

More generally, we might want to address bounds estimation to better cope with multiple source data, e.g., make it the super-set of the bounds of all datasets, rather than just the bounds of a "special" data set.

David-Baddeley: Actually, in the short term, we can probably move this, complete with `FitResults` fallback, to a common `Pipeline._estimate_bounds()` that is shared between here and `OpenFile()`.

David-Baddeley: Longer term, we should add an `.imgBounds` property (lazily computed and cached) to `tabular` datasets to mirror the one in the Image class (or PEP8 the image class and make it `ImageStack.image_bounds`, and replace `pipeline.image_bounds` with a property that computes the bounding box of all datasources.

David-Baddeley:
Have made the long-term change to using the largest bounding box of all datasets in #1543. Needs a little more testing but avoids a lot of guesswork about how we are going to load and structure data.

csoeller:
closed in favour of pull request #1543

PR #1542, opened 2024-08-16
Session paths handling

csoeller:
Addresses paths handling for saved sessions and adds a GUI interface for `load_extra_datasources`.

Enhancement for general session paths handling and (smallish) bug fix to resolve all paths prior to session saving.

**Proposed changes:**

1. Resolve all file paths when requesting session info via `_get_session_datasources`.
2. Add a GUI interface to load extra datasources.
2. Add the ability make session paths (and some paths in recipes) relative to the `.pvs` session file.

**Details:**

1. Originally file paths were returned as stored with the data source. When started from the command line these may be relative to the shell's current working directory (CWD) and this would then fail, for example, when opening the generated `.pvs` file from the Finder window - because `PYMEVis` will generally start with a quite different CWD. Therefore, we now return fully resolved paths when building the session dictionary. But see also 3.
2. In practice it can be quite a bit more convenient to add relevant extra datasources via the GUI (rather than from the command line), tweak the recipe and then save the whole session in a `.pvs` file. Thus this PR also adds a simple GUI interface to the `pipeline` method `load_extra_datasources`.
3. To optionally use relative file paths when writing session files is likely the proposed addition that is the most experimental in this PR. The motivation arises from two scenarios, the first being our prime motivator. (a) Our files and thus also session files often sit in shared directories, for example in OneDrive or Dropbox. As a result absolute paths differ between machines but relative paths stay the same. (b) We sometimes  move files en-bloc with their session files into a subdirectory, different directory etc. This generally preserves relative paths but clearly not absolute paths.

**about 3 - relative filepaths:**

This PR implements the functionality to write relative paths in sessions (and related recipe modules) but enables this only if explicitly requested by setting the `PYME.config` variable `VisGUI-session_paths_relative` to `true` (it is false by default).

The paths translation is careful in the sense that if paths are already in the state requested (relative or absolute) they are left untouched and if errors are encountered when paths are attempted to be rewritten relative then a warning is issued and the original session dict is returned with paths untouched.

We also enable manual addition of modules with their relevant fields to translate absolute paths in `traits.File` fields in recipes. There may be a better (automagic) way, but in practice we have one or two modules that need this.

Perhaps not everybody will want/like this relative paths functionality but at least in our tests everything seems quite robust and in practice makes life a whole lot easier. In addition, as it is not happening by default we can gain some experience with this feature and see if it is worthwhile.

### DIFF
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -509,12 +509,13 @@ class Pipeline(object):
         
         """
         from PYME import warnings
+        from pathlib import Path
         out = {}
         for k in self.recipe.inferred_data:
             ds = self.dataSources[k]
 
             if hasattr(ds, 'filename'):
-                fn = ds.filename
+                fn = str(Path(ds.filename).resolve()) # ensure we return an absolute path!
 
                 q = getattr(ds, 'query', None)
                 if q:
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -508,14 +508,13 @@ class Pipeline(object):
         Return a dictionary of datasources which are suitable for listing in a session file
         
         """
-        from PYME import warnings
-        from pathlib import Path
+        from PYME import warning
         out = {}
         for k in self.recipe.inferred_data:
             ds = self.dataSources[k]
 
             if hasattr(ds, 'filename'):
-                fn = str(Path(ds.filename).resolve()) # ensure we return an absolute path!
+                fn = ds.filename
 
                 q = getattr(ds, 'query', None)
                 if q:
--- a/PYME/LMVis/pipeline.py
+++ b/PYME/LMVis/pipeline.py
@@ -508,7 +508,7 @@ class Pipeline(object):
         Return a dictionary of datasources which are suitable for listing in a session file
         
         """
-        from PYME import warning
+        from PYME import warnings
         out = {}
         for k in self.recipe.inferred_data:
             ds = self.dataSources[k]
--- a/PYME/LMVis/sessionpaths.py
+++ b/PYME/LMVis/sessionpaths.py
@@ -34,7 +34,10 @@ def fnq_string(fn,query):
 def fnstring_relative(fnstring,sessiondir):
     fnamep = Path(fnstring)
     if fnamep.is_absolute():
-        fnamerel = str(fnamep.relative_to(sessiondir)) # now make the filenames relative to the session dir path
+        try:
+            fnamerel = str(fnamep.relative_to(sessiondir)) # now make the filenames relative to the session dir path
+        except ValueError: # value error implies it is not possible to make a "clean" relative path
+            return None # signal via None, we do the further handling in the calling function
     else:
         fnamerel = fnstring
     return fnamerel
@@ -56,20 +59,30 @@ def get_session_dirP(sessionpath):
     
 # note that we do not touch any paths that are already relative
 # this means we can repeatedly call this on session objects without doing any damage
+from PYME import warnings
 def make_session_paths_relative(session,sessionpath):
     sessionrel = session.copy()
     sessiondir = get_session_dirP(sessionpath)
     for ds in session['datasources']:
         fname,query = parse_fnq(session['datasources'][ds])
-        sessionrel['datasources'][ds] = fnq_string(fnstring_relative(fname,sessiondir),query)
+        relstring = fnstring_relative(fname,sessiondir)
+        if relstring is None:
+            warnings.warn("cannot construct relative path from %s to %s, giving up and using absolute paths in session record"
+                          % (fname,sessiondir))
+            return session
+        sessionrel['datasources'][ds] = fnq_string(relstring,query)
     # users can request certain recipe module arguments to be also "path-translated" by registering the module/arguments
     for checkmod in checkmodules:
         for module in sessionrel['recipe']:
             if checkmod in module:
-                # print("PYMEcs.MBMcorrection in module")
                 for field in checkmodules[checkmod]:    
                     if field in module[checkmod]:
-                        module[checkmod][field] = fnstring_relative(module[checkmod][field],sessiondir)
+                        relstring = fnstring_relative(module[checkmod][field],sessiondir)
+                        if relstring is None:
+                            warnings.warn("cannot construct relative path from %s to %s, giving up and using absolute paths in session record"
+                                          % (module[checkmod][field],sessiondir))
+                            return session
+                        module[checkmod][field] = fnstring_relative(relstring,sessiondir)
     return sessionrel
 
 # note that we do not touch any paths that are already absolute
@@ -84,7 +97,6 @@ def make_session_paths_absolute(session,sessionpath):
     for checkmod in checkmodules:
         for module in sessionabs['recipe']:
             if checkmod in module:
-                # print("PYMEcs.MBMcorrection in module")
                 for field in checkmodules[checkmod]:    
                     if field in module[checkmod]:
                         module[checkmod][field] = fnstring_absolute(module[checkmod][field],sessiondir)
--- a/PYME/LMVis/sessionpaths.py
+++ b/PYME/LMVis/sessionpaths.py
@@ -1,19 +1,25 @@
 from pathlib import Path
 
-# in future we could check all traits of type File?
+# the idea of this module is to provide two functions to make paths either relative or absolute in session objects
+# typically this will be relative to the '.pvs' session file path
+# it also provides a method register_path_modulechecks to allow the user/module coder to request path
+# translation for some recipe module arguments
+
+# a dict of recipe module names with arguments to check
 checkmodules = {}
 
-# it seems currently best to reauest this directly from the files in which the modules are defined
+# it seems best to request this directly in the python files in which the modules are defined
 # example:     register_modulecheck('PYMEcs.MBMcorrection','mbmfile','mbmsettings')
+# can this possibly be autoregistered by modifying the Traits.File object suitably?
 def register_path_modulechecks(module,*entries):
     checkmodules[module] = entries
 
-# split finename?query type string in fname and query parts
+# split filename?query type string in fname and query parts
 def parse_fnq(dsfnstr):
     parts = dsfnstr.split('?')
     parts0 = parts[0]
     if len(parts) > 1:
-        parts1 = parts[1] # we assume there are no ?s in the query string
+        parts1 = parts[1] # we assume there are no further '?'s in the query string
     else:
         parts1 = None
     return (parts0,parts1)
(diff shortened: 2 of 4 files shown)


PYME/LMVis/visCore.py
```diff
@@ -421,6 +422,26 @@ def OnViewInterpTriangles(self,event):
...
+        def ask(parent=None, message='', default_value=''):
+            with wx.TextEntryDialog(parent, message, value=default_value) as dlg:
+                dlg.ShowModal()
+                result = dlg.GetValue()
+            if result == '':
+                return None
+            else:
+                return result
+        
+        filename = wx.FileSelector("Choose a file to open",
```
csoeller, line 435: here we should probably add a wildcard flag with all reasonable extensions for which we can currently load datasources (in other words quite a few)...

David-Baddeley:
A few potential issue to consider here:

- we already do some relative path handling in the load logic (e.g. `PYME.IO.unifiedIO.local_or_temp_filename` and `PYME.IO.nameUtils.getFullExistingFilename`). I'd be a bit reluctant to add a further, independent, layer of indirection - both due the the added complexity and the fact that it might not play nicely with the existing path handling logic.
- As it stands, the proposed changes will not play nicely with cluster URIs or with the existing handling of paths relative to `PYMEDATADIR`.
- Re-writing the file input to selected recipe modules is super gross - especially requiring recipe modules to register in a seperate place for this to happen. We should really fix file inputs properly and uniformly rather than hack around this in sessions. In general file inputs are an aberration which limits portability of recipes, but it's hard to see an alternative for calibration data that might be used across several recipes. The right answer is probably something along the lines of doing string substitution in the `FileOrURI` trait using some form of recipe context, although making sure that calibration paths are always relative to `PYMEDATADIR` would be a potential short-term workaround. If we were to do re-writing (which I don't think we should), then scanning each module in the recipe for `FileOrURI` traits and globally re-writing those would be preferable to forcing recipe modules to register a hook. 
- Regardless of the final approach, I don't think we should make this subject to a config hook - we will just get mutually incompatible versions of the session files. 

All in all, it probably needs some further thought as to the best approach. To do this, I probably need a bit more context about what the general use cases for portable sessions are (and whether or not a lot of them could be solved by using, e.g. a network drive which is consistently mounted at the same place across computers).

PYME/LMVis/pipeline.py
```diff
@@ -509,12 +509,13 @@ def _get_session_datasources(self):
...
         """
         from PYME import warnings
+        from pathlib import Path
         out = {}
         for k in self.recipe.inferred_data:
             ds = self.dataSources[k]
 
             if hasattr(ds, 'filename'):
-                fn = ds.filename
+                fn = str(Path(ds.filename).resolve()) # ensure we return an absolute path!
```
David-Baddeley: This will mess with cluster URIs and paths relative to `PYMEDATADIR`

David-Baddeley:
Thinking a bit more, I wonder if this might be the expedient / pragmatic approach:

- don't modify (resolve) paths attached to datasources  (i.e. revert changes in `Pipeline._get_session_datasources`) to prevent issues with cluster URIs and `PYMEDATADIR` relative paths.
- In `VisCore.save_session()` check to see if **every** filename and every `FileOrURI` recipe trait is beneath the proposed session filename. If so, re-write as relative, and add a key to the session indicating that paths are relative.
- In `VisCore.load_session()` check for something like `if session.get('relative_paths', False):` and then rewrite all paths (including those in `FileOrURI` recipe traits) accordingly.

This should default to the current absolute (or non-portable) paths in all situations where it's not obviously clear that the session can be saved in a portable way. Maybe log whether you are saving a portable or non-portable session.

csoeller:
Thanks @David-Baddeley for your thoughts. I admit that cluster URIs and `PYMEDATADIR` had not featured in my thinking. I understand the general concern but just for my understanding, is there a realistic scenario where one would use sessions with cluster URIs and `PYMEDATADIR`? (even if there isn't an obvious one I am not arguing against the need to work with URIs etc to avoid surprises/hard to resolve user errors).

Note that in my code I wasn't changing the datasource filenames, just in `_get_session_datasources` I was resolving the paths that would be returned to build the session info.

I am therefore fully onboard with the proposed solution, we could even be one step more careful, if you think that could be useful, namely:

- don't modify (resolve) paths attached to datasources (i.e. revert changes in Pipeline._get_session_datasources) to prevent issues with cluster URIs and PYMEDATADIR relative paths.
- In VisCore.save_session() check to see if every filename and every FileOrURI recipe trait is beneath the proposed session filename. If so, re-write as relative, and add a key to the session indicating that paths are relative.
   - here we could instead write a session key with the full path to the session filename and leave all paths untouched as absolute
- In VisCore.load_session() check for something like if session.get('relative_paths', False): and then rewrite all paths (including those in FileOrURI recipe traits) accordingly.
   - instead, with paths left absolute in the written session but now a session key present that gives the full paths to the session file as it was written, we could do all checking upon loading, i.e.
      - if paths to datasources are not valid when trying to read, IF all existing paths are relative to (and below) the original file paths, then a final read attempt with all paths now translated relative to the *new* session file path could be attempted
      - if that also fails finally give up (with suitable error/warning message)

What do you think?

csoeller:
PS - the primary use case, and IMHO a valid one: the use of `.pvs` files and recipes with `FileOrURI` recipe traits is extermely annoying when we use OneDrive and DropBox shared folders for work on such files, as we do a lot in our lab.

In this everyday scenario in our use the absolute paths are all different (relative to local user directory) on different machines but the relative paths remain conserved across machines.

David-Baddeley:
I had initially thought of just saving the session path into the .pvs and working everything out on load failure (your suggested modification), but I don't think it's quite as clean ...

As I see it, the main advantage to doing the relative path translation on save is that you know if you have a portable .pvs before you try copy/moving and opening it. I also kind of like having the forced decision on relative or absolute paths rather than just falling back and trying to work out a relative path on failure as it makes it a bit clearer what the correct behaviour is. 

The one downside that I can think of is that we might need to insert some kind of token into the relative paths in the .pvs that we then replace on load i.e. saving as `$session_dir$/data_file.h5r` or `$session_dir$/some_calibration.json`. The upside of using the token approach is that it would make the loading super-easy - we'd  just need to do a `session_yaml.replace('$session_dir$', os.path.dirname(session_filename)` before we loaded it - we wouldn't strictly even need to check the `relative_paths` flag.

csoeller:
I like the session dir token approach quite a bit! Seems pretty clean to me and also human readable, if ever some manual editing were to occur.

csoeller:
With latest commits the PR should now be close to the approach we had discussed:

- pipeline changes reverted
- use a sessiondir token as we had discussed
- only rewrite with session dir token if all data sources files are below session dir
- upon reading a session file automatically translate any present session token
   - note: currently it does not explicitly check for the `session['relative_paths']` flag, but we do record it correctly
- it runs through the list of all registered modules to get a list of all possible `FileOrURI` recipe parameters
   - incidentally a couple of modules could not be instantiated without raising a `TraitsError`: `failedmodules
['measurement.IntensityAtPoints', 'tracking.TrackFeatures']`

The PR should be ready for some initial comments on this round of commits.

**Caveats:**

- it still contains a compatibility function to load session files ok that do not yet contain the session dir token. This will allow me to rewrite the ones I already have sitting around. Will be removed when done...
- a bit more testing will surely be useful

David-Baddeley:
I still don't like having modules have to register to get their paths changed - Is there a reason we need this?

PYME/LMVis/sessionpaths.py
```diff
@@ -0,0 +1,161 @@
...
+        try:
+            fnamerel = str(fnamep.relative_to(sessiondir)) # now make the filenames relative to the session dir path
+        except ValueError: # value error implies it is not possible to make a "clean" relative path, i.e. that the file fnstring is not truely below sessiondir
+            return None # signal via None, we do the further handling in the calling function
+    else:
+        fnamerel = fnstring
+    return fnamerel
+
+def allpaths_relative_to(session,sessiondir):
+    # currently we only check data sources, NOT recipe FileOrURI entries
```
David-Baddeley: We need to check FileOrURI too ....

PYME/LMVis/visCore.py
```diff
@@ -847,8 +870,15 @@ def OnSaveSession(self, event):
...
         import yaml
-        with open(filename, 'r') as f:
-            session = yaml.safe_load(f)
+        from pathlib import Path
+        from PYME.LMVis.sessionpaths import make_session_paths_absolute_compat, SESSIONDIR_TOKEN
+        fpath = Path(filename)
+        sessiondir = fpath.resolve().parent # note that filename could be relative, in that case need to resolve to obtain session dir
+        session = yaml.safe_load(fpath.read_text().replace(SESSIONDIR_TOKEN,str(sessiondir))) # replace any possibly present SESSIONDIR_TOKEN
+
+        # temporary compatibility call
```
David-Baddeley: Need to remove any compat stuff before merge

csoeller:
> I still don't like having modules have to register to get their paths changed - Is there a reason we need this?

No, this can go with the latest changes. It is not used anymore.

PYME/LMVis/sessionpaths.py
csoeller: Here it would be good to agree if one of the recipe FileOrURI arguments turning out to be a URI is reason enough to return False.

Am I incorrect in thinking that having a URI in a session to be saved is a bad idea in any case, paths relative or not?

David-Baddeley:
As it stands, you still walk all the loaded modules and register modules with `FileOrURI` traits - I think this would be cleaner if detection were done on the fly -  potentially assisted by a classmethod on the base recipe module class e.g.

```python
@classmethod
    def file_or_uri_traits(cls):
        """Return a list of FileOrURI traits, as these can require special handling (potentially warn if this list is not empty)"""
        return [k for k, v in cls.class_traits.items() if isinstance(v.trait_type, traits.FileOrURI)]
``` 

and in the session code:
```python
file_inputs = []
for mn, mod_dict in recipe:
    file_traits = base.all_modules[mn].file_or_uri_traits()
    file_inputs.extend([mod_dict[k] for k in file_traits]
```

PYME/LMVis/sessionpaths.py
David-Baddeley: I think any paths that are not relative mean that the session is not portable (and it's much safer to be strict here).

I don't see any major issue in having URIs in sessions however (they are portable by default, assuming other settings are correct). In an ideal world, every path in the session would probably be a URI rather than an filesystem path. This assumes that all your computers are set up to use the same cluster(s), but that is the design intent.

csoeller:
Al the main things should now be in as mentioned above; could possibly coded more cleverly in some of the loops but appears to do the job.

csoeller:
@David-Baddeley , all features we discussed should now be implemented. I have also run some tests with a notebook and some session manipulations. For the cases tested it seems ok. It should be good for another check if it is ready for merging now.

PYME/LMVis/sessionpaths.py
```diff
@@ -0,0 +1,86 @@
...
+    return is_safe
+
+def allpaths_relative_to(session,sessiondir):
+    for ds in session['datasources']:
+        fname,query = parse_fnq(session['datasources'][ds])
+        if not path_is_safe_and_relativeto(fname,sessiondir):
+            return False
+    for module in session['recipe']:
+        [(mn, mod_dict)] = module.items()
+        # we only check registered modules - session loading will anyway fail with an unregistered module
```
David-Baddeley: Is it really possible that there is an un-registered module in the recipe??? If that is the case, is it not better that the session saving fails?

csoeller:
Many thanks @David-Baddeley for review, further tweaks and merging. One issue is to be fixed with your merged PR, details in #1548. Would be cool if that could be merged (with suitable tweaks if required) so that it then works without raising an error.

PR #1543, opened 2024-08-18
image_bounds_superset

David-Baddeley:
Make pipeline imageBounds the spanning set of all datasources. Replaces and closes #1541.

### DIFF
--- a/PYME/IO/image.py
+++ b/PYME/IO/image.py
@@ -434,6 +434,9 @@ class ImageStack(object):
         self.mdh['ImageBounds.z0'] = value.z0
         self.mdh['ImageBounds.z1'] = value.z1
 
+    # PEP8 alias
+    image_bounds = imgBounds
+
     @property
     def metadata(self):
         return self.mdh
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -55,6 +55,8 @@ def deprecated_name(name):
     return _dec
 
 class TabularBase(object):
+    _image_bounds = False
+
     def toDataFrame(self, keys=None):
         warnings.warn('toDataFrame is deprecated, use to_pandas instead', DeprecationWarning)
         self.to_pandas(keys)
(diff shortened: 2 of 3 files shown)


csoeller:
I have provisionally tested it with a number of MINFLUX sessions, several datasets and the notebooks from [PYME-extra-sample-data](https://github.com/csoeller/PYME-extra-sample-data) and has worked without a hitch. With the image bounds working the `Home` icon also works again with sessions (I had been wondering why it did not seem to do anything).

csoeller:
There is at least one scenario where I can still make things fall over, you need to tell me if this is a "valid" scenario or not. If I add a module to move the coordinates sufficiently, e.g. I used a mapping with `{'x': 'x+1000', 'y': 'y+1000'}` to diagonally displace the data by micron. The image bounds do not budge and if I now try to render a region outside the nominal image bounds I get the familiar error:

```
File "/Users/csoe002/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.10-conda_1/python-microscopy-python-310-compat/PYME/LMVis/visHelpers.py", line 174, in rendGauss
    iy = np.absolute(Y - y[i]).argmin()
ValueError: attempt to get argmin of an empty sequence
```

The scenario seems not completely crazy to me as you might try to align two spearately acquired data segments etc.

David-Baddeley:
Hmm - this one is not super easy to solve ... For the alignment use case, if both segments are loaded you should be OK (bounding box will be expanded to cover second segment). Only solution I can think of would be to allow the mapping filter to ignore the metadata bounding box (although this should ideally only be under certain circumstances).

David-Baddeley:
@csoeller - I've added a hook for the mapping filter, which should fix this, but have it defaulting to `False` (specifically not added in the `Mapping` recipe module. The thinking here is that:
- in most cases, the alignment use case should have both data sources present, in which the superset of the bounding boxes applies.
- large shifts are probably best handled in their own recipe module

PR #1544, opened 2024-08-27
update note on ids peak install instructions for 2.10 and later ids peak

barentine:
Addresses issue IDS peak python wheel distribution has changed, and for later versions you can simply install from pypi.

enhancement
**Proposed changes:**
- update ids peak install instructions in comment block at top of ids_peak_cam.py


David-Baddeley:
Many thanks!

PR #1548, opened 2024-09-16
small fix: string conversion of session dir before use in replace method

csoeller:
Needed to avoid error with replace method call. Also consistently adopt `as_posix` conversion which should achieve file name portability.

### DIFF
--- a/PYME/LMVis/sessionpaths.py
+++ b/PYME/LMVis/sessionpaths.py
@@ -1,4 +1,4 @@
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 import os
 from PYME.recipes import base
 from PYME.IO.unifiedIO import is_cluster_uri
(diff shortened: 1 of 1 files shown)


csoeller:
Small fix applies to #1542 ...

PYME/LMVis/sessionpaths.py
```diff
@@ -120,16 +121,16 @@ def substitute_sessiondir(session_txt, session_filename):
...
+        path to the session file
 
     Returns
     -------
     str
         session_txt with SESSIONDIR_TOKEN replaced by sessiondir
     """
     from pathlib import Path
-    sessiondir = Path(session_filename).resolve().parent
+    sessiondir = Path(session_filename).resolve().parent.as_posix()
```
csoeller, line 133: This is the crucial fix, otherwise we get an error from replace as it cannot deal with the path object

David-Baddeley:
@csoeller - Presume this is already tested on windows, so saving posix paths there and not doing explicit sep translation on load is presumably OK? Inclined to merge and deal with any issues if and when they pop up ...

csoeller:
@David-Baddeley - yes, works fine on windows according to my tests. Will get more testing with the lab group on windows very soon, this should reveal any issues in potential edge cases...

Thanks for merging! I also noted that we do not currently allow opening session files through the GUI. Is this desirable or conversely an issue when using a `pymevis` instance that might already have open layers etc? Could be enabled through allowing `.pvs` in `File>Open` or could add `File>Load Session`...

Issue #1463, opened 2024-01-11
[ENHANCEMENT] Point-scan "camera" base classes


barentine:
**Is your feature request related to a problem? Please describe.**
I am building a point-scanning system, and we'll have at least one more being built shortly in the lab. My impulse is to implement something in PYME, which is of course expecting a camera these days. My temptation then is to make a point-scanning camera class. The key components are of course the scanning, reading a detector, and the synchronization between the two.

For especially slow stage scanning, which would satisfy my immediate need, I'm not above using offset piezo classes and raster scanning the offset. This would be controlled by a `StagePointScanner(Camera)` class which would also interface with a DAQ (at the moment, not above using an oscilloscope I'm already talking to over serial, but longer term I would like to use a proper daq). This will have to be pretty slow to avoid botching the synchronization, but as I mentioned, I'm quite OK with that for now.

Is there some long-term synergy here, @David-Baddeley ? I could see a case for having an easier GUI way to toggle continuous vs. software triggered readout modes in PYMEAcquire (i.e. a cam panel that stops/restarts the framewrangler so StartAq gets called and the camera actually changes modes) but otherwise I think we can present a stage scanning system as a camera fairly easily.

**Describe alternatives you've considered**
Other point-scanning acq frameworks exist, but I'd like to keep the integration with automated analysis / feedback capabilities we have in PYME

**Are you prepared to work on it?**
Certainly!

David-Baddeley:
Following on from the comment on #1551, I think any purely software point-scanning will hit limitations pretty quickly (due, amoungst other things, to the windows event loop and the fact that OS code is not fixed latency). The biggest issue is likely to be the variable latency, which in a naive implementation, is going to translate into variations in pixel dwell time (and corresponding irregularity in exposure and bleaching).

Because of this, i think it would be prudent to structure any confocal scanning camera code so that it is scalable to a hardware (or other fixed latency solution - e.g. real time OS or microprocessor). In my mind this looks something like the following:

### Hardware / real time firmware on, e.g. embedded microprocessor
- A master pixel clock is shared between position control and clocking a sample and hold / ADC for  readout
- Low latency position control via either analogue voltage (e.g. output from DAC card, arbitrary function generator, or clocking through a wavetable in the piezo controller) 
- Readout has clocked sample and hold (fixed dwell time), stores results in local memory (probably entire frame)
[...]

PR #1551, opened 2024-10-10
raster scan camera shim

barentine:
Along the lines of #1463 , I am currently working on a system with both a camera and a point detector. Ideally, I would like to be able to shim a stage-scanned image in as a `Camera`, so I can acquire images with either "camera" and/or run one from another instance for simultaneous imaging.

What I'm draft PR'ing for discussion here is a close-to-minimum working example for the approach. In my init script, I'm wrapping each axis of my 3-axis stage in a `base_piezo.SingleAxisWrapper` and then a `offsetPiezoREST.TargetOwningOffsetPiezo` so that I can leave the global microscope position the same, and scan about that position in a `RasterscanCameraShim.scan` call.

At the moment I've tried to set things up to accept a variety of "axes" to scan. Generally, I like the idea of using state handlers so we can scan more than just positions (i.e. scanning positions of a filter wheel would also be viable).

enhancement
**Proposed changes:**
- add voxelsignalprovider base class, a test version which simply records the scan positions at each location, and a dual-phase lock in version which logs both the X and Y channels at each scan position
- add a StageScanner class, with the idea of having it on somewhat equal footing with a mirror scanning approach at a later date
- add a `RasterScanCameraShim` class to mimic a camera while actually raster-scanning a point detector

My init script looks like the following:

<details> <summary> init script </summary>

```
    from PYME.Acquire.Hardware.Piezos import piezo_pipython_gcs, base_piezo
    from PYME.Acquire.Hardware.Piezos.offsetPiezoREST import TargetOwningOffsetPiezo

    nanocube_desc = 'PI E-727 Controller SN 0121089077'
    scope.stage = piezo_pipython_gcs.GCSPiezoThreaded(nanocube_desc, axes=('1', '2', '3'))
    x = base_piezo.SingleAxisWrapper(scope.stage, 0)
    x = TargetOwningOffsetPiezo(x)
    y = base_piezo.SingleAxisWrapper(scope.stage, 2)
    y = TargetOwningOffsetPiezo(y)
    z = base_piezo.SingleAxisWrapper(scope.stage, 1)
    z = TargetOwningOffsetPiezo(z)

    scope.register_piezo(x, 'x', needCamRestart=False, multiplier=-1)
    scope.register_piezo(y, 'y', needCamRestart=False, multiplier=1)
    scope.register_piezo(z, 'z', needCamRestart=False)
    scope.pifoc = z

    from PYME.Acquire.Hardware import raster_scan_shim
    scope.raster_scanner = raster_scan_shim.StageScanner(scope.positioning)
    
    from pymeasure.adapters import VISAAdapter
    from pymeasure.instruments.srs import sr844
    adapter = VISAAdapter("GPIB0::8::INSTR", write_termination='\n')
    scope.lockin = sr844.SR844(adapter=adapter)
    scope.lockin.reference_source = 'External'
    scope.lockin.time_constant = 0.1
    scope.lockin.sensitivity = 300e-6  # [V]
    scope.lockin.phase = 0

    scope.lockin_voxel_provider = raster_scan_shim.LockInSignalProvider(scope.lockin)
    scope.raster_cam = raster_scan_shim.RasterscanCameraShim(scope.raster_scanner, scope.lockin_voxel_provider,
                                        n_pixels_x=9, n_pixels_y=9,
                                        pixel_size_x=0.2, pixel_size_y=0.2)  # [um]

```

</details>


David-Baddeley:
Hi Andrew,

I'm really sorry I didn't respond to the original issue in a timely fashion - it came in when I was on holiday and then dropped out of my attention ... I'll follow up with more detail shortly (probably on the original issue), but the main thrust of my thoughts are as follows: 
- you'll hit the limitations of pure-software point-scanning pretty quickly, and it would be good to structure things so that the architecture continues to work even when you have more deterministic scanning (the simplest here would likely be to have position command voltages and sensor reads running from a constant clock in an DAC card, arduino, or something else with deterministic timing). It's very likely in this case that you would want to buffer an entire frame on the ADC card and read that buffer in one go.
- my gut feeling is that going through the scope.state mechanism might have some undesired wrinkles (notably automatic camera restarting when changing some state variables)
- there seems to be some duplication in the motion logic with the existing point_scanner class currently used for tiling (this was originally written to do half-assed confocal using a small ROI on camera and summing the pixels).

Best wishes,
David

PYME/Acquire/Hardware/raster_scan_camera.py
```diff
@@ -0,0 +1,245 @@
...
+            self._image_buffer[tuple(indices)]= signal
+        
+        return self._image_buffer
+    
+    def frame_rate(self):
+        return self._scan_params['n_pixels_x'] * self._scan_params['n_pixels_y'] \
+              * self._scan_params['n_pixels_z'] / self.signal_provider.pixel_dwell_time
+    
+    def GetIntegTime(self):
+        return self.scanner.frame_rate()
```
barentine: should be 1/frame rate

barentine:
Working towards a more sane/complete implementation. As it stands now the `BaseScanner` class outlines an interface for a mirror/stage scanner to hold a buffer of frames, and a `PointscanCamera` class acts like a ~normal camera class pulling frames from the `BaseScanner` instead of e.g. a camera FPGA.
There are two init_sim_X.py scripts in a `pointscan_shim` folder for testing/debugging.

Followed your suggestions, @David-Baddeley , for multichannel, signaling 1 frame for each channel to be read in by the frameWrangler. The Preview in PYMEAcquire is a bit odd, but for now this is a really nice improvement for me to be able to use the spooling/saving framework. 

Image

I modified PYME/Acquire/SpoolController.py and protocol_acquisition.py to check for the number of channels of the `PointscanCamera` and mofidy the shape accordingly. If there's one thing to review early / provide some feedback on, @David-Baddeley , it'd be nice to know how bad I hacked it there ;)

PYME/Acquire/SpoolController.py
```diff
@@ -552,7 +552,10 @@ def start_spooling(self, fn=None, settings={}, preflight_mode='interactive'):
             backend_kwargs['compression_settings'] = settings.get('pzf_compression_settings', self.pzf_compression_settings)
         elif self.spoolType == 'File':
             backend_kwargs['complevel'] = settings.get('hdf_compression_level', self.hdf_compression_level)
-        
+        # set dimensions and shape
```
David-Baddeley: This will need to be revisited to make sure it plays well with `XYZTCAcquisition` and the like. `XYZTCAcquisition` creates it's backend with `backend(..., shape=shape, dim_order=dim_order, **backend_kwargs)` meaning that you will get repeated `dim_order` and `shape`  keyword arguments to the backend and an error.

David-Baddeley:
@barentine - as noted above, SpoolController changes will need revisiting. Need to think a bit about best way to handle this, but is likely to depend on your exact requirements, specifically whether your colour acquisitions need to be `ProtocolAcquisition` instances (or subclasses), or if they would be better as an `XYZTCAcquisition` subclass. To help answer this there are 2 questions:

- Do you know the sequence length in advance?
- Do you need the protocol features (e.g. modifying laser powers during acquisition etc ...)?

If the answers are *yes* and *no* respectively it would likely be better to use `XYZTCAcquisition` as a base class. Using `ProtocolAcquisition` might also bring some interesting complexity linking events to frame numbers with the emitting multiple frames paradigm.

David-Baddeley:
If you still think that `ProtocolAcquisition` is the correct acquisition type, a short term workaround (although possibly still not a final one) would be to move the logic you have in `SpoolController` into `ProtocolAcquisition`. This would avoid the XYZTC breakage, but would not, e.g., let you do colour z stacks.

barentine:
Really appreciate the help, @David-Baddeley ! 

I don't have the best overview of the acquisition backends, so I appreciate your patience.

In answer to your questions, 
- at some point it would be nice to image until bleach without knowing ahead of time how long that will be. This is a less immediate need, and possibly not urgent to have undetermined series lengths at the same time as Z stepping
- I would like to use protocols for laser shuttering/unshuttering and possibly triggering an acquisition on an ~overview camera in a separate instance

It seems like maybe a reasonable thing to do is move logic out of `SpoolController` into `ProtocolAcquisition`, at least for short term, because I'm comfortable with protocols and might want to use them.

I think I see now how to register my own acquisition type for the subclass  `XYZTCAcquisition`  approach (will try and remember to call some of this out in configuring PYMEAcquire doc at some point). That said, it seems odd for me to subclass just to use an argument, `channel_settings` that already exists in  `XYZTCAcquisition`, but as far as I can tell doesn't get used anywhere. How/where should one be presenting `channel_settings` so they'll get used appropriately?

David-Baddeley:
For now `channel_settings` is just a stub, but we should really flesh it out a bit. It's stubbed in the base class to allow the maximum flexibility for how channels could be implemented, and I haven't put any effort into fleshing it out so far. There are at least 4 different legitimate ways to implement multi-channel acquisitions:
- Monochrome camera + filter wheel and/or control of lasers / LEDs
- Colour camera
- A combination of the above two
- Something else entirely (e.g. OIDIC phase shifts - if you still have access to the BewersdorfLab github, there is an example XYZTC subclass in the pyme-oidic repo).

In the reasonably short term we should probably add a channel implementation to the default XYZTC acquisition, but maybe by making a subclass the default that gets used for z stacks etc  rather than modifying the base.

Issue #1545, opened 2024-08-28
[BUG] ids peak support for uEye-only cameras issues


barentine:
**Describe the bug**
As discussed elsewhere, U3 and up cameras only support IDS peak api, while the older uEye cameras can still operate through a shim. I'm finally getting around to trying a uEye camera with the IDS peak API, and have come across a couple of issues.

Certianly, the easiest way to get around this issue is just to revert back to using the pyuEye implementations for older cameras (SDK stuff is still included in the ids peak installation, just pip install pyueye and use PYME `ueye.UEyeCamera`), but in the hopes of one day maintaining just one I wanted to document what I've run into.

1. 12 bit does not seem to be supported. 
2. You cannot set the framerate by changing the exposure time

12 bit does not seem to be supported. 
==========================
Our 12 bit handling on these cameras could ~always use a little more clarity I think. It looks like we go with the route of packing 12 bit into padded 16bit. The IDS peak api comes with an IPL part of the package with some image format query / buffer allocation / conversion helper functions we might enjoy using at some point in the future. Meanwhile, if I try to run a 12 bit uEye camera (UI327xCP-M) with the 12 bit initialization, I get the lower half of each frame returned as zeros, and the top part of each frame duplicated. Obviously we're missing some pixels with this, since it thinks the size is correct. 
![image](https://github.com/user-attachments/assets/1fbdbdcf-07c6-4ecc-bb84-5082d7938756)
If I try to use the IDS peak queries to ask for the pixel format, I get 'Custom' regardless of whether we are in 12 bit or 8 bit mode. Additionally, the 'payloadsize', height and width suggest they are both just 8 bit. Running in 8 bit mode works 'fine'
```
>>> buffer = scope.cam._data_stream.WaitForFinishedBuffer(100)
>>> buffer.PixelFormat()
17301505
>>> from ids_peak import ids_peak as peak
>>> peak.PixelFormatNamespaceEnumEntryToString(buffer.PixelFormat())
'Custom'

```

You cannot set the framerate by changing the exposure time
==========================

On startup, my camera is getting limited to 39.9 ms integration time, or 25 FPS. I can't set it to a higher integration time, and this is in fact enforced by the IDS peak DLL / camera. In order to adjust the integration time higher, I have to first set the AcquisitionFrameRate higher and then I can freely adjust the intregration time up to the max it should be for that framerate. On second look, this is the same as the uEye API, I just missed it when setting up the ids peak thing.
```
[...]

PR #1552, opened 2024-10-18
ids peak camera bug fixes & adding software triggering

barentine:
Addresses issue #1545 

- enhancement: adding software triggering mode (and optionless hardware triggering)
- bugfix: tell the camera what pixel mode we want
- bugfix: set integration time and then framerate in `SetIntegTime`
**Proposed changes:**

12 bit working OK, including at saturation.

image

image

Might at some point consider allowing packed formats if we need to speed up data transfer.
Will likely follow-up with some hardware triggering option methods e.g. rising/falling edge, delay, etc.


David-Baddeley:
Thanks!

PR #1553, opened 2024-10-23
Enable loading of SMAP files

zacsimile:
[SMAP](https://github.com/jries/SMAP) has these nested .mat files 

image (47)

that `scipy` does not load, due to their format being too new:

```
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/wx/core.py", line 2346, in Notify
    self.notify()
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/wx/core.py", line 3552, in Notify
    self.result = self.callable(*self.args, **self.kwargs)
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 881, in OpenFile
  File "/Users/zachcm/Code/python-microscopy/PYME/LMVis/visCore.py", line 739, in _populate_open_args
    mf = loadmat(filename)
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/scipy/io/matlab/_mio.py", line 226, in loadmat
    MR, _ = mat_reader_factory(f, **kwargs)
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/scipy/io/matlab/_mio.py", line 80, in mat_reader_factory
    raise NotImplementedError('Please use HDF reader for matlab v7.3 '
NotImplementedError: Please use HDF reader for matlab v7.3 files, e.g. h5py
```

This is a first-thought approach to get these files to load into PYME. I think there's a better way to do it (e.g., even using `tables` so as to not introduce another (optional) dependency). Do you have any thoughts on a preferred approach, @David-Baddeley?

### DIFF
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -672,9 +672,18 @@ class MatfileSource(TabularBase):
         for the position data and it's error should ensure that this functions
         with the visualisation backends"""
 
-        import scipy.io
+        try:
+            import scipy.io
 
-        self.res = scipy.io.loadmat(filename)[varName].astype('f4')  # TODO: evaluate why these are cast as floats
+            mf = scipy.io.loadmat(filename)
+        except NotImplementedError:
+            # we have most likely tried to open a MATLAB >7.2 file.
+            import h5py
+
+            # this is SMAP-specific
+            mf = h5py.File(filename)['saveloc']['loc']
+
+        self.res = mf[varName].astype('f4')  # TODO: evaluate why these are cast as floats
         self.res = np.rec.fromarrays(self.res.T, dtype={'names' : columnnames,  'formats' :  ['f4' for i in range(len(columnnames))]})
 
         self._keys = list(columnnames)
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Hi Zach,

a few thoughts ...

- how fixed is the SMAP format (does it always have the same 3 structs? Does saveloc['locs'] have consistent column names?)
- do SMAP files use the .mat extension, or do they customise the extension (so we could easily detect them)?
- the SMAP logic doesn't need to be in both `MatfileSource` and `MatfileColumnSource` (from a quick look it would seem that `MatfileSource` is the closest in expected format).
- we can't hard-coding `['saveloc']['loc']` in the normal mat file loader, as this will break support for all non SMAP matlab formats saved by recent matlab versions (e.g. 4PiSMS data). It would be better to detect SMAP files (e.g. by looking at the struct names) and defer to a custom SMAP loader. This would also let us use a preset set of key translations (if applicable) and make loading more streamlined.
- That said, handling other .mat formats saved from recent Matlab would be good.
- Agree on using pytables rather than h5py to minimise external dependencies (if this works - would need to test), although might not be the end of the world if we need h5py as I think it's a pandas dependency anyway.
- We might be able to use `tabular.HDFSource` directly (or potentially subclass with SMAP specific transkeys), rather than loading into a recarray (might need modification to support nested names, but that should hopefully be simple if not already supported). This would have the advantage of supporting lazy loading for really large datasets, and avoiding a couple of memory copies / allocations, as well as code duplication. 

If you have a sample file I'd be happy to look at seeing if `HDFSource` would work.

Best wishes,
David

zacsimile:
Hi David. Thanks for your reply. I think providing support for new MATLAB files sources through `tabular.HDFSource` is probably the best move. Since I PRed this, I discovered a way to export `.csv` files from SMAP, and I'm now asking my collaborator to use this approach. It's much easier and requires no code changes.

zacsimile:
Closing as the CSV export from SMAP/import to PYME works.

PR #1559, opened 2024-11-26
fix backend_kwargs bug

David-Baddeley:
Tentative fix for error below:

image

@Yujin-Bao

### DIFF
--- a/PYME/Acquire/Hardware/AndorIXon/AndorIXon.py
+++ b/PYME/Acquire/Hardware/AndorIXon/AndorIXon.py
@@ -456,7 +456,7 @@ class iXonCamera(Camera):
         self._GetAcqTimings()
         self._GetBufferSize()
 
-        eventLog.logEvent('StartAq', '')
+        self._log_exposure_start()
         ret = ac.StartAcquisition()
         if not ret == ac.DRV_SUCCESS:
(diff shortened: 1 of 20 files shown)


Yujin-Bao:
Hi David, thanks for the fix! However, I got exactly the same `TypeError` on my side.

barentine:
Chiming in from my phone here quickly - The call happens on both sides of an if/else, and is currently broken in both places. There are many args passed by keyword, not just complevel, all of which will currently throw an error (e.g. 'shape'). Most (all?) of the explicit arg passing by keywords here is just passing in what happens to already be the default of the constructor anyway, so they are not needed (or they are about to get passed through a third time by the **kwargs, can't remember off the top of my head. Or both). 

On my machine, for the side of the control flow I was on, I just removed everything except for the new **kwargs pass.

David-Baddeley:
Should hopefully have hit all the potential `backend_kwargs` duplications now (at least things work fine in the simulator)

Yujin-Bao:
Thanks David and Andrew! It now works well!

David-Baddeley:
@Yujin-Bao - now supports acquisition of tiled z stacks (as a prelude to, and template for tiled OIDIC), with a few TODOs ...

- Analysis (reconstruction) needs to be re-worked - currently untouched and just gives you a mean projection of all z levels
- Might want to add events for the z moves to make reconstruction easier / more general ....

PR #1561, opened 2025-01-13
Tiling fixes yb

Yujin-Bao:
Addresses PR#1599.

This is bugfix and enhancement.

**Proposed changes:**
Some stuff related to PR#1559. It has not been finished yet and still needs some work.


Yujin-Bao:
Hi David @David-Baddeley , I made some changes according to our discussion. There are two remaining issues that might need a second thought.

 - Linking xy stage to the child process by proxy is still not working well. The second axis linked to the client will overwrite the first axis so only one axis is moving although in the child process it shows they are both moving together.
 - Running OI-DIC tiling with z stacking through `RemoteSpoolSeries` in queued actions gives the following error when clicking `Add`:
 ```
Error whilst running Adding action
==============================================
python-microscopy=24.11.14.post0.dev[git]
python=3.7.11, platform=win32
numpy=1.21.6, wx=4.1.1 msw (phoenix) wxWidgets 3.1.5

Traceback
=========
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\site-packages\requests\models.py", line 960, in json
    return complexjson.loads(self.content.decode(encoding), **kwargs)
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\site-packages\simplejson\__init__.py", line 525, in loads
    return _default_decoder.decode(s)
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\site-packages\simplejson\decoder.py", line 370, in decode
    obj, end = self.raw_decode(s)
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\site-packages\simplejson\decoder.py", line 400, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

  File "c:\users\scope_baby\code\python-microscopy\PYME\ui\progress.py", line 153, in func
    return fcn(*args, **kwargs)
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\ui\actionUI.py", line 818, in OnAddActions
    actions = self._pan_action.get_actions()
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\ui\actionUI.py", line 185, in get_actions
    return [self.get_action(i) for i in range(self.num_actions)]
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\ui\actionUI.py", line 185, in <listcomp>
    return [self.get_action(i) for i in range(self.num_actions)]
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\ui\actionUI.py", line 177, in get_action
    action = self._get_action(idx)
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\ui\actionUI.py", line 614, in _get_action
    settings = getattr(self.scope, remote_instance).spooling_info()['settings']
  File "c:\users\scope_baby\code\python-microscopy\PYME\Acquire\acquire_client.py", line 72, in spooling_info
    return requests.get(self.base_url + '/spool_controller/info').json()
  File "C:\Users\scope_baby\miniconda3\envs\pyme\lib\site-packages\requests\models.py", line 968, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

Other than those, everything is working well! Thank you very much for your suggestions!

Best,
Yujin

File: PYME/Acquire/xyztc.py [hunk 1]
```diff
@@ -92,7 +110,8 @@ def __init__(self, scope, dim_order='XYCZT', stack_settings=None, time_settings=
         if channel_settings is None:
             self.shape_c = 1
         else:
-            self.shape_c = channel_settings.get('num_channels', 1)
+            #self.shape_c = channel_settings.get('num_channels', 1)
+            self.shape_c = getattr(channel_settings, 'num_channels', 1)
```
barentine, line 114: @Yujin-Bao  and @David-Baddeley  , `channel_settings` is passed as an argument here in the base class and not stored as an attribute, so I'm finding this change a bit odd.

Was there meant to be a `self.channel_settings = channel_settings` on a previous line in this `__init__`? Or might it make sense to revert this change and patch the subclass that motivated it?

David-Baddeley: No, this makes sense (at least in the absense of any prior definition of what `channel_settings` should look like. It's a `getattr` on the `channel_settings` object, not on the class itself. What we do need is a base class for `channel_settings` to document the interface. The reason for changing from a dict like access for `'num_channels'` in `channel_settings` to an attribute access was so that it could be a computed property rather than a fixed value.

@barentine - for your multi-colour camera use case all we would need is to define a channel settings like this:

```python
class ColourCameraChannelSettings(object):
    @property
    def num_channels(self):
        return scope.cam.num_channels

```

(or similar)

Alternatively, use a `named_tuple` for a static `num_channels`

David-Baddeley: We only need to save `channel_settings` if we will use it later in `set_c()` - seeing as the base class doesn't do anything with `channel_settings` there is no need to remember it.

barentine: Ah, OK. Thanks for the explanation and pointer, @David-Baddeley !

PR #1566, opened 2025-01-27
hoperfully fix flashing in queued action panel

David-Baddeley:
An attempt to fix flashing in the ActionUI on windows. Does this by only updating GUI if acquisition settings have changed.


David-Baddeley:
@Yujin-Bao

barentine:
Thanks for this, @David-Baddeley !

Issue #1567, opened 2025-01-31
[BUG] 'DictFloatEditList' object has no attribute 'makeColumnEditable'


csoeller:
**Describe the bug**

More recent versions of wx do not seem to support the method `makeColumnEditable` for wx.lib.mixins any longer.

```
  File "/Users/csoe002/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.10-conda_1/python-microscopy-python-310-compat/PYME/ui/custom_traits_editors.py", line 363, in __init__

    self.makeColumnEditable(1)

AttributeError: 'DictFloatEditList' object has no attribute 'makeColumnEditable'
```

**To Reproduce**
1. Open visgui and use module that relies on `DictFloatEditList` custom traits editor.
4. See error (also then fails to show GUI representation of modules in column, see picture below)

**Expected behavior**
Not trigger error.

**Screenshots**

<img width="722" alt="Image" src="https://github.com/user-attachments/assets/9bd2e9c3-6c1b-4f61-aae5-d08a436600f8" />

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
Latest python-microscopy master and Python 3.10.

**Additional context**

csoeller:
Fix merged with #1568, although fix somewhat temporary in nature

PR #1568, opened 2025-01-31
work around missing makeColumnEditable method in custom traits editor

csoeller:
Attempts to fix bug #1567.

Issue in custom_traists_editors that is triggered by method not being available in some cases. Just wrapping it into a try-block seems to fix it (but may need checking if anything else needs to be done for full functionality of the traits editor).

### DIFF
--- a/PYME/ui/custom_traits_editors.py
+++ b/PYME/ui/custom_traits_editors.py
@@ -360,7 +360,10 @@ class DictFloatEditList(editList.EditListCtrl):
         self.InsertColumn(0, 'Parameter')
         self.InsertColumn(1, 'Value')
 
-        self.makeColumnEditable(1)
+        try:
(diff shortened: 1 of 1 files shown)


PYME/ui/custom_traits_editors.py
```diff
@@ -360,7 +360,10 @@ def __init__(self, parent, value_dict, editor):
         self.InsertColumn(0, 'Parameter')
         self.InsertColumn(1, 'Value')
 
-        self.makeColumnEditable(1)
+        try:
+            self.makeColumnEditable(1)
```
David-Baddeley, line 367:
```suggestion
            # Temporary workaround for makeColumnEditible dissappearing in recent wx
            # TODO - find a permanent fix
            # TODO - prevent editing in column 0
            self.makeColumnEditable(1)
```

PR #1569, opened 2025-02-05
Channel aware tabular data

David-Baddeley:
As it stands, colour/channel handling in PYMEVis feels a bit like black magic. To do anything on a per-channel basis you currently need to create a `ColourFilter` on top of the current data source. This is generally obfuscated behind the UI, and it's not immediately clear (unless you already know) what to do if you want to create a channel aware recipe module.

The status-quo is largely an artefact of the original fixed pipeline and is optimised for ratiometric approach to multi-colour. To accomodate a wider range of use cases (and to make recipes a bit clearer) the best option is likely to give all tabular sources some minimal channel awareness. This is a test implementation of what that could look like. 

At this point very much a work in progress and could change significantly, but putting it up as a draft to solicit comments/opinions.

### Key points

- every tabular data source now has a `channel_names` attribute (returns an empty list for non-channel data), and a `get_channel_ds(channel_name)` method which returns a single channel tabular source with just that channel.
- you can access a particular channel of a datasource within the recipe namespace / `pipeline.dataSources` by using a key with the channel name appended after a dot e.g. `pipeline.dataSources['filtered_localizations.Chan0']`

### DIFF
--- a/PYME/IO/tabular.py
+++ b/PYME/IO/tabular.py
@@ -54,7 +54,7 @@ def deprecated_name(name):
     
     return _dec
 
-class TabularBase(object):
+class _TabularBase(object):
     _image_bounds = False
 
     def toDataFrame(self, keys=None):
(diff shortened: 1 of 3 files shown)


David-Baddeley:
### TODOs

- make it so that you can use channel suffixes in recipes
- extend channel suffixes to image data (This would remove the need for the `ExtractChannel` recipe module)
- make available suffixes show up in data source selection uI (PYMEVis)
- use something better than looking for `p_XXX` keys to determine channel names (metadata ??)
- [maybe] remove some of the legacy `ColourFilter` stuff
- [maybe] support integer suffixes in addition to descriptive channel names

barentine:
Hi @David-Baddeley - any progress on this? I think it would be nice, and help consolidate some of the multiview / splitter stuff (at least for shiftmapping).

Do I understand correctly that we're looking to suffix such that 'x_chan0' would pull 'x' with 'p_chan0'=1? 
I assume this would be best implemented with separate handling of channel identity post-assignment (tabular base, or subclass?), and continue to use ProcessColour to do any ratiometric assignment?

PR #1573, opened 2025-02-23
backwards compatibel fixes for traits.api in traits 7.X onwards

csoeller:
Allows use of latest traits package releases version 7.x.

7.x has removed a number of deprecated interfaces, notably `HasTraits.get` and `HasTraits.set`, as well as some of the `ListXXX` and `DictXXX` traits.

I semiregularly build complete PYME installs from `conda-forge` using `PYME-test-env` and during latest builds `traits` 7 gets used without explicit pinning. This PR seems to get everything working fine according to my testing so far. The proposed changes seem preferable over pinning `traits` to pre-7.

Compatibility "bugfix" for recent traits package versions.

**Proposed changes:**

- we wrap traits `set` and `get` calls in a try block to fall back to the replacement `trait_get` and `trait_set` methods. 
- define `DictStrFloat`, `ListFloat` etc in `PYME.recipes.traits`.

`trait_get` etc may have been available since traits 4.X, we could alternatively just change the `.get` and `.set` calls to their `trait_x` alternatives without explicit try blocks.

**Checklist:**
Tested with latest git source and `traits` 6.x and also `traits` 7.0.

### DIFF
--- a/PYME/recipes/traits.py
+++ b/PYME/recipes/traits.py
@@ -12,6 +12,7 @@ except ImportError:
 
 # definitions below have been adapted based on https://docs.enthought.com/traits-futures/0.2/_modules/traits/trait_types.html
 # these Dict have been deprecated for a while and need to be defined here for newer traits.api modules
+# FIXME - change usage within code and remove aliases
 DictStrAny = Dict(str, Any)
 DictStrStr = Dict(str, str)
 DictStrList = Dict(str, list)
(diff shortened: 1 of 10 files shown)


csoeller:
Modified patch to  just change the `.get` and `.set` calls to their `trait_[gs]et` alternatives without try blocks. Much simpler and `trait_[gs]et` have been used in other code places for quite a while already.

PYME/recipes/traits.py
```diff
@@ -10,6 +10,21 @@
 except ImportError:
     from traits.api import *
 
+# definitions below have been adapted based on https://docs.enthought.com/traits-futures/0.2/_modules/traits/trait_types.html
+# these Dict have been deprecated for a while and need to be defined here for newer traits.api modules
```
David-Baddeley, line 14: Do we actually use any of these? Would it be better to update out code where they are used?

PR #1580, opened 2025-03-07
pipython gcs piezo updates: analog control option and ontarget tolerance setting

barentine:
Addresses issues: 
- my need for external DAQ/FPGA driven fast sample scanning / signal IO  (see #1463) 
- unset ontarget tolerance in `piezo_pipython_gcs.GCSPiezoThreaded`, defaulting to numpy.isclose defaults which are rather small

one of each
**Proposed changes:**
1. add a method to change control for a given axis from serial to an ADC channel on the controller for analog control. 

> On my 3 axis stage, this lets me control two axes with analog lines from a DAQ while still using serial commands to move the remaining axis and poll positions of all 3 axes (useful because I do not have a coarse and fine stage, only the one stage to use for sample positioning and scanning both). There is an oddity or two remaining for control of some axes with serial commands and some with analog signals, but I imagine most folks will not be in the position I am in of using 1 stage/controller for so many tasks, and I am aware of the current limitations: `PiezoOnTarget` events will not generally fire or will fire on coincidence if any axis is in analog control, and extra move commands will be sent to the controller (and ignored for analog axes).

2. Add an initialization arg `piezo_pipython_gcs.GCSPiezoThreaded` setting a target tolerance for use in `numpy.isclose` as the `atol` parameter, so an on-target tolerance of e.g. 1 nm (new default) can be set, rather than 10^-8 micrometers. 

> This should save extra mov commands being sent presumably regularly by the polling thread (my bad) and potentially resulting `PiezoOnTarget` events. Users will need to set the tolerance to something reasonable for their application. I didn't want to make it too large given that the previous default was so minuscule, but 1 nm is likely ambitious for many stages. I have also set the relative tolerance, rtol to zero, even though it was already quite small, to prevent a position-dependant tolerance on large-travel stages. See [numpy.isclose](https://numpy.org/doc/1.25/reference/generated/numpy.isclose.html) , which comares a and b as |a - b| <= (atol + rtol * |b|)

@csoeller , does this look/sound OK to you? I wouldn't want the on-target tolerance to change without you knowing about it, though practically I don't think this will effect actual stage movement.

**Checklist:**
- [X] Does the PR avoid variable renaming in existing code, whitespace changes, and other forms of tidying? [There is a place for code tidying, but it makes reviewing 

- [X] Does this maintain backwards compatibility with old data?
- [X] Does this change the required dependencies?

### DIFF
--- a/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
@@ -397,6 +397,7 @@ class GCSPiezoThreaded(PiezoBase):
                     self._on_target = np.asarray([on_targets[axis] for axis in self.axes])
                     if not np.all(self._on_target == old_on_target):
                         # FIXME - something to log which axis would be cool?
+                        # FIXME - analog control of even 1 axis will essentially break this
                         logEvent('PiezoOnTarget', '%s' % self.positions, time.time())
                         self._all_on_target = np.all(self._on_target)
                     targets_matched = np.isclose(self.target_positions, self._last_target_positions,
@@ -444,6 +445,9 @@ class GCSPiezoThreaded(PiezoBase):
         
         Notes
         -----
+        OnTarget events are not aware of analog control, and will not
+        generally fire if even one axis is under analog control.
+
         Requires command level 1. which can be enabled with 
         self.pi.gcsdevice.CCL(1, password)
 
--- a/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
+++ b/PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
@@ -401,7 +401,7 @@ class GCSPiezoThreaded(PiezoBase):
                         logEvent('PiezoOnTarget', '%s' % self.positions, time.time())
                         self._all_on_target = np.all(self._on_target)
                     targets_matched = np.isclose(self.target_positions, self._last_target_positions,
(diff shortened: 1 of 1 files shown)


PYME/Acquire/Hardware/Piezos/piezo_pipython_gcs.py
```diff
@@ -32,6 +32,11 @@
 logger = logging.getLogger(__name__)
 
 
+ADC_CHANNEL_FOR_TARGET = 0x06000500
```
David-Baddeley: Are these specific to a particular piezo? If so, would it be better to have them as parameters to __init__ so that they can be set in the relevant hardware initialisation script for a particular setup?

barentine: Fair point, David. They are potentially controller-specific. Have gone ahead and added/tested a change so they're now passed in the init. Thanks for the suggestion!

barentine:
bumping this - look OK, David?

David-Baddeley:
Will merge, and can always revisit. My one remaining concern is whether you should be able to run with default parameters (As these all result in `SPA` calls, it implies that they are saved in the controller flash, and should have defaults already on the piezo - i.e. you might not need to set them, or, you could tweak them in PiMicrMove or similar, close, and then run PYME and have it work).

I'm not sure if it's a real or imagined concern but I have generally tried to minimise SPA calls as the manual at one point claimed a limited number of write cycles to flash.

barentine:
Thanks, @David-Baddeley !

Issue #1588, opened 2025-04-13
[BUG] blobfinding still uses scipy.X calls for numpy methods


csoeller:
**Describe the bug**
`blobfinding` appears to still have `scipy.XXX(...)` calls for numpy methods.

**To Reproduce**
1. Use blobfinding in `dh5view`.
2. See errors

**Bug message**

```
  self.ofd = ObjectIdentifier(np.atleast_3d(self.image.data[:,:,:, chnum]))

Traceback (most recent call last):

  File "/Users/XXXX/miniconda3/envs/test-pyme-3.11-conda_1/lib/python3.11/site-packages/scipy/__init__.py", line 137, in __getattr__

    return globals()[name]

           ~~~~~~~~~^^^^^^

KeyError: 'mgrid'



During handling of the above exception, another exception occurred:



Traceback (most recent call last):

  File "/Users/XXXX/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.11-conda_1/python-microscopy-working-jan-2025/PYME/DSView/modules/blobFinding.py", line 171, in OnFindObjects

    self.ofd.FindObjects(threshold, blurRadius = bs, blurRadiusZ = bs)

  File "/Users/XXXX/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.11-conda_1/python-microscopy-working-jan-2025/PYME/localization/ofind3d.py", line 206, in FindObjects

    X,Y,Z = scipy.mgrid[0:maskedFilteredData.shape[0], 0:maskedFilteredData.shape[1],0:maskedFilteredData.shape[2]]

            ^^^^^^^^^^^

  File "/Users/XXXX/miniconda3/envs/test-pyme-3.11-conda_1/lib/python3.11/site-packages/scipy/__init__.py", line 139, in __getattr__

    raise AttributeError(

AttributeError: Module 'scipy' has no attribute 'mgrid'

DEBUG:PYME.ui.AUIFrame:Creating fold panel
```

David-Baddeley:
should be fixed with merge of #1589

PR #1589, opened 2025-04-13
Ofind3d fix

csoeller:
Addresses issue #1588.

Bugfix.

**Proposed changes:**

Replace `scipy.X(...)` with corresponding numpy `np.X(...)` calls.
Also replace `print` calls with `logger.debug` calls.

**Checklist:**
Tested with latest PYME git master.


David-Baddeley:
Thanks!

Issue #1562, opened 2025-01-13
[BUG] `setup.py develop` fails for for `setuptools>=65`


zacsimile:
With newer versions of `setuptools`, I get the error

```
Traceback (most recent call last):
  File "setup.py", line 64, in <module>
    from numpy.distutils.core import setup
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/numpy/distutils/core.py", line 24, in <module>
    from numpy.distutils.command import config, config_compiler, \
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/numpy/distutils/command/config.py", line 19, in <module>
    from numpy.distutils.mingw32ccompiler import generate_manifest
  File "/Users/zachcm/miniforge3/envs/pyme/lib/python3.8/site-packages/numpy/distutils/mingw32ccompiler.py", line 28, in <module>
    from distutils.msvccompiler import get_build_version as get_build_msvc_version
ModuleNotFoundError: No module named 'distutils.msvccompiler'
```

There's a discussion of the source of this error at https://github.com/pypa/setuptools/pull/3505. 

Installing `setuptools<65` solves the issue. Perhaps we should pin this from now on?

David-Baddeley:
Thanks for spotting this (and working out the correct pin). Between this and the pending `numpy.distutils` deprecation we have a lot of pain headed our way ... I've made some early attempts to switch to `meson` (for now just on `pymecompress`, and not quite working), but that is likely the way forward.

David-Baddeley:
numpy reccomends pinning `setuptools<60.0` ...

https://numpy.org/doc/2.1/reference/distutils_status_migration.html

zacsimile:
Why `meson` over something higher-level like `poetry` or `uv`?

zacsimile:
Ah, never mind, you mean as an alternative to `setuptools`. Sorry.

David-Baddeley:
Considering meson as a replacement for `numpy.distutils`, rather than `setuptools`. It's the only thing which looks like it might be able to give a reasonably clean replacement for our hierarchal `setup.py` files, and have the possibility of compiling for runtime SIMD dispatch.

csoeller:
Not sure if relevant, I have been pinning to `setuptools<=73` for a while now and that seems to work fine for all purposes - even though it emits a good pile of warnings when using the "old" setup commands.

David-Baddeley:
[...]

PR #1590, opened 2025-04-13
Plugin factory - building on PR #1524

csoeller:
This is a debugged and ever so slightly tweaked version of PR #1524.

More like enhancement although mostly a smallish bugfix of PR #1524.

The plugin factory approach presented in PR #1524 seems compelling but would not run. This version has been debugged and very slightly tweaked using `argparse.BooleanOptionalAction` which requires a recent Python version (3.9+ I believe) but at this stage I think we demand versions in this range with latest PYME releases in any case.

The PR could alternatively have been raised against `David_Baddeley/python-microscopy` which is still entirely possible if preferred, important to mention that all credits should go to @David-Baddeley for the idea and implementation. Please provide guidance in the comments if that is the preferred route.

It seems sensible to get this PR integrated into the dist ahead of the bigger task of updating the PYME build system (see also #1562) so that it can be considered during that upgrade for generating plugins compatible with the new build system. 

**Proposed changes:**

#1524 was not running as is and with some testing everything now appears to work in this version.

Usage is

```python
python -m PYME.util.plugin_factory output_directory package_name
```
with a few possible options, i.e.

```
usage: plugin_factory.py [-h] [--pymeimage | --no-pymeimage] [--pymevis | --no-pymevis] [--recipe | --no-recipe]
                         [--fit_factories | --no-fit_factories] [--reports | --no-reports] [--dist | --no-dist]
                         output_dir package_name

Create a template for a PYME plugin

positional arguments:
  output_dir            Directory to write the plugin template to
  package_name          Name of the package

options:
  -h, --help            show this help message and exit
  --pymeimage, --no-pymeimage
                        Create a PYMEImage plugin with the given name
  --pymevis, --no-pymevis
                        Create a PYMEVisualize plugin with the given name
  --recipe, --no-recipe
                        Create a recipe module with the given name
  --fit_factories, --no-fit_factories
                        Create a fit factory with the given name
  --reports, --no-reports
                        Create a reports module with the given name
  --dist, --no-dist     setup to install plugins in dist mode (for other users)
```

**Checklist:**

Checked with recent git versions of PYME. Seems to generate and then install plugins fine with current PYME.

### DIFF
--- a/PYME/util/plugin_factory.py
+++ b/PYME/util/plugin_factory.py
@@ -1,0 +1,465 @@
+def create_pymeimage_plugin():
+    plugin_template = f'''\
+# This is a template for a PYMEImage UI plugin.
+from PYME.DSView.modules._base import Plugin
+
+class MyCoolPluginKlass(Plugin):
+    def __init__(self, dsviewer):
+        super().__init__(dsviewer)
+
+        # Add your code here
+
+        # Example: Add a menu item
+        dsviewer.AddMenuItem('Processing', 'Do something neat', self.OnDoSomethingNeat)
+
+    def OnDoSomethingNeat(self, event=None):
+        # Menu item callback
+        # Add your code here
+        pass
+    
+def Plug(dsviewer):
+    """ This function is called by PYMEImage to create an instance of the plugin """
+    return MyCoolPluginKlass(dsviewer)
+'''
+    return plugin_template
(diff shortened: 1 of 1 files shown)


csoeller:
This would need revisiting with the newer build guidelines for setuptools etc so closing for now...

Issue #1597, opened 2025-05-14
[BUG] default clipping cuts of data with negative coordinates in PYMEVis


csoeller:
**Describe the bug**
The clipping defaults seem to fail with data containing negative coordinates, see example below.

**To Reproduce**
Open dataset with negative coordinates in `PYMEVis`. See (or not see) clipped data.

**Screenshots**

Get this

![Image](https://github.com/user-attachments/assets/8334b077-600a-4b8f-a86a-776359591eda)

but should be

![Image](https://github.com/user-attachments/assets/d331db48-4b61-43ec-addc-e052db43604d)

**Patch**

The following patch seems to fix:

```patch
diff --git a/PYME/LMVis/view_clipping_pane.py b/PYME/LMVis/view_clipping_pane.py
index 09b014b..d63d86c 100644
--- a/PYME/LMVis/view_clipping_pane.py
+++ b/PYME/LMVis/view_clipping_pane.py
@@ -107,7 +107,7 @@ class ClippingPanel(wx.Panel):
         #print 'bbox: ' + repr(bb)
         
         if bb is None:
-            return [1e-9, 1e9]
+            return [-1e9, 1e9]
         
         
         if self.axis == 'x':
```

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: multiple OS

csoeller:
Fixed with merge #1598.

csoeller:
PS - @David-Baddeley questioned if negative coordinates suggest a data issue. Quite a few external imports can have negative coordinates, in our case data from MINFLUX acquisition often does, probably to do with where the coordinate origin is placed WRT to the scan.

PR #1598, opened 2025-05-14
fix to ensure the incitial clipping limits fit the data in PYMEVis

csoeller:
adresses #1597 

- in datasets with negative coordiantes current initial defaults often clip a part of the data
- this fix attempts to correct that


David-Baddeley:
Probably no harm in this change, although curious how often / when we end up with negative coordinates ... (are we overlooking a sign that something else is wrong?).

Issue #1527, opened 2024-07-08
[SUPPORT] How achieve compatibility with latest matplotlib?


csoeller:
Clearly, matplotlib, as most major Python packages, is a moving target. We noticed a first issue with 3.6.x which is quite easy to fix, see #1525 and #1526.

More recent matplotlib packages, i.e. 3.7.x and 3.8.x have a couple of changes that give rise to issues with `PYME/DSView/modules/graphViewPanel.py`. Notably,

 -  in matplotlib 3.7.X onwards `backend_wx.cursord dictionaries` are [removed](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.7.0.html#cursord-in-gtk-and-wx-backends);
 -  3.8.X [removes](https://matplotlib.org/stable/api/prev_api_changes/api_changes_3.7.0.html#cursord-in-gtk-and-wx-backends) `error_msg_wx` function in `backend_wx`

I am not very familiar what these do and how they are used.

**Question**: Does it make sense to make PYME compatible with more recent matplotlibs (I imagine it would)? If so, how can we best replace/alter the existing `graphViewPanel` implementation?

David-Baddeley:
We should certainly try to keep up with reasonably recent matplotlib.

A bunch of the stuff in `graphViewPanel` is copied from the internal code of an older version of wx (and hence uses a bunch of undocumented internal apis, which is probably why it has broken). We could update `graphViewPanel` so that it works with more recent versions, but might also want to consider if we need to continue support for `graphViewPanel` -  or if we might be able to make do with a lighter weight alternative which avoids re-implementing parts of the matplotlib internals. As far as I remember it's not widely used - potentially only with line profiles and spectrometers.

David-Baddeley:
Looking a little more, these are also used in the PSF and OTF plots, with the code being duplicated. The offending code is trying to encapsulate the toolbar functionality (including zoom etc ...) that you get in a standalone matplotlib plot window, but within our frame.

csoeller:
I have recently tested building PYME with python 3.11 and 3.12. At least conda-forge does not support `matplotlib<=3.6` conda install any more with these python versions (one could probably pip install 3.6, I imagine). This suggests to me we should try to loose the dependencies in graphViewPanel etc pretty soon to allow easing the matplotlib version constraints.

[...]

PR #1601, opened 2025-05-22
Numpy 2.X compatibility fixes

csoeller:
Addresses issue of compiling with numpy 2.x mentioned in #1527.

As noted in #1527, numpy 2.X gives compilation errors when attempting to compile several files which stem from subtle changes or dropped deprecated interfaces versus 1.X. The code changes here make PYME compile with latest numpy 2.X while maintaining recent 1.X backwards compatibility.

This PR implements the following changes:

- change all `PyArray_XXTYPE` to `NPY_XXTYPE`, e.g. `NPY_DOUBLE` etc; this appears backwards compatible with 1.X
- in `gohlke/tifffile.c` change access to `dtype->elsize` to `PyDataType_ELSIZE(dtype)` as outlined in https://numpy.org/doc/stable/numpy_2_0_migration_guide.html and include 1.X compatible ifdef'ed macro
- change occasional `ctypedef np.int_t ...` to `ctypedef np.int64_t ...`; apparently this is now required and seems backward compatible

With this everything builds fine as tested using `Python` 3.11, `numpy` 2.2.6 and `matplotlib` 3.10.3.
Also ok (backwards compat test) using `Python` 3.11, `numpy` 1.26.4 and `matplotlib` 3.8.0.

Using numpy 2.X and recent matplotlib (here 3.10.3) most things seem to work fine, I tested briefly `dh5view` line profile plotting which uses the `graphviewpanel` and even that seems fine but there may be matplotlib changes that trigger issues as yet unidentified. Similalrly, there are possible numpy 2.X python level interface changes lurking in some not yet tested part of PYME. At least one can now start testing this further.

bug fix for numpy 2.x compatibility

**Proposed changes:**

As outlined above.

Tested as mentioned above on mac to date.

### DIFF
--- a/PYME/contrib/gohlke/tifffile.c
+++ b/PYME/contrib/gohlke/tifffile.c
@@ -66,6 +66,11 @@ setup(name='_tifffile', ext_modules=[Extension('_tifffile', ['tifffile.c'],
 #include "string.h"
 #include "numpy/arrayobject.h"
 
+/* NPY 1.X backwards compatibility */
+#if NPY_ABI_VERSION < 0x02000000
+  #define PyDataType_ELSIZE(descr) ((descr)->elsize)
+#endif
+
 /* little endian by default */
 #ifndef MSB
 #define MSB 1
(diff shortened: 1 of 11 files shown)


PYME/contrib/gohlke/tifffile.c
David-Baddeley, line 1: Need to think about whether it still makes sense to keep / maintain our customised copy of tifffile. I think most, but potentially not all places in the code now use the main tifffile package if it's installed.

PR #1605, opened 2025-07-24
Change build to meson

David-Baddeley:
Change build from deprecated Numpy distutils to meson 

This should hopefully ensure forwards compatibility of code with new numpy and python versions


David-Baddeley:
Not quite working yet, due to issues with copying static directories

Issue #1205, opened 2022-02-16
[BUG] PZF PCS series loading through ClusterOfOne / 6D data model issue


barentine:
**Describe the bug**
```
filename == pyme-cluster:///Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs
DEBUG:PYME.IO.cluster_directory:Request for b'http://127.0.0.1:56151/Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs/' failed with error: 404
DEBUG:PYME.IO.clusterIO:Could not find Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs/metadata.json, retrying ...
DEBUG:PYME.IO.cluster_directory:Request for b'http://127.0.0.1:56151/Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs/' failed with error: 404
DEBUG:PYME.IO.clusterIO:Could not find Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs/metadata.json, retrying ...
DEBUG:PYME.IO.cluster_directory:Request for b'http://127.0.0.1:56151/Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs/' failed with error: 404
DEBUG:PYME.IO.clusterIO:could not find file Users/Andrew/PYMEData/Andrew/2022_2_16/000/16_2_series_A.pcs/metadata.json on cluster: cluster nodes = ['PYMEDataServer [Andrews-MacBook-Pro-5]:Andrews-MacBook-PID:3158']
Traceback (most recent call last):
  File "/opt/miniconda3/envs/pdev/lib/python3.7/site-packages/wx/core.py", line 3259, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
  File "/Users/Andrew/code/python-microscopy/PYME/DSView/dsviewer.py", line 478, in LoadData
    im = ImageStack(filename=args[0], queueURI=options.queueURI, mdh=options.metadata, haveGUI=True)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/image.py", line 242, in __init__
    self.Load(filename, prompt=load_prompt, haveGUI=self.haveGUI)
  File "/Users/Andrew/code/python-microscopy/PYME/IO/image.py", line 1222, in Load
    self._loadClusterPZF(filename)
[...]

PR #1609, opened 2025-10-22
Alternative fix to localization settings_form and form_recipe text escape

barentine:
Addresses an issue mentioned in #1205, where the "Launch analysis with selected" button does nothing, and results in
"Uncaught TypeError: item.getAttribute(...).text is not a function" on the browser side.
The `.text()` calls were introduced in response to security concerns (#1110), which it seems can alternatively be avoided by using jQuery instead of javascript to construct the list elements.

bugfix
**Proposed changes:**
- avoid typerror, and avoid passing the series name string as markup

**Checklist:**
I have tested the localization. Have not tested the recipe form fix.

### DIFF
--- a/PYME/cluster/clusterUI/localization/templates/localization/settings_form.html
+++ b/PYME/cluster/clusterUI/localization/templates/localization/settings_form.html
@@ -46,10 +46,15 @@
             $('.cu-hidden-seriesname').remove();
             var seriesNames=[];
             $('.cu-selected').each(function(index, item){
-                var sn = item.getAttribute('data-cu-seriesname').text();
+                var sn = item.getAttribute('data-cu-seriesname');
                 if (sn != null) {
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Hi Andrew,

will have to look at this. I think your change re-introduces the security
issue, but suspect the security issue is entirely hypothetical for the
typical usage (especially given where the values are coming from).

David

David-Baddeley:
OK, with the help of copilot I now understand why your fix is better. For the record, it's because the jquery way sets the DOM value element directly, and doesn't require parsing arbitrary HTML in order to set it. Will merge.

I'd previously got caught up on the fact that it's still not doing any sanitisation.

barentine:
Thanks @David-Baddeley! Glad to have you double check it (as you know HTML/etc is out of my usual wheelhouse)

Issue #1563, opened 2025-01-21
[ENHANCEMENT] PYMEAcquire UI improvements


barentine:
**Is your feature request related to a problem? Please describe.**
Running another instrument in PYME, used by an experienced microscopist other than me, who has correctly pointed out some frustrations with the user interface that we could stand to improve.

**Describe the solution you'd like**
- [x] Ability to type in a z position for the positioning UI (currently can type in X and Y values, but Z goes to a slider). Slider for Z is fine, but would be nice to be able to click-into the text value and enter a new position similar to the slider labels worked out for PYMEVis view cropping.
- [ ] UI for setting an end condition to an acquisition without using a protocol or the action queue. I.e., run a time series for 1,000 frames, or 100 seconds, rather than either doing a single-shot acq, or starting and manually clicking to stop a time series. 
- [ ] making ROI size and position adjustable numerically, somewhat like the PYMEImage `File > Export Cropped` UI, so that if a drawn/toggled ROI isn't quite what you want, you can adjust it without cracking open the shell. 
- [ ] if someone types in a series name, the next filename could be <custom name>_B ... _Z rather than going back to the default on the next series
- [ ] if you change aspect ratio / zoom with an ROI selected (closest thing to crosshairs we have by default), it would be nice to center the view on the selected region. 

Will possibly update to include other notes in the coming weeks

David-Baddeley:
Sorry for not commenting earlier - was away when this came in and it fell off the radar. Most of these sound sensible. 
 - **z position:** I'm sure this has had an associated text edit box at some point in the past (might have gone away in a de-cluttering effort). Shouldn't be too hard to add back, although I am slightly wary of additional clutter - the UI is already super busy. In the interim, might the `focusKeys` plugin (see the default simulator init file) be useful? Should the future text-entry be in the main UI, or accessible from a menu?
 - **max_frames:** With the revamped actions and analysis chains, it might make sense to bring max frames out into the *Time/Blinking series* controls and drop it from the *Queued Actions*. A fixed length XYZTC time series mode is also on the cards (potentially allowing timelapse - i.e. decoupling frame rate and integration time.
 - **Entering ROI coordinates**: Suggest putting this under the *Camera* menu, along with *Toggle ROI*, and pop up a dialog. Don't really think this needs to be shown all the time.
 - **Keeping manually entered filename stub:** Probably makes sense, but avoided to date for a couple of reasons:
[...]

PR #1612, opened 2025-11-06
Idle frameWrangler (pause camera when not spooling, stop retriggering single shot mode)

barentine:
Addresses issue #1563 , drafting this for discussion, but it is not complete.

enhancement
**Proposed changes:**
- add an `Idle` mode for the frameWrangler. For a camera in `MODE_CONTINUOUS` it will be stopped in idle mode, while cameras in `MODE_SINGLE_SHOT` will not be retriggered repeatedly.
- add a GUI component which can be added to cam control panels (e.g. `PYME/Acquire/Hardware/pco/pco_sdk_cam_control_panel.py`) with a "stop" button to enter idle, and a "play" button to exit idle. Pairs nicely with the acquisition mode selector such a cam panel might have.

while active:
image

while idle:
image

***Issues***

- [ ] starting spooling from idle mode does not work yet! 
- [ ] currently, the core logic for idle mode is just slapped into frameWrangler/spoolcontroller, rather than living in the microscope object as @David-Baddeley suggested in #1563 .
- [ ] play button doesn't hard restart frameWrangler / cameras. It works nicely if you haven't broken anything, but if you brick a camera with a bad argument and want a hard restart from the GUI, the play button might look like your answer, but it doesn't do what pa.stop, pa.Prepare, pa.start would do for you/camera. 

After spending some time on this, I think what's been highlighted for me is that consistency elsewhere would go a long way. For that, it might help to introduce the acquisition modes we expect to fakeCam.py, and then make sure that works the way we want it to with spooling, tiling, etc. Otherwise I think some cameras are hacking their own way / we're abusing some oddities to get what we want for certain tasks. The TODO @David-Baddeley added in the base class to add a mechanism for cameras to report what modes they support would also be great.

### DIFF
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -924,7 +924,12 @@ class Camera(object):
         bool
             True if in idle mode
         """
-        return self._idle
+        try:
+            return self._idle
+        except AttributeError:
+            logger.warning("Camera idle flag was not initialized - Did custom camera class call super().__init__() ?")
+            self._idle = False
+            return self._idle
 
     def Shutdown(self):
         """Shutdown and clean up the camera"""
--- a/PYME/Acquire/Hardware/Camera.py
+++ b/PYME/Acquire/Hardware/Camera.py
@@ -34,7 +34,7 @@ from PYME.Acquire import eventLog
 from PYME import config
 from . import camera_noise
 import warnings
-
+from PYME.contrib import dispatch
 import numpy as np
 import logging
 logger = logging.getLogger(__name__)
@@ -152,6 +152,10 @@ class Camera(object):
         # this is important so that the camera settings get recorded
         MetaDataHandler.provideStartMetadata.append(self.GenStartMetadata)
 
+        # create flag for idle state, allowing different behavior between spooling
+        self._idle = False
+        self.on_idle_change = dispatch.Signal(['idle'])
+
     def Init(self):
         """
         Optional intialization function. Also called from the init script.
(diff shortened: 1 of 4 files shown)


David-Baddeley:
I now wonder if idle mode logic belongs in the camera, not the frame wrangler.  For most conventional cameras, the status quo probably works well as idle mode. For a scanning camera, idle mode could be implemented as follows:

When a scanning camera is idle:
- `ExpReady()` returns `False`
- `StartExposure()` either no-ops or  throws an exception
- `getFrame()` throws an exception

Having `ExpReady()` return `False` basically shortcuts everything in the polling loop, so you never get to the point where you'd be triggering.

barentine:
Sounds reasonable to put the logic in the camera. The camera will still need to know whether spooling is live or not, but would be simple to do this via an optional `Camera.Idle` state handler that gets set in `SpoolController.start_spooling` and  `SpoolStopped`.

David-Baddeley:
Why would the camera need to know if spooling is live or not? Surely it would just need to know if it was supposed to be idle? I'm envisaging adding a `set_idle()` method to the camera base class that toggles a flag, and is called when starting and stopping spooling. Derived classes can choose to modify their behaviour `ExpReady()` etc behaviour according to the flag, but the default is just to ignore it..

David-Baddeley:
Actually, re-reading We're probably talking about much the same thing.

barentine:
> Actually, re-reading We're probably talking about much the same thing.

Was just going to say as much! Thanks @David-Baddeley

barentine:
any further review on this @David-Baddeley ?

barentine:
actually looks to be an issue with this PR as it sets - I think the `SpoolController` needs to change the idle state before `self.spooler.start` not after.

PR #1617, opened 2025-11-21
Meson

David-Baddeley:
trying to tweak meson based build


David-Baddeley:
@barentine @csoeller  @zacsimile 

Just tagging you to make you aware. Fairly major change to installation process. Now building against python 3.9 to 3.13, and generating pip packages rather than conda packages.

Conda might return once I get this working, but if it does the conda recipe will likely be along the lines of just getting conda to pip-install the package.

PR #1620, opened 2025-12-17
Fix for pco camera crashing/capturing black frames/being stuck

Yujin-Bao:
Addresses issue of pco edge 4.2 LT used on the OI-DIC microscope.

This is a bugfix.

**Proposed changes:**
Recycle buffer when encoutering timeouts or other errors. Flag errors in `pco_sdk_cam.py` for `framewrangler.py` to restart the camera. Add a logging to report which hardware is not ready, which is useful for debugging what causes the delay in the single-shot mode. There might be some redundancy in the type of errors might be encountered because I only have the timeout error in my logging based on my test, but I still keep the code as is to be safe.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
@@ -20,6 +20,7 @@ import ctypes.wintypes
 import queue
 import threading
 import time
+import copy
 
 k32_dll = ctypes.windll.kernel32  # lets us use the recommended WaitForSingleObject call (see pco.sdk)
                                   # instead of the not-recommended-for-polling pco_sdk.get_buffer_status()
@@ -580,7 +581,7 @@ class PcoSdkCam(Camera):
             res = pco_sdk.force_trigger(self._handle)
             # FIFO queue the queable buffers so we don't
             # grab images before a trigger in _poll_loop
-            self._buffers_to_queue.put(self._i)
+            self._buffers_to_queue.put(copy.deepcopy(self._i)) # ensure we put a copy of the index, not a reference
             self._i += 1
             if self._i >= self._n_buffers:
                 self._i = 0
--- a/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
+++ b/PYME/Acquire/Hardware/pco/pco_sdk_cam.py
@@ -519,9 +519,12 @@ class PcoSdkCam(Camera):
 
     @property
     def _timeout(self):
-        # set the time longer for all hardware check
-        # for example, LC settling time is usually set at 300 ms (current minimum is 150 ms)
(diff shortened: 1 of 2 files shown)


PYME/Acquire/Hardware/pco/pco_sdk_cam.py
```diff
@@ -433,7 +489,9 @@ def _init_buffers(self):
 
     @property
     def _timeout(self):
-        return int(max(2*100*self.GetCycleTime(), 100))
+        # set the time longer for all hardware check
```
David-Baddeley: This timeout should be specific to the camera (other hardware and hardware checks are completely unimportant here). It's used in the tight loop which just serves the camera. Setting this too long is likely to be undesirable (makes it more likely that the camera buffer is not kept full).

PYME/Acquire/Hardware/pco/pco_sdk_cam.py
```diff
@@ -155,24 +161,74 @@ def _poll_loop(self):
...
+                            self._consecutive_errors = 0
+
+                            k32_dll.ResetEvent(self._buf_event[_curr_buf])
+
+                            status = self._buffer_status[_curr_buf]
+                            if status != 0:
+                                # Bad buffer -> recycle it, don't deliver
+                                logger.warning(f"PCO: error {status} during check of buffer {_curr_buf}. Recycling.")
+                                self._consecutive_errors += 1
+                                self._buffers_to_queue.put(_curr_buf)
```
David-Baddeley: We might not want to re-queue bad buffers unless we are in continuous mode  (if we want to match the current logic, i.e.:

For non-continuous modes: 
- A single buffer is queued in TriggerAq
- We wait on this buffer, and return once full.

Re-queueing buffers that are bad risks letting the acquisition get out of sync with the triggers (the re-queued buffer might end up waiting on the next trigger)

zacsimile:
FWIW, I recently used the pco. code as-is (prior to this PR) on an edge 4.2 LT and it worked perfectly. All of the timeouts/transfer issues we experienced ended up being due to the cable connecting the camera to the computer. We tried several off the shelf options that in theory matched the required spec with no success. We finally ordered a far too expensive cable directly from Excelitas and the camera buffering worked.

Yujin-Bao:
Thank you @zacsimile ! That's good to know swithing a cable solved the problem. Honestly I'm a bit surprised only their expensive cables work as I also tried troubleshooting some hardware before this PR but wasn't able to completely solve the issue. I updated the pco usb 3.0 driver and the pco firmware and inserted a PCIe board to provide a usb 3.0 controller only for the camera. This solved the crashing issue but there are still some timeouts and black frames. I therefore tried tweeking the software.

Issue #1619, opened 2025-12-11
[BUG] meson editable build issue


csoeller:
**Describe the bug**
Using the new meson install, when selecting a development install (`pip install --no-deps --no-build-isolation --editable .`) `dh5view` and similar commands that involve loading DSView modules fail, see error below.

**To Reproduce**
1. Install latest PYME with meson as editable install
2. open image with `dh5view` or similar
3. see error

**Reason and possible fix**

This can be traced to the way the editable install works which rewrites the `__path__` module variable in non-obvious ways. It can be worked around with a patch as below. This is also backwards compatible with pre-meson installs and using Python>=3.10 in some quick tests that I did.

```patch
diff --git a/PYME/DSView/modules/__init__.py b/PYME/DSView/modules/__init__.py
index c688dd9..cb2460a 100644
--- a/PYME/DSView/modules/__init__.py
+++ b/PYME/DSView/modules/__init__.py
@@ -33,7 +33,17 @@ from PYME import config
 import logging
 logger = logging.getLogger(__name__)
 
-localmodules = [os.path.splitext(os.path.split(p)[-1])[0] for p in glob.glob(__path__[0] + '/[a-zA-Z]*.py')]
+# this fails with meson "develop" installs AKA editable installs
+# with those installs __path__ resolves to some stub that cannot be parsed in the usual way
+# localmodules = [os.path.splitext(os.path.split(p)[-1])[0] for p in glob.glob(__path__[0] + '/[a-zA-Z]*.py')]
+
+# import.resources should be used instead, see also https://mesonbuild.com/meson-python/how-to-guides/editable-installs.html#data-files
+# the code below should be fully backwards compatible with other installs and newer py3s
+import importlib.resources
+localmodules = []
+for file in importlib.resources.files(__package__).iterdir():
+    if not file.name.startswith('_') and file.name.endswith('.py'):
+        localmodules.append(file.stem)
 
 modLocations = {}
 for m in localmodules:
```

**Error message**

```
Traceback (most recent call last):
  File "/XXXX//miniconda3/envs/test-pyme-3.11-conda_meson/lib/python3.11/site-packages/wx/core.py", line 3425, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1621, opened 2025-12-21
Meson importlib fix

csoeller:
Addresses issue #1619 (partially) .

Bugfix/improvement for uses of `__path__` to find modules matching a pattern.

**Proposed changes:**

Instead use `importlib.resources.files` approach to find module names matching a pattern.

This PR does not yet tackle the use of `__file__` throughout the dist. The use of `__file__` is more heterogeneous and may require a range of different fixes. Perhaps more importantly, the `__path__` related issues lead to runtime errors with editable builds. The current use of `__file__` seems by contrast not to break anything in our "standard installs". So I would consider this a lower priority and indeed seeing a case where it DOES break would be instructive (perhaps when installing from a wheel?).

**Checklist:**

Tested to some degree of meson editable builds and using the previous seup.py based installs. Seems to work fine for both.


David-Baddeley:
Thanks. Making slow progress on the ` __file__` ones locally, but agree it's not a priority.

PR #1622, opened 2025-12-21
Meson fixes for windows build

csoeller:
Addresses issue that current PYME does not successfully build on windows with meson.

Bugfix.

**Proposed changes:**

Two issues were encountered, (1) the use of `norm` in `PYME/experimental/triangle_mesh_utils.c` conflicts with some msvc header in `complex.h` or similar and (2) various CFLAGS for gcc/clang break compilation with msvc.

1. was fixed by renaming `norm` to `vnorm` (read "vector norm").
2. was approached by making various conditionals for `cc.get_id()` identifiying the compiler as MSVC. The use of compile flags could presumably be further improved in a platform dependent way using this and related mechanisms. For examples of pretty nice looking usage see also https://github.com/GNOME/gtk/blob/main/meson.build, especially use of `cc.get_supported_arguments(test_cflags)` to validate CFLAGS before passing as a project option.

**Checklist:**
Tested with latest PYME and meson build on win 11 using MSVC. Builds and runs fine with these changes although likely remaining warnings about "ignored" arguments with remaining standard CFLAGS (e.g. `-ffast-math`) where `if cc.get_id() == ...` has not yet been added to the local `meson.build` files.

### DIFF
--- a/PYME/Analysis/points/SoftRend/meson.build
+++ b/PYME/Analysis/points/SoftRend/meson.build
@@ -26,7 +26,7 @@ srcs = ['triRend.c','drawTriang.c', 'triangRend.c']
 
 #message('numpy include dir:', np_include_dir)
 
-if host_machine.system() == 'windows'
+if cc.get_id() == 'msvc'
   copts = ['/fp:fast']
 else
   copts = ['-O3', '-fno-exceptions', '-ffast-math', '-Wno-unused-but-set-variable']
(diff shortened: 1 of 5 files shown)


David-Baddeley:
Thanks! It's probably prudent to check if nanowrap / ch-shrinkwrap uses the `norm()` function before we merge, as it builds off the triangle mesh class.

David-Baddeley:
Seems to be OK on a quick glance - let's go ahead, can always go and fix ch-shrinkwrap if it turns into an issue.

PR #1624, opened 2026-01-06
PYMEAcquire camera dtype support

barentine:
Addresses issue encountered in #1463 and #1551 , which is that PYMEAcquire currently forces uint16. This PR allows cameras to use float (e.g. reading voltages off an NIDaq for my raster scanning project) or different int bit depths. This is split out from #1551 to simplify review.

enhancement
**Proposed changes:**
- add dtype attribute to camera base class
- use camera dtype in frameWrangler
- store dtype as backend_kwarg in `SpoolController`. 
- add dtype as input to `HDFBackend`. Note that `MemoryBackend` already takes dtype as a parameter, and `ClusterBackend` already handles other data types through `PZFFormat.dumps`.
- use numpy min/max in histogram update if we aren't working with uint16

**Notes**

- I have not worked with or tested the tiff acquisition backend.  
- this PR can be tested with the following simulated camera class, which just casts the typical `fakeCam` frames to float.

<details/> <summary/> FakeFloatCamera </summary>

```

class FakeFloatCamera(FakeCamera):
    numpy_frames=1
    order= 'C'
    dtype= 'float32'

    def __init__(self, XVals, YVals, noiseMaker, zPiezo, zOffset=50.0, fluors=None, laserPowers=[0,50], xpiezo=None, ypiezo=None, illumFcn = 'ConstIllum', pixel_size_nm=70.):
        super().__init__(XVals, YVals, noiseMaker, zPiezo, zOffset, fluors, laserPowers, xpiezo=xpiezo, ypiezo=ypiezo, illumFcn=illumFcn, pixel_size_nm=pixel_size_nm)
        self._saturation_threshold = np.finfo(np.float32).max

    def ExtractColor(self, chSlice, mode):
        try:
            d = self.compT.getIm().astype('float32')  # get image from completed computation thread
            # print d.nbytes, chSlice.nbytes
            memcpy(chSlice.ctypes.data_as(ctypes.POINTER(ctypes.c_uint8)),
                   d.ctypes.data_as(ctypes.POINTER(ctypes.c_uint8)), chSlice.nbytes)
            # chSlice[:,:] = d
            # self.compTOld = None #set computation thread to None such that we get an error if we try and obtain the same result twice
        except AttributeError:  # triggered if called with None
            logger.error("Grabbing problem: probably called with 'None' thread")
```

</details>

### DIFF
--- a/PYME/DSView/displaySettingsPanel.py
+++ b/PYME/DSView/displaySettingsPanel.py
@@ -237,7 +237,11 @@ class dispSettingsPanel2(wx.Panel):
         #if self.hlDispMapping.dragging == None:
         dsa = self.do.ds[:,:,0].ravel('F')
 
-        _min, _max = minmax_u16(dsa)
+        # get min and max efficiently if it is uint16
+        if str(dsa.dtype) == 'uint16':
+            _min, _max = minmax_u16(dsa)
+        else:  # otherwise, just use numpy
+            _min, _max = dsa.min(), dsa.max()
 
         #only perform histogramming on a subset of data points to improve performance
         ##note that this may result in strange behaviour of auto-optimise
@@ -250,9 +254,9 @@ class dispSettingsPanel2(wx.Panel):
 
         # if we are saturating, change the background
         try:
-            if _max >= self.scope.cam.SaturationThreshold:
+            if _max >= self.scope.cam.SaturationThreshold or _min <= -self.scope.cam.SaturationThreshold:
                 self.hlDispMapping.SetBackgroundColour(wx.Colour(0xFF, 0x8F, 0x8F))
-            elif _max >= self.scope.cam.SaturationThreshold/2:
+            elif _max >= self.scope.cam.SaturationThreshold/2 or _min <= -self.scope.cam.SaturationThreshold/2:
                 self.hlDispMapping.SetBackgroundColour(wx.YELLOW)
             else:
                 self.hlDispMapping.SetBackgroundColour(wx.WHITE)
--- a/PYME/DSView/displaySettingsPanel.py
+++ b/PYME/DSView/displaySettingsPanel.py
@@ -23,6 +23,7 @@
 
 import wx
 from PYME.ui import histLimits
+import numpy as np
 
 class dispSettingsFrame(wx.Frame):
     def __init__(self, parent, vp):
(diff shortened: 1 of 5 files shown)


David-Baddeley:
If we are adding support for non uint16 data types then it should be across all backends (or, at the very least, we should make sure we raise an informative error if the backend doesn't support it).  We should also have a mechanism for disabling un-supported backends so they don't show in the GUI (or, alternatively potentially doing type coercion - but that has some potential data loss issues). At a minimum I think it we need to build some data-type awareness into the `Backend` base class - e.g. add a `supported_dtypes` attribute to the base class (defaults to `['uint16',]`), make `dtype` a base class kwarg, and raise an exception in the base class constructor if `dtype not in supported_dtypes`.   Derived classes (currently memory and HDF backends) can override `supported_dtypes` as appropriate.

Because (for now) it's just your special case that uses non uint16 dtypes, we might not need to worry to much about the UI, but the `supported_dtypes` will define the interface we need to check to see if a given backend is offered.

David-Baddeley:
Suggest we also use actual numpy dtype objects rather than strings for the dtypes (we can cast in the constructor, so it will accept both) so that comparisons between, e.g. 'u2' and, 'uint16', and '<u2' etc evaluate correctly.

PYME/DSView/displaySettingsPanel.py
```diff
@@ -237,7 +237,11 @@ def RefrData(self, caller=None):
         #if self.hlDispMapping.dragging == None:
         dsa = self.do.ds[:,:,0].ravel('F')
 
-        _min, _max = minmax_u16(dsa)
+        # get min and max efficiently if it is uint16
+        if str(dsa.dtype) == 'uint16':
```
David-Baddeley: A bit wary about string comparisons on dtypes - maybe
```suggestion
        if dsa.dtype == np.uint16:
```

barentine:
Thanks for the comments, @David-Baddeley ! 

I changed the type of the `dtype` attributes from str to dtype. Note that PZFFormat.py makes string comparisons if we want to tweak that at some point. 

I also added `supported_dtypes` attribute for the backends. At the moment I would say it is more restrictive than it needs to be for e.g. MemoryBackend, but at least gets the framework there.

As an aside - if Time/Blinking series is selected we probably want to grey out 'Memory' backend in the UI unless we change its support.

Issue #1625, opened 2026-01-09
[BUG] with Python 3.13 visgui fails with mappings


csoeller:
**Describe the bug**
When using PYME built with Python 3.13 we invariable trigger a bug when resolving mappings from a `MappingFilter`. In other words when opening any `. h5r` and similar.

This can be traced to the following change in changed semantics for locals: https://docs.python.org/3/whatsnew/3.13.html#defined-mutation-semantics-for-locals.

PR forthcoming...

**To Reproduce**
1. Try opening file in `visgui`
2. See error

**Error message**
```
Traceback (most recent call last):
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/wx/core.py", line 2338, in Notify
    self.notify()
    ~~~~~~~~~~~^^
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/wx/core.py", line 3550, in Notify
    self.result = self.callable(*self.args, **self.kwargs)
                  ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/PYME/LMVis/visCore.py", line 767, in _create_base_layer
    l = self.add_pointcloud_layer(**layer_defaults.new_layer_settings('points', ds_keys=list(self.pipeline.keys())))
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/PYME/LMVis/visCore.py", line 470, in add_pointcloud_layer
    l = PointCloudRenderLayer(self.pipeline, method=method, dsname=ds_name, **kwargs)
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/PYME/LMVis/layers/pointcloud.py", line 203, in __init__
    self._set_method()
    ~~~~~~~~~~~~~~~~^^
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/PYME/LMVis/layers/pointcloud.py", line 224, in _set_method
    self.update()
    ~~~~~~~~~~~^^
  File "/Uxxx/XXX/miniconda3/envs/pip-py313/lib/python3.13/site-packages/PYME/LMVis/layers/pointcloud.py", line 248, in update
    self.update_from_datasource(self.datasource)
[...]

PR #1626, opened 2026-01-09
Eval fix py313

csoeller:
Addresses issue #1625 .

bugfix.

**Proposed changes:**
Supply locals and globals directly in eval call, as necessitated by changed semantics for locals: https://docs.python.org/3/whatsnew/3.13.html#defined-mutation-semantics-for-locals.

There should be no side effects and the changes are backwards compatible with (at least) Python >= 3.9 which we require anyway (and which is actually out of support now). Already successfully tested with 3.0 and 3.11.


David-Baddeley:
Thanks, looks OK.

David

Issue #1631, opened 2026-02-03
[BUG] Opening 3D image data with latest PYME builds and Numpy triggers error


csoeller:
**Describe the bug**
Latest PYME builds will pull in Numpy 2.4.x upon installation which triggers an issue in `BufferedDataSource.py` when opening 3D image data.

Numpy has become more stringent with conversion from numpy array to scalar. The numpy array now has to be 0D.

Adding a `squeeze` fixes the behaviour (PR forthcoming).

**To Reproduce**
1. Use recent PYME build that uses Numpy >= 2.4 (happens by default with current pip modules and python 3.12, possibly others)
2. Open 3D data with `PYMEImage`

**Expected behavior**
No error.

**Error message**

```
Traceback (most recent call last):
  File "/Users/XXXX/miniconda3/envs/test-pyme-3.11-conda_t/lib/python3.11/site-packages/wx/core.py", line 3425, in <lambda>
    lambda event: event.callable(*event.args, **event.kw) )
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/XXXX/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.11-conda_t/python-microscopy-recipe-macos-dpi/PYME/DSView/dsviewer.py", line 475, in LoadData
    vframe = DSViewFrame(im, None, im.filename, mode = mode)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/XXXX/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.11-conda_t/python-microscopy-recipe-macos-dpi/PYME/DSView/dsviewer.py", line 120, in __init__
    self.do.Optimise()
  File "/Users/XXXX/Documents/src/PYMEsrc/PYME-test-env/build-test-py3.11-conda_t/python-microscopy-recipe-macos-dpi/PYME/DSView/displayOptions.py", line 443, in Optimise
    bds = [self._optimal_display_range(c, method=method) for c in self.get_hist_data(subsample_threshold=5e8)]
                                                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1632, opened 2026-02-03
fix for BufferedDataSource triggering TypeError 0d with numpy >= 2.4

csoeller:
Addresses issue #1631.

Bugfix.

**Proposed changes:**
Add `squeeze` call to ensure 0D array.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/IO/DataSources/BufferedDataSource.py
+++ b/PYME/IO/DataSources/BufferedDataSource.py
@@ -51,7 +51,8 @@ class DataSource(XYZTCDataSource): #buffer our io to avoid decompressing multipl
             if ind in self.bufferedSlices: #return from buffer
                 #print int(numpy.where(self.bufferedSlices == ind)[0])
                 #print self.bufferedSlices
-                ret = self.buffer[int(numpy.where(self.bufferedSlices == ind)[0]), :, :].copy()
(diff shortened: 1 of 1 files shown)


barentine:
Looks good to me! This bug is currently producing at least one unit test error.

PYME/IO/DataSources/BufferedDataSource.py
```diff
@@ -51,7 +51,8 @@ def getSlice(self,ind):
             if ind in self.bufferedSlices: #return from buffer
                 #print int(numpy.where(self.bufferedSlices == ind)[0])
                 #print self.bufferedSlices
-                ret = self.buffer[int(numpy.where(self.bufferedSlices == ind)[0]), :, :].copy()
+                # add explicit squeeze to hopefully ensure 0d (required for int conversion with numpy >= 2.4 or so)
```
David-Baddeley, line 54:
```suggestion
                # add explicit squeeze to hopefully ensure 0d (required for int conversion with numpy >= 2.4 or so)
                # TODO? np.where is notoriously slow. Is there a better way of managing the index? Probably doesn't matter if we keep buffer sizes small (say < 50 frames).
```

David-Baddeley:
Looks good. See added comment (we might want to revisit the buffer index at some point)

David-Baddeley:
This might be somewhere where it makes more sense to put the buffered indices into a python dict than a numpy array.

Issue #1627, opened 2026-01-10
[BUG] Recipe tab only displaying in 1/4 of window size on macOS and matplotlib 3.10.8


csoeller:
**Describe the bug**
With latest `PYME`, `matplotlib` 3.10.x, the `Recipe` tab only occupies 1/4 of the available canvas size.
Conditions under which this manifests:

1. `matplotlib` 10.x (all ok on `matplotlib` 3.8, otherwise identical wx versions)
2. macOS (definitely an issue  on Sequoia and Tahoe, all ok on windows, even high DPI monitors etc)
3. retina display or high dpi monitor (ok on lower DPI monitor)

As mentioned, seems `matplotlib` related, thus probably not too similar to #1112, but not sure.

At this stage I do not know where to look for the source of the issue, pointers appreciated. In practice pretty annoying for me as it makes the Recipe tab effectively unusable.

**To Reproduce**
1. Open `visgui` (or `dh5view`)
2. Under conditions as above, see only 1/4 in `Recipe` tab occupied

**Screenshots**

<img width="2314" height="1828" alt="Image" src="https://github.com/user-attachments/assets/0b6816cf-083b-40f4-bbef-96c8707f2c89" />

csoeller:
The following ugly hack seems to fix things on macOS; still to test on windows and external high DPI displays on macOS. Basically, "newer" matplolib (3.8.X and later definitely) seems to implement figures (only wxagg backend?) on macOS higher dpi devices with `.get_dpi()` returning 200.0 instead of 100.0 otherwise. Rendering is visibly crisper on those devices compared to matplotlib 3.6.X. However, for the `size_inches` calculation one still needs to use the dpi of 100.0 for canvas filling results??!!

A better fix may be desirable but this work for me now.

```patch
diff --git a/PYME/contrib/wxPlotPanel.py b/PYME/contrib/wxPlotPanel.py
index 7148962..7797a0d 100755
--- a/PYME/contrib/wxPlotPanel.py
+++ b/PYME/contrib/wxPlotPanel.py
@@ -17,6 +17,9 @@ matplotlib.interactive( True )
 import numpy as num
 import wx
 
+import logging
+logger = logging.getLogger(__name__)
+
 class PlotPanel (wx.Panel):
     """The PlotPanel has a Figure and a Canvas. OnSize events simply set a
 flag, and the actual resizing of the figure is triggered by an Idle event."""
[...]

PR #1633, opened 2026-02-03
Fix recipe tab only displaying in 1/4 of window size on macOS issue

csoeller:
Addresses issue #1627 .

Bugfix.

**Proposed changes:**
Force effective dpi for calculation to 100 in `wxPlotPanel._SetSize()`. By inspection it is backwards compatible with older matplolibs.

**Rationale**
The fix could be regarded as a hack.

In defence of this approach:

1. It works and changes working in the recipe tab on macOS with high DPI from unworkable to very nice (recent numpy backends support dpi=200 by default which gives nice high res on such monitors including the retina displays).
2. The original calculation and implementation may be a hack anyway, if you change the DPI by request to anything other than 100 (e.g. using `rcParams`) the plot in the wxWindow gets screwed up, either too small or too big. So then why not force it to 100 for the sake of the calculation.
3. Happy if somebody comes up with a better solution. The PR includes `logger.debug` calls so you see what is going on if future issues were to arise.

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/contrib/wxPlotPanel.py
+++ b/PYME/contrib/wxPlotPanel.py
@@ -83,16 +83,18 @@ flag, and the actual resizing of the figure is triggered by an Idle event."""
     def _SetSize( self ):
         pixels = tuple( self.GetClientSize() )
 
-        # currently just a hack which (somehow) achieves that
-        # the 'Pipeline Recipe' Tab scales properly with post 3.6 (3.8+?) matplotlib on macOS high DPI displays
+        # currently mainly a hack which (somehow) achieves that
+        # the 'Pipeline Recipe' Tab scales properly with > 3.8.X (3.10.X+?) matplotlib
+        # on macOS high DPI displays
         # (without it the 'Pipeline Recipe' plot is only a quarter of the canvas size on
-        #     macOS high DPI displays and matplotlib>=3.8)
-        # hack seems to work fine with matplotlib <= 3.6
-        # needs testing on windows
+        #     macOS high DPI displays and matplotlib>=3.10)
+        # hack seems to work fine with matplotlib <= 3.8 as dpi seems firm on 100.0
+        # tests ok on windows as well
         logger.debug("pixels[0] %d" % pixels[0])
         logger.debug("dpi %.1f" % self.figure.get_dpi())
+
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Hmm, looks like a matplotlib bug. Happy in principle with the solution as a hack, but if we are going to effectively force 100 dpi then having the `if  (dpi != 100) dpi = 100` logic seems superfluous. Might re-work comments a bit.

David-Baddeley:
Actually fairly nervous about this - feel like it's something that could come to bite us 100 seems awfully arbitrary, standard screen resolution is normally 72 or 140.

David-Baddeley:
yeah, macbook screen dpi is 144. the 100 is some internal matplotlib rubbish (interestingly on older matplotlib, it's not an issue - still returns 100 dpi on high pixel count screens).

csoeller:
The observations I have made are
- on older matplotlibs we always get dpi 100 and even on high DPI displays things then look a little blurred
- on newer matplotlibs (I think it is that and not wx per se, but perhaps the matplotlib wx backend?) on macOS ONLY a dpi of 200 is autoselected on high DPI displays (internal and external but not on external low DPI displays)
- when the DPI of 200 is autoselected on high DPI displays, plots and the recipe tab look noticably crisper which is REALLY NICE
- for the size calculation in `_SetSize` ONLY a nominal value of 100 in the `self.figure.set_size_inches`  call gives the intended tab-filling result, REGARDLESS what `figure.get_dpi()` reports

My conclusion is that the matplotlib dpi concept is a little strange and I am not keen to get into it more deeply.

Accordingly I suggest to hardcode a value of 100 for the calculation in `self.figure.set_size_inches`

```python
    self.figure.set_size_inches( float( pixels[0] )/100.0,
                                         float( pixels[1] )/100.0)
```

and flag this for potential future issues with suitable explanatory comments.

It would be great to merge this soonish in some fashion as otherwise on macOS the recipe tab is kind of unusable when you install from current PyPi etc.

David-Baddeley:
Have added comment saying that we expect this might break in future matplotlib releases and will merge. Comment should be enough to make it easy for copilot or similar to guess if you ask it why it's broken.

PR #1634, opened 2026-02-06
PYMEImage half-pixel display fix

barentine:
Addresses issue of drawing array view from the center of the ulc pixel, rather than from the edge. With my stage-scanning stuff it is more noticable than normal since I'm looking at small images more often than not.

bugfix
**Proposed changes:**
- start drawing from the edge of the ulc pixel in arrayview
- shift the crosshair lines to draw through the center of the pixel

Left: previous behavior, Right: behavior of this PR
image

**Checklist:**
I don't immediately know if there are other side effects to this change. The pixel hit seems to be the same (i.e. x: X, y: Y) display at the bottom is still functioning correctly.
I have not checked XZ or YZ projections, in part because that is erroring for my test dataset anyway pre-fix

### DIFF
--- a/PYME/DSView/arrayViewPanel.py
+++ b/PYME/DSView/arrayViewPanel.py
@@ -457,7 +457,8 @@ class ArrayViewPanel(scrolledImagePanel.ScrolledImagePanel):
 
         
         im2 = wx_compat.BitmapFromImage(im)
-        dc.DrawBitmap(im2,int(-sc2/2),int(-sc2/2))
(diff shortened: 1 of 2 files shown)


David-Baddeley:
I'd want to double check that clicking(and all annotation functions) behave as expected - The previous bitmap offset was an alternative way of getting the relative position of crosshairs and pixels correct, and means that drawing reference frame corresponds to pixel centres (without needing manual half-pixel offsets in all drawing routines, and still working when drawing from a non-pixel frame (e.g. proper distance units). I definitely want to avoid having to duplicate this logic (practically this might mean altering `pixel_to_screen`, rather than than the manual offsets here, or keeping the offset on the bitmap.

PR #1639, opened 2026-03-03
Numpy 2 compatibility

barentine:
Addresses issue failing tests:
- \tests\PYME\IO\test_pzf.py
- \tests\PYME\localization\FitFactories\Interpolators\test_interpolators.py
- \tests\PYME\test_imports.py
- me opening an h5 file in PYMEImage 

compatibility enhancement / bugfix for later numpy
**Proposed changes:**

# PZFFormat.py

- switch numpy.fromstring (deprecated alternative) to numpy.frombuffer. 
-- copy buffers which are passed to bcl.HuffmanDecompress where we need it to be writable (otherwise we get ValueError: buffer source array is read-only)
- use tobytes instead of [its deprecated/removed alias tostring
](https://numpy.org/doc/2.1/reference/generated/numpy.ndarray.tostring.html)
- index 0-dimensional arrays from the header

# buffer_helpers.pyx

- remove deprecated/removed numpy.int in favor of int64 (which here matches the int64_t type being used)

# CSInterpolator.py

- index 0 dimensional array Z position.

# LinearInterpolator

- index 0 dimensional array Z position
- change numpy import so we don't clobber the built-in `min` method. This might seem superfluous, but is a clean fix to the current error on later numpy around the `dx = min(int((interpolator.shape[1] - len(X))/2), xm) - 2` lines. These otherwise blow up on a `numpy.exceptions.AxisError: axis 5 is out of bounds for array of dimension 0`. 

# BGSDataSource

- same squeeze fix as #1632, otherwise opening an h5 file in PYMEImage fails

# Notes
- I have only tested with unit tests, and opening PYMEImage / running a test frame localization.

### DIFF
--- a/PYME/localization/FitFactories/Interpolators/CSInterpolator.py
+++ b/PYME/localization/FitFactories/Interpolators/CSInterpolator.py
@@ -50,7 +50,7 @@ class CSInterpolator(__interpolator):
 
         ox = X[0] - 0.5*self.PSF2Offset
         oy = Y[0]
-        oz = Z #[0]
+        oz = Z[0]
 
         #rx = (ox % self.dx)/self.dx
         #ry = (oy % self.dy)/self.dy
@@ -87,7 +87,7 @@ class CSInterpolator(__interpolator):
 
         ox = X[0] - 0.5*self.PSF2Offset
         oy = Y[0]
-        oz = Z #[0]
+        oz = Z[0]
 
         #rx = (ox % self.dx)/self.dx
         #ry = (oy % self.dy)/self.dy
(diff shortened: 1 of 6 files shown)


PYME/localization/FitFactories/Interpolators/CSInterpolator.py
```diff
@@ -50,7 +50,7 @@ def interp(self, X, Y, Z):
 
         ox = X[0] - 0.5*self.PSF2Offset
         oy = Y[0]
-        oz = Z #[0]
```
David-Baddeley, line 53: This will likely break some use cases (think we are often supplying Z as a scalar not an array)

PR #1640, opened 2026-03-03
CI: add a pull request check action using meson / pip / etc

barentine:
Addresses issue #our tests don't run.

enhancement, and a small meson bugfix.
**Proposed changes:**
- capitalize the s in Scripts for PYME/Acquire/Scripts, otherwise the build failed. 
- add a Action/Workflow triggered on PR's to master branch
- skip installation of wxpython
- install pymecompress after numpy/cython/etc but before the rest of PYME and with no build isolation.

Runs tests on Python 3.11 (see [here where I tested on my fork](https://github.com/barentine/python-microscopy/actions/runs/22605841072/job/65497575770?pr=3)).
Fails on Python 3.12 at the moment, which might be a pymecompress issue.


David-Baddeley:
Guess it makes sense to switch over from the current azure hosting for the tests so all our CI is in one place. We should probably disable the azure pipeline while we are at it (azure-pipelines.yml) - we can likely disable by changing the trigger clause.

PR #1644, opened 2026-03-06
Pymeacquire webui

David-Baddeley:
Trying to make the pymeacquire webui a bit saner and better structured. Very much a work in progress.

Making a draft PR so progress and direction can be seen (e.g. @barentine ).


barentine:
Thanks for sharing, David! Really looking forward to checking it out

PR #1651, opened 2026-03-09
Update fit progress plot for numpy 2 compat, and resize it to show labels

barentine:
Addresses issue: progGraph is non-op on numpy2, and adding labels while we are here
<details> <summary> Traceback </summary> 

```

DEBUG:PYME.DSView.modules.LMAnalysis:use_cluster: newStyle=True, tq=None
ERROR:PYME.ui.mytimer:Error in timer callback
  File "C:\Userfiles\andrew\code\python-microscopy\PYME\ui\mytimer.py", line 48, in Notify
    a()
  File "C:\Userfiles\andrew\code\python-microscopy\PYME\DSView\modules\LMAnalysis.py", line 724, in refresh_analysis
    self.progPan.draw()
  File "C:\Userfiles\andrew\code\python-microscopy\PYME\LMVis\progGraph.py", line 46, in draw
    print((float(numpy.diff(ed[:2]))))
           ^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: only 0-dimensional arrays can be converted to Python scalars
DEBUG:PYME.DSView.modules.LMAnalysis:use_cluster: newStyle=True, tq=None

````

</details> 

both
**Proposed changes:**
- un-hardcode plot size. Plot looks visually worse since it is smaller, but this makes room for labels and adds the ability to read the right ticks
- add y labels
- remove print statement

image

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/LMVis/progGraph.py
+++ b/PYME/LMVis/progGraph.py
@@ -36,24 +36,24 @@ class progPanel(wxPlotPanel.PlotPanel):
                 return
             
             if not hasattr( self, 'subplot1' ):
-                #self.subplot1 = self.figure.add_axes([.14,.55,.85,.44])#self.figure.add_subplot( 211 )
-                #self.subplot2 = self.figure.add_axes([.14,.05,.85,.44])#self.figure.add_subplot( 212 )
-                
-                self.subplot1 = self.figure.add_axes([.14,.05,.85,.9])#self.figure.add_subplot( 211 )
-                self.subplot2 = self.subplot1.twinx()#self.figure.add_subplot( 212 )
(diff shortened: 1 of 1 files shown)


David-Baddeley:
My only concern (which is why it's stayed broken this long) is that the graph is significantly less useful when shrunk down that much. I wonder if we could find some way to make it a bit more compact ... Maybe putting the axis labels in the title or a legend (although this is also likely to cause issues).

barentine:
I know what you mean. I remember messing with it a bit and not being totally satisfied. But at least with this someone can see what it means and expand it if they're interested. Also clears up the stderr/stdout

PR #1652, opened 2026-03-11
add a bffile datasource and bioformats load path in image.py

barentine:
Addresses issue #bioformats installation. Posting as a draft for discussion - unclear if this is a direction we're interested in or not, but it was helpful for me to load some nd2 files from a collaborator. 

enhancement
**Proposed changes:**
- add option to load using [bffile](https://github.com/imaging-formats/bffile)

**loading tested with**
- single-channel nd2 file
- multi-channel nd2 file

**untested**
-time series
- other file types

### DIFF
--- a/PYME/IO/DataSources/BioformatsBFFDataSource.py
+++ b/PYME/IO/DataSources/BioformatsBFFDataSource.py
@@ -1,0 +1,74 @@
+
+import numpy as np
+from bffile import BioFile
+from .BaseDataSource import XYTCDataSource
+import logging
+
+logger = logging.getLogger(__name__)
+
+
+class DataSource(XYTCDataSource):
+    moduleName = 'BioformatsBFFDataSource'
+
+    def __init__(self, filename, series=0):
+        self.filename = filename
+        self.series_num = series
+
+        # Keep the file open for the lifetime of the DataSource so the lazy
+        # array can read
+        self._bf = BioFile(filename)
+        self._bf.open()
+        self._arr = self._bf.as_array(series=series)  # LazyBioArray (T,C,Z,Y,X), squeezed
+
+        # Use OME metadata for dimension sizes (avoids squeezing ambiguity)
+        pixels = self._bf.ome_metadata.images[series].pixels
+        self.sizeX = pixels.size_x
+        self.sizeY = pixels.size_y
+        self.sizeZ = pixels.size_z
+        self.sizeT = pixels.size_t
+        self.sizeC = pixels.size_c
+
+        self.additionalDims = 'TC'
+
+    def _tczyxs_index(self, t, c, z):
+        """Build an index tuple for the lazy array, skipping squeezed dimensions.
+
+        ``LazyBioArray.dims`` contains only the non-squeezed dimension names in
+        TCZYXS order.  We include an integer index for each of T, C, Z that is
+        actually present; Y and X are left as full slices so the result is a 2-D
+        view.
+        """
+        dim_to_idx = {'T': t, 'C': c, 'Z': z}
+        return tuple(dim_to_idx[d] for d in self._arr.dims if d in dim_to_idx)
+
+    # ------------------------------------------------------------------
+    # XYTCDataSource interface
+    # ------------------------------------------------------------------
+
+    def getSlice(self, ind):
+        c = int(ind // (self.sizeZ * self.sizeT))
+        t = int((ind // self.sizeZ) % self.sizeT)
+        z = int(ind % self.sizeZ)
+
+        key = self._tczyxs_index(t, c, z)
+        plane = self._arr[key] if key else self._arr
+        return np.asarray(plane).squeeze()
+
+    def getSliceShape(self):
+        return (self.sizeY, self.sizeX)
+
+    def getNumSlices(self):
+        return self.sizeC * self.sizeT * self.sizeZ
+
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Sounds good to me. Is installing bffile relatively painless? Is it painless enough that w might consider adding it to our dependencies?

David-Baddeley:
I especially like the fact that it has a lazy loader (although the terminology is slightly misleading - `asarray` would usually imply forcing into memory ...)

barentine:
Super painless to install over my recent 3.11 numpy 2.4 install via `pip install bffile`. If memory serves the install was fast, but the first time you call it from code it pulls java jars. Doesn't take long, but small oddity to note.
I think it would be pretty reasonable to add as a dependency.

csoeller:
Hi @barentine, I have played around with this a little, very nice! Trying to use it to get metadata for confocal planes/stacks loaded from `.msr` files and it works in principle. I wanted to note a couple of small caveats I came across when (a) installing and (b) using it.

On the mac I tried `pip install bffile` which initially failed, first because of missing java JDK which I solved with `sudo port install openjdk17` (I am a macport user) and then it complained about missing `ant` which I similar could solve with `sudo port install apache-ant`. Both of these were triggerd by the dependency build of `JPype1`. Not sure if this can be bypassed with a conda install or if you have other tips.

csoeller:
Second small issue I came across is the orientation of loaded images/stacks. Here I get some swapping of x/y axes and orientations as compared to the MINFLUX coordinate data (x and y coordinates).

Interestingly, this was done "correctly" by using the existing `_loadOBF` functionality which uses the `obf_support.py` we have in the dist and is a pure-python version to open some parts of OBFs and `.msr`s. Not sure which of the implementations does it right by the book but certainly for the MINFLUX/PYME coordinate conventions, `obf_support.py` seems to do the right thing (possibly by chance?).

Finally, I do need to fall back to `bffile` to get at the metadata. I believe the current `_loadBioformatsBFF` has a small issue there since using `MetaDataHandler.OMEXMLMDHandler` to get at the OME metadata as this always seems to grab the info from the first image in the long `bf.ome_xml` block if containing an image series, regardless which one you load.

For my own use I need to get at the per image metadata which I can apparently get with `bf.core_metadata(series=ds.stack_number)` and indeed this works but the metadata may be non standard.

Not sure how what is the right way here to go forward, looking for suggestions and would be happy to assist if useful.

barentine:
Hi @csoeller - sorry for the delay! It would be great to have some help on fixing up the datasource. 

X/Y swaps: For OBF, I would say that was an accident - when I wrote the shim it was to analyze some STED stacks for a collaborator and I didn't check any PYMEVis coordinate compatibility (though I'm glad it works correctly!). I am currently working with .nd2 files, and haven't checked the X/Y ordering. I would suggest we go ahead and add a transpose to the datasource, and we can optionally add a dictionary to the file to keep track of which filetypes need it if we find that some do not. 

Metadata: Your metadata fix sounds like a good one to me! I haven't worked with more complicated files yet. I do think it would be nice to pull in all metadata, into an `OriginalMetadata` tree, which we translate certain keys from into PYME metadata, so that if we e.g. save a partially analyzed file to .h5, it has the original metadata with it still.

OSX pre-reqs: I didn't run into these on Windows, and haven't tried on OSX yet. Glad you found that out. May be helpful info for @David-Baddeley as he considers whether to make bffile a dependency or leave it as optional.

As to how to go about adding your fixes - you can certainly submit a PR against my fork:branch (barentine:bffile), or perhaps it's at a stage where @David-Baddeley is comfortable pulling it into the main repository to take your fixes from there? I also don't mind if you just clobber this PR with another containing your improvements, but that can also get messy if it doesn't merge quickly.

csoeller:
Hi @barentine, no worries, thanks for the reply.

First re installation, a simple conda install (`conda install bffile`) seems to succeed on both windows and mac without needing anything else installed. So this could be the painless variant. As any code falls back to the older bioformats interface if `bffile` is not found I would suggest to include the PR, nothing is lost if no `bffile` is present, lots is gained if it is.

In the `pip` install we could indicate `bffile` under the optional dependencies, i.e. adding

```toml
[project.optional-dependencies]
bffile=["bffile"]
```

in `pyproject.toml`. This is then also shown as option on PyPi etc.

The metadata and image origin bits warrant a little extra discussion, I'll tackle that in a followup message when I find the time.

David-Baddeley:
Looking at the pip package, it's just the python code only. I'm kind of surprised pip install works on windows (usually harder for java stuff than mac)! In any case I don't think we can have it as a hard dependency (at least for pip installs), which might force us back to conda if we want it by default (which is a bit of a shame as I'm thoroughly frustrated with conda).

David-Baddeley:
Re metadata - The multi-series in one file use case is something we generally handle pretty poorly in PYME (although it's probably only bioformats and bffile that actuallly have any real support for it at present).

David-Baddeley:
Re merging - What is going to work better for you guys? We could merge and then fix issues as patches against main, or carry on in this branch ...

barentine:
Merging to main and continuing to PR against that would work best for me. I'm juggling a couple of branches for various projects and these commits are starting to be used by several of them. Thanks for asking!

David-Baddeley:
Will merge, under the understanding that there are still a few bugs to be fixed but useful in current state.

PR #1653, opened 2026-03-12
20260312 LMAnalysis detection overlay roisize

abalaji0827:
Bug Fix: In PYMEImage, when detection is tested on a frame, the green box identifying detections is not the same size ast the analysis roi size. It appears that the box was drawn with a default value of 11 pixels

I addedd 3 lines of code to _update_points_overlay to get Analysis.ROISize from the analysis metadata and then use that size to determine size of the box to draw, assuming the roi size is a ROI Half Size (box size of roi_size * 2 + 1).

No variable names are changed. A new variable is created within _update_points_overlay, roi_size.


David-Baddeley:
Makes sense, many thanks.

David-Baddeley:
Sorry I've taken so long to get to this.

PR #1654, opened 2026-03-19
ExtractChannel and FlatFieldDatasource 5D / multichannel datasource fixes

barentine:
Addresses issues:
- `ExtractChannel._pickChannel` uses legacy `image.data` access and collapses the wrong index if you pass it a XYZTC shape [1024, 1024, 1, 1, 4] datasource and try to pick out one of the 4 channels
- FlatFieldDataSource doesn't preserve parent datasource t/z info in its current state, and can also mess up channel extraction. 

bugfix
**Proposed changes:**
- use `image.data_xyztc` access in `ExtractChannel` module
- add units tests for ExtractChannel
- store parent datasource t/z info in FlatFieldDataSource

FlatfieldDataSource currently inherits from `XYTCDataSource`, i.e. the old-style datasource base. Certainly it would be better to do a full re-jig, but this was the fastest fix for me. 

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/recipes/base.py
+++ b/PYME/recipes/base.py
@@ -1021,21 +1021,13 @@ class ExtractChannel(ModuleBase):
     channelToExtract = Int(0)
     
     def _pickChannel(self, image):
+        from PYME.IO.DataSources.CropDataSource import crop_image
         c = self.channelToExtract
-        # Use data_xyztc (5D, X,Y,Z,T,C) with a slice (c:c+1) rather than an integer index
-        # so the channel axis is preserved, avoiding collapse to a 4D array and re-promotion ambiguity.
-        # chan = image.data_xyztc[:,:,:,:,self.channelToExtract]  # fails to preserve channel axis
-        chan = image.data_xyztc[:,:,:,:,c:c+1]
-        
-        im = ImageStack(chan, titleStub = 'Filtered Image')
-        im.mdh.copyEntriesFrom(image.mdh)
+        im = crop_image(image, crange=(c, c+1))
         try:
-            im.mdh['ChannelNames'] = [image.names[self.channelToExtract],]
+            im.mdh['ChannelNames'] = [image.names[c],]
         except (KeyError, AttributeError):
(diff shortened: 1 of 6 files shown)


PYME/recipes/base.py
```diff
@@ -1021,7 +1021,11 @@ class ExtractChannel(ModuleBase):
...
     
     def _pickChannel(self, image):
-        chan = image.data[:,:,:,self.channelToExtract]
+        c = self.channelToExtract
+        # Use data_xyztc (5D, X,Y,Z,T,C) with a slice (c:c+1) rather than an integer index
+        # so the channel axis is preserved, avoiding collapse to a 4D array and re-promotion ambiguity.
+        # chan = image.data_xyztc[:,:,:,:,self.channelToExtract]  # fails to preserve channel axis
+        chan = image.data_xyztc[:,:,:,:,c:c+1]
         
         im = ImageStack(chan, titleStub = 'Filtered Image')
```
David-Baddeley: We should **really** go a step further and try and make this lazy (if practical for the dataset). Otherwise we're going to be forcing lots of stuff into memory un-necessarily.

barentine: I'll have a go at making it lazy, though I don't love letting perfection be the enemy of bug fixing ;)

David-Baddeley:
Sorry, I didn't mean it had to happen. Before merging ... Was just what I
thought when looking at it.

barentine:
No apology needed, David! 

I added a color crop to `CropDataSource`, which looks like a lazy slicer, and made `ExtractChannel` just a color crop. Can back it out if you prefer an alternative.

barentine:
I've added a change to move FlatFieldDataSource to inherit from the 5D XYZTCDataSource. @David-Baddeley let me know if you'd like this PR refactored into a couple smaller ones.

barentine:
Note that the test failure appears unrelated (and passes on my machine). We probably want to look at whether `test_fiducial_matching.py` is doing what we want it to.

PR #1660, opened 2026-04-23
fix hx/hy assignment regression for line profiles

barentine:
Addresses regression in arrayViewPanel ([commit here](https://github.com/python-microscopy/python-microscopy/commit/f6e5bdd6748af39ae3d1c7d774841246a631cba1)) where hx, hy never get assigned for line profiles (width=1) .

bugfix
**Proposed changes:**
- call do.GetSliceSelection() for line profiles (width=1)
- round to get pixel identity rather than casting to int, which is a floor operation. 

image
produces

```
>>> lx, ly, hx, hy = do.GetSliceSelection()
>>> do.GetSliceSelection()
(0, 0, 0, 6)
```

while
image

yields

```
>>> (xi, yi, xf, yf)
(0, 0, 1, 1)
>>> image.data_xytc[xi:xf+1, yi:yf+1,0,0]
array([[ 0.25463662,  2.1478719 ],
       [ 1.99672588, -1.52296621]])
```

There are some remaining quirks:
- `displayOptions.GetSliceSelection` returning start, finish, and us calling things lx, hx is fairly misleading and requires a +1 for the finish to get the full extent.
- There appears to be a bug in `PYME.Analysis.profile_extraction import extract_profile` when drawing along the edge of the frame.

**tested on**
wxPython '4.2.5 msw (phoenix) wxWidgets 3.2.9'
python 3.11
Windows 11

### DIFF
--- a/PYME/DSView/arrayViewPanel.py
+++ b/PYME/DSView/arrayViewPanel.py
@@ -765,7 +765,7 @@ class ArrayViewPanel(scrolledImagePanel.ScrolledImagePanel):
 
         self.do.inOnChange = True
         try:
-            self.do.xp, self.do.yp, self.do.zp = [int(p) for p in pos_3d]
+            self.do.xp, self.do.yp, self.do.zp = [int(round(p)) for p in pos_3d]
         finally:
             self.do.inOnChange = False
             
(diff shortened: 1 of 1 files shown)


David-Baddeley:
I need to understand the implications of round vs floor a bit better. Naively I'd think the correct behaviour for square selections at least is probably floor on the top left and ceil on the bottom right. For lines we want the same behaviour for top left and bottom right (round, floor, or ceil) (but I'd like to avoid selection type logic if I can in anything but display).

PR #1662, opened 2026-04-28
Bffile datasource / loading tweaks

barentine:
Addresses issues:
- channel names not getting pulled in `OMEXMLMDHandler`
- base datasource was 4D, not 5D (my bad), and was subject to autopromotion which (in my case) assumed the sub-100 frame was a stack rather than a short time series. 

enhancement
**Proposed changes:**

- inherit from `XYZTCDataSource` for `BioformatsBFFDataSource`
- check for channel names in OME metadata handler

### DIFF
--- a/PYME/IO/DataSources/BioformatsBFFDataSource.py
+++ b/PYME/IO/DataSources/BioformatsBFFDataSource.py
@@ -1,13 +1,13 @@
 
 import numpy as np
 from bffile import BioFile
-from .BaseDataSource import XYTCDataSource
+from .BaseDataSource import XYZTCDataSource
 import logging
 
 logger = logging.getLogger(__name__)
 
 
-class DataSource(XYTCDataSource):
+class DataSource(XYZTCDataSource):
     moduleName = 'BioformatsBFFDataSource'
 
     def __init__(self, filename, series=0):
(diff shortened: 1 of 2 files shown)


PYME/IO/DataSources/BioformatsBFFDataSource.py
```diff
@@ -1,34 +1,35 @@
...
 
 logger = logging.getLogger(__name__)
 
 
-class DataSource(XYTCDataSource):
+class DataSource(XYZTCDataSource):
     moduleName = 'BioformatsBFFDataSource'
 
-    def __init__(self, filename, series=0):
+    def __init__(self, filename, taskQueue=None, series=0):
```
David-Baddeley, line 13: Any reason for the `taskQueue=None`? I know several other data sources have it, but it might be something we should remove across the board.

David-Baddeley: OK, the taskQueue argument is only used in `remFitBuf`, where it is relic of the defunct (since we left python 2) task queue. Will create a new issue for removal across the project.

David-Baddeley: #1680

PR #1663, opened 2026-04-28
add test for recipe localization modules coordinate conventions

barentine:
I'm PR'ing this as a draft to show a test I think we should have, and to raise as an issue the coordinate convention headache that one might hit if they do a couple of localization processes using recipe modules. 
Namely:
- DetectPoints2D output is in units of pixels, not units of nanometers
- FitPoints requires input positions to be in units of nanometers
- Between density mapping and FitPoints one picks up a half-pixel shift.

Would love to hear some thoughts on preferred way to fix some of these @David-Baddeley ! 
My thought would be to:
- multiply DetectPoints2D by voxelsize_nm.x and voxelsize_nm.y
- fix the fixme in visHelpers.rendGauss about im[0,0] centering
- write a documentation page about coordinate conventions in PYMEVis / PYMEImage

### DIFF
--- a/tests/PYME/recipes/test_measurement.py
+++ b/tests/PYME/recipes/test_measurement.py
@@ -66,6 +66,56 @@ def test_ChunkedTravelingSalesman():
     ordered = recipe.execute()
     assert ordered.mdh['TravelingSalesperson.Distance'] < ordered.mdh['TravelingSalesperson.OriginalDistance']
 
+def test_DensityMapping_DetectPoints2D_and_FitPoints():
+    from PYME.recipes.measurement import DetectPoints2D, FitPoints
+    from PYME.recipes.localisations import DensityMapping
+    from PYME.IO.tabular import DictSource
+    from PYME.IO.MetaDataHandler import DictMDHandler
+    from PYME.recipes.tablefilters import Mapping
+    # generate gt points
+    n = 15
+    voxelsize_nm = 100
+    im_size = 256
+    # generate points, avoiding edges
+    x_px_gt = np.linspace(20, im_size - 20, n).astype('f')
+    y_px_gt = np.linspace(20, im_size - 20, n).astype('f')
+
+    res = DictSource({'x': x_px_gt *voxelsize_nm, 'y': y_px_gt * voxelsize_nm,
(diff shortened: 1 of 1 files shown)


David-Baddeley:
Coordinates in general could use a bit of a work-over  - it's a bit more than just nm vs pixels, it's nm (roi referenced), nm (origin referenced), or pixels. Pixels is the most forgiving if the data comes from a source where it might be missing some metadata , but is also the least powerful.

David-Baddeley:
I wonder about adding some sort of units awareness to tabular, and letting it handle conversions (so if you had a column `x`, you could query e.g. `x_nm` or `x_pixels`), but this could be adding complexity where it's not needed.

barentine:
I do think baking in more unit awareness would be helpful rather than harmful. For example the change to have `voxelsize_nm` instead of just assuming voxelsize.x is in um has been quite nice. 

Separately, I wonder whether there are other ways to bring `DetectPoints2D` into alignment with the `FitPoints` module - right now it is super frustrating to e.g. develop a recipe using localization, do the flat fielding yourself etc. and then if you forget to roll your own mapping filter (which is tough because you have to add voxelsize_nm to the mapping namespace somehow or magic number it), you end up with a major bug.

David-Baddeley:
Having a bit of a think about what we'd really need to make this work in tabular.

To be robust we'd need to track:

type: (initially, None or "coordinate" / position)
units: (pixels, nm)
origin: (camera, ROI, stage, other?)
applied transformations: (has this been re-mapped from an image-splitting devce, or drift correction etc ...)

This feels quite heavy for the tabular class when it applies to 3 columns but not to others. 

Serialising to hdf will be fine (or rather we can find a way to make it work), serialising to csv / excel will loose all the metadata.

I'm not saying we shouldn't do it, but just that we should think a bit about how.

One possible approach (albeit with high breaking change potential) is to actually change anything that currently writes `x` to write e.g. `x_cam_pixels`, `x_roi_pixels`, `x_roi_nm` instead and to implement a light-weight wrapper that does backwards compatible translation to `x` (advantage of this is that the unit intent is not lost if writing to e.g. csv, even if the metadata needed to translate the units is). Asking specifically for `x_roi_pixels` (rather than x) in a downstream module will fail in a fairly explainable way if that can't be computed.

PR #1672, opened 2026-05-20
remove the positionless EVT_COMMAND_RIGHT_CLICK

barentine:
Addresses issue AttributeError when trying to right click a filter key in PYMEVis:

```
AttributeError: 'CommandEvent' object has no attribute 'GetX'. Did you mean: 'GetId'?
  File "C:\Userfiles\andrew\code\python-microscopy\PYME\ui\filterPane.py", line 130, in OnFilterListRightClick
    x = event.GetX()
        ^^^^^^^^^^
```

bugfix for compatibility with wxpython 4+
**Proposed changes:**
- remove binding `EVT_COMMAND_RIGHT_CLICK` binding which fires first and throws an attribute error.
- change to `wx.EVT_CONTEXT_MENU`instead of `wx.EVT_RIGHT_UP` so we can right click without requiring the item to be selected first (let us do single right click instead of e.g. double right click)
- Use `event.GetPosition()`

image

### DIFF
--- a/PYME/DSView/modules/measure/filterPan.py
+++ b/PYME/DSView/modules/measure/filterPan.py
@@ -72,11 +72,7 @@ class FilterPane(wx.Panel):
             self.Bind(wx.EVT_MENU, self.OnFilterDelete, id=self.ID_FILT_DELETE)
             self.Bind(wx.EVT_MENU, self.OnFilterEdit, id=self.ID_FILT_EDIT)
 
-        # for wxMSW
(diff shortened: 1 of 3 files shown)


David-Baddeley:
Is this tested across platforms? I note that the code it's replacing registers 2 different handlers for different platforms (it's also likely written several wx versions ago, so the platform difference might have disappeared).

barentine:
Yep, works on Windows and OSX.
Screenshot 2026-05-21 at 10 43 28 AM

Issue #1673, opened 2026-05-21
[SUPPORT] Can I change the color of the "Scale Bar" in the GUI?


boydcpeters:
Note: Support is also available by emailing support@python-microscopy.org, choose this option if you would like the question and answer to be publicly searchable.

**What are you trying to do?**
Is there a way to change the color of the scale bar in the GUI?

**Where are you stuck?**
When changing the background color to white, the scale bar is no longer visible. Is there a way to change this easily in the GUI?

barentine:
Not through the GUI at the moment. Your GUI option would be Extras>Scalebox, and then you can set the color of the scale box.

There is also not a particularly easy way to hack it through the Shell, unfortunately. Would be
```
glCanvas.ScaleBarOverlayLayer._color = [0.0, 0.0, 0.0] 
glCanvas.Refresh()
```
but the blending mode is additive, so if your background is white you'll never see a scalebar regardless of its color.

If you're on a development install and want a workaround ASAP you can pull #1674

boydcpeters:
Thanks @barentine for the quick reply and even a possible fix!

I am currently not on a development install and as I don't need it ASAP I will wait for the next PyPi release. 

Once again, thanks for the help!

barentine:
Should be all set @boydcpeters with the latest pypi release (yesterday). Thanks for your patience!

boydcpeters:
Amazing, thanks for the help! I will make sure to check it out next week.

Issue #1258, opened 2022-07-15
[ENHANCEMENT] PYMEImage scalebar color options


barentine:
**Is your feature request related to a problem? Please describe.**
I know we've discussed e.g. movable scalebars before, etc. At the moment I'm most plagued by its color - sometimes I show subtractions of two images, where I want to use the pyplot bwr colormap (blue to white to red) to white at zero. Putting the zero of my image, if there are a lot of zeros, washes out the white scalebar & label. 

**Describe the solution you'd like**
dropdown selector with a couple scalebar color options (black, white, yellow, for example, would probably go far). 
Ideally maybe also easily toggle scalebar size - not an immediate issue for me, but I can imagine it becoming one

David-Baddeley:
Certainly not opposed, also not opposed to configurable position (potentially just selecting one of the 4 corners, potentially manually specifying a position as text / a % of the image size), but don't think it should be draggable. 

My objections to a draggable scale bar were (and are):
- difficult to implement cleanly (high implementation cost relative to gains)
- difficult/impossible to position consistently between images
- not really suitable for the stated use case of measuring feature sizes, due to fixed orientation and projection errors. In practice using a moveable scalebar for feature measurement is not going to be any better than eyeballing it and/or using the scalebar to calibrate your thumb and moving it over the image. 

In terms of implementing colour/length options there are a few things to keep in mind.
- there is a length option already, but it's not exposed in the GUI
- we eventually want to unify overlays with PYMEVis (common overlay class with common options and config interface, separate wx and OpenGL render methods). Trying to keep options etc consistent between these is a good start.
- I'm a bit torn between a small set of fixed colours (better for consistency) and just using the OS colour chooser (familiar). A small set of options is probably easier to implement.

In the short term, adding a colour option as a trait to the scale bar overlay is probably the way to go. Adding an e.g. `View->Scalebar options` menu item which called `.configure_traits()` on the scale bar overlay would be a super simple option that would let you set length and colour (and would automatically expand to allowing position to be set if we added an appropriate trait). In the somewhat longer term, we could think about making the "Overlays" panel where you currently toggle overlay visibility more like the PYMEVis Layers control where you could expand each overlay to configure it's options.

PR #1674, opened 2026-05-21
add scalebar color picker to PYMEVis

barentine:
Addresses issues #1673

enhancement
**Proposed changes:**
- change blend style of scalebar render so a dark scalebar can show up on a light background
- add a colourpicker box for the scalebar color so one can change it through the GUI

**notes**
- only tested on windows at the moment
- @David-Baddeley  I do agree with conversations in  #1258 that it would be great to replace the GUI change here with something that brings PYMEImage and PYMEVis closer into alignment, but maybe a second button here is workable in the meantime.

image

### DIFF
--- a/PYME/LMVis/layers/ScaleBarOverlayLayer.py
+++ b/PYME/LMVis/layers/ScaleBarOverlayLayer.py
@@ -93,5 +93,6 @@ class ScaleBarOverlayLayer(OverlayLayer):
                 self._bind_data('scalebar', self._verts, 0*self._verts, self._cols, sp, core_profile=core_profile)
 
                 glPolygonMode(GL_FRONT_AND_BACK, GL_FILL)
+                glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)  # standard blending so dark can be visible on light background
                 glDrawArrays(GL_TRIANGLES, 0, 6)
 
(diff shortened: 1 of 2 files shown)


David-Baddeley:
Sometimes blend func usage propagates through the opengl stack and shows up in the wrong place - let me  just double check

David-Baddeley:
Sorry, miss- click there, didn't mean to close

David-Baddeley:
Should be safe, as the only thing called after the scalebar is the LUT (and it disables blending entirely). Do we want to allow alpha on the scale bar, or would it be simpler to just call `glDisable(GL_BLEND)`?

PYME/LMVis/layers/ScaleBarOverlayLayer.py
```diff
@@ -93,5 +93,6 @@ def render(self, gl_canvas):
                 self._bind_data('scalebar', self._verts, 0*self._verts, self._cols, sp, core_profile=core_profile)
 
                 glPolygonMode(GL_FRONT_AND_BACK, GL_FILL)
+                glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)  # standard blending so dark can be visible on light background
```
David-Baddeley, line 96:
```suggestion
                glBlendFunc(GL_SRC_ALPHA, GL_ONE_MINUS_SRC_ALPHA)  # standard blending so dark can be visible on light background
                # Should we just disable blend instead? or should we make all layers disable blend on exit (and be responsible for enabling it on entry if needed?)
```

PR #1683, opened 2026-06-29
add analysis metadata parameter to set initial sigma value used in LatGaussFitFR fitting

barentine:
Addresses issue: if you are using LatGaussFitFR.py to fit e.g. large beads imaged on low NA setup, rather than single molecules on high NA oil immersion, you'd like to change the initial sigma value.

enhancement
**Proposed changes:**
- introduce Analysis.SigmaGuess metadata parameter to LatGaussFitFR.py

**Checklist:**
of the TODO list).

### DIFF
--- a/PYME/localization/FitFactories/LatGaussFitFR.py
+++ b/PYME/localization/FitFactories/LatGaussFitFR.py
@@ -168,12 +168,12 @@ class GaussianFitFactory(FFBase.FitFactory):
         fitBackground = self.metadata.getOrDefault('Analysis.FitBackground', True)
         
         if fitBackground:
-            startParameters = [A, x0, y0, 250/2.35, dataMean.min(), .001, .001]
+            startParameters = [A, x0, y0, self.metadata['Analysis.SigmaGuess'], dataMean.min(), .001, .001]
     
             #do the fit
             (res, cov_x, infodict, mesg, resCode) = self.solver(self.fitfcn, startParameters, dataMean, sigma, X, Y)
         else:
-            startParameters = [A, x0, y0, 250 / 2.35]
+            startParameters = [A, x0, y0, self.metadata['Analysis.SigmaGuess']]
     
             #do the fit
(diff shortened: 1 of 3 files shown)


PYME/localization/FitFactories/LatGaussFitFR.py
```diff
@@ -168,12 +168,12 @@ def FromPoint(self, x, y, z=None, roiHalfSize=5, axialHalfSize=15):
         fitBackground = self.metadata.getOrDefault('Analysis.FitBackground', True)
         
         if fitBackground:
-            startParameters = [A, x0, y0, 250/2.35, dataMean.min(), .001, .001]
+            startParameters = [A, x0, y0, self.metadata['Analysis.SigmaGuess'], dataMean.min(), .001, .001]
```
David-Baddeley:
```suggestion
            startParameters = [A, x0, y0, self.metadata.get('Analysis.PSFSigmaGuess', 250/2.35), dataMean.min(), .001, .001]
```

- keep backwards compatibility for re-analysis with old metadata (`PARAMETERS` is mostly a GUI niceity)
- make naming line up with, e.g. `GaussMultifit`

David-Baddeley: Obviously needs corresponding changes to other usages

David-Baddeley:
I wonder if we should add this to other Gaussian fits at the same time (ie `AstigGaussFit`, `SplitterFitFNR`)

PYME/localization/FitFactories/LatGaussFitFR.py
barentine: I think we'll want to do `metadata.getOrDefault`, but otherwise appreciate the suggestion to switch style, stick with the naming convention, etc.

barentine:
Good to go for the 3 mentioned FitFactories.
image

barentine:
@David-Baddeley I'm going to suggest @csoeller or @zacsimile takes a look at this, and add a new tag to suggest as much. (We can discuss its name, description, and how to automate it in the Zulip chat)

David-Baddeley:
Sorry for being asleep, I think we can go ahead and merge

Issue #1636, opened 2026-02-25
[BUG] can't import rules on editable meson installs


barentine:
**Describe the bug**

<details>  <summary> Traceback </summary>

```

In [1]: from PYME.cluster.rules import LocalisationRule
using an old version of zeroconf (0.148.0), consider upgrading
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[1], line 1
----> 1 from PYME.cluster.rules import LocalisationRule

File C:\Userfiles\andrew\code\python-microscopy\PYME\cluster\rules.py:118
    115             self._t_watch.start()
    117 import jinja2
--> 118 jenv = jinja2.Environment(loader=jinja2.PackageLoader('PYME.resources', 'web'))
    119 rg_template = jenv.get_template('rule_inline.html')
    120 class RuleGroupWatcher(object):

File ~\AppData\Local\miniforge3\envs\pipyme311\Lib\site-packages\jinja2\loaders.py:359, in PackageLoader.__init__(self, package_name, package_path, encoding)
    357             break
    358     else:
--> 359         raise ValueError(
    360             f"PackageLoader could not find a {package_path!r} directory"
    361             f" in the {package_name!r} package."
    362         )
    364 self._template_root = template_root

ValueError: PackageLoader could not find a 'web' directory in the 'PYME.resources' package.

```

</details>

It seems like the jinja2 PackageLoader does not work with meson editable installations (see [here](https://github.com/mesonbuild/meson-python/issues/716) ), and like it won't work until jinja2 v3.2 

I'm not immediately having luck swapping out the `PackageLoader` for `FileSystemLoader`.

**To Reproduce**
1. install PYME for development via pip / meson
2. `from PYME.cluster.rules import LocalisationRule`
3. See error

**Platform (please complete the following information - or copy-paste from error dialog if shown):**
 - OS: [e.g. Win10] win11
 - Python version: 3.11 
 - numpy version: 2.4.2
 - wxpython version [if a UI bug]: 4.2.4

barentine:
one potential workaround is
```
import importlib.resources
import PYME.resources
rg_template = jinja2.Template(importlib.resources.files(PYME.resources).joinpath('web').joinpath('rule_inline.html').read_text())
```
though jinja2 API suggests it is usually better to instantiate through an environment w/ loader rather than directly

barentine:
fixed with #1687

PR #1687, opened 2026-07-30
Misc deprecation and regression fixes

David-Baddeley:
A bunch of deprecation and regression fixes (mostly not actually related to python 3.11). Mostly centred around getting the PYMEAcquire simulator running again, but have tried to fix other bits where the problem


barentine:
Thanks @David-Baddeley - I recognize the fakeCam inheritance and jinja2 template issues (can likely close #1636 after this is merged) . Sorry I didn't submit an issue or PR when I hit the fakeCam issue.

PR #1691, opened 2026-08-04
Extract Positioning.Stage_X, Stage_Y, and PIFoc, from OME XML

barentine:
Addresses issue wanting to know stage position after loading nd2 file through bffiledatasource.

Screenshot 2026-08-03 at 5 09 35 PM

I think PIFoc is a little odd to use by default for Z, but for now I think Positioning.Stage_Z is unused.

### DIFF
--- a/PYME/IO/DataSources/BioformatsBFFDataSource.py
+++ b/PYME/IO/DataSources/BioformatsBFFDataSource.py
@@ -73,3 +73,7 @@ class DataSource(XYZTCDataSource):
     def __del__(self):
         self.release()
 
+    @property
+    def ome_series_metadata(self):
+        return self._bf.ome_metadata.images[self.series_num]
+
--- a/PYME/IO/MetaDataHandler.py
+++ b/PYME/IO/MetaDataHandler.py
@@ -917,7 +917,7 @@ class OMEXMLMDHandler(XMLMDHandler):
             
         return ps
         
-    def __init__(self, XMLData = None, mdToCopy=None):
+    def __init__(self, XMLData = None, mdToCopy=None, series=0):
         if not XMLData is None:
             #loading an existing file
             self.doc = parseString(XMLData)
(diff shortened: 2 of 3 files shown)


barentine:
This will need a bit more work for e.g. an nd2 file with multiple series stored in it.

David-Baddeley:
The canonical keys are `Positioning.x`, `Positioning.y`, `Positioning.z`. We also save the `Positioning.<stage name>` because a) - that was what we started doing,  and b) we still want to persist positioning for non-canonical stages - e.g. a phase piezo in a SIM setup, or stacked stages (Stage_X for the stepper motor stage, Piezo_x for the piezo on top). When reading metadata from systems which weren't running PYME we should probably just populate Positioning.x etc ...

David-Baddeley:
IE there is usually redundancy in the positioning metadata - x, y, z, and then each stage by it's assigned display name.

David-Baddeley:
In the metadata handler accessing `Positioning.PiFoc` only occurs if we can't find `Positioning.z` (ie it's a backwards compatibility fallback). In eventView we don't look for the cannonical ones (which is a bug in eventView, although likely not one which will bite us as imported data doesn't usually have a concept of events)

barentine:
Good catch. Thanks, @David-Baddeley ! I've switched the keys to `positioning.x` and the like.

David-Baddeley:
Cool, will merge. Note that there will likely be some caveats with actual alignment due to camera rotation, mirrors in the optical path, and sign conventions. We have ways of dealing with this in PYMEAcquire, although they are fragile. On incoming data we have no way of guessing, so don't expect to be able to tile (or similar). Even overlaying different sized FOVs is likely to prove challenging.

David-Baddeley:
Maybe we should add a coordinate-transform option to the metadata, although this would need some thought (should support rotation and mirroring, not sure about scaling, which feels like it might not be such a good idea.

PR #1700, opened 2026-08-27
Metadata plugins

David-Baddeley:
Addresses issue #1899

Adds a new and hopefully simpler method of registering plugins by specifying entry points in the plugins pyproject.toml. This means that you can just pip-install the plugin and PYME will find it.

@csoeller @barentine - this should nicely handle the plugin installation workflow for pip based plugins discussed in email.

@zacsimile  - Have already ported ch-shrinkwrap
@barentine - Haven't done anything to nep-fitting

Docs below:

Declare entry points in your package's ``pyproject.toml``. PYME discovers them automatically on startup with no
file copying. This works identically for pip, conda, and editable installs.

```toml

    [project.entry-points."pyme.plugins.recipes"]
    myplugin        = "mypackage.recipe_modules"
    myplugin_extra  = "mypackage.extra_modules"

    [project.entry-points."pyme.plugins.visgui"]
    myplugin = "mypackage.visgui_modules"

    [project.entry-points."pyme.plugins.dsviewer"]
    myplugin = "mypackage.dsviewer_modules"

    [project.entry-points."pyme.plugins.fit_factories"]
    myplugin = "mypackage.fit_factories"
```

The entry point *name* (left-hand side) is arbitrary and ignored by PYME; the *value* is the fully qualified module
path. An optional ``:attr`` suffix (e.g. ``mypackage.module:some_func``) is accepted but only the module path is used.

### DIFF
--- a/PYME/config.py
+++ b/PYME/config.py
@@ -104,7 +104,7 @@ method of plugiin registration is supported for backwards compatibility only - n
 <plugin-name>.yaml config file as detailed below.
 
 
-pyproject.toml entry points (preferred for pip/wheel packages)
+pyproject.toml entry points (recommended for all packages)
 -------------------------------------------------------------
 
 The recommended approach for pip-installable plugins is to declare ``importlib.metadata`` entry points in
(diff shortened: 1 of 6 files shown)


David-Baddeley:
going to go ahead and merge this and re-release as there's a user waiting on the ch-shrinkwrap and the new ch-shrinkwrap installer already uses these endpoints.

PR #1703, opened 2026-09-03
Simulation tidying (and new functionality)

David-Baddeley:
A bunch of tidying up of simulation code, with the slightly longer term plan on making it easier to simulate without running through PYMEAcquire.

Also has a few deprecation fixes, and some incomplete packaging fixes (we can now build a standalone windows installer which doesn't require doing a pip fetch at the far end, albeit a rather bloated one (somehow we get 2 copies of PYME, python, and all the modules installed)

Some new functionality (mostly a novelty) in that I've added confocal and STED PSFs to the simulator (really only sensible if you just use the 405 laser)

Sted mode ... 
image

vs widefield
image


David-Baddeley:
Going to merge early as I really want some of the bugfixes in a new install .. (and this shouldn't break anything, fingers crossed)

Issue #1706, opened 2026-09-12
[BUG] widefield shift map dialog doesn't open


barentine:
**Describe the bug**
<details> <summary> Traceback </summary>

'''
 Error whilst running Processing>Apply vector shifts
==============================================
python-microscopy=26.09.03.post0.dev[git]
python=3.11.15, platform=darwin
numpy=2.4.3, wx=4.2.5 osx-cocoa (phoenix) wxWidgets 3.2.9

Traceback
=========
Traceback (most recent call last):
  File "/Users/andy/code/python-microscopy/PYME/ui/progress.py", line 153, in func
    return fcn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/code/python-microscopy/PYME/DSView/modules/composite.py", line 625, in OnApplyShiftmap
    dlg = ShiftmapSelectionDialog(self.dsviewer, self.image)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/code/python-microscopy/PYME/DSView/modules/composite.py", line 181, in __init__
    sizer2.Add(wx.StaticText(self, -1, image.names[chan]), 0, wx.ALIGN_CENTER_VERTICAL | wx.ALL, 5)
wx._core.wxAssertionError: C++ assertion ""Assert failure"" failed at ./src/common/sizer.cpp(1606) in DoInsert(): too many items (5 > 2*2) in grid sizer (maybe you should omit the number of either rows or columns?)

'''

</details>

**To Reproduce**
1. PYMEImage -t
2. Click on 'Processing>Apply shift vectors'

PR #1705, opened 2026-09-10
remove invalid flag from dictchoice dialog

barentine:
Addresses issue #invalid wx alignment flags stoped the dictstringeditor from coming up when trying to e.g. use `Output.HDFOutput` recipe module in the recipe editor.

<details> <summary> Traceback </summary>

'''
Traceback (most recent call last):
  File "/Users/andy/code/python-microscopy/PYME/ui/custom_traits_editors.py", line 277, in OnFilterAdd
    dlg = DictChoiceStrEditDialog(self, mode='new', 
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/andy/code/python-microscopy/PYME/ui/custom_traits_editors.py", line 134, in __init__
    vsizer.Add(hsizer, 0, wx.ALIGN_CENTER_VERTICAL | wx.ALL, 5)
wx._core.wxAssertionError: C++ assertion "CheckSizerFlags(!((flags) & (wxALIGN_CENTRE_VERTICAL)))" failed at ./src/common/sizer.cpp(2273) in DoInsert(): wxALIGN_CENTRE_VERTICAL will be ignored in this sizer: only horizontal alignment flags can be used in vertical sizers

DO NOT PANIC !!

If you're an end user running a program not developed by you, please ignore this message, it is harmless, and please try reporting the problem to the program developers.

You may also set WXSUPPRESS_SIZER_FLAGS_CHECK environment variable to suppress all such checks when running this program.

If you're the developer, simply remove this flag from your code to avoid getting this message. You can also call wxSizerFlags::DisableConsistencyChecks() to globally disable all such checks, but this is strongly not recommended.
'''

</details>

**Is this a bugfix or an enhancement?**
maintenance
**Proposed changes:**
remove flags which were ignored anyway

**Checklist:**

wxpython                  4.2.5           py311h8325047_2    conda-forge
wxwidgets                 3.2.9                h34c1157_0    conda-forge

Screenshot 2026-09-10 at 2 40 47 PM

PR #1707, opened 2026-09-12
remove invalid wx sizer flags

barentine:
Addresses issue #1706 and follow-ons from #1705. Also capitalized an `L` in `wx.HORIZONTAL` so it resolves even though it isn't used.

bugfix on later wxwidgets/wxpython

**Proposed changes:**
follow-on from #1705 , using a Claude-coded parse of our source code to look for invalid wx sizer flags. Added the check as a unit test to prevent us from doing this in the future.

Remove horizontal sizer flags on horizontal sizers, align with expand, etc.

**Checklist:**
Not all of these have been tested, but I spot checked several and they crash us out like #1705 does on 
wxpython                  4.2.5           py311h8325047_2    conda-forge
wxwidgets                 3.2.9                h34c1157_0    conda-forge

(spot checked 3 files fixes, mentioned in commit messages, by removing the fix and hitting the bug through the GUI)

**Notes:**
Some of these are tough to test at the moment because they exist in code which is never called. Some things we might want to just remove include `LaserControlFrame.py` and `IntegrationSliders.IntegrationSliders_`. Several others under e.g. if False statements. But with the test hitting them as failures I added them as fixes.

### DIFF
--- a/PYME/Acquire/Hardware/AndorNeo/AndorNeoControlFrame.py
+++ b/PYME/Acquire/Hardware/AndorNeo/AndorNeoControlFrame.py
@@ -69,7 +69,7 @@ class AndorNeoPanel(wx.Panel):
 
         self.tCCDTemp = wx.TextCtrl(self, -1, '0', size=(30, -1))
         hsizer2.Add(self.tCCDTemp, 1, wx.wx.RIGHT|wx.ALIGN_CENTER_VERTICAL, 5)
-        sbCooling.Add(hsizer2, 0, wx.EXPAND|wx.ALIGN_CENTER_HORIZONTAL, 0)
+        sbCooling.Add(hsizer2, 0, wx.EXPAND, 0)
 
         self.bSetTemp = wx.Button(self, -1, 'Set', style=wx.BU_EXACTFIT)
         self.bSetTemp.Bind(wx.EVT_BUTTON, self.OnBSetTempButton)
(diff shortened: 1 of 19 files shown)


David-Baddeley:
Looks like Claude's got a bunch of the wrong ones. `ALIGN_CENTRE_VERTICAL` is valid and often desired on a horizontal slider (e.g. to make labels align with their corresponding text fields), and vice-versa.

David-Baddeley:
On closer inspection, they might be ones with `EXPAND`, but I think this unfortunately needs a case-by case walkthrough. It might be the `EXPAND` rather than the `ALIGN_CENTRE_VERTICAL`  which should go.

barentine:
I wasn't grossly offended by the way the ones I spot checked rendered:
`PYME/ui/progress.py`
`PYME/DSView/modules/segmentation.py`
`PYME/DSView/modules/composite.py`
I also verified that they failed/crashed as described in #1706 without the fix, so I think the alternative to some ugliness is a lack of function.

Are there any others that you think are worth manually checking @David-Baddeley ? It is difficult to test many of them because they are in completely unused code (which explains why we hadn't caught them).

David-Baddeley:
Just had a look through - agree the exposed surface is pretty small. Will go ahead and merge, and if anything bugs us can loop back and fix it if and when it becomes an issue.

David-Baddeley:
I think `wx.EXPAND` is probably over-used in the codebase - there are places where we've used `wx.EXPAND` where we really should have just set the priority.

